PR #181, opened 2014-12-27
Removed a redundant call to clif_party_info on party loading.

Lemongrass3110:
As you can see here the info messages get displayed twice:
screenrathena001

Here is the call hirachy in a diagram:
preview

How to reproduce?
1) Start your server
2) Log in with a user that has a party
3) The server will load the party and notify all party player[this is the first message that will be sent to the player]
4) When the client finished loading and requests the party information the function will be called again[this is the second time the player will see the info]

After this everything will be only displayed once as long as the server is up and running(at least for this party).

### DIFF
--- a/src/map/party.c
+++ b/src/map/party.c
@@ -319,7 +319,8 @@ int party_recv_info(struct party* sp, uint32 char_id)
 
 		clif_charnameupdate(sd); //Update other people's display. [Skotlex]
 		clif_party_member_info(p,sd);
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
Here is additional information provided by ziu:
screenshot-ziu

@spawn @lighta 
I think you should merge this and we create another issue for the zone server issue.

lighta:
Quick testing of change right now and merge.

PR #191, opened 2015-01-01
A little cleanup for Pet DB

cydh:
* Separated pet_db for Renewal & Pre-Renewal to avoid non-existant monster in different mode.
* Cleaned script command in pet_db.txt file, just point it to doc/script_commands.txt. Read them there!
* Separated damage value for petskillattack2 to its own var 'damage', that previously use 'lv' var to store the value.
* Added config 'pet_ignore_infinite_def' in conf/battle/pet.conf,
  * If disabled (val:"no"), fixed damage from pet script 'petskillattack2' will be adjusted to 1. Example, just like player attacks Shining Plant.
  * Enabled by default. Previously, fixed damage ignore the infinite defense, that's why it's enabled by default.
  * TODO: Correcting the logic and maybe also confirm the correct behavior!
* Added validation for pet script commands,
  * Skill for petskillattack, petskillattack2, & petskillsupport
  * SC range for petrecovery.
* Cleaned up some source documentation.

### DIFF
--- a/conf/battle/pet.conf
+++ b/conf/battle/pet.conf
@@ -75,3 +75,6 @@ pet_max_atk2: 1000
 // If set to yes, pets are automatically returned to egg when entering castles during WoE times
 // and hatching is forbidden within as well.
 pet_disable_in_gvg: no
+
+// Will does petskillattack2 fixed damage ignore plant infnite defense? (Note 1)
+pet_ignore_infinite_def: yes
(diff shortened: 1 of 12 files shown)


aleos89:
Looks like you beat me to separating the pet database versus renewal and pre-renewal. :P I'll test this out shortly.

cydh:
*sigh* can't auto merge

cydh:
ehh, @aleos did u test to merge manually before merge this? github said cannot be merge because will cause some conflicts.

aleos89:
Yup, I manually merged it over.

PR #206, opened 2015-01-06
Import support for map data

Lemongrass3110:
This pull should provide the rAthena users with the possibility to add maps by putting them into import files.


src/common/mapindex.c
```diff
@@ -134,29 +132,44 @@ void mapindex_init(void) {
 	int last_index = -1;
 	int index;
 	char map_name[MAP_NAME_LENGTH];
+	char* mapindex_cfgfile[80] = {
+		"db/map_index.txt",
+		"db/import/map_index.txt"
```
cydh: why don't `db_path` & `DBIMPORT`?

cydh:
then, "how can I remove map by using the import file?" :P

Lemongrass3110:
You can do that in conf/import/map_athena.txt by writing: "delmap: mapname" at least that is ought to work. Did not test to confirm this though.
Deleting something from the cache or index list is unnecassary I guess. You are also able to override data if necessary.

src/common/mapindex.c
Lemongrass3110: As you can see in b5e30e0 it was hardcoded before too. I will adjust this later today. ;)

cydh:
if tested and good, just merge it. I have nothing to test (yet)

Lemongrass3110:
I tested it with Arsinoe from Aerie. :)

cydh:
tested also for __replacing__ only __a__ mapindex or mapcache?

PR #215, opened 2015-01-10
Fixed missing NPC IDs in KVM scripts

Patskie:
Made by Patskie

### DIFF
--- a/npc/battleground/kvm/kvm01.txt
+++ b/npc/battleground/kvm/kvm01.txt
@@ -3,7 +3,7 @@
 //===== By: ================================================== 
 //= ???, L0ne_W0lf
 //===== Current Version: ===================================== 
-//= 1.2
+//= 1.3
 //===== Compatible With: ===================================== 
 //= rAthena Project
 //===== Description: ========================================= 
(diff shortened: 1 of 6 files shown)


cydh:
what's __"Work"__?

Patskie:
I am not so good at pull requests. My first or second time using it. So i never thought that git commit -m "meesage" would be my title LOL

Lemongrass3110:
lol I just created a PR for this, because I did not see this one.

I'll adjust your title. ;)

euphyy:
Why is this wrong?  The label name is unique, and is intended to be executed on each duplicate.

Atemo:
> If the supplied event label has the form "NpcName::OnLabel", then only given
NPC's event label will be invoked (much like 'goto' into another NPC). If the
form is "::OnLabel" (NPC name omitted), the event code of all NPCs with given
label will be invoked, one after another.
PR invalid however I think the NPC for kvm2 and kvm3 should call the label OnKvM02One, OnKvM02Two, OnKvM03One, OnKvM03Two

Atemo:
Thanks for the PR though.

Lemongrass3110:
Sorry Atemo, but does not your explanation tell us, that we have something wrong at the moment?
We currently have ::OnKvM01One which would trigger all OnKvM01One labels on the whole mapserver. Obviously this would trigger the NPC event on all 3 KvM files or am I wrong?
Instead of renaming them, why not adress them directly as it is intended?

Correct me if I am wrong please!

PR #222, opened 2015-01-12
Fixed #204

### LINKED ISSUE #204: Boss monsters not using skills when setted on import/mob_db.txt
Boss monsters aren't using skills when they are setted on import/mob_db.txt, for some reason this only affects boss monsters and not the normal ones. Even if you use the same exact line from pre-re or re mob_db.txt

cydh:
* Separated how monster's skills to be assigned to each monster to `mob_skill_db_set()` function
* Added temporary `DBMap *mob_skill_db` to store monster skills by using `struct s_mob_skill`

### DIFF
--- a/src/map/mob.c
+++ b/src/map/mob.c
@@ -53,10 +53,11 @@
 #define ACTIVEPATHSEARCH
 
 //Dynamic mob database, allows saving of memory when there's big gaps in the mob_db [Skotlex]
+//!NOTE: Mob ID is used also as index in mob db
 struct mob_db *mob_db_data[MAX_MOB_DB+1];
 struct mob_db *mob_dummy = NULL;	//Dummy mob to be returned when a non-existant one is requested.
 
-struct mob_db *mob_db(int index) { if (index < 0 || index > MAX_MOB_DB || mob_db_data[index] == NULL) return mob_dummy; return mob_db_data[index]; }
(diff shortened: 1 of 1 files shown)


src/map/mob.c
```diff
@@ -4719,6 +4712,85 @@ static void mob_drop_ratio_adjust(void){
...
+	uint8 i;
+
+	nullpo_retv(mob);
+	nullpo_retv(skill);
+
+	for (i = 0; mob->maxskill < MAX_MOBSKILL && i < skill->count; i++) {
+		mob->skill[mob->maxskill++] = skill->skill[i];
+	}
+
+	if (i < skill->count)
```
lighta: using i here it's correct but weird. it's prone to create an error one day. I think (MAX_MOBSKILL < skill->count) would have been a easier choice.
And beside how could i be superior if ( i < skill->count) is a out condition for the loop above..

cydh: how if u try to import something? somehow my mood is bad today. :P

currently MAX_MOBSKILL is 50, only 2 mobs officially have 50 skills. then try u import some skills for those 2 mobs (up to u the count) or import skills for -1/-2/-3
try debugging it.

lighta: alright make sense now. (u wake up on the wrong foot ?)
I couldn't test live right now so I did it in my head but ye you was right the import will cause issue about my proposition =)
just a note for thoses wondering
loop 1st cond is to ensure outofbound, 2nd is to ensure we loop tought all defined skill.
then we have the check when the skill-count is over MAX_MOBSKILL, like for say 50+2. causing us to have that cond false (mob->maxskill < MAX_MOBSKILL) so the check for skipping to act.

PR #228, opened 2015-01-14
Eclage update r3

Skormie:


npc/re/cities/eclage.txt
```diff
@@ -1,469 +1,89 @@
...
+//= 1.0.1
 //===== Compatible With: ===================================== 
 //= rAthena Project
 //===== Description: ========================================= 
-//= Eclage Town Script
+//= Eclage Town NPC
 //===== Additional Comments: ================================= 
-//= 1.0 Nearly complete iRO town NPCs. [Euphy]
+//= 1.0 Generic NPC place holders. [Euphy]
+//= 1.0.1 Converted from the official Aegis script. [Dastgir]
```
cydh: 2.0 since almost a whole script are different with before

npc/re/merchants/coin_exchange.txt
```diff
@@ -13,9 +13,10 @@
 //= 1.0 First version. [Euphy/Lemongrass]
 //= 1.1 Added remaining Malangdo traders. [Euphy/Lemongrass]
 //= 1.2 Added Eclage traders (not fully complete). [Euphy]
+//= 1.2.1 Completed Eclage Traders. [Dastgir]
```
cydh: just make it `1.3` since scripting almost never the 3rd version number

npc/re/warps/dungeons/ecl_dun.txt
```diff
@@ -0,0 +1,26 @@
...
+//===== Current Version: =====================================
+//= 1.0a
+//===== Compatible With: ===================================== 
+//= rAthena Project
+//===== Description: =========================================
+//= Warp Portals for Eclage Dungeon
+//===== Additional Comments: =================================
+//= 1.0 First Version.
+//= 1.0a Corrected Warp Co-ordinates and 
+//=      name of Warp Portal. [Dastgir]
```
cydh: 1.1

cydh:
for this, https://github.com/Skormie/rathena/blob/eclage_update_r1/npc/re/quests/quests_eclage.txt
make it 2.0

Skormie:
Does it really matter that much... Honestly for these version numbers to be modified so much. It can only be a matter of opinion, and if you want my opinion as long as the increments are increasing and a point is made then they're fine.

Atemo:
We just follow semver from now on but I only care about the number for major change.

npc/re/cities/eclage.txt
Skormie: I agree with this...

npc/re/merchants/coin_exchange.txt
Skormie: I think it should be 1.3 but not because the third version number isn't used.

PR #244, opened 2015-01-23
Light of Regene

RagnarokNova:
Due to the way the source handles hom intimacy, 251 actually sets intimacy to 2 (Hate with a passion). The two extra 0s make it the appropriate value.

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -10171,7 +10171,7 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case MH_LIGHT_OF_REGENE: //self
 		if(hd) {
 			struct block_list *s_bl = battle_get_master(src);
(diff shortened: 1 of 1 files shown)


src/map/skill.c
```diff
@@ -10170,8 +10170,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	     break;
 	case MH_LIGHT_OF_REGENE: //self
 		if(hd) {
+			struct block_list *s_bl = battle_get_master(src);
+			if(s_bl) sc_start2(src, s_bl, type, 100, skill_lv, hd->homunculus.level, skill_get_time(skill_id, skill_lv));
```
cydh: just need sc_start(), the val2 is unused. can u do a follow up?
````
sc_start(src, s_bl, type, 100, skill_lv, skill_get_time(skill_id, skill_lv));
````

PR #245, opened 2015-01-23
groups.conf import support

cydh:
* Moved hardcoded groups.conf filename on `src/map/pc_groups.c` to `conf/map_athena.conf`
* Defined the path with `groups_conf_file: conf/groups.conf`. So in `conf/import/maps_conf.txt` file (imported map_athena.conf file) can be used to redefine the path, `groups_conf_file: conf/import/groups.conf`

### DIFF
--- a/conf/import-tmpl/groups.conf
+++ b/conf/import-tmpl/groups.conf
@@ -1,0 +1,1 @@
+// See conf/groups.conf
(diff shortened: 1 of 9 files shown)


cydh:
somehow the comment in `conf/import-tmpl/groups.conf` is weird

aleos89:
Is a comment really needed in it since all the other import files are empty?

cydh:
xD I like comments.
I'll remove it and add import for atcommand_athena.conf

aleos89:
:) Sounds good.

cydh:
I'll drop this. lol
I'll include this change with #252

cydh:
cloossseee

PR #250, opened 2015-01-25
3rd Job Quest compatibility fix

Lemongrass3110:
The job quests now (all) support servers where the 2nd classes can be over baselevel 99.
Most of them were correct anyway, because they check for BaseLevel > 98 and JobLevel > 49.


cydh:
sounds good /slur

Atemo:
Seems correct

PR #251, opened 2015-01-26
Moved skillratio calculations to battle_get_skill_ratio()

cydh:
* Moved skillratio calculations to battle_get_skill_ratio()
  * From battle_calc_magic_attack()
  * From battle_calc_attack_skill_ratio(), called from battle_calc_weapon_attack()
  * TODO from battle_calc_misc_attack()
* Clean up some macros on battle.c
* Moved magic base damage calculation to battle_calc_magic_attack_base() from battle_calc_magic_attack()

* NOTE: Not tested all yet. :D


Stolao:
this is now over 6 months old, shouldn't we do something with this?

PR #265, opened 2015-02-02
Wind Walk not affecting party members.

RagnarokNova:
For whatever reason, Wind Walk wasn't affecting party members on my server. After swapping the weapon to magic, it now magically works. Figured I'd share it.


cydh:
whatever the reason, seems it's working properly

Issue #254, opened 2015-01-27
RG King's Grace cooldown


raynra:
Git Hash/Date: 1f147f2 / Jan-26-2015
Client Date: 2013-08-07a
Description:

http://youtu.be/xFJeUxULtx4
http://forums.warpportal.com/index.php?/topic/65542-hayroh-phoenixs-royal-guard-17560-kings-grace-update-buildsguide-paladin-woe-te-iro/
https://rathena.org/board/topic/94989-new-3rd-job-skill-effect/?p=255230

Seems like there is a 1 minute cooldown on iRO/kRO and a higher cooldown on ruRO.
I'll say our current no cooldown is likely incorrect.

If it's a flat 1 minute cooldown regardless of level, then in skill_cast_db:
5013,1000,0,0,5000,0,60000,-1

PR #266, opened 2015-02-02
King's Grace Cooldown (Issue #254)

RagnarokNova:
As per https://github.com/rathena/rathena/issues/254


aleos89:
Shouldn't this also be applied to pre-re?

aleos89:
Fixed in 083cf5d.

PR #268, opened 2015-02-03
Follow up to #266

RagnarokNova:
Added the CD to pre-re as well.

https://github.com/rathena/rathena/pull/266


db/pre-re/skill_cast_db.txt
```diff
@@ -1410,7 +1410,7 @@
 2325,2000,2000,0,30000:45000:60000:75000:90000,0,540000:480000:420000:360000:300000
 
 //-- LG_KINGS_GRACE
-5013,1000,0,0,5000,0,0
+5013,1000,0,0,5000,0,60000,-1
```
aleos89: This is wrong, there shouldn't be a 8th value as that only applies to renewal for fixed cast.

cydh:
why do u have to make new PR for follow up ur own active PR? D:
u just need to edit the files in same branch in ur active PR #266, then commit & push

aleos89:
Fixed in 083cf5d.

PR #287, opened 2015-02-12
Clean up Skill DB

### LINKED ISSUE #277: Quest Skill Randomly Lost
So here's my problem, last commit that i used is this : 
https://github.com/rathena/rathena/commit/b43b855d210c627f06863c50a817ba8797a7db55
(no big problem, working as intended)

Then i pull this one
https://github.com/rathena/rathena/commit/90d9da18e1d1980c52e70b5baeaa9bcd9e769390

And quest skills (such as Lord Exclamation, Cart Change, etc) are randomly lost when re-log, or using healer npc (the one with SC_START blablabla).

First i thought it was my customization that cause it, but then i tried again on my other server that has no customization at all (i'm only using custom npc from rA default like healer, reset, etc), but the skills keeps randomly lost everytime i logout. Lord exclamation was worse, it acts like scroll-based skills, one time usage then its lost.

So, i switch to this commit : 
https://github.com/rathena/rathena/commit/a427dbf5132de1351caa8c337389a92e3072df3a

And all the problems are gone. Is it because of the skill db clean up?

cydh:
* Changed skill_db array to better memory allocation.
* Reduced MAX_SKILL_LEVEL from 100 to 10 (part of Hercules https://github.com/HerculesWS/Hercules/commit/0f4a50d13538c3e5d3ca4d3822f92217c8da0479), also has changes on skill_get checks for skill level that more than 10.
* Reduced MAX_SKILL from 5020 to 1200 (actually only 1109 skills are used).
* Added macros for checking Homunculus, Guild, Mercenary, & Elemental skill ranges.
* Added skill check & index validation when player logged in and when @reloadskilldb.
* Corrected `enum e_skill_flag` order for SKILL_FLAG_REPLACED_LV_0's sake.
* Merged 'addtoskill' script command just as alias of 'skill' script command.
* Fixed #277
* Changed 'skill' script command flag to constant value

NOTE:
* Decreased memory usage by map-server reduced up to 50 MB.
* Decreased memory for each player because of mmo_charstatus::skill[] only has 1200 array, before is 5020.
* Please use skill_get_index() for accessing sd->status.skill[] or skill_db[], don't reckless use skill_id as array index.
* Please import upgrade_20150211_skillset.sql


aleos89:
Been testing this for the past few days and everything seems to be fine now.

PR #288, opened 2015-02-12
Fixed #139 Vacuum Extreme effect

### LINKED ISSUE #139: Vacum extreme
Client Date: **2012-04-10**
Src Mods: **NONE**
General settings: 
- **Full Renewal**
- **1/1/1/1/1 rate**

How to reproduce: **Use the skill**
Description:

The recurring issue has returned once again. Latest vers, re, pk server, 20140410 client, vacuum extreme does not stop monsters. Neither moving ones, nor static ones from starting to move.

cydh:


cydh:
I need testers

reigneil:
can you also modified the duration of this to reflect with the official.
https://docs.google.com/document/pub?id=1I2byIqDij02jwl6Tn4ekgHff0E501JP16iwYLc9vhDE
it is stated that the duration is only affected by BASE str.
Duration: [(2 + 2 * Skill Level) - (Target’s Base STR / 20)] seconds
something like this...
tick -= (max(status_get_base_status(bl)->str,1) / 20) * 1000;
correct me if i'm wrong.

RagnarokNova:
You can currently keep players and enemies permanently under the status effect, which ends up being pretty broken as a mechanic.

cydh:
merged manually @ cb2ffa24c8fbd4c0170fc4325683beffe0a7d4d8

PR #293, opened 2015-02-16
Init `warp` support for other map in other map-server.

cydh:
* Added warp timer if the map destination is in other map-server, all opened script will be terminated immediately!
* Make sure your 'warp' is always in the end of NPC, like `... close2; warp "payon",0,0; end; ...'

TODO:
* `warpparty`
* `warppguild`
* `areawarp`
* `warppartner`
* `warpwaitingpc`
* `warpchar` (eh?? why don't just merge this to `warp` and add optional param for char_id?)

### DIFF
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -1739,6 +1739,9 @@ int map_quit(struct map_session_data *sd) {
 	if (sd->autotrade_tid != INVALID_TIMER)
 		delete_timer(sd->autotrade_tid, pc_autotrade_timer);
 
+	if (sd->warp.tid != INVALID_TIMER)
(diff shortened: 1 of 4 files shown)


cydh:
hmm, don't merge this. still messed up from past test

cydh:
why don't u try use normal 'warp' on script to move u to other map on other map-server?

Lemongrass3110:
I just wanted to tell you, that you can copy/compare the code from there.

icxbb-xx:
atcommand @warp @goto not support multiple mapserver too

Lemongrass3110:
The atcommand @warp does support it (partially). Mapflag checks etc cant work, but the rest should be ok.
@goto does not work yes.

PR #294, opened 2015-02-16
Moved Bank Vault to char-server

cydh:
Follow up first Bank feature implementation at 1290826
* Moved `bank_vault` from `login` (login-server access) to char-server access by default, avoiding to do transaction between 2 different kind of char-server.
* Dropped `bank_vault` field from `login` table, moved the value to `global_reg_value` with default var name #BANKVAULT
* Defined bank vault variable in pc.h::#define BANK_VAULT_VAR
* Added `BankVault` as script constant.
* Removed `bank_vault` related from doc/packet_interserv.txt

NOTE:
* Please read a note in `upgrade_20150216.sql` before import it!


icxbb-xx:
I have test this ok : ) working

but in chrif.c
line 30 not use ?
int chrif_save_bankdata(struct map_session_data *sd);

cydh:
leftover, u know that, just remove it

RadianFord:
char_mapif.c:684: warning: unused variable ‘val2’

cydh:
how come it becomes unused, it's clearly used in https://github.com/rathena/rathena/pull/294/files#diff-23da1d3326d24b8a3555248cb239f90bR684
````
		int val2 = RFIFOL(fd,40);
````

**Comment by RadianFord** [comment]
@cydh that came out when i recompile my server..

**Comment by cydh** [comment]
oh rite, confirmed that it's really unused

**Comment by RadianFord** [comment]
Its okay to delete it?

**Comment by cydh** [comment]
just comment the line

**Comment by RagnarokNova** [comment]
Any updates on this? :] Currently, when my mapserver crashes, sometimes some character's bank balances are deleted, I'm wondering if this will fix that.

File: src/map/pc.c [hunk 1]
```diff
@@ -8814,6 +8823,26 @@ bool pc_setregistry_str(struct map_session_data *sd,const char *reg,const char *
...
+
+	val = cap_value(val, 0, INT_MAX);
+
+	return (reg[1] == '#') ? pc_setaccountreg2(sd, reg, val) : pc_setaccountreg(sd, reg, val);
+}
+
+int pc_readaccountreg_(struct map_session_data *sd, const char *reg) {
+	nullpo_ret(sd);
+
+	if (reg[0] != '#' || reg[strlen(reg)-1] == '$')
```
cydh: should be
````
	if (reg[0] == '.' || reg[0] == '$' || reg[0] == '\'' || reg[strlen(reg)-1] == '$')
````

PR #300, opened 2015-02-21
Hotfix for issue 198

### LINKED ISSUE #198: Walkpath Problem
I found an infinity loop when trying to loot an item.

How to reproduce?
1) Drop an item.
2) Put a character on the right cell beside the dropped item.
3) Walk to the right cell next to the chracter.
4) Try to grab the item.
5) You will be stuck in an infinity loop walking against the character and back to the cell where you stood.

I think that might be because of the cell stack limit(1 unit per cell) and when trying to grab the item it might want to place my character on the same cell as the other character.

screenrathena007

Lemongrass3110:
Ignores incoming walk request as on official servers

Fixes #198

### DIFF
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -81,6 +81,16 @@ int unit_walktoxy_sub(struct block_list *bl)
 		return 0;
 
 #ifdef OFFICIAL_WALKPATH
+	if( battle_config.official_cell_stack_limit == 1 && check_distance( ud->to_x - ud->bl->x, ud->to_y - ud->bl->y, 1 ) ){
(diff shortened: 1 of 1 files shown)


RadianFord:
Cool can i test this one?

Playtester:
I have written a slightly different solution to this one which I think is better, I already discussed with LemonGrass about it.

The main question is if this movement prevention should only happen on official settings or also if you increase the number of possible units on the same cell as in theory even there the loop could occur.

Lemongrass3110:
Closing this one, because @Playtester fixed it in d4aa4c70a7

PR #354, opened 2015-03-20
Corrected Tiucknuc mob name into Tiyanak in /npc/re/quest/quest_malaya.txt

musashi076:
To avoid confusion in Tiyanak Mob Hunt.

### DIFF
--- a/npc/re/quests/quests_malaya.txt
+++ b/npc/re/quests/quests_malaya.txt
@@ -5682,7 +5682,7 @@ malaya,283,265,4	script	Village Chief#bako2	580,{
 		mes "[Village Chief]";
 		mes "^0000FFLost Belongings^000000";
 		mes "are said to be carried by";
-		mes "^FF0000Tiucknuc^000000 that roams Baryo Mahiwaga.";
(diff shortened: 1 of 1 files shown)


cydh:
@Atemo @nanakiwurtz @Luciar can u check on official/Aegis?

Atemo:
It's Tiucknuc in aegis, maybe a mistake since it doesn't follow the mob name..?

musashi076:
based in pRO and iRO. I think aegis pronounce it in their native pronounciation. 
Tiucknuc = Tiyanak
Mongkukuromu = Mangkukulam (Witch here in PH)
Bongisungisu= Bungisngis

i was doing the Port malaya quest in normal gameplay, also can we put the Lost Belongings Drop of Tiyanak/Tiucknuc in quest since it doesn't drop normally? After i saw the official gameplay, they only drop when in quest and put directly in your inventory. I tried the hercules script and modified it to limit drop per quest npc (1st npc=2, 2nd=7, 3rd=3, 4th=5)
based in here.
http://irowiki.org/wiki/Bakonawa_Extermination

cydh:
it makes sense. xD just like write English in Japanese

PR #375, opened 2015-04-01
Event label load order issue

cydh:


cydh:
afaik @emistry ever _'crying'_ because of this :P then I found current way to executes event label (eg. OnInit) is annoying when I made some scripts :cry:

lighta:
Hmm I'm a bit worry about memory leak for that one.
Otherwise I didn't get the ordering part. npc_event_doall vs npc_event_data_doall ehh, I don't see quite the difference.
They will both do all "OnInit" at once without order, (well the order is actually the one in which they was include in memory, So the order of scripts_athena.conf + txt order.
A way to force this will be to add an argument in the label.
e.g : OnInit::1, then you'll have to check if they're wasn't another  OnInit::1 in parse time, and do something to return the min on the queue, (or order them like a priority queue)

cydh:
`npc_event_doall` executes labels from `ev_db` (`ev_db = strdb_alloc(...)`), it's not a linked list.
`npc_event_doall` calls `npc_event_doall_id` that uses `ev_db->foreach(ev_db,npc_event_doall_sub,&c,buf,rid);` (even using db_iter, it won't be in order, u know the case by atcommand ordering)

PR #377, opened 2015-04-03
Makes 'default' = date2version(PACKETVER)

cydh:
* If `packet_db_ver` is 'default', use packet version based on PACKETVER on mmo.h
* Shows warning if packet version is not match.

### DIFF
--- a/db/import-tmpl/packet_db.txt
+++ b/db/import-tmpl/packet_db.txt
@@ -15,21 +15,29 @@
 // The packet database allows you to add support for new clients,
 // because packets change every release.
 //
-// Note: Every packet version needs a wanttoconnection specification, since
+// NOTE: Every packet version needs a wanttoconnection specification, since
 // that is the packet used to identify a client's version.
 // If multiple versions have the same connection packet, the higher version
 // will be used (unless the lower one is specified as the default)
 //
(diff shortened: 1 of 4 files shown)


cydh:
@aleos89 @lighta

aleos89:
I like this. It's how it should have been from the start. There's no reason for PACKETVER and packet_db_ver to be different.

lighta:
I would accept this.
Now for your question @Cydh, the check seem to be quite recent. (I'm sure he wasn't there before).
So basicaly I think you know this but we have the tab of packet_db to allow multiple client version to access the server in the same time.
Then the way I see it clif, char_clif etc serve as interpreter. basically with your client_packet_version they could have determined they will construct the correct packet for you.
So it's dynamic per client. Now PACKETVER in this approach should tend to disappear. He's here only as historic version and cause not all packet was using the dynamic lookup or as a fallback. If no version was found or the function not yet set as dynamic well we will use PACKETVER that all.

Now if you want the Herc approach where the serv is compiled for only a specific version (PACKETVER). ye you no longer need that struct and that dynamic interpreter approach.
You're clif only format packet the same way and enqueu them. (that really all what they do, take different argument prepare a struct and enqueue into a FIFO).

PR #393, opened 2015-04-11
HAMI_CASTLE fix

### LINKED ISSUE #274: Castling mob count overflow
Client Date: **2012-04-10**
Src Mods: **NONE**
General settings: 
- **Full Renewal**
- **1/1/1/1/1 rate**

How to reproduce & Description:

~~I was farming Mi Gaos, and I got hit multiple times, which caused my garment (Uflheiden) to activate NPC_STONESKIN. Looks like the bonus got totally glitched, because I ended up receiving 1000%+++ more damage from monsters, while others were not affected. When I removed all my equips I received a lot less damage, but it was still ridiculously high. When I re-equipped my stuff I received even much more damage. Even my raydric card effect got reversed and when I equipped my garment I got more damage.~~

~~I'm not entirely sure if this was the result of the NPC_STONESKIN, but it is my only suspect. Something is pretty messed up here.....~~

Just watch, this is a simple hunter fly hitting me:
http://youtu.be/JMPH6rYvDDg

Okay, I figured it out. It's Castling. Every time I use Castling the mobs that were hitting me starts to target the homunculus, but still they count in my character's mob count like they were hitting me, and if I use castling many times I will end up hundreds of virtual mobs "targeting" me, thus greatly increasing my defense penalty, but really none of those are attacking me, and if one mob hits me, it will take the bonus damage.

cydh:
* Fixed #274 Castling (HAMI_CASTLE) makes a whole area attack you!
* Changed `skill_changetarget` to `unit_changetarget`, also fixed the validation before switching the `bl`'s attacking target from current `src` to new `target`.

[ci skip]

### DIFF
--- a/src/map/unit.c
+++ b/src/map/unit.c
@@ -2759,6 +2759,42 @@ int unit_changeviewsize(struct block_list *bl,short size)
 }
 
 /**
+ * Makes 'bl' that attacking 'src' switch to 'target'
+ * @param bl
+ * @param ap
+ * @param src Current target
+ * @param target New target
+ * @param flag 0x1 Force to attack although 'bl' is idle
+ *             0x2 Force to attack although 'bl' has different target (not targetting 'src')
+ **/
+int unit_changetarget(struct block_list *bl, va_list ap) {
+	struct unit_data *ud = unit_bl2ud(bl);
+	struct block_list *src = va_arg(ap,struct block_list *);
+	struct block_list *target = va_arg(ap,struct block_list *);
+	int flag = va_arg(ap,int);
+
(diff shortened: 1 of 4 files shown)


src/map/unit.c
```diff
@@ -2759,6 +2759,42 @@ int unit_changeviewsize(struct block_list *bl,short size)
 }
 
 /**
+ * Makes 'bl' that attacking 'src' switch to 'target'
+ * @param bl
+ * @param ap
+ * @param src Current target
+ * @param target New target
+ * @param flag 0x1 Force to attack although 'bl' is idle
```
cydh: these such a waste

PR #399, opened 2015-04-18
Status Effect Damage

RagnarokNova:
Fixed a few numbers that made it so that certain status effects were adding an additional 100% damage boost than what the formula intended.


cydh:
I saw @aleos89 also in this. wait him

RagnarokNova:
Anyone :( This is a pretty big renewal bug.

aleos89:
Fixed in 583d079. Thanks and sorry for the long wait!

PR #432, opened 2015-05-12
OnPickEvent Label

Zellukas:
This label only works in map with mapflag 'pickevent'.


cydh:
I don't see reason to add this.
but if u willing to add this event for urself, maybe add `picked_item` and `picked_amount` will be good (like killedrid does).
and besides, I failed to understand the state of `pick` that u meant, because I thought *pick* means pick item from the floor, `pc_additem` is from everywhere, storages, cart, getitem script, etc etc

PR #457, opened 2015-06-11
Script Engine Upgrade

aleos89:
* Upgraded the script engine variables to no longer be limited.
* Increased array limit from 127 to 2 billion.
* Improvement in overall speed and storage size.
* All variable types now support arrays.
- Merge from HerculesWS/Hercules@82b583b. Thanks to those who worked on it.


Stolao:
Increased array limit from 128* to 2 billion.
forgetting 0 :)

also downloaded branch will respond with results tomorrow

Stolao:
@aleos89 did several test today before work, everything seemed fine. I'll do some more complicated test when home.

aleos89:
Thanks for the testing. There are a few memleaks on server close which I will be fixing shortly. :)

Stolao:
I tested again using each variable type (NPC,Player,Account,Server) for both strings and int, up to variable array length of 30000ish. I lost the account stinged Array after server restart once (might have not saved the variable before the reset is my guess)

when you update again ill try again, im already loving this feature for my mining system and login rewards scripts,

PR #482, opened 2015-06-23
Waiting room utils:

anubisK:
Two script commands to be used in scripts, that bring a bit more of flexibility when scripting.

    * waitingroomkick "<NPC object name>" , "<character name>"; Kicks one player, there was just a method for several players
    * getwaitingroomusers "<NPC object name>"; get the mumber of players inside a waitingroom and their rids.

The original code was taken form here: http://herc.ws/board/topic/9876-help-my-old-3ceam-src-to-latest-hercules/#entry58710

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6447,6 +6447,20 @@ example.
 
 ---------------------------------------
 
+*waitingroomkick "<NPC object name>" , "<character name>";
+
+This command kicks the given character from the waiting room attached to the given NPC.
+
+---------------------------------------
+
+*getwaitingroomusers "<NPC object name>";
+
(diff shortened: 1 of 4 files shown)


cydh:
so...... where are these script/db dev? really..!!
@Atemo @euphyy @nanakiwurtz

nanakiwurtz:
I can't C... really... -_-'a

cydh:
even just give some thoughts? :disappointed:

nanakiwurtz:
I always opened for new improvements and if it doesn't do any harm to current code, then I'll say "Okay let's merge this" (After getting approval from another Script/DB or Core dev)
:banana:

PR #495, opened 2015-06-29
Item db

ElsaEve:
Need more info on item scripts
//  New Daggers
28703
28704
28705
28706

//  New Shields
28900
28901
28902


cydh:
what is this?

PR #571, opened 2015-08-12
Merged main.sql and logs.sql script to increase database setup simplicity.

GammaCrucis:
These modifications were made in the hopes of simplifying the database setup process by eliminating several steps from the installation guide on the wiki. Additional modifications have been made to increase redundancy in case of an incomplete initial run of the script.

- Modified main.sql and merged it with logs.sql, so that the user just needs to run a single script to create the 'ragnarok' and 'log' schemas and have the proper tables created in their respective schemas. Users no longer have to manually create these schemas, assign them as default, and run the main.sql/logs.sql scripts separately.

- Modified the table creation statement for the login table/the insert statement for the default server communication 'user' to check for the prior existence of that data before inserting it with an INSERT IGNORE statement.

- Did the same as above with the db_roulette table creation statement/insert statements.

- Modified db_roulette tables creation statement to have IF NOT EXISTS.

- Rewrote the db_roulette table creation statement (to fix a bizarre error I was having with it/the rewritten statement is identical). Note that the client which supports the roulette feature (2014-10-22b) is still under stability testing and this should probably be removed from the main.sql script.

- Organized the create table statements by alphabetical order.

Additional:
Cashlog table "types" should be documented as the Picklog, Chatlog, and Zenylog types have been.


GammaCrucis:
Got some feedback over on Hercules of why most of this is a terrible idea. There's some stuff of value done here (added redundancy to the SQL statements, organization of statements by alphabetic order), which I will resubmit as a separate pull request.

I'm going to take a proper look at automating the schema/table creation process using methods that can allow for user input and multi-server setups.

PR #572, opened 2015-08-12
Organized and added redundancy to SQL statements.

GammaCrucis:
- Organized the table creation statements by alphabetical order.

- Added some redundancy to some SQL statements.

### DIFF
--- a/sql-files/main.sql
+++ b/sql-files/main.sql
@@ -1,17 +1,3 @@
-
---
--- Table structure for table `skillcooldown`
---
-
-CREATE TABLE IF NOT EXISTS `skillcooldown` (
-  `account_id` int(11) unsigned NOT NULL,
-  `char_id` int(11) unsigned NOT NULL,
-  `skill` smallint(11) unsigned NOT NULL DEFAULT '0',
-  `tick` int(11) NOT NULL,
-  KEY `account_id` (`account_id`),
-  KEY `char_id` (`char_id`)
-) ENGINE=MyISAM;
-
 --
 -- Table structure for table `auction`
 --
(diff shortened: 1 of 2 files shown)


Freyja-Folkvangr:
Do we need to run an SQL query into an old db to upgrade or something?

anacondaqq:
hmm, how to do backup for 1 db with logs & main data inside on the live server with ~ 200 online where is database size ~ 2-3GB?

GammaCrucis:
I had made a separate pull request for combining the two script into one (though the dbs would be separate, it was an attempt at automating setup a bit more). That turned out to be a terrible idea as pointed out. I closed that pull request for that very reason. My apologies.

This pull request has minor changes to both the main.sql and logs.sql scripts which modifies the insert statements on the login and db_roulette tables to prevent script failure if the data in their respective INSERT statements is already present. The table creation statements already have the functionality to check if the tables exist, so I figured it would be handy to have it on the insert statements as well in case the scripts were to end prematurely for whatever reason.

It also puts the table creation statements in alphabetical order in those two scripts. It was looking kind of messy the way that it was.

sql-files/main.sql
```diff
@@ -814,50 +860,4 @@ CREATE TABLE IF NOT EXISTS `vendings` (
...
-
-CREATE TABLE IF NOT EXISTS `market` (
-  `name` varchar(32) NOT NULL DEFAULT '',
-  `nameid` SMALLINT(5) UNSIGNED NOT NULL,
-  `price` INT(11) UNSIGNED NOT NULL,
-  `amount` SMALLINT(5) UNSIGNED NOT NULL,
-  `flag` TINYINT(2) UNSIGNED NOT NULL DEFAULT '0',
-  PRIMARY KEY  (`name`,`nameid`)
-) ENGINE = MyISAM;
+) ENGINE=MyISAM;
```
cydh: can you recheck all again and add the newline here before we merging this

aleos89:
Thanks for organizing the tables alphabetically! As @cydh stated, do you mind adding the return to the bottom of the file, that way this can be merged.

lighta:
I would actually separate the main into they're supposed realm.
char_db.sql <= most will go here anyway
login_db.sql
map_db.sql

Then main.sql ={
source char_db.sql
source map_db.sql
source login_db.sql
}
And this is without seeing for the log db. like charlog_db, but I didn't investigate those yet.
Why all that micmac ? Well cause it seem it was designed in a distributed manner. Which mean the 3 didn't had to be on the same host. (and then ofc by extension schema).

Why does this matter ? 
Well the current setup make it hard to have that separation setup right away unless you know your stuff. But as I just show it would be easy to keep it for people who would want all in 1 place. (so basically we just preventing an opportunity).
More troublesome is that due to that setup many newcomer and even veteran dev do stuff that break that separation. (which mean in the end without having 3small raspberry to run the serv you need something which costing way more..).
Beside the price issue we end up in a very bad situation where you have a DB designed for distribution with lot of redundancy but used it as single. so in short the bad points of boths design without their advantages...

Atemo:
The owner is inactif I closed this pull and open a new #757

PR #669, opened 2015-10-08
Fixed questlog display for clients from 2014-10-22 on

Lemongrass3110:
@icxbb-xx @aleos89

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -15608,40 +15608,81 @@ void clif_parse_PartyTick(int fd, struct map_session_data* sd)
 
 /// Sends list of all quest states (ZC_ALL_QUEST_LIST).
 /// 02b1 <packet len>.W <num>.L { <quest id>.L <active>.B }*num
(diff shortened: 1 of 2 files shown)


src/map/clif.c
```diff
@@ -15608,40 +15608,81 @@ void clif_parse_PartyTick(int fd, struct map_session_data* sd)
...
 {
+#if PACKETVER >= 20141022
+	#define INFOLEN 15
 	int fd = sd->fd;
 	int i;
-#if PACKETVER >= 20141022
-	int info_len = 15;
-	int len = sd->avail_quests*info_len+8;
-	WFIFOHEAD(fd,len);
+	int offset;
```
icxbb-xx: int offset = 0;

src/map/quest.c
```diff
@@ -48,11 +48,14 @@ int quest_pc_login(TBL_PC *sd)
 		return 1;
 
 	clif_quest_send_list(sd);
+
+#if PACKETVER < 20141022
 	clif_quest_send_mission(sd);
```
icxbb-xx: move int i; to this

PR #714, opened 2015-10-23
Warg Dash Bug Fix

RagnarokNova:
Fixes a bug where you're able to spam Warg Dash while next to a mob
based on your ASPD.
Tested the skill live on iRO, and the skill does not cast if a unit is
right in front of you already.


RagnarokNova:
Re-written.

PR #748, opened 2015-11-09
Feature/pincode extension

Lemongrass3110:
Some additional features for the pincode system. On aegis they use a table to store prohibited values but we came to the conclusion that it would only cause more confusion if each server would prohibit other values.


anacondaqq:
awesome. Thank you! :+1:

PR #749, opened 2015-11-09
Fixing sql upgrade 2014 06 12 losing data when having negative values in card columns

Baalberith6:
Fixing sql upgrade 2014 06 12 losing data when having negative data (e.g. char ids) in card slots. Easiest way to reproduce is on a Named Wedding Ring. After the upgrade, names were not recognized and the item showed as Unnamed.
This was caused by commits 42b29ee and 7cd82d0.

### DIFF
--- a/sql-files/upgrades/upgrade_20140612.sql
+++ b/sql-files/upgrades/upgrade_20140612.sql
@@ -5,35 +5,29 @@ UPDATE `inventory` SET `card0` = 256 WHERE `card0` = -256;
 UPDATE `mail` SET `card0` = 256 WHERE `card0` = -256;
 UPDATE `storage` SET `card0` = 256 WHERE `card0` = -256;
 
-UPDATE `auction` SET `card0` = (65536 + `card0`) WHERE `card0` < 0;
-UPDATE `auction` SET `card1` = (65536 + `card1`) WHERE `card1` < 0;
-UPDATE `auction` SET `card2` = (65536 + `card2`) WHERE `card2` < 0;
-UPDATE `auction` SET `card3` = (65536 + `card3`) WHERE `card3` < 0;
+UPDATE `auction` SET `card1` = (65536 + `card1`) WHERE `card1` < 0 AND `card0` IN(254,255);
+UPDATE `auction` SET `card2` = (65536 + `card2`) WHERE `card2` < 0 AND `card0` IN(254,255);
+UPDATE `auction` SET `card3` = (65536 + `card3`) WHERE `card3` < 0 AND `card0` IN(254,255);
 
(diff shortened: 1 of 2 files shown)


aleos89:
This was followed up in upgrade_20140616.sql but it looks like the log was skipped.

Baalberith6:
Upgrade_20140616.sql deals with a different problem.
Look at these DB rows for Wedding Rings from pre-2014-06-12:
mmsql-inventory-negative-card
Now, applying both 2014-06-12 and 2014-06-16 will cause card0, card1, card2 and card3 to be 254, 0, 0, 2.

lighta:
For me those update only concerne card0 so not sure I'm on track.
I think he's right for named_item and such. (Why not do *-1 btw ?)
My concern is that if people have used other -charid as special kind well doing this would give them item. (But I think that would be special case and if so they should know what they did)

Baalberith6:
To understand, how negative numbers got there, read atcommand.c#getring(sd):
	item_tmp.nameid = item_id;
	item_tmp.identify = 1;
	item_tmp.card[0] = 255;
	item_tmp.card[2] = sd->status.partner_id;
	item_tmp.card[3] = sd->status.partner_id >> 16;
This code just takes the first 16b and store it in card slot 3, then move the pointer by16b and stores it in slot 4.

The only thing we want here is to represent 16b signed int as 16b unsigned int before changing the column types. How do we store negative numbers as 16b signed type? The first 1 is the sign and remaining 15b are for storing the actual number.
Positive numbers start from 0001, 0010, 0011 etc. negative numbers start as 1111, 1110, 1101 etc. So to convert bit representation correctly, we have to do a complement, which is the proposed 65536 - (negative number).
If the explanation is still not enough, I'll be hanging on the IRC channel.

lighta:
Ye I know how to make something signed or not. NB: (you could also do val&0x10000000, that would perform a bit faster then the addition)
Ok so it's from this shift I see thx. 

Still the issue about if someone as decide to do store (-itemid), for fake item the offset fix you propose will make them create "real item".
So imo I'd put the "fix' but I would say to people to apply it carefully, that they should know what they have store in db before this. (They should have use CARD0 as a marker like us but whatever).

Baalberith6:
I guess if they decide to store -itemid in card slots it is their problem, but naming items is a supported rAthena feature. We can play it safe here and convert only those values, which start with card0 = 254 or 255.

lighta:
hmm I think that would be preferable altough yes is their problem.
perhaps a " where CARD0 IN(254,255,...)" //all those we use in CARD0

@aleos89 waiting your final input for this, as I will merge once that little wher is added otherwise

aleos89:
If someone decides to store custom values (item IDs) in those locations then it's on their end to make the proper changes.

I agree with your changes!

lighta:
tested, working fine.

PR #762, opened 2015-11-17
Changed the regeneration timer to not run, if a BL is dead.

Lemongrass3110:
Also removed an unused parameter at the same time.

### DIFF
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -2101,7 +2101,7 @@ void map_foreachnpc(int (*func)(struct npc_data* nd, va_list args), ...)
 
 /// Applies func to everything in the db.
 /// Stops iterating if func returns -1.
-void map_foreachregen(int (*func)(struct block_list* bl, va_list args), ...)
+void map_foreachregen(int (*func)(struct block_list* bl), ...)
 {
 	DBIterator* iter;
 	struct block_list* bl;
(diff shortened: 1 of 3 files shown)


Lemongrass3110:
summoning @Playtester @cydh @aleos89 @lighta for comments

src/map/map.c
```diff
@@ -2109,14 +2109,12 @@ void map_foreachregen(int (*func)(struct block_list* bl, va_list args), ...)
...
-		va_start(args, func);
-		ret = func(bl, args);
-		va_end(args);
-		if( ret == -1 )
-			break;// stop iterating
+		// Dead BLs don't regenerate
+		// TODO: BLs other than PC should issue a warning/error since they might refer to bad code(not removed bls)
+		if( !status_isdead(bl) ){
+			if( func(bl) == -1 )
+				break;// stop iterating
```
lighta: I'm wondering why we stop iterating ?
I mean ok we shouldn't have error but on this case, half of bl will be healed and other won't since an error happen ?
There place where the iteration should stop but that one look odd imo

Lemongrass3110: I totally agree with you. I just kept it like it was.
We never ever return -1 anyway, so I guess we should really remove it after all. :)

Lemongrass3110:
So to further optimize that, we would have to remove BL_PC and BL_HOM from regen_db on death and add them back in on revival/spawn.
Anyone up to that?

lighta:
Doesn't agree with the new "optimization".
This mostly depend on the DB format.
To be confirmed, but if this is a Red&Black tree like I think it is, adding stuff is a costly operation, (due to node rebalancing), so that wouldn't be a good solution.
All this depend on what db where using obviously. (I'm using db as a src name term here not a real db).

Regen_db, should probably be just be a map, or simply a vector.
Once this is confirmed ye we could do your optimisation. (why only those BL ?, ye perhaps only those could die actually but I don't think the check is necessary no one can regen on death could they ??)

Lemongrass3110:
The other BL types are all completely removed from memory on death it seems. At least my tests showed that.

lighta:
K so I just checked, db_regen is a DB_Map which is a R&B tree, and as you can see if add (put) operation we perform the rebalance, so no I'm not for the new optimisation:
db_obj_put: https://github.com/rathena/rathena/blob/master/src/common/db.c#L1857
If we want a new optimisation please change this as a vector, since the rebalancing is really not needed here. (This only apply if we need to grad fast a specific ID, which we never do so...)

Beside this I'll try to validate your change this weekend and apply the merge.

PR #786, opened 2015-11-29
Updating HP/SP tables for 3rd levels 150-175 

Baalberith6:
Updating HP/SP tables for 3rd levels 150-175 according to Hercules tables and irocalc. We may need to ask someone with access to kro for confirmation.


cydh:
>irocalc

irocalc 'again'?

Baalberith6:
I do believe irocalc vaues are still better, then the extrapolated ones we have on rA. Also checked with a few iro players and they are much closer to what you actually get, then the current rA ones.

Playtester:
I already updated the HP/SP tables now based on data from @RagnarokNova because I was working on fixing the summoner HP/SP and improving the structure.

Sorry completely forgot about this pull request. I guess it can be closed now.

PR #787, opened 2015-11-29
Fixing quest objectives being shown by client incorrectly on monster kill.

Baalberith6:
Fixing quest objectives being shown by client incorrectly on monster kill. Client only shows the first monster received, but we want to show the killed monster.


cydh:
seems u create a new branch with ur hpsp_table_udpdate branch instead master.
the trick, before creating new branch, checkout ur master branch first. ^_^

cydh:
then about the fix, yes must be like that,  ONLY the 'just killed' mob must be shown, not the first mob on quest objective. (afaik, that 'bug' is like that since ages)

Baalberith6:
Oh thanks, still getting confused between perforce at work and git at home:)

Lemongrass3110:
Have you confirmed this behaviour on kRO?
The way you implemented it the packet would look like header, many many zeros, the updated quest and many many zeros again.

Also please consider creating a fresh pull request and please also use the correct account that is connected to @Baalberith6 or attach the other emails to your account. Your recent commits were all with different accounts: nsbuild@bq.corp.netsuite.com and matej.pristak@gmail.com
Thank you!

Baalberith6:
@Lemongrass3110: Let me check my account setting and recreate the PR.
As for the change, my 2013 and 2014 clients are OK with that packet. Maybe I can try to send just the change... let me try it.
I haven't check the behavior with kRO, but even if this was consistent with kRO, it wouldn't make much sense and we could provide at least a preference for sending the correct data.

Lemongrass3110:
It makes sense if the client would just take that packet data and overwrites everything with zeros in its current memory. So please recheck if all your other quests are still OK after one of those update packets. Gravity does some stupid things from time to time(client and/or serverside).

Baalberith6:
I'm using this change for about 2 years on my production server with 50-250 players online and no complaints, I think it is quite safe:)

Lemongrass3110:
Well thats an answer I like to hear ;-) So then please recheck your account settings and recreate the PR from a clean branch. Thank you!

Baalberith6:
@Lemongrass3110: Would you mind deleting my mails from your post? Thanks!

Lemongrass3110:
Deleted! Thanks again!

PR #822, opened 2015-12-09
Avoiding conditional directives that break statements

RomeroMalaquias:


lighta:
Sorry for the delais.
The simplier is the best so having not to retain the MACRO is probably good. (I may not do it for all but here it sound good)

PR #847, opened 2015-12-18
Fixed #772.

### LINKED ISSUE #772: mob_avail option not showing cart
Git Hash: 'a47af467f3241c6e2227c4dcc89570c7644eb712'
Pre-renewal
2013-12-23 client

I tried to create a new mob based from a class and try to put a cart in the option. but all the carts isn't appearing. can anyone confirm this if this is also happen to yours? or is it just me? thanks.

mob_avail.txt
```
3129,4019,1,0,0,0,0,0,0,0,8,387
3130,4019,1,0,0,0,0,0,0,0,128,387
3131,4019,1,0,0,0,0,0,0,0,256,387
3132,4019,1,0,0,0,0,0,0,0,512,387
3133,4019,1,0,0,0,0,0,0,0,1024,387
```

Lemongrass3110:
This is not a real fix but rather a fix which issues a warning for this kind of problem to show clearly that it is working as "intended".


lighta:
@aleos89 say he has a bigger fix for this, but even so the detection of db issue is good.

PR #857, opened 2015-12-29
Feature/client 2015 tests

lighta:
Add fews check to facilitate configurations.
Remove some useless PACKETVER gate in char-serv, (the way is was done was like PACKETVER>2012 mean necessary 0x970 which could be very true but then we don't need to add that case in chclif_parse in 1st place.
Separate them to facilitate flow.

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -1358,101 +1358,116 @@ int char_rename_char_sql(struct char_session_data *sd, uint32 char_id)
 	return 0;
 }
 
-int char_check_char_name(char * name, char * esc_name)
+/**
+* Function that check if a given name is valide within config
+* currently more oriented to for newchar creation
+* @param inout name : Char name to check
+return e_makechar_error
+*/
+e_makechar_error char_check_char_name(char * name)
 {
-	int i;
+	normalize_name(name, TRIM_CHARS);
 
 	// check length of character name
 	if( name[0] == '\0' )
-		return -2; // empty character name
+		return ERROR_NAME_LEN; // empty character name
+
+	 // remove special characthere before continuing check
+	if (remove_control_chars(name))
+		return ERROR_NAME_INVALID_CHAR; // control chars in name
+
 	/**
 	 * The client does not allow you to create names with less than 4 characters, however,
 	 * the use of WPE can bypass this, and this fixes the exploit.
 	 **/
 	if( strlen( name ) < 4 )
-		return -2;
-	// check content of character name
-	if( remove_control_chars(name) )
-		return -2; // control chars in name
+		return ERROR_NAME_LEN;
 
 	// check for reserved names
 	if( strcmpi(name, charserv_config.wisp_server_name) == 0 )
-		return -1; // nick reserved for internal server messages
+		return ERROR_NAME_RESERVED; // nick reserved for internal server messages
 
 	// Check Authorised letters/symbols in the name of the character
+
 	if( charserv_config.char_config.char_name_option == 1 )
 	{ // only letters/symbols in char_name_letters are authorised
+		int i;
 		for( i = 0; i < NAME_LENGTH && name[i]; i++ )
 			if( strchr(charserv_config.char_config.char_name_letters, name[i]) == NULL )
-				return -2;
+				return ERROR_NAME_INVALID_CHAR;
 	}
 	else if( charserv_config.char_config.char_name_option == 2 )
 	{ // letters/symbols in char_name_letters are forbidden
+		int i;
 		for( i = 0; i < NAME_LENGTH && name[i]; i++ )
 			if( strchr(charserv_config.char_config.char_name_letters, name[i]) != NULL )
-				return -2;
+				return ERROR_NAME_INVALID_CHAR;
 	}
-	if( charserv_config.char_config.name_ignoring_case ) {
-		if( SQL_ERROR == Sql_Query(sql_handle, "SELECT 1 FROM `%s` WHERE BINARY `name` = '%s' LIMIT 1", schema_config.char_db, esc_name) ) {
-			Sql_ShowDebug(sql_handle);
-			return -2;
-		}
-	} else {
-		if( SQL_ERROR == Sql_Query(sql_handle, "SELECT 1 FROM `%s` WHERE `name` = '%s' LIMIT 1", schema_config.char_db, esc_name) ) {
-			Sql_ShowDebug(sql_handle);
-			return -2;
-		}
-	}
-	if( Sql_NumRows(sql_handle) > 0 )
-		return -1; // name already exists
 
-	return 0;
+	{ //check duplicate
+		char esc_name[NAME_LENGTH * 2 + 1];
+		Sql_EscapeStringLen(sql_handle, esc_name, name, strnlen(name, NAME_LENGTH));
+		if (charserv_config.char_config.name_ignoring_case) {
+			if (SQL_ERROR == Sql_Query(sql_handle, "SELECT 1 FROM `%s` WHERE BINARY `name` = '%s' LIMIT 1", schema_config.char_db, esc_name)) {
+				Sql_ShowDebug(sql_handle);
+				return ERROR_MK_SQL;
+			}
+		}
+		else {
+			if (SQL_ERROR == Sql_Query(sql_handle, "SELECT 1 FROM `%s` WHERE `name` = '%s' LIMIT 1", schema_config.char_db, esc_name)) {
+				Sql_ShowDebug(sql_handle);
+				return ERROR_MK_SQL;
+			}
+		}
+		if (Sql_NumRows(sql_handle) > 0)
+			return ERROR_NAME_DUPLICATE; // name already exists
+	}
+
+	return ERROR_NO_ERROR;
 }
 
-//-----------------------------------
-// Function to create a new character
-//-----------------------------------
-int char_make_new_char_sql_67(struct char_session_data* sd, char* name_, int str, int agi, int vit, int int_, int dex, int luk, int slot, int hair_color, int hair_style) {
+/**
+* Function to create a new character. (and store it into SQL)
+* @param sd: char_session_data of the client
+* @param name_: name requested
+* @param str: str requested
+* @param agi: agi requested
+* @param vit: vit requested
+* @param int_: int_ requested
+* @param dex: dex requested
+* @param luk: luk requested
+* @param slot: characthere slot requested
+* @param hair_color: hair color requested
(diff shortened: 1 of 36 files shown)


src/char/char.h
```diff
@@ -29,6 +29,19 @@ enum {
...
+	ERROR_INVALID_SLOT,
+	ERROR_MAXLIMIT,
+	ERROR_NAME_RESERVED,
+	ERROR_NAME_LEN,
+	ERROR_NAME_INVALID_CHAR,
+	ERROR_NAME_DUPLICATE,
+	ERROR_CREATION_DISABLE,
+
+	ERROR_NO_ERROR = 0,
+} e_makechar_error;
```
cydh: any better prefix than `ERROR`? just because `ERROR` is too common. replace it with `MAKECHAR` maybe?

lighta: I'll replace it with MAKECHAR as prefix, does anyone have other input on this : @Lemongrass3110, @aleos89 ?

src/char/char.c
```diff
@@ -1358,106 +1358,116 @@ int char_rename_char_sql(struct char_session_data *sd, uint32 char_id)
 	return 0;
 }
 
-int char_check_char_name(char * name, char * esc_name)
+/**
+* Function that check if a given name is valide within config
+* currently more oriented to for newchar creation
+* @param inout name : Char name to check
+return e_makechar_error
+*/
```
aleos89: Function that checks if the given name is within the valid parameters.
'@param' name : Character name to check
'@return' e_makechar_error

-- Obviously don't include the quote, I just had to put them so we don't tag users. :)

src/char/char.h
aleos89: I agree with @cydh, using `ERROR` is too common of a name for this. What Cydh suggested as the new name is fine with me.

src/char/char.h
```diff
@@ -29,6 +29,19 @@ enum {
 	TABLE_GUILD_STORAGE,
 };
 
+typedef enum e_makechar_error {
+	ERROR_MK_SQL = -255,
```
aleos89: `MAKECHAR_MAKE_SQL`. Is "MK" supposed to be short for "MAKE"? Best to type it all out.

src/char/char.h
```diff
@@ -29,6 +29,19 @@ enum {
...
+	ERROR_INVALID_SLOT,
+	ERROR_MAXLIMIT,
```
aleos89: `MAKECHAR_MAX_LIMIT`. Should include the underscore since it's between the rest of the other values.

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
 	return 1;
 }
 
+enum e_charclif_error {							//tested with 0x6e in 20150914
+	ERROR_SAMENAME_EXIST = 0x0,					//Character name already exist
```
aleos89: `MAKECHAR_NAME_EXISTS`

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
...
+	ERROR_LIMIT_AGE = 0x1,						//You are underaged
+	//ERROR_LIMIT_CHAR_DELETE = 0x2,			
+	ERROR_INVALID_SYMBOLS = 0x2,				//Symbols in Character Names are forbidden
+	ERROR_MAKECHAR_INVALID_SLOT = 0x3,			//You are not elegible to open the Character Slot
```
aleos89: `MAKECHAR_INVALID_SLOT` -- Obviously, just drop it off since you are going to rename it to `MAKECHAR` :)

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
...
+	ERROR_NEED_PREMIUM_SERVICE = 0xb,			//This service is only available for premium users
+	ERROR_NONPERMISSION_NAME = 0xc,				//Failed to move char slot
```
aleos89: If the player failed to move the character to another slot why is the name `ERROR_NONPERMISSION_NAME`? That doesn't seem like it fits this use case. Something like `MAKECHAR_SLOT_MOVE_FAIL` or something along those lines.

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
...
+	//ERROR_DELETE_LIMIT_LEVEL = 0x64,			//
+
+	ERROR_MAKECHAR_DENIED = 0xFF					//actually default message
+};
+/**
+* Acknoledge the client that the car creation failed
+* @param fd: client file descriptor
+* @param err: error that occur to inform client
+*/
```
aleos89: Inform the client that the character creation failed.
`@param` fd: client file descriptor
`@param err`: error that occurred

-- Can we make this a void function since all we do is return 1?

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
...
+int chclif_createnewchar_ack_error(int fd, e_makechar_error err) {
```
aleos89: Can we make this a void function since all we do is return 1?

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
...
+
+	return 1;
+}
+
+/**
+* Acknoledge the client that the car creation was a success
+* @param fd: client file descriptor
+* @param sd: client char_session_data
+* @param sd: new char_id created
+*/
```
aleos89: Inform the client that the character creation succeeded.
`@param` fd: client file descriptor
`@param` sd: client char_session_data
`@param` char_id: new character ID created

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
...
+int chclif_createnewchar_ack_success(int fd, struct char_session_data* sd, uint32 char_id){
```
aleos89: Can we make this a void function since all we do is return 1?

src/char/char_clif.c
```diff
@@ -885,61 +888,143 @@ int chclif_parse_charselect(int fd, struct char_session_data* sd,uint32 ipl){
...
+
 // S 0970 <name>.24B <slot>.B <hair color>.W <hair style>.W
 // S 0067 <name>.24B <str>.B <agi>.B <vit>.B <int>.B <dex>.B <luk>.B <slot>.B <hair color>.W <hair style>.W
-int chclif_parse_createnewchar(int fd, struct char_session_data* sd,int cmd){
-	int i = 0;
-
-	if (cmd == 0x970) FIFOSD_CHECK(31) //>=20120307
-	else if (cmd == 0x67) FIFOSD_CHECK(37)
-	else return 0;
+int chclif_parse_createnewchar(int fd, struct char_session_data* sd) {
```
aleos89: This function will always return 1. Are there no more fail cases?

src/common/mmo.h
```diff
@@ -21,6 +21,7 @@
 // see conf/battle/client.conf for other version
 
 #ifndef PACKETVER
+	//#define PACKETVER 20150916
```
aleos89: Maybe not add this since we are not officially supporting this client.

src/map/clif.c
```diff
@@ -9851,21 +9851,37 @@ static int clif_guess_PacketVer(int fd, int get_previous, int *error)
 		err = 0;\
 		if( error )\
 			*error = 0;\
+		if( diskey ) ShowWarning("Client use no encryption while server do, please fix me\n"); \
```
aleos89: Client is not sending encrypted packets but the server is setup to accept encrypted packets. Please disable PACKET_OBFUSCATION.

src/char/char.c
lighta: I think parameters is to generic. I was refering as config = char_name_option in char_athena.conf
I should also probably specify the other check :
min name len, special character, reserved name and check duplicate

src/char/char.h
lighta: Yes but I don't like make, as it's more make as DO stuff, so I would name it MAKECHAR_DO_SQL

src/char/char_clif.c
lighta: That enum for clif are extracted from aegis. I prefer kinda keep their name there, just to have a kinda sync.

src/char/char_clif.c
lighta: same as below, from aegis, that why enum actually. those defined by them kinda static used by client, and the makechar_one which is more internal, more specific to our implementation

src/map/clif.c
lighta: No, well it's misleading. What I want to inform here is that 1 client is in mistmatch with the server.
To fix it you can either use a client without obfuscation
or recompile without PACKET_OBFUSCATION.
Now with the later the client that currently well setup will have an issue.
I originally tought about changing this at runtime before realising that I shouldn't. What I want to achieve here is to help detect the mistmatch not to tell him which way to go.

Lemongrass3110: I think we should not add this after all. It helps us to see if we used a wrong client during development, but for a production server that one is a no-go. If a server-owner wants players to use packet obfuscation and players try to connect with a custom diffed client to their server, the server gets flooded by those messages. So either we do it like hercules, that we add an option for supporting obfuscated and plain connections at the same time(which must be possible to be disabled by a setting) or just leave it out in general.

Lemongrass3110:
Again I would recommend to split this up, since you mixed a lot of different things in here and none of them really have anything to do with "Feature/client 2015 tests".

src/char/char_clif.c
aleos89: Ah, ok. I wasn't aware it was from Aegis.

src/map/clif.c
lighta: OK, is only 1 message per connection attempt but you're right.
1 question tough, should I keep it with a define set as off by default or completly remove it ?
(I'd prefer the define option if you ask me)

src/char/char_clif.c
lighta: yes, but I'll review all oh them in same time

lighta:
Gonna do separate PR

PR #864, opened 2016-01-02
Error correction in the pre-renewal table.

admkakaroto:
Small mistake, missing a small supplement in pre-renewal table.

http://i.imgur.com/JPHFVNw.png


Lemongrass3110:
Thanks, but I could not auto merge it, because I cannot see any change for GD_EMERGENCYCALL

PR #881, opened 2016-01-06
Cleanup/const.txt

Lemongrass3110:
Any further input @aleos89 @cydh @lighta?
From my point of view it is good enough to be merged into master now.


cydh:
just maybe, after the merge, someone need to look into items & scripts for constants usages, atleast after this 90dcd845e2ff4dc3f536c8811183494bb9750a81.
or maybe summon @Atemo to do it here?

Lemongrass3110:
I would rather do it after the merge, since we already behind master regarding item updates in here. This way we don't produce conflicts and it's easier to merge this. :)
But yeah we should definitely use these constants whereever possible!

aleos89:
Things still in const that can be merged over:
* PET_* is used for *getpetinfo()
* MOB_* is used for *getmonsterinfo()
* QTYPE_* is used for *questinfo()
* FW_* is used for *announce() or other broadcasting types
* VAR_* is used for *getlook -- script_commands doesn't have it listed as valid values.
* DIR_* is used for *pushpc() -- Differs from source values but only this command uses these values.
* MOBG_* is used in item_db for *monster()
* False and True

Lemongrass3110:
@aleos89 pushed all of the remaining ones you stated, except the ones for the mob groups and pushpc. We should do an overhaul later for that command, so that it uses values we maybe could use for the NPC directions and other source side ones too.
And for the mob groups. Source side there are some of them missing, should also be looked into before working with it here.

aleos89:
@Lemongrass3110 Yeah that's perfectly fine. I know those two need a bit more work for them so what you've done is great!

PR #885, opened 2016-01-08
Mob DB Update

secretdataz:
Added card drops for
- Lighthalzen Biolab F4
- Hall of Abyss
- Malaya Port
- Eclage
- Old Glast Heim


secretdataz:
@nanakiwurtz Any updates?

PR #888, opened 2016-01-08
Add master id check in setunitdata script command. Fixes #887

### LINKED ISSUE #887: Map server crash when using script command setunitdata
Git Hash :  315056b

When you use setunitdata <mobid>,UMOB_SLAVECPYMSTRMD,1; script command on a monster while it doesn't have any master, the map server will crash with following exception.
```
Unhandled exception at 0x007B9DD9 in map-server.exe: 0xC0000005: Access violation reading location 0x0000025C.
```

Relevant code : map/script.c Line 17179

secretdataz:


src/map/script.c
```diff
@@ -17172,7 +17172,17 @@ BUILDIN_FUNC(setunitdata)
 			case UMOB_INT: md->status.int_ = (unsigned short)value; status_calc_misc(bl, &md->status, md->level); break;
 			case UMOB_DEX: md->status.dex = (unsigned short)value; status_calc_misc(bl, &md->status, md->level); break;
 			case UMOB_LUK: md->status.luk = (unsigned short)value; status_calc_misc(bl, &md->status, md->level); break;
-			case UMOB_SLAVECPYMSTRMD: md->state.copy_master_mode = value > 0 ? 1 : 0; if (value > 0) { TBL_MOB *md2 = map_id2md(md->master_id); md->status.mode = md2->status.mode; } break;
+			case UMOB_SLAVECPYMSTRMD:
+				md->state.copy_master_mode = value > 0 ? 1 : 0;
```
aleos89: I think this line should be moved below. There's no reason we should set the value for this if it will fail later on if the monster has no master.

src/map/script.c
```diff
@@ -17172,7 +17172,17 @@ BUILDIN_FUNC(setunitdata)
...
+				if (value > 0) {
+					if (!md->master_id) {
+						ShowWarning("buildin_setunitdata: Trying to set UMOB_SLAVECPYMSTRMD on mob without master!\n");
+						break;
+					}
+					TBL_MOB *md2 = map_id2md(md->master_id);
+					md->status.mode = md2->status.mode;
```
aleos89:
```
md->state.copy_master_mode = 1;
```

src/map/script.c
```diff
@@ -17172,7 +17172,17 @@ BUILDIN_FUNC(setunitdata)
...
+				}
```
aleos89:
```
} else
    md->state.copy_master_mode = 0;
```

aleos89:
Thank you for the patch!

secretdataz:
Edited the patch according to your suggestion.

PR #889, opened 2016-01-08
Randomized Start Point

### LINKED ISSUE #805: Server Startpoint
Hi, in official server, re and pre-re. The start point is shuffled, reference http://herc.ws/board/tracker/issue-7750-official-server-startpoint/

aleos89:
* Fixes #805
* Character's will follow somewhat of the official method of a 'random' start point at their first login.
* Official servers use load balancing for setting the start point dynamically according to user count, whereas we at rAthena wanted to stay simple since we don't believe you guys will need this kind of load balancing for new characters.

### DIFF
--- a/src/char/char.c
+++ b/src/char/char.c
@@ -2670,6 +2670,8 @@ static void char_config_split_startpoint(char *w2_value)
 #endif
 
 	fields = (char **)aMalloc(fields_length * sizeof(char *));
+	if (fields == NULL)
+		return; // Failed to allocate memory.
 	lineitem = strtok(w2_value, ":");
 
 	while (lineitem != NULL) {
@@ -2715,6 +2717,8 @@ static void char_config_split_startitem(char *w2_value)
 	strcat(config_name, "start_items");
 
 	fields = (char **)aMalloc(fields_length * sizeof(char *));
+	if (fields == NULL)
+		return; // Failed to allocate memory.
 	lineitem = strtok(w2_value, ":");
 
 	while (lineitem != NULL) {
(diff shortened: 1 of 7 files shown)


src/char/char.c
```diff
@@ -2747,17 +2748,34 @@ bool char_config_read(const char* cfgName, bool normal){
...
-			if (sscanf(w2, "%15[^,],%6hd,%6hd", map, &x, &y) < 3){
-				ShowWarning( "Specified start_point has an invalid format.\n" );
-				continue;
+			int i = 0, fields_length = 3 + 1;
+			char *lineitem, **fields;
+
+			fields = (char**)aMalloc(fields_length * sizeof(char*));
+			lineitem = strtok(w2, ":");
+
+			while (lineitem != NULL) {
```
lighta: a sub_function for parsing this would be better for clarity. we don't want to clobber char_config_read, with parsing code

src/char/char.h
```diff
@@ -142,7 +144,8 @@ struct CharServ_Config {
 	int log_inter;	// loggin inter or not [devil]
 	int char_check_db;	///cheking sql-table at begining ?
 
-	struct point start_point; // Initial position the player will spawn on server
+	struct point start_point[MAX_STARTPOINT]; // Initial position the player will spawn on server
```
lighta: try ti change this for a dynamic array if possible. You have start_point_count to mark the limit.

aleos89: I'd like to do this in a later commit. It would be nice to refactor some things into vectors.

PR #952, opened 2016-02-08
Use sv_readdb to read CashShop Database

CairoLee:
Prevent the use of incorrect data in the CashShop database caused crash map-server.

### DIFF
--- a/src/map/cashshop.c
+++ b/src/map/cashshop.c
@@ -77,7 +77,6 @@ static void cashshop_read_db_txt( void ){
 		char* dbsubpath1 = (char*)aMalloc(n1+1);
 		char* dbsubpath2 = (char*)aMalloc(n2+1);
 
-
 		if(fi==0) {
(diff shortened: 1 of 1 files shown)


src/map/cashshop.c
```diff
@@ -162,7 +121,7 @@ static int cashshop_read_db_sql( void ){
 				}
 			}
 
-			if( !cashshop_parse_dbrow( str, cash_db_name[fi], lines ) )
+			if( !cashshop_parse_dbrow( str, 3, lines ) )
```
cydh: well, then add the info on fail here, tells the SQL table name, since in previous function it parses file or sql name.
something like
````
if( !cashshop_parse_dbrow( str, 3, lines ) ) {
    ShowError("cashshop_read_db_sql: Cannot process table '%s' at line '%d'.\n", what_is_sql_table_var, lines);
    continue;
}
````

CairoLee:
Hi @cydh 
Could you help me review this again? Thank you XD

PR #978, opened 2016-02-23
Removed the displayed name in from of the message displayed by npctal…

Atemo:
…k and unittalk scripts commands (so we can append whatever we want to).

Signed-off-by: Atemo <capucrath@gmail.com>


aleos89:
This works out fine. I think I like what you did more rather than adding an optional flag to show the name since the name has to be written out anyways.

cydh:
hmmmmmmmm
so, technically `npctalk` and `unittalk` are same (both need the name manually as an input)
I just think, need depreciate the `npctalk` then /hmm

Atemo:
Yes we could depreciate npctalk but I don't like unittalk because there is no default unit id (uggly..)

PR #981, opened 2016-02-24
Woe:TE

cydh:
WOE TE (Training Edition) Implementation

Todo:
* [x] Monsters for Godly Quest
* [x] Monsters for Treasure Boxes
* [x] Monster for Missions (the mob is known as 'walking' plant-type, affected by status changes, always deals 1 damage, can attack)
* [x] Monsters for Guild dungeons (optional, since not as 1 part for TE, mostly for Siege Coin quests)
* [x] `NoCostume` mapflag from #1137 (can be merge/update later on master as follow up)
* [x] Recheck NPC Talk scripts (as follow up of #978)
* [x] Recheck mobs spawn coordinates in god item script


cydh:
@atemo, yes you can remove the script function, I just "ah I'll do it later".

and as u can see, the script name maybe a bit off, just suggest a better name for it.

db/re/skill_nocast_db.txt
```diff
@@ -99,10 +100,32 @@
...
+491,16	// CR_CULTIVATION
+530,16	// NJ_KIRIKAGE
+691,16	// CASH_ASSUMPTIO
+//!TODO: More 3rd Class skills
+2284,16	// SC_FATALMENACE
+2300,16	// SC_DIMENSIONDOOR
+2293,16	// SC_GROOMY
+2296,16	// SC_UNLUCKY
+2494,16	// GN_CHANGEMATERIAL
+2462,16	// SO_EL_ANALYSIS
```
aleos89: Is there a reason we specifically block 3rd class skills on WoE:TE maps?

aleos89:
Overall it's looking good! Just a few more things, and I'm assuming @Atemo's other PR (#978) will tick off those `unittalk` parts in the To-Do list.

I'm still testing the stuff in-game but I'll throw anything here if I run into problems.

cydh:
dying, waiting for the experts convert mob db @aleos89

db/re/item_noequip.txt
```diff
@@ -172,11 +172,8 @@
...
 // Misc
 605,16		// Anodyne
-2365,16		// Orleans_Gown
-4077,16		// Phen_Card
 4128,16		// Golden_Bug_Card
 4198,16		// Maya_Puple_Card
-4327,16		// Blood_Butterfly_Card
 11503,16	// Siege_White_Potion
 11504,16	// Siege_Blue_Potion
 11547,16	// Woe_Violet_Potion
```
cydh: the fact, we still need it.
we did wrong implementation of `bNoCastCancel`. technically just like other blocked items, the script will no-run on restricted map.
So if it's Phen Card, the whole effect of Phen (casttime and nocancel) won't works, but currently, without noequip flag, the casttime penalty still applied.

db/status_disabled.txt
```diff
@@ -29,14 +29,281 @@
...
+SC_GENSOU,16
+SC_AKAITSUKI,16
+
+// Others
+SC_FEAR,16
+SC_BURNING,16
+SC_FREEZING,16
+SC_TEARGAS_SOB,16
+SC__FEINTBOMB,16
+SC__CHAOS,16
```
cydh: Cool

cydh:
Maybe, we need to change the way `getpermissionwoete` works?
Maybe like `item_noequip` flag, we create `job_restricted_maps.txt` file, with stucture
```
//<Job ID>,<Zone>
JOB_GENETIC,4112 // 16: GVG TE, 4096: Zone 8 for teg_dun01 and teg_dun02
```
Then change the `getpermissionwoete` script becomes `is_job_allowed("<mapname>"{,<JobID>});`

poke again @Atemo @Lemongrass3110 @aleos89

aleos89:
That definitely makes it more dynamic (better for other cases, possibly in the future?) than the current implementation.

cydh:
Here
```
// Defines Job(s) that resitriced to enter map (by flag/zones)
//
// Structure of Database:
// JobID,FlagZone,GroupLevelBypass
//
// JobID: See JOB_* constants or use job number
//
// Legend for 'Flag' field (bitmask):
// 1    - restricted in normal maps
// 2    - restricted in PVP
// 4    - restricted in GVG
// 8    - restricted in Battlegrounds
// 16   - restricted in WOE:TE castles
// Restricted zones - configured by 'restricted <number>' mapflag
// 32   - restricted in zone 1
// 64   - restricted in zone 2
// 128  - restricted in zone 3
// 256  - restricted in zone 4
// 512  - restricted in zone 5
// 1024 - restricted in zone 6
// 2048 - restricted in zone 7
// 4096 - restricted in zone 8
//
// GroupLevelBypass: Group Level (groups.conf) to ignore the restriction
//
// NOTES:
// - Restriction will be overwrite for multiple lines with same Job ID.
// - The flag is used by 'jobcanentermap' script.
```

and the script
```
*jobcanentermap("<mapname>"{,<JobID>});

Return 1 if player (decided by job) can enter the map, 0 otherwise.

For optional 'JobID', see constant of Job_*, or use player's Class, BaseJob,
and BaseClass can be used. If no player attached, this param must has a value.

See also db/[pre-]re/job_noenter_map.txt
```

If you guys think it's good, I'll push the changes.

aleos89:
Looks good! Might be best to include this as a separate commit like the `status_disabled` database.

Atemo:
I didn't understand at first. It's a good idea push it

cydh:
@atemo can u update makes Thief Poring & Suspicious Baby won't be removed when Emperium is broken?

cydh:
Job restriction is #1526

src/map/script.c
```diff
@@ -11949,6 +11953,17 @@ BUILDIN_FUNC(setmapflag)
...
+			case MF_GVG_TE_CASTLE:		map[m].flag.gvg_te_castle = 1; break;
+			case MF_GVG_TE:
+				{
+					struct block_list bl;
+					map[m].flag.gvg_te = 1;
+					clif_map_property_mapall(m, MAPPROPERTY_AGITZONE);
+					bl.type = BL_NUL;
+					bl.m = m;
+					clif_maptypeproperty2(&bl,ALL_SAMEMAP);
+				}
```
Lemongrass3110: A little dirty hack I think... Maybe we should resend the mapflags in general, if we do not do this already.

cydh: I don't get what u mean

npc/re/warps/guildcastles.txt
```diff
@@ -78,16 +78,17 @@ OnTouch:
 	end;
 
 S_Warp:
-	if (getpermissionwoete() == 0) {
+	.@castle$ = getarg(0);
+	if (jobcanentermap(.@castle$) == 0) {
 		mes "3rd and expansion job level can't enter in WoE TE.";
```
Lemongrass3110: Is that one an official dialog? Otherwise it would be good to adjust it.

cydh: from the script

cydh:
I will add these
https://gist.github.com/cydh/61b5fc1bd5e262deeb20a4e83aca1761
https://gist.github.com/cydh/2e4688f48bec9626560fe9931f36b64b
https://gist.github.com/cydh/854689476ebd7d19aa8f50919f9a39f7

admkakaroto:
if use Urgent Call, crash map_server.

cydh:
@admkakaroto that's all u can tell?

admkakaroto:
Sorry!


I can not explain it, only when someone uses urgent call inside the castle the server crash.

img: http://imgur.com/2npABtN

PR #1001, opened 2016-03-03
Fix chat related issue on 2015-10-29 clients. Fixes #1000.

### LINKED ISSUE #1000: rAthena adds null to already null-terminated chat
the bug causes this code in `clif_process_message` to trigger and prevent the chat message from displaying.
```
	// the declared length must match real length
	if( messagelen != strnlen(message, messagelen)+1 ) {
		ShowWarning("clif_process_message: Received malformed packet from player '%s' (length is incorrect)!\n", sd->status.name);
		return false;
	}
```

secretdataz:
Some special chat messages e.g. Frost Joke are already null-terminated. We don't have to add more null byte.

Fix last letter in party chat get cut.

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -9803,7 +9803,10 @@ static bool clif_process_message(struct map_session_data* sd, int format, char**
 	}
 
 #if PACKETVER >= 20151001
-	message[messagelen++] = '\0';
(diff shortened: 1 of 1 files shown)


secretdataz:
Removed #1003 fix because it's fixed by @Lemongrass3110.

Issue #1014, opened 2016-03-05
EXP overflow


Lemongrass3110:
If you set your rate to something super high and the base or job exp become something in the range of UINT32_MAX and it gets added to the current base exp there will be an overflow.

This will mainly be a problem for HR or SHR server since others usually do not get as much experience.

The affected additions are in the function pc.c:pc_gainexp
```
	// Give EXP for Base Level
	if (base_exp) {
		if(sd->status.base_exp > nextb - base_exp)
			sd->status.base_exp = nextb;
		else
			sd->status.base_exp += base_exp;
		if (!pc_checkbaselevelup(sd))
			clif_updatestatus(sd,SP_BASEEXP);
	}

	// Give EXP for Job Level
	if (job_exp) {
		if(sd->status.job_exp > nextj - job_exp)
			sd->status.job_exp = nextj;
		else
			sd->status.job_exp += job_exp;
		if (!pc_checkjoblevelup(sd))
			clif_updatestatus(sd,SP_JOBEXP);
	}
```

Git hash I confirmed this on is 2b7eb9f

Lemongrass3110:
@MishimaHaruna this one is also on HerculesWS/Hercules@ee60ae3

jezznar:
I've been looking at this one earlier because of Issue #1013 forgot to post it here. 

I discovered it while testing the issue and killed an MVP but only got  the MVP exp and not the base/job exp that it gives.

MishimaHaruna:
Thanks @Lemongrass3110, I'll check it asap. I glanced at it and saw the word `unsigned` appear too many times in that function. Scary.

PR #1015, opened 2016-03-05
Fix for issue #1014

jezznar:
Fix for issue #1014
by handling exp calculations greater than the maximum limit of uint32 through iteration


Lemongrass3110:
As you stated yourself this one had a syntax error.

PR #1016, opened 2016-03-05
Handle EXP overflow through iterative deductions Fix for issue #1014

jezznar:
#1014
Just a syntax error correction from the previous commit which lacked a closing parentheses on the first while statement that I added

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -6497,24 +6497,29 @@ int pc_gainexp(struct map_session_data *sd, struct block_list *src, unsigned int
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
Thank you for submitting a pull request, but I fixed the issue on b70b259 by just using uint64 instead of any iterative deductions. From my point of view this makes it easier to read and understand.

I'm looking forward to seeing you contributing to rAthena in the future again though. :+1:

jezznar:
It's actually great,  I didn't know typecasting would work for this. thanks!

PR #1043, opened 2016-03-11
Fix to NPC [Larissa#mos_01] in script [npc/quests/quests_moscovia.txt]

AtlantisRO:
Fix to NPC [Larissa#mos_01] in script [npc/quests/quests_moscovia.txt] from [Help Mikhail] Quest.

delitem script uses ".@di" array instead of ".@n" array.


cydh:
Thank you. :D

AtlantisRO:
Thank you too. We have a few fixes we have applied to npc scripts to solve bugs and problems like this one, and we want to share it with emulator community to give back and help with the project. We were testing how to make pull request in this pull.

About the optimization, you mean check all the ingredientes right before doing the delitem/getitem part?

PR #1054, opened 2016-03-13
Fixes to some quests and typo errors.

AtlantisRO:
- Fix to NPC [Iromo's Mother#ep3_2] to fix a bug that restarts the quest.

- A minor fix to some typo error.

- Pre-RE: Fixed the position of NPC [Promotional Staff#iz] from izlude before renewal.

### DIFF
--- a/npc/cities/izlude.txt
+++ b/npc/cities/izlude.txt
@@ -186,10 +186,10 @@
 		mes "Well, of course, they are all very cute, but Ghostring is an EXCEPTION. It is very very dangerous.";
 		next;
 		mes "[Dega]";
-		mes "If you are lucky enough, you might even bump into Angelring, the Poring with Angel wings.";
+		mes "If you are lucky enough, you might even bump into Angeling, the Poring with Angel wings.";
 		next;
 		while(1) {
-			switch(select("Ghostring?:Angelring?:End Conversation.")) {
+			switch(select("Ghostring?:Angeling?:End Conversation.")) {
 			case 1:
 				mes "[Dega]";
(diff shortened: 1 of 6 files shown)


Lemongrass3110:
- [x] c5cd672 is good to merge
- [ ] 3cd6774 needs confirmation from our leaks
- [ ] 438e33a looks good, but we need to check how the official script handles this, since it might only be a conversion problem.
- [ ] 9c99c5c sounds reasonable and should also be recheck with our aegis sources.

Lemongrass3110:
438e33a is good to merge, this bug comes from strict conversion of the aegis script. It also has this error.

Lemongrass3110:
9c99c5c confirmed with our aegis sources.

AtlantisRO:
The position of the npc [Promotional Staff#iz] was changed in commit [a1c1238](https://github.com/rathena/rathena/commit/a1c12384fa63078e12c43e89ffdb181c517ec30d) to fit with the redisign of izlude map.

After that were differenced the localization from re and pre-re version, but without changing the position back.

However, the correct facing is 5 (according to the previous position of rAthena's script), not 6 as in the commit. Sorry about that.

Lemongrass3110:
Thanks for looking that up! Would you mind to create another small follow up so that this can be merged together with it? It's your work after all. :-)

AtlantisRO:
I added a small commit f3fa99f to change the position from 6 to 5. Would you explain how to do a follow up?

aleos89:
@AtlantisRO You did exactly what he meant. :+1:

AtlantisRO:
Thank you both :)

Lemongrass3110:
Thanks @AtlantisRO :-)

PR #1089, opened 2016-03-22
Fixing @at venders being stuck when map->char connection gets reset.

Baalberith6:
Reproduction steps:
1) Have at least 1 `@autotrade` vender in vendings/vending_items table
2) start the server, make sure the `@at` merchant is loaded
3) kill the connection between char-map server and wait for map to reconnect
 (Note that log says [22/04:11:20][Status]: Done loading '1' vending autotraders with '1' items. even though the trader was already online)
4) wait 60 seconds

Map server log: 
[22/04:12:39][Info]: auth_db: Node (state Login) timed out for 2013551:183567

I believe venders shouldn't be re-processed OnInterInit, just once (OnInterInitOnce)

### DIFF
--- a/src/map/chrif.c
+++ b/src/map/chrif.c
@@ -30,6 +30,7 @@ static int check_connect_char_server(int tid, unsigned int tick, int id, intptr_
 
 static struct eri *auth_db_ers; //For reutilizing player login structures.
 static DBMap* auth_db; // int id -> struct auth_node*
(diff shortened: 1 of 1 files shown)


CairoLee:
i can confirm this problem, and this patch is work!
Thank you @Baalberith6 .

Badarosk0:
Confirmed this problem too.

PR #1094, opened 2016-03-24
Add getweaponlv() script command.

secretdataz:
This command returns weapon level of the item that called it.
This command can be useful on cards that requires its attached weapon to be certain level.
e.g.
```
4696,True_Alphoccio_Card,True Alphoccio Card,6,20,,10,,,,,,,,2,,,,,{ .@dmg = 20; if(getrefine() > 9) .@dmg += 20; if(getweaponlv() == 4) .@dmg += 20; bonus2 bSkillAtk,"WM_REVERBERATION",.@dmg;  },{},{}
```


cydh:
this is good, just I think for something better

src/map/script.c
```diff
@@ -21172,6 +21172,21 @@ BUILDIN_FUNC(adopt)
...
 
+/*=======================================================
+* Returns weapon level of item calling this command.
+* Returns 0 when caller isn't weapon or error.
+* This command accepts no argument. [secretdataz]
+*-------------------------------------------------------*/
+BUILDIN_FUNC(getweaponlv)
+{
+	TBL_PC *sd;
+	if ((sd = script_rid2sd(st)) != NULL && sd->inventory_data[current_equip_item_index])
```
cydh: must be value check before using `current_equip_item_index`,
`current_equip_item_index > -1 && current_equip_item_index < EQI_MAX`

secretdataz: @cydh It'll just return 0 when caller isn't weapon.

Atemo:
hmm.. don't we get the same result with getequipweaponlv ?[https://github.com/rathena/rathena/blob/master/src/map/script.c#L8397](url)

secretdataz:
`getequipweaponlv` accepts equipment position as argument. In some case like my example script `getequipweaponlv(EQI_HAND_R);` can be abused by using a level 4 weapon in the right hand and a Main Gauche[4] with 4 cards because we don't know which hand is the card in.

PR #1096, opened 2016-03-24
Allow Doram with New Basic Skill to use Kafra storage.

Kolkka:
Added check in functions_kafras.txt


Kolkka:
Looks like there is another more complete pull request.

PR #1101, opened 2016-03-25
Cleaned up Inventory, Cart Inventory, and Storage implementation

### LINKED ISSUE #441: Cart Items need be saved before vending
11251288_820368931372402_74167565396148733_n 1

aleos89:
* These storage types now mimic Guild Storage and can be loaded/saved whenever needed.
* Relieves mmo_charstatus from having to send storage types and lets the char-server handle it.
* All storage types now have a max of 1637 items each.
* Fixes #441 - Players will no longer be required to log out to resync cart item data before opening a Vending Store.
* Added script commands guildstoragecountitem[2] and guildstoragedelitem[2].

Thanks to @lighta and @cydh for their help with it!


cydh:
> All storage types now have a max of **1637** items each.

That number maybe can be lower now. :P
iirc, when we started to make it, the `unique_id` wasn't `uint64`

src/map/vending.c
```diff
@@ -315,28 +317,16 @@ int8 vending_openvending(struct map_session_data* sd, const char* message, const
...
-		// Player just moved item to cart and we don't have the correct cart ID yet.
-		if (sd->status.cart[sd->vending[i].index].id == 0) {
-			struct item_data *idb = itemdb_search(sd->status.cart[index].nameid);
-			char msg[256];
-
-			sprintf(msg, msg_txt(sd, 733), idb->jname);
-			clif_displaymessage(sd->fd, msg);
-			clif_skill_fail(sd, MC_VENDING, USESKILL_FAIL_LEVEL, 0);
-			return 4;
-		}
```
cydh: hmm, maybe u forgot that we talked about this (or maybe I overlook something on inter for cart items x mysql)
even cart/inventory/storage items are saved from map-server to char/inter-server, the `id` is never be sent back by char/inter-server to map-server.
because `id` is `AUTO_INCREMENT` by MySQL table and we cannot get the recent items were added to attach the `id`.

imho, instead make this message is set as default, (again) I prefer to tell everyone that "this is the **price** you have to 'pay' if u want to use autotrade persistance.
```c++
if (battle_config.feature_autotrade && sd->status.cart[sd->vending[i].index].id == 0) {
//...
//...
}
```

aleos89: Hm, true. I could have sworn it was resolved with this. I might look at it again then. I have a follow up to resolve some other things.

aleos89:
Sorry about this. Closed this one because of the horrible merging process.

PR #1113, opened 2016-03-29
Abount Runes and NPC

nehpetskie:
... Mam Sir Can you please?fix the rune issue on Rune Knight The Reusable and The rune must Cannot be Dispell Like "RAIDO" cuz on official you cannot dispell the rune specially that skill .. and my other issue about the Old glast hiem Count of the victim to proceed to next map ...

thank you and .. please notice this for more balance in 3rd job system


aleos89:
Can you please create an [Issue](https://github.com/rathena/rathena/issues). This is a Pull Request and is completely unrelated to what you are asking about.

PR #1120, opened 2016-03-30
Make arguments in getequipweaponlv optional

secretdataz:


doc/script_commands.txt
```diff
@@ -2642,12 +2642,15 @@ this is +10:
 
 ---------------------------------------
 
-*getequipweaponlv(<equipment slot>{,<char_id>})
+*getequipweaponlv({<equipment slot>,<char_id>})
```
Atemo: It should be
```
getequipweaponlv({<equipment slot>{,<char_id>}})
```

src/map/script.c
```diff
@@ -8392,14 +8392,25 @@ BUILDIN_FUNC(getequiprefinerycnt)
...
+ * getequipweaponlv({<equipment slot>,<char_id>})
  *------------------------------------------*/
 BUILDIN_FUNC(getequipweaponlv)
 {
 	int i = -1,num;
 	TBL_PC *sd;
 
-	num = script_getnum(st,2);
+	if (!script_hasdata(st, 2)) {
+		if ((sd = script_rid2sd(st)) != NULL && current_equip_item_index < MAX_INVENTORY && sd->inventory_data[current_equip_item_index])
```
aleos89: If we set `sd = st` here we nullify the optional `char_id` parameter.

doc/script_commands.txt
```diff
@@ -2642,14 +2642,14 @@ this is +10:
...
+*getequipweaponlv(<equipment slot>{,<char_id>})
 
 This function returns the weapon level for the weapon equipped in the specified 
 equipment slot on the invoking character. For a list of equipment slots see 
 'getequipid'.
 
-If no arguments are provided. Weapon level for the item calling this function
-assuming it is called by an item script will be returned. Otherwise, 0 will be returned.
+If -1 is passed as <equipment slot> argument. Weapon level for the item calling this function, 
+assuming it is called by an item script, will be returned. Otherwise, 0 will be returned.
```
aleos89: If -1 is passed as the equipment slot then use the weapon level for the item calling
this function, assuming it is called by an item script, will be returned. Otherwise 0
will be returned.

aleos89: Would be nice to later adjust the -1 to a constant value, but this sort of falls into the `EQP_` and `EQI_` issue.

Issue #1119, opened 2016-03-30
Watery Evasion X Aqua Benedicta, Water Ball


cydh:
As I experienced in idRO, Priestess only can cast Aqua Benedicta (AL_HOLYWATER) once in same cell of Ninja's Watery Evasion (NJ_SUITON) area. The 2nd attempts (and so) will be failed. Except the Ninja recast the Watery Evasion.

[As iRO Wiki notes](http://irowiki.org/wiki/Watery_Evasion#Notes)
> This skill acts as shallow water for Water Ball and Aqua Benedicta skills, though any cells used for these skills will be removed after casting.

Maybe different on kRO?

Playtester:
Just tested and can confirm. The cell is completely used up just like water ball.

Playtester:
Further testing:
1. When using Aqua Benedicta on LP on water, then the skill will fail at castbegin. It won't even start casting. Waterball is the same except its search area is larger. No error message. This must happen before any units are deployed.
2. Using Aqua Benedicta on Water Evasion will only remove the Watery Evasion cell at castend. This happens even when standing on water.
3. Aqua Benedicta does NOT remove Deluge. You can cast it unlimited while standing on a Deluge cell without it being used up.
4. Aqua Benedicta does NOT fail at castend when standing on LP (Waterball does!). This is the proof that it's not a unit skill like Waterball even though it checks for LP at castbegin.

cydh:
Well, it's fail on LP area. I didn't test it yet, I just realized that I don't have enough Mage classes on idRO.

Playtester:
Yeah but you are correct it's not a unit skill, because it does NOT fail if you place LP during cast of Aqua Benedicta. So at castend, it doesn't need to place any unit to be successful.

The castbegin check for water simply includes checking for LP.

PR #1125, opened 2016-04-01
AL_HOLYWATER: Set skill unit and Delete unit when overlapped with NJ_SUITON/SA_DELUGE

jezznar:
Fix for Issue #1119
- AL_HOLYWATER set skill unit
- Detects overlap with NJ_SUITON/SA_DELUGE
- Deletes unit when overlapped

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -7332,7 +7332,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case AL_HOLYWATER:
 		if(sd) {
 			if (skill_produce_mix(sd, skill_id, ITEMID_HOLY_WATER, 0, 0, 0, 1, -1))
+			{
 				clif_skill_nodamage(src,bl,skill_id,skill_lv,1);
+				skill_unitsetting(src, skill_id, skill_lv, src->x, src->y, 0); 	//Fix for Issue #1119
+			}
 			else
 				clif_skill_fail(sd,skill_id,USESKILL_FAIL_LEVEL,0);
 		}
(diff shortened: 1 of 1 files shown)


jezznar:
First Skill Use:
screenrathena040
Second Skill Use: Suiton Cell consumed from first use, skill fails.
screenrathena041

cydh:
well, it can be work like that. but it doesn't

jezznar:
@cydh what do you mean? can you please clarify? :)

Playtester:
I see your approach, but you are doing some assumptions here which are:
1. Aqua Benedictio also removes Deluge cells.
2. Aqua Benedictio is a unit skill.

I think both should be tested first.

Also, assuming it's a unit skill, I'd prefer we implement it properly with an entry in skill_unit_db and a unit duration in skill_db.

cydh:
It's not as unit. It just simply "if this Watery Evasion is occupied by Aqua Benedicta (or Water Ball), its cell effect is removed".
"Removed" means just simply lost its effect, cannot be used for another Aqua Benedicta or Water Ball, also the debuff effect is removed.
Just
````
if (casted(AL_HOLYWATER) || casted(WL_WATERBALL)) {
   if (cell_is_active()) {
      do_cast_the_skill();
      cell_deactive();
   }
   else {
       cast_failed();
   }
}
````

I worked on that issue already /ene

Playtester:
Well to check if it should be a unit skill, we should just test if you can use it on LP when LP is on water. If you can still use it on LP as long as you are standing on water it's not a unit skill. If you can't cast it on LP, then it should be implemented as a unit skill, similar to waterball.

Playtester:
Hi, I added more tests on issue #1119 and I'm afraid Cydh is correct. Even though it checks for LP at castbegin, it does not fail when LP is there on castend. This is proof that, unlike Waterball, it's not a unit skill and shouldn't call unitsetting. Instead it should simply remove the Watery Evasion cell.

Also, it only removes Watery Evasion but not Deluge.

jezznar:
Ahhh I see. Alright then. I guess I lacked that part of the process. Thanks y'all

cydh:
That's why it's not as unit skill, my openkore will tells me if it's Area spell.
So, @jezznar, I'll close this PR first.

jezznar:
ok. thanks.

PR #1170, opened 2016-04-13
Updated ZLIB to 1.2.8

Lemongrass3110:
Upgraded from version 1.2.7 to the latest version 1.2.8

Source:
http://www.zlib.net/


wilkemeyer:
As the distributed ZLib is used for Windows builds only, i'd suggest removing the precompiled dll && distribute the source of the zlib && alter the build-process so athena gets statically linked against zlib.

There won't be a license Problem either as zlib's license is BSD like, so the produced athena-binarys won't violate anything.

**Pros:**
 - no binary bloat @ git repo
 - Support for 32 && 64 Bit Builds 
 - Proper debugging support too, when build debug.

**Cons:**
 - Probably 5 seconds longer build-times :)

Lemongrass3110:
Hi sirius black!

Nice to have you around again. I was thinking of that possibility as well.
It would be nice if you could join us on discord so we can discuss this issue.
Here's an invite link(please do not wonder if you are not able to read or post anything at the start, we have to put you into a usergroup first):  https://discord.gg/0flQ268FQEeB0H4X

Thanks for your help!

PR #1175, opened 2016-04-16
Correct docs and script follow up the refactor of EQI

Atemo:
Don't merge yet please.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -2489,27 +2489,27 @@ This function returns the item ID of the item equipped in the equipment slot
 specified on the invoking character. If nothing is equipped there, it returns -1. 
 Valid equipment slots are:
 
-EQI_HEAD_TOP (1)          - Upper Headgear
-EQI_ARMOR (2)             - Armor (jackets, robes)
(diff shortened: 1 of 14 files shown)


Lemongrass3110:
Hmm seeing where this stuff is used makes me think that we maybe should create a function which returns the name of the place of the equipment.
That way we would only have the EQI constants stored in the scripts an the names would also be dynamically generated.

Also I think we should remove the values of the constants completely from the documentation, so that noone uses them in the future(and creates issues with hardcoded values like we have now). What do you think @Atemo and @aleos89?

Atemo:
>Hmm seeing where this stuff is used makes me think that we maybe should create a function which returns the name of the place of the equipment.
That way we would only have the EQI constants stored in the scripts an the names would also be dynamically generated.

Hmm yes we could make it with script function and add it in global function file.

>Also I think we should remove the values of the constants completely from the documentation, so that noone uses them in the future(and creates issues with hardcoded values like we have now). What do you think @Atemo and @aleos89?

I'm against. I want to encourage the user to use constant (maybe with a note in the doc) but I want to leave the value if need (for example you need the values of eqi if you want to make a loop)

PR #1183, opened 2016-04-17
Aqua Benedicta/Water Ball official behavior. Fixes #1119

### LINKED ISSUE #1119: Watery Evasion X Aqua Benedicta, Water Ball

secretdataz:
This PR corrects Aqua Benedicta and Water Ball's behavior.
- Both skills will fail if casted on Land Protector cell.
- Aqua Benedicta will consume a Watery Evasion cell on castend.  
Thanks to @Playtester for advices

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -15367,12 +15367,6 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 				return false;
 			}
 			break;
-		case AL_HOLYWATER:
-			if (map_getcell(sd->bl.m, sd->bl.x, sd->bl.y, CELL_CHKLANDPROTECTOR)) {
-				clif_skill_fail(sd,skill_id,USESKILL_FAIL_LEVEL,0);
-				return false; // Aqua Benedicta will not cast on LP [secretdataz]
-			}
-			break;
 	}
 
 	/* check state required */
@@ -15425,7 +15419,7 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 		case ST_WATER:
 			if (sc && (sc->data[SC_DELUGE] || sc->data[SC_SUITON]))
 				break;
-			if (map_getcell(sd->bl.m,sd->bl.x,sd->bl.y,CELL_CHKWATER))
+			if (map_getcell(sd->bl.m,sd->bl.x,sd->bl.y,CELL_CHKWATER) && !map_getcell(sd->bl.m,sd->bl.x,sd->bl.y,CELL_CHKLANDPROTECTOR))
 				break;
 			clif_skill_fail(sd,skill_id,USESKILL_FAIL_LEVEL,0);
 			return false;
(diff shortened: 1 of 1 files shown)


src/map/skill.c
```diff
@@ -7347,8 +7347,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 
 	case AL_HOLYWATER:
 		if(sd) {
-			if (skill_produce_mix(sd, skill_id, ITEMID_HOLY_WATER, 0, 0, 0, 1, -1))
-				clif_skill_nodamage(src,bl,skill_id,skill_lv,1);
+			if (skill_produce_mix(sd, skill_id, ITEMID_HOLY_WATER, 0, 0, 0, 1, -1)) {
+				struct skill_unit* su;
+				if ((su = map_find_skill_unit_oncell(bl, bl->x, bl->y, NJ_SUITON, NULL, 0)) != NULL)
+					skill_delunit(su);
```
cydh: del unit will remove the unit entirely, which it's wrong. I was mentioned only the effect that will be nulled, the 'cell display effect' is still remain.

secretdataz: @cydh hmm. how about adding a flag in val1 of the units on castend, marking it as "used"? Then we can add a check in ST_WATER check for that flag.

Playtester: Actually that's not true, it removes both the effect (only on that one cell) as well as its effect.

cydh: fck off @Playtester. U need add ur knowledge.
first I said, Aqua Benedicta wasn't skill unit, u didn't believe. fkc off, I told the fact, the truth how it works.
then now, u don't believe again? whatever, u can test with stucked brain.

@secretdataz yes just need flag, I was mentioned this on previous PR (or in issue I made)
yep just use unit flag, flagged used if Aqua Benedicta was casted there, the 'cell display effect' is remain. The unit won't be deleted, I confirm it my self. Thanks to Openkore :D

Playtester: I know that the effect disappears because I tested that.

Lemongrass3110: @cydh instead of swearing around you could show @Playtester the confirmation you gathered through Openkore or whatever tool you used. If you do not add any proof it is statement against statement.
As you know for sure aegis implementations differ from server to server from time to time, so it might be good to gather all the facts from both sides before swearing around in public.

As for me personally if it really is statement against statement you might as well consider, that @Playtester is an official team member and therefore his statement might be consider more worthy than the statement of an "inofficial" team member who only uses his commit permissions when he feels like without talking to anyone and additionally only insults and offends people from the team or others.

To sum it up, if I were you I would really reconsider who is the one who has a stuck brain and who has the right to tell someone to fuck off. It would also be a pleasure for all of us to see your manner in public getting a little more respectful. I do not care how you talk about us in private or directly to us, but writing comments like these in public while being in your position is absolutely not okay.

Best regards,
Lemon

PR #1199, opened 2016-04-23
small cleanup && optimization @ skill.c / skilldb_id2idx lookup

wilkemeyer:
As mentioned in the commit message:
Before this change skilldb_id2idx was a DBMap (int-db)
which is quite costly for a static lookup table which has an 16-bit int as index.

Also should be noted that the lookup via id2idx is done quite often during runtime .. so this change makes sense.

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -57,7 +57,7 @@
 	#error GD_SKILLRANGEMAX is greater than 999
 #endif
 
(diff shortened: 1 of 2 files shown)


wilkemeyer:
As the code was'nt compatible with C90/C89... i'll create a new PR which will be.

PR #1208, opened 2016-04-25
Refactored/Cleaned Visual-Studio Solutions

wilkemeyer:
Removed:
 old Visual Studio 2008, 2010, 2012, 2013, 2015 Projects.

Added: 
new & clean Visual Studio 2013 Solution

The new Soltution is a way more cleaner than the old were.
Each Component has its own Project.
The Core (common) is also a stand-alone project (static-lib) = no more redundant source-refs in all projects.

AMD64/x64 Support Prepared,
if this PR gets accepted i'll add missing dependencies for x64-builds.

Why 2013 & not 2015?
Because 2013 was the first edition that became fully available for free(limited) as Community Edition.
i assume 2013 is still more common than 2015 (even if i personally prefer 2015 over 2013 :)), so the Solution & all Project's are created with/for 2013, the Platform Toolset of all Project's is also 2013 by default.

When working with Visual Studio 2015 - just open The Solution & just don't upgrade it - you'll be fine.
If you're running VS2015 without 2013 Installed, you'll need to change the platform Toolset in all projects to their 2015 counterpart.

I'd suggest keeping just one VS Version supported (at least as solution/project-files)
as the previous system was a bit messy in terms of code-maintenance:
When adding/removing a new Source-File you had to add it @ 6 Build Systems:

VS 2008, 2010, 2012, 2013, 2015,  autoconf && cmake.

Now just VS2013 && autoconf/cmake.

Just a Personal note/opinion:
i'd suggest supporting just one VS Version at all, maybe we should also switch to VS2015, it's also freely available as Community Edition & works on any NT 6 Machine.
Supporting several Major Versions of a compiler for one platform just increases the complexity of providing support for the compiled-release.

Also i think if someone want's to Provide a Ragnarok Server, we can require him/her to install a most-recent compiler & IDE which is freely available.

If someone just does'nt want to install Visual-Studio fully he/she can even use the most recent Windows Platform SDK && compile it via cmd/msbuild.

For further discussions: I'm also available @ IRC/Discord (Sirius_Black)


cydh:
then,
I think u have to make/update the compiling guides for Windows for this point
> If someone just does'nt want to install Visual-Studio fully he/she can even use the most recent Windows Platform SDK && compile it via cmd/msbuild.

Just a Personal note/opinion:
> Also i think if someone want's to Provide a Ragnarok Server, we can require him/her to install a most-recent compiler & IDE which is freely available.

this point just so much hurting for people who has 'old' IDE that cannot be freely downloaded because of Internet connection issue :D

wilkemeyer:
@cydh about guides - no problem :D

& About Bandwidth /  slow internet connections:
In the current state (Using Select on Windows) athena won't even run fine on windows if there're really users would playing on it.

So i assume the most affected user's are developers right now :D

As i wrote before - at some point a project needs to make a cut in terms of supporting old Operating Systems, as well as in terms of which IDE/Compiler is supported.

If someone really needs to use VS 2008 or even an older version - it's up to him/her to maintain the older Solution, The Project can be forked at any time.


I already wrote to lemongrass that i'll setup a mirror for pre-compiled releases of rAthena (probably today .. or tomorrow ~ )

secretdataz:
Merged in C++ refactor branch.

secretdataz:
For further code addition and discussion please refer to #2089.

PR #1232, opened 2016-05-02
fix return values in pc_setpos 

wilkemeyer:
Use correct enum constants instead of int's for returning from pc_setpos

### DIFF
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -5277,7 +5277,7 @@ enum e_setpos pc_setpos(struct map_session_data* sd, unsigned short mapindex, in
 		sd->state.pmap = sd->bl.m;
 		if (sd->sc.count) { // Cancel some map related stuff.
 			if (sd->sc.data[SC_JAILED])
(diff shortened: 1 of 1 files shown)


src/map/pc.c
```diff
@@ -5277,7 +5277,7 @@ enum e_setpos pc_setpos(struct map_session_data* sd, unsigned short mapindex, in
 		sd->state.pmap = sd->bl.m;
 		if (sd->sc.count) { // Cancel some map related stuff.
 			if (sd->sc.data[SC_JAILED])
-				return 1; //You may not get out!
+				return SETPOS_MAPINDEX; //You may not get out!
```
Lemongrass3110: Maybe this one should also get it's own return value, what do you think @aleos89?

aleos89: If we specifically check somewhere for when a player is jailed and return a unique jailed message, then yes. If not we shouldn't really need any unique value for it.

Issue #1155, opened 2016-04-09
Arrow equip


Atemo:
Hash 2305a6610103b1b475017bef516b53268cdd825e
On official renewal you can't equip arrow without bow equipped (so on rArthena you can. Assassin class can't equip them ; skill fail for ranger if no bow equipped).
Also on official if you unequip the bow the arrow is automatically unequipped

cydh:
Depends on which official server u refers to :P

But maybe, this behavior was added in the MT when Assassin cannot equip an arrow any longer.

Atemo:
lol you're right, I refer to iro we need to test on kro

RagnarokNova:
Confirmed on kRO. Unequipping a bow removes your arrows. It's really uncomfortable.

secretdataz:
This should be made optional with official behavior as default imo.

Playtester:
If we made an option for that please make also one for ammo of gunslingers. There the official behavior is that you can't even equip weapons when the ammo doesn't match, very annoying. I'd rather have that auto-unequip that arrows have lol.

MrAntares:
What happens on bow switch? Will it de-equip the arrows?

Atemo:
No it doesn't unequip the arrow when you switch the bow.

stealth62:
this will break pvp mechanics when you switch bow and arrow automatically unequip i remember iro revert this changes a lot of player complain switching bow will unequip arrow.

Playtester:
Even more reason to make it optional.

MrAntares:
And default off ;)

Playtester:
Default is always kRO. =3

joelolopez:
i cant equip new type of guns coz the bullets is still equipped, such a hassle

Playtester:
Yeah that's official behavior though. I'd like to see it configurable. Should just automatically unequip the bullets and arrows when no suitable weapon equipped.

joelolopez:
oh i see , lets keep it like that ahahha i love official behaviors , always keep messing ur game ahaha xo

MrAntares:
Much work for a feature that will rarely be used <3

benching:
On assasin , baby and parent jobs,
is venom knife(1771) can be equipped even without weapon?

PR #1248, opened 2016-05-06
Fixed Issue #1155

benching:
applied config for arrow/ammo unequip on weapon unequip.

### DIFF
--- a/conf/battle/battle.conf
+++ b/conf/battle/battle.conf
@@ -139,7 +139,14 @@ delay_battle_damage: yes
 // 2 = Yes even for skills that do not specify arrow consumption when said 
 //     skill is weapon-based and used with ranged weapons (auto-guesses which 
 //     skills should consume ammo when it's acquired via a card or plagiarize)
-arrow_decrement: 1
+arrow_decrement: 0
+
+// Unequip arrows/ammo when the weapon is unequipped
+// 0 = No
+// 1 = Unequip arrow only
+// 2 = Unequip ammo/bullets only
+// 3 = Unequip arrow and ammo
+weapon_ammo_unequip: 3
 
 // Should the item script bonus 'Autospell' check for range/obstacles before casting?
 // Official behavior is "no", setting this to "yes" will make skills use their defined
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -7979,6 +7979,7 @@ static const struct _battle_data {
 	{ "natural_heal_skill_interval",        &battle_config.natural_heal_skill_interval,     10000,  NATURAL_HEAL_INTERVAL, INT_MAX, },
 	{ "natural_heal_weight_rate",           &battle_config.natural_heal_weight_rate,        50,     50,     101             },
 	{ "arrow_decrement",                    &battle_config.arrow_decrement,                 1,      0,      2,              },
+	{ "weapon_ammo_unequip",				&battle_config.weapon_ammo_unequip,				3,		0,		3,				},
 	{ "max_aspd",                           &battle_config.max_aspd,                        190,    100,    199,            },
 	{ "max_third_aspd",                     &battle_config.max_third_aspd,                  193,    100,    199,            },
 	{ "max_walk_speed",                     &battle_config.max_walk_speed,                  300,    100,    100*DEFAULT_WALK_SPEED, },
@@ -8519,6 +8520,9 @@ void battle_adjust_conf()
 	battle_config.max_extended_aspd = 2000 - battle_config.max_extended_aspd * 10;
 	battle_config.max_walk_speed = 100 * DEFAULT_WALK_SPEED / battle_config.max_walk_speed;
 	battle_config.max_cart_weight *= 10;
+
+	if (battle_config.weapon_ammo_unequip < 0 || battle_config.weapon_ammo_unequip > 3)
+		battle_config.weapon_ammo_unequip = 3;
 
 	if (battle_config.max_def > 100 && !battle_config.weapon_defense_type) // added by [Skotlex]
 		battle_config.max_def = 100;
--- a/src/map/pc.c
+++ b/src/map/pc.c
@@ -9456,6 +9456,18 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos)
 		return false;
 	}
 
+	// dont equip arrow if no bow is equipped
+	if ((sd->class_&MAPID_BASEMASK) == MAPID_ARCHER || (sd->class_&MAPID_BASEMASK) == MAPID_ARCHER_HIGH || (sd->class_&MAPID_BASEMASK) == MAPID_SNIPER || (sd->class_&MAPID_BASEMASK) == MAPID_HUNTER || (sd->class_&MAPID_BASEMASK) == MAPID_RANGER){
+		if (id->type == IT_AMMO){
+			int w_idx = sd->equip_index[EQI_HAND_R];
+			enum weapon_type w_type = (w_idx != -1) ? (enum weapon_type)sd->inventory_data[w_idx]->look : W_FIST;
+			if (w_idx == -1 || id->look == W_BOW){
+				clif_equipitemack(sd, 0, 0, ITEM_EQUIP_ACK_FAIL);
+				return false;
+			}
+		}
+	}
+
 	if ((sd->class_&MAPID_BASEMASK) == MAPID_GUNSLINGER) {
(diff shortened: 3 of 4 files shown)


src/map/battle.c
```diff
@@ -8520,6 +8521,9 @@ void battle_adjust_conf()
 	battle_config.max_walk_speed = 100 * DEFAULT_WALK_SPEED / battle_config.max_walk_speed;
 	battle_config.max_cart_weight *= 10;
 
+	if (battle_config.weapon_ammo_unequip < 0 || battle_config.weapon_ammo_unequip > 3)
+		battle_config.weapon_ammo_unequip = 3;
```
aleos89: This check isn't needed since you have the define just above in `_battle_data`.

conf/battle/battle.conf
```diff
@@ -139,7 +139,14 @@ delay_battle_damage: yes
...
 //     skill is weapon-based and used with ranged weapons (auto-guesses which 
 //     skills should consume ammo when it's acquired via a card or plagiarize)
-arrow_decrement: 1
+arrow_decrement: 0
+
+// Unequip arrows/ammo when the weapon is unequipped
+// 0 = No
+// 1 = Unequip arrow only
+// 2 = Unequip ammo/bullets only
+// 3 = Unequip arrow and ammo
```
aleos89: Maybe put in parenthesis here that this option is the official setting.

src/map/pc.c
```diff
@@ -9456,6 +9456,18 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos)
 		return false;
 	}
 
+	// dont equip arrow if no bow is equipped
+	if ((sd->class_&MAPID_BASEMASK) == MAPID_ARCHER || (sd->class_&MAPID_BASEMASK) == MAPID_ARCHER_HIGH || (sd->class_&MAPID_BASEMASK) == MAPID_SNIPER || (sd->class_&MAPID_BASEMASK) == MAPID_HUNTER || (sd->class_&MAPID_BASEMASK) == MAPID_RANGER){
```
aleos89: This only applies to the Hunter class? What about the Stalker or Dancer/Bard classes?

conf/battle/battle.conf
```diff
@@ -139,7 +139,14 @@ delay_battle_damage: yes
 // 2 = Yes even for skills that do not specify arrow consumption when said 
 //     skill is weapon-based and used with ranged weapons (auto-guesses which 
 //     skills should consume ammo when it's acquired via a card or plagiarize)
-arrow_decrement: 1
+arrow_decrement: 0
+
+// Unequip arrows/ammo when the weapon is unequipped
+// 0 = No
```
aleos89: Instead of "No", state that it doesn't remove arrow and ammo.

src/map/pc.c
```diff
@@ -9678,44 +9690,68 @@ bool pc_unequipitem(struct map_session_data *sd, int n, int flag) {
...
 
-	if(sd->status.inventory[n].equip & EQP_HAND_R) {
+	if (sd->status.inventory[n].equip & EQP_HAND_R) {
 		sd->weapontype1 = 0;
 		sd->status.weapon = sd->weapontype2;
 		pc_calcweapontype(sd);
-		clif_changelook(&sd->bl,LOOK_WEAPON,sd->status.weapon);
-		if( !battle_config.dancing_weaponswitch_fix )
+		clif_changelook(&sd->bl, LOOK_WEAPON, sd->status.weapon);
+		if (!battle_config.dancing_weaponswitch_fix)
```
aleos89: This comment applies to the above changes. It would be nice to keep style like this out of this since it's not related to this pull request.

conf/battle/battle.conf
```diff
@@ -139,7 +139,14 @@ delay_battle_damage: yes
 // 2 = Yes even for skills that do not specify arrow consumption when said 
 //     skill is weapon-based and used with ranged weapons (auto-guesses which 
 //     skills should consume ammo when it's acquired via a card or plagiarize)
-arrow_decrement: 1
+arrow_decrement: 0
```
cydh: why 0?

benching: oh god, i forgot about that

src/map/pc.c
```diff
@@ -9456,12 +9456,12 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos)
...
 
-	// dont equip arrow if no bow is equipped
-	if ((sd->class_&MAPID_BASEMASK) == MAPID_ARCHER || (sd->class_&MAPID_BASEMASK) == MAPID_ARCHER_HIGH || (sd->class_&MAPID_BASEMASK) == MAPID_SNIPER || (sd->class_&MAPID_BASEMASK) == MAPID_HUNTER || (sd->class_&MAPID_BASEMASK) == MAPID_RANGER){
+	// dont equip arrow if weapon equipped is not bow / guitar / whip.
+	if (sd->status.weapon != W_BOW && sd->status.weapon != W_MUSICAL && sd->status.weapon != W_WHIP){
 		if (id->type == IT_AMMO){
 			int w_idx = sd->equip_index[EQI_HAND_R];
 			enum weapon_type w_type = (w_idx != -1) ? (enum weapon_type)sd->inventory_data[w_idx]->look : W_FIST;
-			if (w_idx == -1 || id->look == W_BOW){
+			if (w_idx == -1 || id->look != W_BOW && id->look != W_MUSICAL && id->look != W_WHIP){
```
aleos89: You will need some parenthesis around your &&. If not you will be getting incorrect results.

```
if (w_idx || (id->look && id->look && id->look))
```

src/map/pc.c
```diff
@@ -9678,44 +9693,68 @@ bool pc_unequipitem(struct map_session_data *sd, int n, int flag) {
...
 		sd->status.weapon = sd->weapontype2;
 		pc_calcweapontype(sd);
-		clif_changelook(&sd->bl,LOOK_WEAPON,sd->status.weapon);
-		if( !battle_config.dancing_weaponswitch_fix )
+		clif_changelook(&sd->bl, LOOK_WEAPON, sd->status.weapon);
+		if (!battle_config.dancing_weaponswitch_fix)
 			status_change_end(&sd->bl, SC_DANCING, INVALID_TIMER); // Unequipping => stop dancing.
+
+		// [benching] 
+		switch (battle_config.weapon_ammo_unequip){
```
lighta: the switch here is not really extendable with a bitmask.
I mean the code is correct don't get me wrong. 
But I want to add an option to that weapon_ammo_unequip, I may need to add 3 more case to fill all the combinaison, and so on...
Imo I'd just have writting this with 2 independant if.

src/map/pc.c
```diff
@@ -9456,6 +9456,21 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos)
...
+	// dont equip arrow if weapon equipped is not bow / guitar / whip.
+	// allow assassin, baby and parent jobs to equip ammuntions without weapons (need official info)
+	if ((sd->status.weapon != W_BOW && sd->status.weapon != W_MUSICAL && sd->status.weapon != W_WHIP)
+		&& (sd->class_&MAPID_BASEMASK != MAPID_ASSASSIN || sd->class_&MAPID_BASEMASK != MAPID_ASSASSIN_CROSS || sd->class_&MAPID_BASEMASK != MAPID_BABY_ASSASSIN
+		|| sd->class_&MAPID_BASEMASK != MAPID_GUILLOTINE_CROSS || sd->class_&MAPID_BASEMASK != MAPID_GUILLOTINE_CROSS_T || sd->class_&MAPID_BASEMASK != MAPID_BABY_CROSS)){
+		if (id->type == IT_AMMO){
+			int w_idx = sd->equip_index[EQI_HAND_R];
+			enum weapon_type w_type = (w_idx != -1) ? (enum weapon_type)sd->inventory_data[w_idx]->look : W_FIST;
+			if (w_idx == -1 || (id->look != W_BOW && id->look != W_MUSICAL && id->look != W_WHIP)){
+				clif_equipitemack(sd, 0, 0, ITEM_EQUIP_ACK_FAIL);
```
lighta: all good. altough I'm now wondering why pc_equipitem, have some direct call on clif.
I mean it's working but I feel that clif call would be an issue to use that fonction on a script or some other internal.
It's not at all in the scope of that change, but still I'm just droping this here. ;)

ecdarreola:
Can this be updated?

PR #1255, opened 2016-05-08
Classic_Hat_J and Cowboy_Hat_J names swapped

Daegaladh:
Fix for Classic_Hat_J and Cowboy_Hat_J names swapped


zackdreaver:
Why not include Purple Cowboy Hat item script:
ATK +15
Flee -5

PR #1264, opened 2016-05-14
Expanded the instance system to support new modes 

aleos89:
New modes include: No player attached, single player, and guild.
Expanded script command instance_create to take in an owner ID and an optional mode.
Added a new script command instance_check_guild (works the same as instance_check_party).
Adjusted official instance scripts to support new format.
Adjusted the instance_db to allow instance idle time to be adjustable.
Refactored all instance_id to unsigned short.

### DIFF
--- a/db/import-tmpl/instance_db.txt
+++ b/db/import-tmpl/instance_db.txt
@@ -1,5 +1,5 @@
 // Instance Database
 //
 // Structure of Database:
-// ID,Name,LimitTime,EnterMap,EnterX,EnterY,Map1,Map2,Map3,Map4,Map5,Map6,Map7,Map8
+// ID,Name,LimitTime,IdleTimeout,EnterMap,EnterX,EnterY,Map1,Map2,Map3,Map4,Map5,Map6,Map7,Map8
 
--- a/db/import-tmpl/instance_db.txt
+++ b/db/import-tmpl/instance_db.txt
@@ -1,5 +1,6 @@
 // Instance Database
 //
 // Structure of Database:
-// ID,Name,LimitTime,IdleTimeout,EnterMap,EnterX,EnterY,Map1,Map2,Map3,Map4,Map5,Map6,Map7,Map8
-
+// ID,Name,LimitTime,IdleTimeOut,EnterMap,EnterX,EnterY,Map2,Map3,Map4,Map5,Map6,Map7,Map8,Map9,Map10
+//
+// EnterMap is considered as Map1
--- a/db/import-tmpl/instance_db.txt
+++ b/db/import-tmpl/instance_db.txt
@@ -1,6 +1,6 @@
 // Instance Database
 //
 // Structure of Database:
-// ID,Name,LimitTime,IdleTimeOut,EnterMap,EnterX,EnterY,Map2,Map3,Map4,Map5,Map6,Map7,Map8,Map9,Map10
+// ID,Name,LimitTime,IdleTimeOut,EnterMap,EnterX,EnterY,Map2,Map3,...,Map255
 //
 // EnterMap is considered as Map1
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3,7 +3,7 @@
 //===== By: ==================================================
 //= rAthena Dev Team
 //===== Last Updated: ========================================
-//= 20150610
+//= 20160523
 //===== Description: =========================================
 //= A reference manual for the rAthena scripting language.
 //= Commands are sorted depending on their functionality.
@@ -8113,9 +8113,9 @@ This command will open a book item at the specified page.
 ========================
 ---------------------------------------
 
-*instance_create("<instance name>",<owner_id>{,<mode>});
+*instance_create("<instance name>"{,<instance mode>{,<owner id>}});
 
-Creates an instance for the <owner_id> of <mode>. The instance name, along with
+Creates an instance for the <owner id> of <mode>. The instance name, along with
 all other instance data, is read from 'db/(pre-)re/instance_db.txt'. Upon success,
 the command generates a unique instance ID, duplicates all listed maps and NPCs,
 sets the alive time, and triggers the "OnInstanceInit" label in all NPCs inside
@@ -8129,7 +8129,7 @@ Instance Mode options:
 
 The command returns the instance ID upon success, and these values upon failure:
  -1: Invalid type.
- -2: Party not found.
+ -2: Character/Party/Guild not found.
  -3: Instance already exists.
  -4: No free instances (MAX_INSTANCE exceeded).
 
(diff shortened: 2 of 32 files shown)


anacondaqq:
This is so good! I'm going to try *_*

doc/script_commands.txt
```diff
@@ -8131,8 +8139,9 @@ The command returns the instance ID upon success, and these values upon failure:
 
 Destroys instance with the ID <instance id>. If no ID is specified, the instance
 the script is attached to is used. If the script is not attached to an instance,
-the instance of the currently attached player's party is used. If that fails,
-the script will come to a halt.
+the instance of the currently attached player is used (if it is a party or guild
```
Lemongrass3110: not for char mode?

src/map/instance.c
```diff
@@ -478,76 +653,93 @@ int instance_destroy(short instance_id)
 /*==========================================
  * Allows a user to enter an instance
  *------------------------------------------*/
-int instance_enter(struct map_session_data *sd, const char *name)
+int instance_enter(struct map_session_data *sd, unsigned short instance_id, const char *name)
 {
 	struct instance_db *db = instance_searchname_db(name);
```
Lemongrass3110: If we already have the instance_id, what do we need to do a name lookup for?

src/map/instance.c
```diff
@@ -478,76 +653,93 @@ int instance_destroy(short instance_id)
...
 
 /*==========================================
  * Warp a user into instance
  *------------------------------------------*/
-int instance_enter_position(struct map_session_data *sd, const char *name, short x, short y)
+int instance_enter_position(struct map_session_data *sd, unsigned short instance_id, const char *name, short x, short y)
 {
-	struct instance_data *im;
+	struct instance_data *im = &instance_data[instance_id];
 	struct instance_db *db = instance_searchname_db(name);
```
Lemongrass3110: Same as for instance_enter.

src/map/clif.c
```diff
@@ -16572,80 +16572,99 @@ void clif_font(struct map_session_data *sd)
...
 
-	if(!sd) return; //party_getavailablesd can return NULL
+	instance_getsd(instance_id, &sd, &target);
+
+	if (!sd)
+		return;
 
 	WBUFW(buf,0) = 0x2cd;
-	memcpy( WBUFP(buf,2), name, 62 );
+	memcpy(WBUFP(buf,2), instance_data[instance_id].name, 62);
```
Lemongrass3110: Please change this to safestrncpy and also use INSTANCE_NAME_LENGTH for the length to copy.
See the documentation on the packet you added. 61B instance name.
It was wrong before too, but it got overwritten by the next instruction, so noone noticed.

src/map/script.c
```diff
@@ -18868,15 +18884,15 @@ BUILDIN_FUNC(instance_create)
...
 
 	if( script_hasdata(st,2) )
 		instance_id = script_getnum(st,2);
 	else
 		instance_id = script_instancegetid(st);
 
 	if( instance_id <= 0 || instance_id >= MAX_MAP_PER_SERVER ) {
-		ShowError("script:instance_destroy: Trying to destroy invalid instance %d.\n", instance_id);
+		ShowError("buildin_instance_destroy: Trying to destroy invalid instance %d.\n", instance_id);
 		return SCRIPT_CMD_SUCCESS;
```
Lemongrass3110: Shouldnt that be SCRIPT_CMD_FAILURE?

src/map/script.c
```diff
@@ -19008,16 +19041,30 @@ BUILDIN_FUNC(instance_warpall)
...
-		return SCRIPT_CMD_SUCCESS;
-
-	if( !(p = party_search(instance_data[instance_id].party_id)) )
-		return SCRIPT_CMD_SUCCESS;
+		return SCRIPT_CMD_FAILURE;
 
-	for( i = 0; i < MAX_PARTY; i++ ) {
-		struct map_session_data *pl_sd;
-		if( (pl_sd = p->data[i].sd) && map[pl_sd->bl.m].instance_id == instance_id ) pc_setpos(pl_sd,map_id2index(m),x,y,CLR_TELEPORT);
+	switch(instance_data[instance_id].mode) {
```
Lemongrass3110: This block should be moved into buildin_instance_warpall_sub since you need to check if each session you find meets the criteria before you warp it, just like the code checked if the party matched before. Now you only try to check if the owner exists in general.

src/map/script.c
```diff
@@ -19115,6 +19161,68 @@ BUILDIN_FUNC(instance_check_party)
 }
 
 /*==========================================
+ * instance_check_guild
+ * Values:
+ * guild_id : Guild ID of the invoking character. [Required Parameter]
```
Lemongrass3110: If it's the guild id of the invoking character, why do we have it marked as required?
We could just read it from the attached player and have it optional.

src/map/script.c
```diff
@@ -19115,6 +19161,68 @@ BUILDIN_FUNC(instance_check_party)
...
+		return SCRIPT_CMD_FAILURE;
+	} else if (max < 1 || max > MAX_LEVEL) {
+		ShowError("buildin_instance_check_guild: Invalid max level, %d\n", max);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (script_hasdata(st,2))
+		guild_id = script_getnum(st,2);
+	else
+		return SCRIPT_CMD_SUCCESS;
```
Lemongrass3110: Should be FAILURE, if we someone did not provide a required parameter and the script engine let that call through.

Atemo:
Nice :+1: 
I thought about to add a command to change the instance timer before it starts (allowing dynamical instance timer), similiar to the removed command `instance_set_timeout`
```
*instance_set_timeout <alive timeout>,<idle timeout>{,<instance id>};

Sets the timeout values for an instance given by <instance id>. If no ID 
is specified, the instance the script is attached to is used. If in the end,
no instance_id is found the command halts the script execution.

Parameter <alive timeout> specifies the total amount of time the instance 
will exist. Parameter <idle timeout> specifies how long players have, when 
they are outside of the instance, until it is destroyed.

Both timeout values are in seconds.
```
Any thought?

aleos89:
@Atemo That is possible! I see no problem with expanding it further, especially now since the Instance TimeOut has been added to the DB.

secretdataz:
I'd go for
```
*instance_set_timeout <timeout>,<type>{,<instance id>};
Sets the timeout values for an instance given by <instance id>. If no ID 
is specified, the instance the script is attached to is used. If in the end,
no instance_id is found the command halts the script execution.

If <type> is 0, parameter <timeout> specifies the total amount of time the instance 
will exist. 
If <type> is 1, parameter <timeout> specifies how long players have, when 
they are outside of the instance, until it is destroyed.

Both timeout values are in seconds.
```
And a command to retrieve instance information from instance db would be great.

db/import-tmpl/instance_db.txt
```diff
@@ -1,5 +1,6 @@
 // Instance Database
 //
 // Structure of Database:
-// ID,Name,LimitTime,IdleTimeout,EnterMap,EnterX,EnterY,Map1,Map2,Map3,Map4,Map5,Map6,Map7,Map8
-
+// ID,Name,LimitTime,IdleTimeOut,EnterMap,EnterX,EnterY,Map2,Map3,Map4,Map5,Map6,Map7,Map8,Map9,Map10
```
Lemongrass3110: Should be Map2,...,Map255 now

src/map/clif.c
```diff
@@ -16587,8 +16588,9 @@ void clif_instance_create(unsigned short instance_id, int num)
 	if (!sd)
 		return;
 
+	db = instance_searchtype_db(instance_data[instance_id].type);
```
Lemongrass3110: if( !db )
 return;

src/map/clif.c
```diff
@@ -16632,8 +16635,9 @@ void clif_instance_status(unsigned short instance_id, unsigned int limit1, unsig
...
```
Lemongrass3110:
```
if( !db )
 return;
```

MidasTch:
.diff

Issue #1049, opened 2016-03-13
Unknown packets on 2015-10-29


admkakaroto:
Hello good evening, the client 2015-10-29 is giving desconect the players randomly, but this only occurs if it is performing some task, whether walking, talking or anything like that, if the same stand there it does not take dc.

Darkelfen:
I'm using the same client and we got ""massive"" random disconnections (30 ppl or so at the same time). Nevertheless, I dunno whether it is my hosting provider or my internet connection or a buggy OnTimer from custom script or the client itself.

Lemongrass3110:
This could be due to some packet's we do not have yet. That is why this client is not yet marked as stable.
If you can pinpoint when those disconnects happen we can look into it further and try to get a hold of the newer packet(s) or the structure changes of old ones we do not have yet.

Darkelfen:
As soon as I can isolate any factor, I will post it here.

Lemongrass3110:
@RagnarokNova any input from your server on this?

admkakaroto:
Console: http://i.imgur.com/qzoVTTg.png

Lemongrass3110:
@admkakaroto could you monitor this and ask your users what they do, before they get kicked by the server? But I think we are on the right track already.

admkakaroto:
@Lemongrass3110, I was talking to them, and no one knows for sure, the problem occurs in very different situations, one thing is quite certain, if the player is idle without performing any action, he has no problem with that.

Lemongrass3110:
Could you uncomment this line, so that we at least get to know which data is sent:
https://github.com/rathena/rathena/blob/master/src/map/clif.c#L58
Through this you should get the data which is sent from the client to the server on your console. :-)

Darkelfen:
Also packet 0x037b is causing disconnections.
I'm trying to pinpoint it cause.

![asdasd](https://cloud.githubusercontent.com/assets/7315177/13766649/2e630c92-ea3a-11e5-8927-69c802799906.png)

Lemongrass3110:
Which clients do both of you use?
The Ragexe or the RagexeRE?
If you use the Ragexe: Have you customized your packet keys?

Darkelfen:
I'm using: 
"2015-10-29aRagexeRE"

I will copy the packet_db for ver 54 I'm using.

```
//2015-10-29Ragexe
packet_ver: 54
packet_keys: ..., ..., ... //I don't want any of my players reading this get intel on those.
[...]

PR #1277, opened 2016-05-17
Cleaned and merged the chat functions

Lemongrass3110:
This adresses the issue that clients from late 2015 will not zero terminate most of their messages anymore.
It should also fix the problem with the unknown packets that were most likely caused by adding the zero termination directly into the packet buffer.


Lemongrass3110:
@admkakaroto @Darkelfen @Badarosk0 this could be interessting for you to test, because of #1049.

admkakaroto:
I will put in the emulator and see if there is a difference, if it improves, or something.

spinzaku:
is this related to people disconnecting via random chat ?

Lemongrass3110:
It most likely is. Since I was not able to reproduce these issues myself I can not promise, but it is most likely related to the wrong handling of the zero termination.

spinzaku:
my players have reported if they chat too much, they get sudden "disconnected from server" ><

Badarosk0:
Yes . This is happening in client 2015+

spinzaku:
Any estimate on this? my players disconnect so fast.. just talking in party chat. haha

Lemongrass3110:
Test it and give us feedback then it will be merged into master faster.

admkakaroto:
I'm with personal problems, so I'm away from ragnarok for some time ... I can only re-test and help within two weeks.

PR #1286, opened 2016-05-20
Adds var 'killedrid' to script monster OnKillEvents. (Closes #1278)

### LINKED ISSUE #1278: Killedrid on special Labels
recently i was making a script that needed had many monsters with on death event label, trying to reduce server needs i didn't use OnNPCKillEvent but instead made the mobs use OnKillTestEvent: which i attached to every mob manually.

when killing the mob it i notice the check for killedrid was not working

so i tested with

```
OnKillTestEvent:
	message strcharinfo(0),"Target "+ Target +" vs Killed "+killedrid;
end;
```

and it returned 

`Target 1002 vs Killed 0`

then when i killed a BigFoot then killed a special mob again i revived this

`Target 1002 vs Killed 1060`

Jeybla:
For more information, see Issue  #1278.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5628,7 +5628,8 @@ The only very special thing about this command is an event label, which is an
 optional parameter. This label is written like '<NPC object name>::<label name>' 
 and upon the monster being killed, it will execute the script inside of the 
 specified NPC object starting from the label given. The RID of the player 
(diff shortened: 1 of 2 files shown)


Lemongrass3110:
@Stolao could you please test if this fixes your problems?

PR #1298, opened 2016-05-24
Checkweight Ring of Wise King (Fixes #910)

### LINKED ISSUE #910: 13.2 quests: getitem vs player overweight
In the npc/quests/quests_13_2.txt we have some `getitem`s for items that weigh > 0 like `getitem 2782,1; //Ring_Of_Wise_King` and `getitem 7574+.@num,1;` at `jewel_13_2` function. But we do also need a player's current weight check there. When a player has no free weight (especially if he was overweighted during the dialog) and is to receive a quest item, an item isn't received and the quest continues, leaving a player no opportunity to attempt to get that item again. 

Latest commit: 6a62aab 
Mode: Renewal AND Pre-Renewal

Jeybla:
Issue #910, just a simple fix.

### DIFF
--- a/npc/quests/quests_13_2.txt
+++ b/npc/quests/quests_13_2.txt
@@ -3701,6 +3701,12 @@ mid_camp,197,237,3	script	Schwartzvalt Mechanic#1	869,{
 				mes "Hmm...";
 				mes "Anyways, now we've got the idea, I will give you this first piece of work.";
 				next;
+				if( !checkweight(2782,1) )
+				{
+					mes "[Mechanic Engineer Dorance]";
+					mes "I'm sorry but you're carrying too many items.";
+					close;
+				}
 				mes "[Mechanic Engineer Dorance]";
 				mes "Thank you for helping us out.";
 				mes "I hope this ring would be a big help for your journey through the other world.";


Lemongrass3110:
I kind of dislike to do this check silently without telling the user why he did not get his reward.
Maybe adding a custom dialog would be a good solution.
What do you guys think @rathena/rathena-script-db-developers?

cydh:
Just remain your players to make sure that there's empty slot in inventory & not overweight :v

secretdataz:
@Lemongrass3110 I agree, adding a custom dialog is the best way while we don't have any global solution for this imo.

Jeybla:
@Lemongrass3110 you're absolutely right.

Lemongrass3110:
Sorry to tell you, but I think the check you added should be before the "thank you, here have this ring" stuff.
Then I guess it could be merged.

Atemo:
Please move the check at the beginning of the npc, like the others check in official npc.

aleos89:
@Jeybla : If you want to update this, then it can be merged. :+1:

secretdataz:
Merged in 055ee42b3d6bfe035c235b7aa7d9ac7bc935feef.

PR #1299, opened 2016-05-25
Clean up `char_del_option` checks

cydh:
* That was removed accidently.
* Added comment for `char_del_option`, that only work for clients that send delete request by using 0x0068 or 0x01fb


src/char/char_clif.c
```diff
@@ -523,6 +527,36 @@ int chclif_parse_char_delete2_req(int fd, struct char_session_data* sd) {
...
+			))
+			;
+		else {
+			ShowInfo(""CL_RED"Char Deleted"CL_RESET" "CL_GREEN"(E-Mail)"CL_RESET".\n");
+			return 1;
+		}
+	}
+	// Birthdate (YYMMDD)
+	if (flag&2) {
+		if (!strcmp(sd->birthdate+2, delcode)) { // +2 to cut off the century
```
Lemongrass3110: Could you use && to merge those two ifs?

src/char/char_clif.c
```diff
@@ -523,6 +527,36 @@ int chclif_parse_char_delete2_req(int fd, struct char_session_data* sd) {
...
+ * @param flag Delete flag
+ * @return 1:Success, 0:Failure
+ **/
+static int chclif_delchar_check(struct char_session_data *sd, char *delcode, uint8 flag) {
+	// E-Mail check
+	if (flag&1) {
+		if (strcmp(delcode, sd->email) && //email does not matches and
+			(strcmp("a@a.com", sd->email) || //it is not default email, or
+			(strcmp("a@a.com", delcode) && strcmp("", delcode)) //email sent does not matches default
+			))
```
Lemongrass3110: I know this check is not coming from you, but it would also be nice to have that one changed check for a match and not a "not match" and having a no operation inside the if block. Also same as below the two ifs should be merged.
I would suggest that way:
```
!strcmp(delcode, sd->email) || //email does match or
(
	!strcmp("a@a.com", sd->email) && //it is default email and
	(
		!strcmp("a@a.com", delcode) || // user also sent the default email 
		!strcmp("", delcode)) // user sent an empty email
	)
)
```

conf/char_athena.conf
```diff
@@ -172,11 +172,14 @@ char_del_delay: 86400
...
 // This restricts players from changing the langtype and deleting characters.
-// For birthdate, the client must be 20100803 or newer.
 // Defaults based on client date.
 // 1: Email address
 // 2: Birthdate
 // 3: Email address or Birthdate
+// IMPORTANT!
+// - This config only works for clients that send 0x0068 or 0x01fb for delete request.
+// - Try use langtype 1 for newer clients (2013+), to use 0x01fb.
+// - Clients that are not using 0x0068 or 0x01fb, only use birthdate (YYMMDD) as default.
```
Lemongrass3110: This might be a bit to technical for the comment here.
Maybe we/you should hand over the packet id to your newly created function chclif_delchar_check and include those checks in there.
For example:
The user has set his config to email address, but the client sends us packet that are not 0x0068 or 0x01fb, then you could add a warning which is displayed on the console.

As for birthdate in general we could add checks like we did for auction and battle conf in the mapserver, to check if the packet version is higher then 20100803 when the user has the option set to &2.

src/char/char_clif.c
cydh: yea, bit confused with that code. I wanted just check
 * 1. Is it match
 * 2. Not default email
for empty email/birthdate, must always be checked first, tbh.

Lemongrass3110:
Any chance that you do a follow up @cydh or should @aleos89 or I do it?

conf/char_athena.conf
cydh: hmm I just want to put it there, so people will read it.

cydh:
Hmm. just merge it maybe?

PR #1308, opened 2016-05-26
Cleaned up item group parsing

Lemongrass3110:


Everade:
You've just broken all my custom item boxes.
And i've no idea how this new thing shall work.

Could you guys please start to document such drastic changes?

This happens all the time.
[Error]: sv_readdb: Insufficient columns in line 1 of "db/import/item_group_db.txt" (found 1, need at least 2).

my item_group_db.txt includes:
import: db/item_custom.txt

That's it

Akkarinage:
@Everade it's generally prudent to have a look at the changes you'll receive before updating. I would also move your import: to another file, perhaps one capable of parsing an import statement correctly.

Lemongrass3110:
Hey everade,
Sorry that we broke things for you.
We generally asume people look in the changes when updating and since this was the only commit regarding this we asumed anyone would find it out really fast. We expect that people running a server are able to find out where their problem might be coming from and are able to resolve them in most of the cases.

As for your problem it would be best to put your customs into /db/import/item_group_db.txt because the import statement was dropped for item random databases with this commit to follow the rest of the import behaviour.

I will talk with @akkarinage and @aleos today and it is very likely that we create a topic on this one and monster random group parsing, since I understand that this might cause problems for you and other users too.

Everade:
@Akkarinage 
@Lemongrass3110 
Yes you can assume that.
And i did that.

I ALWAYS read through every single update, so i won't miss something important.
And i've seen this group item, as you guys stated within the commit *Docs have been updated* so i checked the docs, but there was NO update on it.

So i was kinda confused, so i also took the time to check what exactly you've changed, and i found that you removed the import stuff, yet i wondered how the hell my customs are going to be found.
The only logical solution i've had was, i still have to use the import function.

I did some tests, however gave up as nothing seemed to work and you guys haven't made any doc update about this matter.

lighta:
heya @Everade, yeah they forgot the doc but I can understand it since they have tons of work and exactly how thing work so they just forget about it. Just need a little help sometime...

Anyway back to your issue, putting back the import command shouldn't be too much work I think but it's not the way I would recommend you to go since new commit will expect you won't have it.
So the question is "Do you really need the 'import' ?" 

If I take you example of "import: db/item_custom.txt"
The equivalent would be to move that file into : db/import/item_db.txt
The way it's work is that **every** file in db/ folder have an equivalent in db/import/ which work like the import command. (mean it override the base file if id conflict).

Finally I'm the one faulty here for the "db/import/ stuff not having done some doc, but we had a little announcement for my defence :p
https://rathena.org/board/topic/92095-january-digest-2014/

PR #1321, opened 2016-05-27
map-server: corrected intif_create_party sent-buffer-allocation size.

wilkemeyer:
Yet another 'too small buffer space reserveration' case.

### DIFF
--- a/src/map/intif.c
+++ b/src/map/intif.c
@@ -543,7 +543,7 @@ int intif_create_party(struct party_member *member,char *name,int item,int item2
 		return 0;
 	nullpo_ret(member);
 
-	WFIFOHEAD(inter_fd,64);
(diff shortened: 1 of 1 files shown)


src/map/intif.c
```diff
@@ -543,7 +543,7 @@ int intif_create_party(struct party_member *member,char *name,int item,int item2
 		return 0;
 	nullpo_ret(member);
 
-	WFIFOHEAD(inter_fd,64);
+	WFIFOHEAD(inter_fd, 30+sizeof(struct party_member));
```
Lemongrass3110: If we already touch it this should be 6 + NAME_LENGTH + sizeof(struct party_member)
Dont you think so?
Same should be applied to the actual size of the packet at offset 2.

aleos89: I agree with @Lemongrass3110. We should include the `NAME_LENGTH` in the size.

PR #1334, opened 2016-06-02
case IM_CHAR:

Atemo:
case IM_CHAR:
    script.c instance_create use sd->status.char_id
    instance.c use sd->status.account_id and map_id2sd check

### DIFF
--- a/src/map/instance.c
+++ b/src/map/instance.c
@@ -78,7 +78,7 @@ void instance_getsd(unsigned short instance_id, struct map_session_data **sd, en
 			(*target) = PARTY;
 			break;
 		case IM_CHAR:
-			(*sd) = map_id2sd(instance_data[instance_id].owner_id);
+			(*sd) = map_charid2sd(instance_data[instance_id].owner_id);
 			(*target) = SELF;
 			break;
 	}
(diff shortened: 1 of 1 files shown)


aleos89:
This method works as well. The only difference is that if the player logs onto another character, they will essentially be able to create another instance with that character (while the previous one still exists/times outs).

Atemo:
>The only difference is that if the player logs onto another character, they will essentially be able to create another instance with that character

Each character of the account can create his own instance before and after my commit from my test.

This commit resolve an issue with IM_CHAR and still allow to use the 'char_id' as owner id for ```instance_create```command :
In [https://github.com/rathena/rathena/blob/master/src/map/script.c#L18886](url) ```owner_id``` store the char_id but ```instance_create``` in instance.c check the account id [https://github.com/rathena/rathena/blob/master/src/map/instance.c#L341](url) resulting the perma ShowError
```ShowError("instance_create: character %d not found for instance '%s'.\n", owner_id, name);```

Lemongrass3110:
@aleos89 we could take this as a feature improvement and add another one: IM_ACC(OUNT).
One instance per account and make IM_CHAR once per character rather than once per account, which could be misleading as it is now.

Atemo:
I clarify the point
>>The only difference is that if the player logs onto another character, they will essentially be able to create another instance with that character

>Each character of the account can create his own instance before and after my commit from my test.

If the player logs onto another character of his account, he will be able to create another instance with that character before and after my commit from my test. It's not currently one instance per account for IM_CHAR.

cydh:
@Atemo go click the merge button :v

PR #1092, opened 2016-03-23
Add support for the new basic skill & Clean-ups

secretdataz:

PR #1380, opened 2016-06-21
Added support to Para_Team_Mark_ item in Eden Group

AtlantisRO:
In 08/01/2012 Maintenance was added an improved version for Eden Group Mark:

"- Added exchange function for 'Lime Evenor' NPC where you can exchange  'Eden Group Mark' for a consumable item 'Eden Group Mark' with  butterflywing-like function that can be used infinite number of times."
Source: http://forums.irowiki.org/index.php?topic=95103.0

This commit gives support to this new Eden Group Mark. All dialogues are official and have been taken from iRO.

If you need any reference about dialogues, can provide a replay using this npc and receiving this item in iRO.


secretdataz:
I think you should unify the item checking part into one function like what I did in [this PR](https://github.com/rathena/rathena/pull/1092/files).

Lemongrass3110:
His changes look good.
We could do that as a follow up too?
I would suggest adding that one as a global function in npc/re/quests/eden/eden_common.txt

Thanks for submitting this pull request @AtlantisRO!

AtlantisRO:
Thanks for the merge and the suggestion. We were traying to be the less invasive possible with changes so, only changed the strictly needed.

I can do that function, here are some suggestions:

"F_HasEdenMark"
"F_EdenMarkCheck"
"F_IsEdenMember"
"F_EdenMember"

Please tell me the one that looks better or give me another different.

AtlantisRO:
Sended function as "F_HasEdenGroupMark".

PR #1417, opened 2016-07-05
Used defined variable names for constant exports

Lemongrass3110:
Same should maybe be done for bank vault, but since the define has a `#` we should be more careful about that one maybe.


cydh:
anyway, maybe we should start normalize the constants usage in DB or Scripts, since we simply export it from `script_constants.h`, I mean about the letter case.
Such `Class_All -> CLASS_ALL`

Lemongrass3110:
Would totally make sense, but since we support those constants case insensitive it might be problematic to enforce that rule, but we could try to use them in caps whereever possible on the scripts in the repository.

cydh:
I don't mean to enforce, just change all our db/scripts that use the
constants.

PR #1437, opened 2016-07-19
Added script bonuses to add player's drop rate

cydh:
Added script bonuses in order to support official `AddReceiveItem_Equip n`

* `bonus2 bDropAddRace,r,x;` : Adds x% to player's drop rate when killing a monster with race r.
* `bonus2 bDropAddClass,c,x;` : Adds x% to player's drop rate when killing a monster with race c.

----------

cc @rathena/rathena-core-developers 
The drop rates are calculated by
* 1. Normal server/mob rate
* 2. (Custom) Size influence (multiplication)
* 3. (Custom rate by) Luk (multiplication)
* 4. Bonus from PK server (multiplication)
* 5. **This** item bonus (additive)
* 6. `SC_ITEMBOOST` (multiplication)
* 7. VIP (multiplication)

I wonder if **this** item bonus must be before the size influence. xD


src/map/mob.c
```diff
@@ -2527,14 +2527,29 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
 				(int)(md->level - sd->status.base_level) >= 20)
 				drop_rate = (int)(drop_rate*1.25); // pk_mode increase drops if 20 level difference [Valaris]
 
-			// Increase drop rate if user has SC_ITEMBOOST
-			if (sd && sd->sc.data[SC_ITEMBOOST]) // now rig the drop rate to never be over 90% unless it is originally >90%.
-				drop_rate = max(drop_rate,cap_value((int)(0.5+drop_rate*(sd->sc.data[SC_ITEMBOOST]->val1)/100.),0,9000));
+			if (sd) {
+				int drop_rate_ = 0;
+
+				if (src) {
```
Lemongrass3110: Why do you check for src here?

cydh: I think I removed it already, maybe I pop wrong stash.

cydh:
@Lemongrass3110 thank you for the follow up

PR #1461, opened 2016-07-29
Fixes MC_CHANGECART skill

jgoog999:
- adjusted the skill requirements in to correct values


src/map/clif.c
```diff
@@ -11700,20 +11700,20 @@ void clif_parse_ChangeCart(int fd,struct map_session_data *sd)
 
 	type = (int)RFIFOW(fd,packet_db[sd->packet_ver][RFIFOW(fd,0)].pos[0]);
 #ifdef NEW_CARTS
-	if( (type == 9 && sd->status.base_level > 131) ||
-		(type == 8 && sd->status.base_level > 121) ||
-		(type == 7 && sd->status.base_level > 111) ||
-		(type == 6 && sd->status.base_level > 101) ||
```
aleos89: It may be more wise to just make these 4 cases above -1 rather than adjusting the rest below to `>=`.

PR #1469, opened 2016-07-31
Updated .gitattributes

cydh:
Updated for:
* `*.c          diff=cpp eol=crlf`
* `*.h          diff=cpp eol=crlf`
* `*.sql        text eol=crlf`

Set .sql end-line to CRLF (since it always be like that) and to prevent any future ugly changes.
Happened twice already, by @Playtester and @rAthenaAPI.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,12 +1,14 @@
-*        text=auto
-*.c      diff=cpp
-*.sln    merge=union
-*.vcproj merge=union
-*.vcxproj merge=union
-*.sln text eol=crlf
-*.bat text eol=crlf
-configure text eol=lf
+*            text=auto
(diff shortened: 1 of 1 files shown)


aleos89:
Won't this ruin line endings for Linux users now?

cydh:
and because we don't normalize it, there's always be clash.
but since it doesn't work, this guy @rAthenaAPI still generates LF. will revert this then.
something changed? previously this guy generates CRLF.
so, someone to ASK everyone to don't generate .sql files anymore, just leave @rAthenaAPI generate the files.

Lemongrass3110:
gitattributes will not change how things are commited only how they are checked out.
https://help.github.com/articles/dealing-with-line-endings/

PR #1481, opened 2016-08-04
Status Change display in status window

### LINKED ISSUE #554: RE: Weapon attack bonus from Strength/Dexterity Display Issue
this is also similar to Critical Display wearing a Katar Weapon but this time it is Weapon Attack bonus from stats that must be hidden from Status Window (Alt+A or Alt+Q)

note:
to reproduce you must wear weapon to display additional weapon attack.

aleos89:
- Adjusted the way some status changes display in the player's status window. (Fixes #554)
  - Some status changes are hidden from the display.
- Katar's critical bonus is no longer shown in the status window.
- SC_INCCRI now has doubled critical rate while wearing Katars. Thanks to @Limestone!
- Refactored some status_calc_\* functions to no longer be static and moved them to the header.


cydh:
What I found for now,
`SC_ASSUMPTIO` is only called by `status_calc_def(..., false)` and `status_calc_mdef(..., false)` manually, from battle.c.
It doesn't trigger the calculation on `display` part of status_calc_mdef/def, so it doesn't require `SCB_DEF|SCB_MDEF`, or it will triggers calculation when status changes for nothing.
So does `SC_FORCEOFVANGUARD`.

aleos89:
You're right, those flags for those SC's won't be needed in the end.

PR #1509, opened 2016-08-21
Implemented status reduction potion support

secretdataz:
* New script command `needed_status_point` added.
* The items are not yet implemented.

### DIFF
--- a/npc/other/CashShop_Functions.txt
+++ b/npc/other/CashShop_Functions.txt
@@ -319,11 +319,12 @@ function	script	F_CashReduceStat	{
 	
 	if((readparam(.@type) + .@amount) < 1) return;
 	
-	if(.@itemid)
+	if(.@itemid) {
 		if(countitem(.@itemid))
 			delitem .@itemid,1;
 		else
 			return;
+	}
 	StatusPoint += needed_status_point(.@type, .@amount);
 	statusup2 .@type,.@amount;
 	return;
--- a/npc/other/CashShop_Functions.txt
+++ b/npc/other/CashShop_Functions.txt
@@ -307,20 +307,24 @@ function	script	F_Snowball	{
 
 // Status reduction potion
 //============================================================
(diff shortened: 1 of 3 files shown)


secretdataz:
@zackdreaver You may want to check if this is your expected behavior of stat reduction potion.

npc/other/CashShop_Functions.txt
```diff
@@ -304,3 +304,23 @@ function	script	F_Snowball	{
...
+//   that stat to original value.
+// - Doesn't work if base status <type> is already 1.
+// * callfunc("F_CashReduceStat",<type>{,<itemid>});
+function	script	F_CashReduceStat	{
+	.@type = getarg(0);
+	.@itemid = getarg(1, 0);
+	
+	if(readparam(.@type) < 2) return;
+	
+	if(.@itemid)
```
Lemongrass3110: You get the status points before and check if that status can be decreased at all. Because then the item should not be deleted. Would also be nice to add the hardcoded -1 as parameter instead, so that customs can be done that also reduce by -5 for example.

zackdreaver:
Apologies for late reply, I'm not with pc installed with VS and Ro for weeks.
will test tomorrow, when i'm back home.

zackdreaver:
Alright, I'm done with testing.
This is the item script that I use:
`22702,STR_Reduction_Potion,STR Reduction Potion,2,20,,0,,,,,0xFFFFFFFF,63,2,,,,,,{ callfunc "F_CashReduceStat",bStr,-1,22702; },{},{}`

The effect is working fine,
but the item quantity shouldn't be consumed/decreased when used with base stat 1.

secretdataz:
@zackdreaver I think you could use item type 11 for that.

zackdreaver:
Then, this PR is good.
Just need the required item's db.

zackdreaver:
thank @secretdataz

PR #1511, opened 2016-08-23
Initial release of the clan system

### LINKED ISSUE #1317: Clan System

Lemongrass3110:
Thanks to @RagnarokNova for his help.

We have a few things that are not yet in there. I will open up some issues for it and try to solve them as soon as possible.

Fixes #1317


sql-files/main.sql
```diff
@@ -286,6 +286,26 @@ CREATE TABLE IF NOT EXISTS `charlog` (
...
+  PRIMARY KEY (`clan_id`)
+) ENGINE=MyISAM AUTO_INCREMENT=5 DEFAULT CHARSET=latin1;
+
+-- ----------------------------
+-- Records of clan
+-- ----------------------------
+INSERT INTO `clan` VALUES ('1', 'Swordman Clan', 'Raffam Oranpere', 'prontera', '500');
+INSERT INTO `clan` VALUES ('2', 'Arcwand Clan', 'Devon Aire', 'geffen', '500');
+INSERT INTO `clan` VALUES ('3', 'Golden Mace Clan', 'Berman Aire', 'prontera', '500');
+INSERT INTO `clan` VALUES ('4', 'Crossbow Clan', 'Shaam Rumi', 'payon', '500');
```
cydh: the `clan_id`s, not using the constants?

src/char/int_clan.c
```diff
@@ -0,0 +1,228 @@
...
+
+		aFree(clan_ids);
+	}
+
+	return 0;
+}
+
+void inter_clan_final(){
+	db_destroy(clan_db);
+}
```
cydh: empty endline is needed here

anacondaqq:
Dear @Lemongrass3110 what minimal exe date and exe we should use to try the feature?
2015-11-04a-Ragexe is ok for it?

db/packet_db.txt
```diff
@@ -2330,6 +2330,9 @@ packet_keys: 0x631C511C,0x111C111C,0x111C111C // [Shakto]
 0x09D6,-1,npcmarketpurchase,2:4:6
 0x09D7,-1
 0x09D8,2,npcmarketclosed,0
+0x098D,-1,clanchat,2:4
```
spinzaku: will this packets work with 2015-11-04aRagexe ?

Lemongrass3110: Yes.
@anacondaqq everything from 2013-12-23 will work.

sql-files/main.sql
Lemongrass3110: Thanks seems like I forgot to add that update... >.<

I did not use constants because because the char server does not have a constant resolving feature in it yet. Any suggestions?

src/char/int_clan.c
Lemongrass3110: Thanks I will add it with the next follow up.

src/map/script.c
```diff
@@ -20967,6 +20970,38 @@ BUILDIN_FUNC(npcshopupdate) {
...
+	if( !script_charid2sd( 2, sd ) ){
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if( clan_member_leave( sd, sd->status.clan_id, sd->status.account_id, sd->status.char_id ) ){
+		return SCRIPT_CMD_SUCCESS;
+	}else{
+		return SCRIPT_CMD_FAILURE;
+	}
+}
```
cydh: shouldn't the `else` part, just let it be end of function control?
(and for script above)

src/map/script_constants.h
```diff
@@ -1401,6 +1401,11 @@
...
 	export_constant(SC_TIME_ACCESSORY);
 	export_constant(SC_MAGICAL_FEATHER);
+	export_constant(SC_CLAN_INFO);
+	export_constant(SC_SWORDCLAN);
+	export_constant(SC_ARCWANDCLAN);
+	export_constant(SC_GOLDENMACECLAN);
+	export_constant(SC_CROSSBOWCLAN);
 #ifdef RENEWAL
 	export_constant(SC_EXTREMITYFIST2);
 #endif
```
cydh: `export_constant(CLAN);` too? from `enum send_target`?

sql-files/main.sql
cydh: forgot, well. just leave it be.

sql-files/upgrades/upgrade_20160823.sql
```diff
@@ -0,0 +1,22 @@
...
+-- ----------------------------
+-- Records of clan
+-- ----------------------------
+INSERT INTO `clan` VALUES ('1', 'Swordman Clan', 'Raffam Oranpere', 'prontera', '500');
+INSERT INTO `clan` VALUES ('2', 'Arcwand Clan', 'Devon Aire', 'geffen', '500');
+INSERT INTO `clan` VALUES ('3', 'Golden Mace Clan', 'Berman Aire', 'prontera', '500');
+INSERT INTO `clan` VALUES ('4', 'Crossbow Clan', 'Shaam Rumi', 'payon', '500');
+
+ALTER TABLE `char`
+	ADD COLUMN `clan_id` INT(5) UNSIGNED NOT NULL DEFAULT '0' AFTER `hotkey_rowshift`;
```
cydh: must be `INT(11)`

cydh:
Need the doc updates for `packet_interserv.txt`

src/map/script.c
Lemongrass3110: I was thinking that it might be safer to terminate the script at this point, that is why I set it up like this.

src/map/clif.c
```diff
@@ -18117,6 +18144,94 @@ void clif_party_leaderchanged(struct map_session_data *sd, int prev_leader_aid,
...
+	clan_send_message( sd, (char*)RFIFOP(fd,4), RFIFOW(fd,2) - 4 );
+#endif
+}
+
+void clif_clan_basicinfo( struct map_session_data *sd ){
+#if PACKETVER >= 20131223
+	int fd = sd->fd, offset, length;
+	struct clan* clan = sd->clan;
+	char mapname[MAP_NAME_LENGTH_EXT];
+
```
cydh: sure doesn't want to check the `sd` first?
and also the `fd` (`clif_session_isValid()` wow, forgot about this actually)
just incase if player is autorade, they has no `fd` set.

src/map/script.c
cydh: /ic

Lemongrass3110: Might also be an option to make it return true/false to the caller so that they can check it.
What would you prefer/suggest?

src/map/script_constants.h
Lemongrass3110: done in 2cf39c0

src/map/script.c
aleos89: Having a return value can be beneficial for `clan_join` and `clan_leave` to let the script know if the player was properly entered/removed from the clan.

src/char/int_clan.c
```diff
@@ -0,0 +1,228 @@
...
+	return clan;
+}
+
+int mapif_clan_info( int fd ){
+	DBIterator *iter;
+	int offset;
+	struct clan* clan;
+	int length;
+
+	length = 4 + db_size(clan_db) * sizeof( struct clan );
```
cydh: make it
```int clan_len = sizeof( struct clan )```

File: src/map/clan.h [hunk 9]
```diff
@@ -0,0 +1,22 @@
...
+	void do_final_clan();
+	struct clan* clan_search( int id );
+	void clan_load_clandata( int count, struct clan* clans );
+	void clan_member_joined( struct map_session_data* sd );
+	void clan_member_left( struct map_session_data* sd );
+	int clan_member_join( struct map_session_data *sd, int clan_id, uint32 account_id, uint32 char_id );
+	int clan_member_leave( struct map_session_data* sd, int clan_id, uint32 account_id, uint32 char_id );
+	void clan_send_message( struct map_session_data *sd, const char *mes, int len );
+	void clan_recv_message(int clan_id,uint32 account_id,const char *mes,int len);
+	struct map_session_data* clan_getavailablesd( struct clan* clan );
```
cydh: what a indent, just remove it like other `.h` files

cydh:
1. Need alliance & enemies table
  2. Prevent player create guild while in clan by using `@guild`
  3. Why? ```// Should display a clientside message "You are currently joined in Clan !!" so we ignore it```
  4. `Because you need guildleaders to WoE, clansmen cannot participate in WoE`. Maybe add a fun point, add config to dual mode (can join both Guild and Clan).
  5. There's **enemies**, why don't add custom `cvc` (Clan vs Clan) mapflag?

Lemongrass3110:
1. This is why it is called initial release. Since we do not have enough information about what influence alliances and enemies have I just left them out for now.
2. Will be added shortly. Thanks for bringing that up. I thought I did that already.
3. Because this is clientside functionality. Ask Gravity why they did it like that.
4. This will create really troublesome problems, but if you want to try it you can do that. The serverside logic is easy, but having it displayed correctly on clientside will be a funny thing to do.
5. I wanted to add that when we add alliances and enemies.

admkakaroto:
The clan checking npc is not checking properly, even if not in a clan he will not let you proceed when I get home from work I will try to fix and open a pull.

admkakaroto:
@Lemongrass3110 

In the topic NPCs show the emblem of the clan under each one, I may have applied the wrong patch, everything is working, less clan emblem that does not appear below the npc, see:

Topic: https://rathena.org/board/topic/105463-clan-system/

Server: http://imgur.com/tZZFEzq

secretdataz:
@admkakaroto What is the date of the client you are using?

Lemongrass3110:
@admkakaroto this is a known issue at the moment. I know how to get it working, but we did not implement it properly yet.

admkakaroto:
@secretdataz 2015-11-04

@Lemongrass3110 i understand, but the system is working perfectly, congratulations.

src/char/inter.c
```diff
@@ -54,6 +55,7 @@ int inter_recv_packet_length[] = {
 	-1,10, 6,-1,  0, 0, 0, 0,  0, 0, 0, 0, -1,10,  6,-1,	// 3070-  Mercenary packets [Zephyrus], Elemental packets [pakpil]
 	48,14,-1, 6,  0, 0, 0, 0,  0, 0, 0, 0,  0, 0,  0, 0,	// 3080-
 	-1,10,-1, 6,  0, 0, 0, 0,  0, 0, 0, 0,  0, 0,  0, 0,	// 3090-  Homunculus packets [albator]
+	 2,-1, 6, 6,  0, 0, 0, 0,  0, 0, 0, 0,  0, 0,  0, 0,	// 30A0-  Clan packets
```
cydh: I wonder, maybe can we move the clan packets to empty packets there. We have lot unused packets

Lemongrass3110: Yeah would not be a problem. I just though i'll create a new range because it seemed like that for all other things too.

Grantea:
My character's clan emblem disappears when I use `@refresh`, but I'm still able to see other's emblem and when I change maps it comes back. it's not a big issue.

Client: 2015-11-04.

Also, I don't know how the official behavior is, but would be great if we have a conf like "display_party_name" to show the clan name.

Lemongrass3110:
From looking into the code I think there are more things missing when you use refresh at the moment. All kind of visual status changes and hat effects. I will test it later today.

About the display_party_name feature:
So you would want it to only display the party name, if a player is a guild OR in a clan, correct?
In officials you cannot be in a clan and a guild at the same time and we mimiced that behaviour for now. It's unsure if we will support being in both anytime soon.
Also the clan name is not being displayed, you only see the clan emblem beside the member characters.

Grantea:
@Lemongrass3110 Exactly. Thank you for your quick response and amazing work!

admkakaroto:
@Lemongrass3110 

When an apprentice enters a Clan after logout and login again, the HP of the character is reduced, is that correct?

Before: http://imgur.com/XVHqI07
After: http://imgur.com/yamQOD3

not yet applied the changes you made on the 10th, I will do it tomorrow and test.

Lemongrass3110:
@admkakaroto just confirmed this and looking into it. Thanks for reporting.

PR #1516, opened 2016-08-25
Prevented Phantasmic Arrow to knock back targets that would die from …

cyilcode:
…its attack. Added party and guild checks for character delete process.


secretdataz:
Your commit has another unrelated change bloated. Please remove that first.
Also, please refrain from using your master branch to commit changes. If this PR gets rejected, your fork will be screwed.

aleos89:
As @secretdataz stated, you should create this in separate branches on your fork. Also, separate the Phantasmic Arrow fix from the Guild/Party fix into two branches and PRs rather than being in the same PR.

Issue #1480, opened 2016-08-04
Prevent character deletion if player is in a party or guild


cydh:
Just another forgotten official features.
`char_clif.c::chclif_parse_char_delete2_req`
```c++
	/*
		// Aegis imposes these checks probably to avoid dead member
		// entries in guilds/parties, otherwise they are not required.
		// TODO: Figure out how these are enforced during waiting.
		if( guild_id )
		{// character in guild
			chclif_char_delete2_ack(fd, char_id, 4, 0);
			return 1;
		}

		if( party_id )
		{// character in party
			chclif_char_delete2_ack(fd, char_id, 5, 0);
			return 1;
		}
	*/
```

PR #1519, opened 2016-08-26
Added guild/party checks for character deletion process #1480

cyilcode:


cydh:
Ok, looks good. This is that I wanted.

Lemongrass3110:
I think this should have a configuration to enable/disable this check.

Additionally shouldn't it be added to `chclif_parse_delchar` too?

Lemongrass3110:
Could you please enable edit from maintainers?
https://rathena.org/board/topic/107284-improving-collaboration-with-forks/
Thank you.

cydh:
what's the point to enable it, in the end u squash it LOL
just merge (don't push it yet) it locally, edit as u like, then push it ckckck

Lemongrass3110:
The difference is that others can review my changes before i push it to master too. And yes I will squash it because I dont need credits for 2 lines of changes to feel better ;)

PR #1530, opened 2016-08-31
Itemskill script command expansion

secretdataz:
This PR expands the `itemskill` script command to allow skill requirements checking to be done in some items that require skill requirements to be checked normally.


secretdataz:
Potential side effects testing done.

PR #1546, opened 2016-09-06
Added `skill_duration` mapflag.

cydh:
- This mapflag to set skill unit time limit to `n%` of original duration.
- Implemented mapflags for Hunter's Traps in GVG, BG, and Novice Guild maps.

Signed-off-by: Cydh Ramdh cydh@pservero.com


cydh:
I'm f*cked since #2089 was merged

PR #1550, opened 2016-09-08
Update client.conf

Sirique:


Sirique:
oops sorry wrong place haha

PR #1556, opened 2016-09-13
Added English comments in nullpo.h

secretdataz:
Should we also remove _f suffixed macros?

### DIFF
--- a/src/common/nullpo.h
+++ b/src/common/nullpo.h
@@ -53,7 +53,7 @@
 	if (nullpo_chk(NLP_MARK, (void *)(t))) {break;}
 
 // Different C compilers uses different argument formats
-#if __STDC_VERSION__ >= 199901L
+#if __STDC_VERSION__ >= 199901L || defined(_MSC_VER)
 /* C99 standard */
 /**
  * Return 0 and display additional information if pointer is not found.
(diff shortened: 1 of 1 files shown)


src/common/nullpo.h
```diff
@@ -19,52 +19,48 @@
  *----------------------------------------------------------------------------
  */
 /*======================================
- * Null�`�F�b�N �y�� ���o�͌� return
- *�E�W�J�����if�Ƃ�return�����o��̂�
- *  ��s�P�̂Ŏg��Ă��������B
- *�Enullpo_ret(x = func());
- *  �̂悤�Ȏg�p�@��z�肵�Ă��܂��B
+ * Check for NULL pointer and output information
```
Lemongrass3110: Should have "after return" at the end as you can see in the japanese text
The additional comment that was lost here is a recommendation to use a single line of code for this kind of checks and a note that this also works for/with functions.

src/common/nullpo.h
```diff
@@ -194,27 +189,28 @@ int nullpo_chk_f(const char *file, int line, const char *func, const void *targe
 
 /*======================================
  * nullpo_info
- *   nullpo���o��
- * [��]
+ *   Display information of the code that cause this function to trigger
```
Lemongrass3110: Diplay+s
cause+d

PR #1574, opened 2016-09-28
map-server: clif: bugfixes in clif_getareachar_skillunit 

wilkemeyer:
Updated clif_getareachar_skillunit:

Fixed packet usage: packet ZC_SKILL_ENTRY3 is available since 2011-04-20.
Also ZC_SKILL_ENTRY4 member: range is not WORD its byte.

Also ZC_SKILL_ENTRY3 & up is dynamic not fixed length packet.

Additionally i've split the method into 2 as i'm working on optimizing unit group spawns.
Ragnarok's protocoll allows since ZC_SKILL_ENTRY3 multiple unit's being sent with one packet:
This will save some cpu time as on serverside fewer loops & checks are required.
On clientside there's less parsing overhead && fewer pps overall.

Maybe someone can make a review before merging it (by testing it against packetversions that use ZC_SKILL_ENTRY4 & ZC_SKILL_ENTRY5, i did only some testing against ZC_SKILL_ENTRY3.)


lighta:
This is already merge in master.
If anything only need to uncomment:

[8c7.diff.txt](https://github.com/rathena/rathena/files/1484442/8c7.diff.txt)

Issue #1561, opened 2016-09-18
Random Options going haywire


Darkelfen:
So not so long ago I've decided to update my live server to "latest" (hash: https://github.com/rathena/rathena/commit/e820ef9090a02e8483ef672147e159210dd09c16) and updated the SQL tables.

I'm using a storage size of 850 and a guild storage size of 999, consequently having the mmo_charstatus issue, so I've reduced the number of random options to 1 in order to avoid that issue.

So far everything was ok, but I've noticed that items with random options were appearing out of the blue. Some of my players reported having items with 4 zeroed options. What scared me the most was that I've created an item through `@item ` and it came with 4 random options but one of them were not zeroed, it was SP recovery + -3261%, then I've relogged and the options disapeared as they came.

**Edit**: Every single armor/weapon on storage got 4 zeroed options.

I dunno how this issue is generated or how it can be replicated.

admkakaroto:
I'm having the same problem , I could solve this by using a patch in rathena , but I'm afraid it is likely to cause problems in the emulator, see about it in the link:

link: https://github.com/rathena/rathena/issues/1523

Darkelfen:
What is concerning me is that items with random options are appearing at random without a pattern or whatsoever.... Today I've found this and it really freaked me out.

![image](https://cloud.githubusercontent.com/assets/7315177/18624007/f31759d0-7e19-11e6-8b0d-3ad669e9931d.png)

Hopefully it's just a "visual" problem, because on the SQL everything is zeroed. So, my best guess is that maybe the client isn't properly showing the items with/without random options.
 
As a side note, I've set random options to 1 on mmo.h so I'm not having the mmo_charstatus size issue.

Lemongrass3110:
If you reduce the define to 1 you have to make sure that all of the clif functions still send out 4, because this is the packet size the client expects. The rest will possibly cause visual glitches like the ones you describe because of uninitialized memory on clientside or parts of the packets after the item packets being read into your random options.

admkakaroto:
@Lemongrass3110 
You advised not to use this diff ?

Diff; https://raw.githubusercontent.com/vstumpf/ragnarok/master/mmo_charstatuspacketlength.diff

Darkelfen:
@Lemongrass3110  Thank you, it was indeed solved. I'll be monitoring if an item apears but I doubt so.

PR #1580, opened 2016-09-29
Fixed random options with MAX_ITEM_RDM_OPT < 5

Lemongrass3110:
Fixed a possible issue on 2015 clients with random options, if you define an amount less than 5, because the client expects that much.

Related issue: #1561

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -2445,6 +2445,14 @@ void clif_add_random_options(unsigned char* buf, struct item *it) {
 		WBUFW(buf, i*5 + 2) = it->option[i].value;	// Value
 		WBUFB(buf, i*5 + 4) = it->option[i].param;	// Param1
 	}
+	
+#if MAX_ITEM_RDM_OPT < 5
+	for ( ; i < MAX_ITEM_RDM_OPT; i++) {
+		WBUFW(buf, i*5 + 0) = 0;		// OptIndex
+		WBUFW(buf, i*5 + 2) = 0;	// Value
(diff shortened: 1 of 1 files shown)


src/map/clif.c
```diff
@@ -2445,6 +2445,14 @@ void clif_add_random_options(unsigned char* buf, struct item *it) {
 		WBUFW(buf, i*5 + 2) = it->option[i].value;	// Value
 		WBUFB(buf, i*5 + 4) = it->option[i].param;	// Param1
 	}
+	
+#if MAX_ITEM_RDM_OPT < 5
+	for ( ; i < MAX_ITEM_RDM_OPT; i++) {
```
secretdataz: Condition statement should be `i < 5;`.

src/map/clif.c
```diff
@@ -2445,6 +2445,14 @@ void clif_add_random_options(unsigned char* buf, struct item *it) {
...
 		WBUFB(buf, i*5 + 4) = it->option[i].param;	// Param1
 	}
+	
+#if MAX_ITEM_RDM_OPT < 5
+	for ( ; i < 5; i++) {
+		WBUFW(buf, i*5 + 0) = 0;		// OptIndex
+		WBUFW(buf, i*5 + 2) = 0;	// Value
+		WBUFB(buf, i*5 + 4) = 0;	// Param1
+	}
+#endif
```
anacondaqq: two `#endif?`

secretdataz: Another `#endif` is for the `#if PACKETVER >` block.

PR #1584, opened 2016-10-02
Removed redundant buildin_warpchar function

secretdataz:
* the warpchar script command is still usable but not recommended
* removed warpchar documentation from script_commands.txt to let it fade from people's memory
* `warp` script command now accepts char id as an optional parameter. `warp "<map name>",<x>,<y>{,<char id>};`

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3836,9 +3836,9 @@ will be seen to everyone around but not in chat window.
 
 ---------------------------------------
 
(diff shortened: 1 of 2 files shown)


anacondaqq:
why are you scaring me and have not added to comments, that "warp" command was a little bit improved:
> warp "<map name>",<x>,<y>{,<char id>};

secretdataz:
@anacondaqq sorry for freaking you out <3
added the comment

PR #1595, opened 2016-10-07
Fixed #1594

### LINKED ISSUE #1594: Compile Issues from Party Window Cleanup
This specific line is causing warnings when compiling:
https://github.com/rathena/rathena/commit/9a52768f6fb35637edac233f85d7a857f852376e#diff-324c3750c88c4e33364ba5f770e9fc5dR724

```
        CC      duel.c
In file included from duel.c:8:0:
clif.h:771:134: warning: ‘enum e_party_member_withdraw’ declared inside parameter list [enabled by default]
 void clif_party_withdraw(struct map_session_data *sd, uint32 account_id, const char* name, enum e_party_member_withdraw result, enum send_target target);
                                                                                                                                      ^
clif.h:771:134: warning: its scope is only this definition or declaration, which is probably not what you want [enabled by default]
```

Because duel.c doesn't import mmo.h, and is a very sad confused c file ):

It also happens in instance.c, trade.c and vending.c.

cydh:
Follow up 9a52768f6fb35637edac233f85d7a857f852376e
* Fixed #1594, compile warning

### DIFF
--- a/src/map/clif.h
+++ b/src/map/clif.h
@@ -30,6 +30,7 @@ struct guild;
 struct battleground_data;
 struct quest;
 struct party_booking_ad_info;
+enum e_party_member_withdraw;
 #include <stdarg.h>
 
 enum { // packet DB


cydh:
because I'm using MSVS only, just need to test the result on travis first.

PR #1616, opened 2016-10-12
Added documentation for item links in mes

### LINKED ISSUE #1614: Missing Item Links in script_commands.txt
Yesterday I was reading the script_commands.txt to try to find the Item Links documentation and I noticed that we don't have any description for that yet.

Can you update the description to cover Item Links too!? Here is a post that I found that pretty much explains how it works and which syntax must be used according to the client version.

https://rathena.org/board/topic/105065-calling-item-description-in-script/

Thank you very much!

Lemongrass3110:
Fixes #1614

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -1094,6 +1094,27 @@ to alberta (98,154) when clicked.
 
 See also 'navigateto', which can be used for certain NPC events.
 
+Items
+-----
+You can refer to items by using HTML-like links to certain items:
+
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
It would be nice if we could get the exact client dates when this feature was added and when it was changed/broken in 2015.

secretdataz:
It was broken in 2015-11-04. I'll try to unpack clients after that date to see if it's fixed shortly after that. However, I'm sure the feature works fine in 2016 clients

PR #1617, opened 2016-10-13
Updated Manhole skill behavior

### LINKED ISSUE #1601: Manhole & Emperium
The emperium and guardian stones at bg are target-able by Manhole making impossible to kill

aleos89:
* Fixes #1601
* GvG and Battleground monster objects are not able to be targeted by Manhole.
* Removed the usage of Strip Accessory on targets who are affected by Manhole.
Thanks to @ignizh!

### DIFF
--- a/src/map/status.c
+++ b/src/map/status.c
@@ -2014,6 +2014,10 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
 				&& (src->type != BL_PC || ((TBL_PC*)src)->skillitem != skill_id))
 				return false;
 			break;
+		case SC_MANHOLE:
+			// Skill is disabled against special racial grouped monsters(GvG and Battleground)
+			if (status_get_race2(target) == RC2_GVG || status_get_race2(target) == RC2_BATTLEFIELD)
+				return false;
 		default:
 			break;
 	}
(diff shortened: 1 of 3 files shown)


src/map/status.c
```diff
@@ -2188,6 +2188,14 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
 				if( tsc->option&hide_flag && !status_has_mode(status,MD_DETECTOR))
 					return false;
 			}
+			if (status_get_race2(target) == RC2_GVG || status_get_race2(target) == RC2_BATTLEFIELD) { // Skills disabled in GvG and Battleground against special racial grouped monsters
```
Lemongrass3110: The comment sounds a little misleading, since it is making it sound like it would be mapflag related, but this is a race only condition.
Also would it no be better to move it up to ~1996 where we have a switch by skill id already and just check the target race there?

alxalmora:
Sorry for write here, but I think I found and Issue about this. With these diff (and of course adding the miss `if(target && ...)` you can`t cast manhole in Emperium or BG Crystals but if you cast in the cell next-to, they still get in Manhole, I mean I can't see the visual effect but I can't attack the Crystal and Players doesn`t seem to be affected by the Manhole (but visual is an active manhole). I put this "issue" here because, maybe was fixed. I apply manual diffs since RndOpt (that crash my server and I make a revert, then I make manual commits only for SRC skills bugs).

PR #1618, opened 2016-10-13
Updated Ranger Traps behavior

### LINKED ISSUE #1611: Ranger Traps
Hash: f51a1c7
Mode: Both

The Ranger traps (Magenta, Cobalt, Maize, and Verdure) should all change the element of the target permanently (until death).

Source: kRO

aleos89:
* Fixes #1611.
* Magenta, Cobalt, Maize, and Verdure Traps should permanently change the element of the target.


src/map/skill.c
```diff
@@ -17279,8 +17279,14 @@ static int skill_trap_splash(struct block_list *bl, va_list ap)
 		case UNT_COBALTTRAP:
 		case UNT_MAIZETRAP:
 		case UNT_VERDURETRAP:
-			if( bl->type != BL_PC && status_get_class_(bl) != CLASS_BOSS )
-				sc_start2(ss,bl,SC_ELEMENTALCHANGE,100,sg->skill_lv,skill_get_ele(sg->skill_id,sg->skill_lv),skill_get_time2(sg->skill_id,sg->skill_lv));
+			if( bl->type != BL_PC && status_get_class_(bl) != CLASS_BOSS ) {
+				struct status_data *status = status_get_status_data(bl);
+
+				if (status) {
```
Lemongrass3110: can never be null. it will at least return dummy_status.

PR #1632, opened 2016-10-18
Added a configure option for vip features

Lemongrass3110:
Added the new option to the CI matrix


cydh:
When I added pre-re and re matrix to Travis, I felt guilty. 4 builds every commits, and now it becomes 8 builds for free service.
Give them some donations maybe :S

PR #1640, opened 2016-10-19
Fixed bonus bAddEff success rate

### LINKED ISSUE #1638: AddEff Bonus
Hash: b2f0ffb
Mode: Both

The AddEff bonus is not behaving as it does on Aegis. The effect should still trigger even if the target has more resistance.

rAthena Behavior:
- Checks if you have an AddEff bonus.
- If you do, call `status_change_start(..., rate,...);`
- This never happens because the rate gets reduced to a negative number due to resistances, so the bAddEff has a 0% chance of success.

Aegis Behavior:
- Checks if you have an AddEff bonus.
- Does a `(rnd() % 10000 >= rate) ? continue : else`...
- Call `status_change_start(..., 10000,...);`

Thanks to @Tokeiburu for confirming!

aleos89:
- Fixes #1638.
- Adjusted bonus bAddEff to match Aegis' effect.
- Success rate is not applied to the SC but a separate random check.
- This fixes statuses that were being resisted when they shouldn't have been.
  Thanks to @Tokeiburu!

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -1019,7 +1019,7 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
 				rate = sd->addeff[i].rate;
 				if( attack_type&BF_LONG ) // Any ranged physical attack takes status arrows into account (Grimtooth...) [DracoRPG]
 					rate += sd->addeff[i].arrow_rate;
-				if( !rate )
+				if( rnd() % 10000 >= rate )
 					continue;
 
 				if( (sd->addeff[i].flag&(ATF_WEAPON|ATF_MAGIC|ATF_MISC)) != (ATF_WEAPON|ATF_MAGIC|ATF_MISC) ) {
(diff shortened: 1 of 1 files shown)


cydh:
it check the rate separately, what I guess, the SCSTART flag must be changed too, that ignore rate adjustment

src/map/skill.c
```diff
@@ -1064,9 +1064,9 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
 				time = sd->addeff[i].duration;
 
 				if( sd->addeff_onskill[i].target&ATF_TARGET )
-					status_change_start(src,bl,type,10000,7,0,0,0,time,SCSTART_NONE);
+					status_change_start(src,bl,type,10000,7,0,0,0,time,SCSTART_NORATEDEF);
 				if( sd->addeff_onskill[i].target&ATF_SELF )
-					status_change_start(src,src,type,10000,7,0,0,0,time,SCSTART_NONE);
+					status_change_start(src,src,type,10000,7,0,0,0,time,SCSTART_NORATEDEF);
```
cydh: I said "I guess" not "I suggest" D:

Issue #1073, opened 2016-03-19
Emergency Escape Animation


stealth62:
im using latest revision renewal mode 20130807 client i notice emergency escape copycat animation.

stealth62:
I share my modification
in skill.c
static int skill_timerskill(int tid, unsigned int tick, int id, intptr_t data)

find this line
	case CH_PALMSTRIKE:
					{
						struct status_change* tsc = status_get_sc(target);
						struct status_change* sc = status_get_sc(src);
						if( ( tsc && tsc->option&OPTION_HIDE ) ||
							( sc && sc->option&OPTION_HIDE ) ){
							skill_blown(src,target,skill_get_blewcount(skl->skill_id, skl->skill_lv), -1, 0);
							break;
						}
						skill_attack(skl->type,src,src,target,skl->skill_id,skl->skill_lv,tick,skl->flag);
						break;
					}

then add this line
case SC_ESCAPE:
					if (skl->type < 4 + skl->skill_lv){
						clif_skill_damage(src, src, tick, 0, 0, -30000, 1, skl->skill_id, skl->skill_lv, 0);
						skill_blown(src, src, 1, unit_getdir(src), 0);
						skill_addtimerskill(src, tick + 100, src->id, 0, 0, skl->skill_id, skl->skill_lv, skl->type + 1, 0);
					}
					break;

again in skill.c
int skill_castend_pos2(struct block_list* src, int x, int y, uint16 skill_id, uint16 skill_lv, unsigned int tick, int flag)

find this line
	case SC_ESCAPE:
		skill_unitsetting(src, skill_id, skill_lv, x, y, 0);
		skill_blown(src, src, skill_get_blewcount(skill_id, skill_lv), unit_getdir(src), 0);
		clif_skill_nodamage(src,src,skill_id,skill_lv,1);
		flag |= 1;
		break;

replace to this line
case SC_ESCAPE:
		clif_skill_nodamage(src, src, skill_id, -1, 1);
		skill_unitsetting(src, skill_id, skill_lv, x, y, 2);
		skill_addtimerskill(src, tick, src->id, 0, 0, skill_id, skill_lv, 0, 0);
		flag |= 1;
		break;

this will correct emergency escape of shadow chaser

Lemongrass3110:
Hi @stealth62.
First of all thanks for reporting and commenting a lot of issues lately. But please try to read https://guides.github.com/features/mastering-markdown/ and use it for putting code parts here on github.
[...]

PR #1669, opened 2016-10-29
Shadow Chaser Skill Escape

Akkarinage:
Applied partial fix for #1073 using code in same Issue. Staggers the movement correctly but animation of "shadows" left over by the character aren't evenly spread across the 10 cells during movement when they should be. Tested using 20151104 client.

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -4106,6 +4106,13 @@ static int skill_timerskill(int tid, unsigned int tick, int id, intptr_t data)
 						skill_attack(skl->type,src,src,target,skl->skill_id,skl->skill_lv,tick,skl->flag);
 						break;
 					}
+				case SC_ESCAPE:
+					if (skl->type < 4 + skl->skill_lv) {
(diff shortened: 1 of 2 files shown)


src/map/skill.c
```diff
@@ -4106,6 +4106,13 @@ static int skill_timerskill(int tid, unsigned int tick, int id, intptr_t data)
 						skill_attack(skl->type,src,src,target,skl->skill_id,skl->skill_lv,tick,skl->flag);
 						break;
 					}
+				case SC_ESCAPE:
+					if (skl->type < 4 + skl->skill_lv) {
```
Lemongrass3110: Wouldn't it be nicer to read that from the skill_db column you removed the entries from instead of hardcoding it?

aleos89: I agree. This should use the values that were removed from the skill_db.

secretdataz:
~~Stolen~~ Re-pushed in #1771.

PR #1670, opened 2016-10-29
Updated default PACKETVER

Akkarinage:
It makes no sense to use 20130807 as the default PACKETVER when we use 55 (20151104) as the default packet_ver in /db/packet_db.txt

Signed-off-by: Akkarinage akkarin@rathena.org

### DIFF
--- a/conf/battle/feature.conf
+++ b/conf/battle/feature.conf
@@ -23,7 +23,7 @@ feature.atcommand_suggestions: off
 // Feature became unstable on clients 2012 onwards (exact date not known),
 // it has been fixed on clients 2013-05-15 onwards however.
 // Feature was removed again on clients 2014-11-12 and later.
-feature.auction: on
+feature.auction: off
 
 // Warp suggestions (Note 1)
 // Show suggestions when attempting to @warp to a non-existent map?
--- a/src/common/mmo.h
+++ b/src/common/mmo.h
@@ -21,7 +21,7 @@
 // see conf/battle/client.conf for other version
 
 #ifndef PACKETVER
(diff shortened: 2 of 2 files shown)


CairoLee:
New mail system(RODEX) and the

CairoLee:
RODEX New Mail System Feature
Clan System Feature
Pet Evolution System Feature
Achievement System (New Extra Quest) Feature
Title System Feature

they are not implemented

Akkarinage:
We fully support the client and it doesn't (afaik) break anything we currently have, which is why we already have default_packet_ver set for this client. Setting PACKETVER to this client data doesn't mean a great deal, as most users change it anyway.

Some of the above features would make their way into the master branch quicker if people bothered to test the PRs that are available.

cydh:
The part that must be change is `packet_db_ver: default` in packet_db instead change the PACKETVER.
Say, if follow every latest packet supported, PACKETVER must be changed.
Well, it's not significant for me, I put many local gitignores for some files. xD

CairoLee:
We can let the packet_db_ver support auto mode.
if set packet_db_ver to auto, then map-server will read PACKETVER and set the match packet_db_ver value.

Sorry for my bad english XD

aleos89:
I see no problem updating the PACKETVER to 2015+. The clients are stable for play which is the main point. Just because all of the features of the client aren't fully supported by rAthena doesn't seem (to me at least) to be a reason to hold it back from updating it. If we hold to this standard, then that means we shouldn't be on 2013 since we don't support the BG Queue.

PR #1692, opened 2016-11-04
Updated Rebellion class

### LINKED ISSUE #859: rebellion fallen angel animation
the problem is the rebellion jumps while casting, then he falls to early into the ground even though he's not done casting the skill, and when he's done he'll jump once again then fall onto the target.  its too obvious when ur rebellion has 1 dex (slow casting time)

correct me if im wrong, the animation should be like this

-cast then skill and jump , after done casting then falls onto the target

### LINKED ISSUE #1449: Rebellion skill Binding trap does 1 damage when using  Flicker
I think the title pretty much explains it. I'm using rAthena latest version. 

Steps to reproduce: Use binding trap, someone else steps on it, use Flicker and it does 1 HP damage. Also it slows down the caster (but i don't know if this is the correct behavior or not) and it seems like you can't cast on the same cell of someone else or yourself.

aleos89:
* Information based on Rebellion Awakening patch from kRO.
  - http://ro.gnjoy.com/news/update/View.asp?seq=183&curpage=1
* Fixes #859 and fixes #1449.

TODO:
- [ ] New SP values.
- [x] New EXP values. (new max level is 175 see #2267)
- [x] Confirm Slug Shot damage modifiers.
- Confirm Bind Trap:
  - [x] damage
  - [ ] move speed rate
  - [ ] duration reduction
- [ ] Confirm Shatter Storm equipment break chance.
- [x] Implement new skill fail message for Platinum Alter.
- [x] Implement official job change quest.

Thanks to @RagnarokNova and Fyrus for their information!


cydh:
> * Information based on Rebellion Awakening patch from kRO.
    * http://ro.gnjoy.com/news/update/View.asp?seq=183&curpage=1

just like my previous comment, need the English or Engrish ver, so I can test this PR

Felleonel:
If I'm not mistaken, as well as the "Slug Shot's Ammunitions" is no longer treated as ammunition, now Dragon Tail Missile and  Projection Landmines aren't ammunition too, but are consumed  while using their specifics skills.

Heat Barrel increases attack, I don't know if this was removed from the skill in the new upgrade, but in this video, apparently the damage is increased: 

https://www.youtube.com/watch?v=ty8h1IC7Bco

Sorry for my bad english.

aleos89:
@Felleonel: Those changes have already been accounted for.

db/re/skill_db.txt
```diff
@@ -1267,24 +1267,24 @@
 //2551,0,0,0,0,0,0,10,0,no,0,0,0,none,0,0,	RL_GLITTERING_GREED,Flip The Coin Greed
 2552,0,6,4,0,0x1,0,1,1,no,0,0,0,none,0,0x0,	RL_RICHS_COIN,Rich's Coin
 2553,-9,6,1,-1,0,0,5,1,no,0,0,0,weapon,0,0x0,	RL_MASS_SPIRAL,Mass Spiral
-2554,-9,6,1,-1,0x2,2,5,1,no,0,0,0,weapon,0,0x10020,	RL_BANISHING_BUSTER,Banishing Buster
+2554,-9,6,1,-1,0,0,5,1,no,0,0,0,weapon,0,0x10020,	RL_BANISHING_BUSTER,Banishing Buster
 2555,3:4:4:5:5,6,2,0,0x1,1,5,1,no,0,0,1,misc,0,0x0,	RL_B_TRAP,Bind Trap
```
Darkelfen: I guess that 15% of HP + Dex should be 0x81 instead, so that the damage is applied regardless of target's def cards, can someone confirm in kRo if def cards affects this skill?

aleos89: Just confirmed that Bind Trap does not ignore target's DEF cards.

Darkelfen: So Thara Frog, etc is applied then, ty.

cydh:
@aleos89 about new bullets, u didn't add those yet?

db/re/skill_db.txt
```diff
@@ -729,7 +729,7 @@
 504,0,6,4,0,0x1,0,1,1,no,0,0,0,weapon,0,0x0,		GS_MADNESSCANCEL,Madness Canceller
 505,0,6,4,0,0x1,0,1,1,no,0,0,0,weapon,0,0x0,		GS_ADJUSTMENT,AdJustment
 506,0,6,4,0,0x1,0,1,1,no,0,0,0,weapon,0,0x0,		GS_INCREASING,Increasing Accuracy
-507,-9,6,1,8,0x20,0,1,1,no,0,0,0,misc,0,0x0,		GS_MAGICALBULLET,Magical Bullet
+507,0,6,4,0,0x1,0,1,1,no,0,0,0,magic,0,0x0,		GS_MAGICALBULLET,Magical Bullet
```
cydh: 0 range?

src/map/battle.c
```diff
@@ -3304,6 +3321,9 @@ static struct Damage battle_calc_multi_attack(struct Damage wd, struct block_lis
 			if( tsc && (tsc->data[SC_BITE] || tsc->data[SC_ANKLE] || tsc->data[SC_ELECTRICSHOCKER]) )
 				wd.div_ = tstatus->size + 2 + ( (rnd()%100 < 50-tstatus->size*10) ? 1 : 0 );
 			break;
+		case RL_QD_SHOT:
+			wd.div_ = 1 + (sd ? sd->status.job_level : 1) / 20 + (tsc && tsc->data[SC_C_MARKER] ? 2 : 0);
```
cydh: hmm interesting

cydh:
Slugs are not removed after using Slug Shot.

db/re/skill_db.txt
aleos89: It's a self cast skill now.

LunarSM:
Quick Draw Shot damages self instead of mobs

aleos89:
@LunarSM 
5 Coins:
image

4 Coins remaining after Fallen Angel:
image

LunarSM:
Yeah saw, i forgot to update that.

But something is weird...
Quick Draw Shot damages me instead of mobs.

I compared and i can`t see that i copied wrong somewhere :(

aleos89:
Quick Draw Shot should damage yourself as well as enemies.

rizz2016:
Are Platinum Alter & Heat Barrel ATK bonus nerfed in latest update? I have tested and I think they get significant change, especially Heat Barrel

src/map/battle.c
```diff
@@ -1407,6 +1407,17 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 				status_change_end(bl, SC_KYRIE, INVALID_TIMER);
 		}
 
+		if ((sce = sc->data[SC_P_ALTER]) && damage > 0) {
+			clif_specialeffect(bl, 336, AREA);
+			sce->val3 -= (int)cap_value(damage, INT_MIN, INT_MAX);
```
Jeybla: Maybe `sce->val3 = (int)cap_value(sce->val3 - damage, INT_MIN, INT_MAX);`
so the followed if-else isn't needed?

ZakosDryiakk:
After a bit of looking into, checked that Heat Barrel still applies closer to the 16% atk bonus **per coin**, so 160% max (just search for "Ragnarok レベリオン" or "리벨리온" on youtube and you'll get to check that)

erosama28:
heat barrel formula is also wrong i tried to match it with nova and kRO the closest i get is using the code

`		if (sc->data[SC_HEAT_BARREL]) {
			ATK_ADDRATE(wd->equipAtk, wd->equipAtk2, sc->data[SC_HEAT_BARREL]->val3);
			ATK_RATE(wd->weaponAtk, wd->weaponAtk2, sc->data[SC_HEAT_BARREL]->val3);
`
also missing minimum coin check for platinum alter and heat barrel

`		case SC_HEAT_BARREL:
		case LG_RAGEBURST:
		case SC_P_ALTER:
			if( sd->spiritball == 0 ) {
				clif_skill_fail(sd,skill_id,USESKILL_FAIL_SKILLINTERVAL,0);
				return false;
			}
			sd->spiritball_old = require.spiritball = sd->spiritball;
			break;`

so it has minimum requirement.

PR #1703, opened 2016-11-08
Mora Enchants: Arch Bishop Rings should be exchangeable by ABs since …

Jeybla:
…they can only bought by them.

This is a suggestion! It's not possible to exchange the rings currently.

http://irowiki.org/wiki/Mora_Enchants

> Only an Arch Bishop can trade in Mora Coins for the gear with the Keeper of Secrets. Enchanting with the Master of Relics has a chance of breaking the gear.

> The Master of Relics is located near the middle of town by the lake (mora 96, 74). Any of the gears may be purchased from the Keeper of Secrets (88, 89) at a cost of 10 Mora Coins each; however, the player is only allowed to select which equipment type they desire (shoes, armor, weapon, accessory), and they are given one of the 2 or 3 possibilities at random. However, you are not allowed to purchase an accessory when you already have one or more in your inventory. Also, it's possible to trade any of the three accessories in to obtain 10 Mora coins, to do so choose the third option from the Keeper of Secrets. 

Anyway even with this change it sounds strange to me. Why do you get a random accessory, when it's possible to exchange them for the same buying price?

(needs official information)

### DIFF
--- a/npc/re/merchants/enchan_mora.txt
+++ b/npc/re/merchants/enchan_mora.txt
@@ -106,15 +106,15 @@ mora,88,89,5	script	Keeper of Secrets#pa082	495,{
 		mes "You have obtained ^aa00aa"+getitemname(.@item)+"^000000. May the Blessing of the Light be with you.";
 		close;
 	case 3:
-		if (Class == Job_Arch_Bishop || Class == Job_Arch_Bishop_T || Class == Job_Baby_Bishop) {
+		if (Class != Job_Arch_Bishop && Class != Job_Arch_Bishop_T && Class != Job_Baby_Bishop) {
 			mes "[Artifice]";
-			mes "We exchange rings into coins for those who cannot control the ring.";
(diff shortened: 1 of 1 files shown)


Darkelfen:
If I recall correctly, the way in which mora equipment is obtained got changed a while ago:

http://forums.irowiki.org/index.php?topic=107149.0

Nevertheless I couldn't find more intel on this.

Jeybla:
> - The Mora accessories for Warlock and Archbishop are now dropped by monsters, as with other jobs

If that's the case exchanging these items by non AB classes would make sense and this PR should not be merged.

Darkelfen:
As far as I know the whole system got revamped, but the specs regarding those changes are alien to me.

Atemo:
According to ours datas (and tested on iro) only non-AB classes can exchange AB rings to mora coins (so we won't change it). We will wait for more infos about the drop/exchange update.

PR #1714, opened 2016-11-14
Added windows specific socket timeout code

### LINKED ISSUE #794: socket timeout options on WinAPI use DWORD, not struct timeval
In https://github.com/rathena/rathena/commit/f67f82c472585bb5fdb4860deedf32cccb508f98 a cascade of 'timeout' parameters was added, eventually ending at setsocketopts:

```
    if(delay_timeout){
        struct timeval timeout;
        timeout.tv_sec = delay_timeout;
        timeout.tv_usec = 0;

        if (sSetsockopt (fd, SOL_SOCKET, SO_RCVTIMEO, (char *)&timeout,sizeof(timeout)) < 0)
            ShowError("setsocketopts: Unable to set SO_RCVTIMEO timeout for connection #%d!\n");
        if (sSetsockopt (fd, SOL_SOCKET, SO_SNDTIMEO, (char *)&timeout,sizeof(timeout)) < 0)
            ShowError("setsocketopts: Unable to set SO_SNDTIMEO timeout for connection #%d!\n");
    }
```

According to MSDN docs:

```
SO_RCVTIMEO DWORD   Sets the timeout, in milliseconds, for blocking receive calls. 
SO_SNDTIMEO DWORD   The timeout, in milliseconds, for blocking send calls. 
```

So there is a platform incompatibility, which results in a parameter type mismatch. This is the cause why in https://github.com/rathena/rathena/issues/793 I saw a 21-second delay.

Lemongrass3110:
Added socket code for windows that respects our given timeout.
Also added some more ifndef checks to some stat reporting functions.
Updated the MSDN reference link.
Fixed the delay being milliseconds on windows, thanks to @theultramage
Fixes #794 on windows.

### DIFF
--- a/conf/inter_athena.conf
+++ b/conf/inter_athena.conf
@@ -93,7 +93,6 @@ scdata_db: sc_data
 cart_db: cart_inventory
 inventory_db: inventory
 charlog_db: charlog
-storage_db: storage
 skill_db: skill
 interlog_db: interlog
 memo_db: memo
@@ -149,5 +148,7 @@ db_roulette_table: db_roulette
 
 // Use SQL item_db, mob_db and mob_skill_db for the map server? (yes/no)
 use_sql_db: no
+
+inter_server_conf: conf/inter_server.conf
 
 import: conf/import/inter_conf.txt
--- a/conf/inter_server.conf
+++ b/conf/inter_server.conf
@@ -1,0 +1,26 @@
+/**
+ * Config for Storages
+ *
+ * To access the premium storage, use script command 'openstorage2'.
+ * If premium storages are added, copy the structure of the storage table and match the table name in this config.
+ * The 'max' of premium storages are not adjusted by 'vip_storage_increase' config nor MIN_STORAGE.
+ *
+ * Structure:
+{
+	id: <storage_id>          // (int) Storage ID will be used for script command 'openstorage2'.
+	name: "<storage name>"    // (string) Storage name will be sent to the client to display on the title bar.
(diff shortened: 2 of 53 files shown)


theultramage:
Ouch, that windows case looks as painful as I imagined. You could help it a bit by flattening the if-return-else error handling code (remove 'else' scope and unindent, because the 'if' scope ends with a 'return').

theultramage:
https://gist.github.com/theultramage/e22b0013407bf85e0ab2cd755ceb1973 is what I had in mind in my previous post. You chose to do a fake while() scope and break; out of it in the positive case, and folded the three error cases into one block of code. That's quite rough, I prefer to stick with the duplicated code for more clarity. I do remember that we used to do do { ... } while( 0 ); to indicate dummy scope. Just some extra thoughts.

Lemongrass3110:
Yeah I understood what you were getting at. That was just my solution of removing the duplicate code, since I personally prefer not duplicating code if it is useless. If it would not be fitting onto one screen I would possibly also duplicate the code but like this I think it's quite useless.
I also though about the dummy scope you mentioned, but I thought that I could just use the first if as while entry condition to get the same effect and without having to add one level of identitation for the pseudo scope again.

theultramage:
Okay, I see. I was just worried about the while() thing, because I had to read the whole code listing twice to realize it's not intended to loop and just forms a dummy breakable scope.

Lemongrass3110:
Yeah you are right it might be misleading. At least I should (have) put a comment above it to make it clear for anyone reading the code.

PR #1715, opened 2016-11-15
Update name for all party members on rename

mkbu95:
Currently you have to restart the server for the changed name in the party list to update.


src/char/int_party.c
```diff
@@ -889,3 +889,29 @@ int inter_party_CharOffline(uint32 char_id, int party_id) {
...
+		ShowError("inter_party_charname_changed: Can't find character %d in party %d.\n", char_id, party_id);
+		return 0;
+	}
+
+	safestrncpy(p->party.member[i].name, name, NAME_LENGTH);
+
+	mapif_party_info(-1, &p->party, char_id);
+	
+	return 0;
+}
```
Lemongrass3110: You lost the newline at the end of the file.

PR #1720, opened 2016-11-16
Added an additional parameter to instance_enter

Lemongrass3110:
This allows easier usage for instances in IM_NONE, where a player will never be attached to the instance.
All you have to do is remember the created instance id and supply this command with it. That way you do not have to use any workaround like the following anymore:
```
warp instance_mapname( <mapname>, <instance id> ), <hardcoded enter x coordinate>, <hardcoded enter y coordinate>;
```

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8276,7 +8276,7 @@ that fails, the script will come to a halt.
 
 ---------------------------------------
 
-*instance_enter("<instance name>",{<x>,<y>,<char_id>});
+*instance_enter("<instance name>",{<x>,<y>,<char_id>,<instance id>});
 
 Warps player to the specified instance after the script terminates. The map and
 coordinates are located in 'db/(pre-)re/instance_db.txt'.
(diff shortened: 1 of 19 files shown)


doc/script_commands.txt
```diff
@@ -8288,6 +8288,9 @@ The command returns 0 upon success, and these values upon failure:
 
 Put -1 for x and y if want to warp player with default entrance coordinates.
 
+Note:
+Instance name is a legacy parameter and will be ignored internally.
```
secretdataz: Instance name is currently used for retrieving `instance_db` struct pointer.

Lemongrass3110: Woops you're right. Maybe I'll clean this up in a follow up too.
That could be done by `instance_searchtype_db(instance_data[instance_id]->type)` so these parameters really could be ignored for real in the future.

Lemongrass3110:
@Atemo could you have a look at these:

- hazy forest:
	default case
- octopus cave:
	default case
- buwaya cave:
	default case
- eclage interior:
	case 1(IE_NOMEMBER) is missing
	default case

Those default cases were possibly converted from aegis and actually should have some meaning. Right now they are unreachable.

PR #1730, opened 2016-11-19
Replaced hardcoded genders with constants

Lemongrass3110:
Based on @Emistry's idea from HerculesWS/Hercules@05a2538

### DIFF
--- a/npc/quests/skills/swordman_skills.txt
+++ b/npc/quests/skills/swordman_skills.txt
@@ -98,7 +98,7 @@ OnTouch:
 		mes "Being a swordsman must come with a lot of responsibility and sacrifice.";
 		next;
 		mes "[De Thomas]";
-		if (Sex == 0)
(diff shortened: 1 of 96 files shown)


npc/quests/skills/swordman_skills.txt
```diff
@@ -98,7 +98,7 @@ OnTouch:
 		mes "Being a swordsman must come with a lot of responsibility and sacrifice.";
 		next;
 		mes "[De Thomas]";
-		if (Sex == 0)
+		if (Sex == SEX_FEMALE)
 			mes "For these swordsmen and knights, there is a wonderful skill.";
 		else
 			mes "For these swordswomen and knights, there is a wonderful skill young lady.";
```
secretdataz: I think these dialogues are swapped

aleos89: Agreed, this check needs to be flipped.

PR #1739, opened 2016-11-22
Item DB update and script command adjustment

secretdataz:
* Made getequipid script command accept -1 for current equipment slot when running in item script.
* Updated renewal item database.


cydh:
that `-1` for
> When compounded on Royal Guard Ring, Atk + 5% and Matk + 5%.

even it works, it's kinda.... suck
maybe new EQI enum for `EQI_COMPOUND_ON`? or maybe new command `getcompoundid` /hmm

Lemongrass3110:
How about making the slot argument for getequipid optional?
Like how we have getrefine(btw when I just looked into getrefine, that looks dangerous if it is called from outside of an item script xD

PR #1741, opened 2016-11-22
Cleaned up Item Groups

aleos89:
* Moved constants from const.txt to script_constants.h.
* Removes hard coded item groups into the IG_* database to easily be adjusted/created.
* Renamed IG_FingingOre -> IG_FindingOre.


src/map/itemdb.c
```diff
@@ -40,6 +40,27 @@ struct s_item_group_db *itemdb_group_exists(unsigned short group_id) {
...
 
 /**
+ * Check if an item exists in a group
+ * @param nameid: Item to check for in group
+ * @return True if item is in group, else false
+ */
+bool itemdb_group_item_exists(unsigned short group_id, unsigned short nameid)
+{
+	struct s_item_group_db *group = (struct s_item_group_db *)uidb_get(itemdb_group, group_id);
+	uint8 i, j;
```
cydh: the `j` must be `unsigned short`

aleos89: Hah! I was literally looking at that yesterday too and thought the same thing.

PR #1747, opened 2016-11-25
correct bad npc name format in buwaya cave instance

CitrusAroma:
IMO. The first char of npc name should not be put in with '#' .

### DIFF
--- a/npc/re/instances/BuwayaCave.txt
+++ b/npc/re/instances/BuwayaCave.txt
@@ -156,20 +156,20 @@ OnTouch:
 
 // Instance Scripts :: buwaya_in
 //============================================================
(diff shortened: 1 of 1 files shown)


secretdataz:
NPC names with `#` as prefix are meant to be hidden.
Your PR is meaningless since every NPC you edited its name has 139 (`HIDDEN_WARP_NPC`) as sprite ID.

secretdataz:
This pull request has been rejected.

CitrusAroma:
Every NPC has  its hidden sprite ID why we need '#' ?

secretdataz:
It was converted from the official script and kept there.

PR #1767, opened 2016-12-05
Refactored hardcoded npc event names

Lemongrass3110:
While refactoring those events the following other changes were made:

Introducing OnInstanceDestroy event. This event can be used to hook script code right in front of the instance destruction. This can be useful if you have any stored references to an instance id for example.

The script command cmdothernpc will now check if the target event exists and report failures. Therefore it now returns true or false.

All agit(FE,SE,TE) start and end commands from atcommand and script commands have been merged in their respective guild function which now returns a bool value of true for successful actions and false if it did not succeed(if the specific WoE was [not] running).

All global triggered events with mapserver status output now call the same function and therefore have the same mapserver output(including their respective event name of course).

### DIFF
--- a/src/map/instance.c
+++ b/src/map/instance.c
@@ -291,6 +291,20 @@ static int instance_npcinit(struct block_list *bl, va_list ap)
 }
 
 /*==========================================
+ * Run the OnInstanceDestroy events for duplicated NPCs
+ *------------------------------------------*/
+static int instance_npcdestroy(struct block_list *bl, va_list ap)
+{
+	struct npc_data* nd;
+
+	nullpo_retr(0, bl);
+	nullpo_retr(0, ap);
+	nullpo_retr(0, nd = (struct npc_data *)bl);
+
+	return npc_instancedestroy(nd);
+}
+
+/*==========================================
  * Add an NPC to an instance
  *------------------------------------------*/
 static int instance_addnpc_sub(struct block_list *bl, va_list ap)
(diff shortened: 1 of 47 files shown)


Atemo:
__OnInstanceDestroy__ docs is missing !

src/map/instance.c
```diff
@@ -291,6 +291,20 @@ static int instance_npcinit(struct block_list *bl, va_list ap)
...
 
 /*==========================================
+ * Run the OnInstanceDestroy events for duplicated NPCs
+ *------------------------------------------*/
+static int instance_npcdestroy(struct block_list *bl, va_list ap)
+{
+	struct npc_data* nd;
+
+	nullpo_retr(0, bl);
+	nullpo_retr(0, ap);
```
lighta: why checking ap, it doesn't seem required atm

src/map/npc.c
```diff
@@ -3256,7 +3267,20 @@ int npc_instanceinit(struct npc_data* nd)
...
+		run_script(nd->u.scr.script,ev->pos,0,nd->bl.id);
+
+	return 0;
+}
+
+int npc_instancedestroy(struct npc_data* nd)
+{
+	struct event_data *ev;
+	char evname[EVENT_NAME_LENGTH];
+
```
lighta: I wonder if a nullpo on nd wouldn't be safer here. Or it may be totally useless..

Lemongrass3110:
Those two functions are copies of the instance init ones. I did not change any logic at all. ;P
Of course we could adjust it, but if we do, we should do it for both cases.

PR #1771, opened 2016-12-07
Fixed SC_ESCAPE animation

secretdataz:
This PR fixes issue #1073.

> Staggers the movement correctly but animation of "shadows" left over by the character aren't evenly spread across the 10 cells during movement when they should be. Tested using 20151104 client.


secretdataz:
Closing this since we confirmed the changed behavior in this PR as incorrect.

PR #1800, opened 2016-12-15
Unit command consistency changes

Locien:
Made it so that all unit commands(except for unitexists) take 0 as the unit ID argument which will reference the activator of the script. Unitwarp has since previously had this functionality and it's very useful. Example use: When a monster walks over an npc script with the OnTouchNPC label unitkill 0; will kill the monster.


Locien:
Redoing this once I've actually got the hold of git and GitTortoise. Sorry I've never been part of a git project before.

PR #1803, opened 2016-12-16
Added party config `change_party_leader_samemap`

cydh:
* Prevent to change party leader if not in same map with the specified player (Note 1)
* Default is `yes` as official does
* This config will be ignored by `party_changeleader` script command


Lemongrass3110:
Squash and merge please :)

PR #1813, opened 2016-12-19
Horror toy factory

Atemo:
TODO:
- [ ] Implement NPC_HELLBURNING


db/re/skill_cast_db.txt
```diff
@@ -1042,6 +1042,9 @@
 698,0,20000,0,0,0,0,-1
 //==========================================
 
+//===== New Monster Skills (12.1) ==========
```
secretdataz: I'm sorry >_>
should be ep 15.?? instead

PR #1827, opened 2016-12-25
Made prevent_logout effect on log-in optional

secretdataz:
kRO does not prevent players from logging out after connecting to its zone servers, so I adjusted this behavior on rAthena to follow kRO.

There is a new battle config called `prevent_logout_on_login` to opt-in this behavior again.


anacondaqq:
Looks nice now =) You are great!

PR #1831, opened 2016-12-26
Clean-up permanent monster spawn:

Atemo:
- ```<delay1>,<delay2>,<event>``` are actually optionnal field
- ```<event>``` value 0 and 1 are meaningless (actually event name with length < 4 are skipped)


Lemongrass3110:
Would be nice to make xs and ys optional.
Additionally adding support for constants for AI and size would be nice if we already touch the format! :)

src/map/npc.c
```diff
@@ -3751,11 +3751,16 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
...
 	if (mob_lv > 0 && mob_lv <= MAX_LEVEL)
 		mob.level = mob_lv;
-	if (size > 0 && size <= 2)
+	if (size > SZ_SMALL && size <= SZ_BIG)
 		mob.state.size = size;
 	if (ai > AI_NONE && ai <= AI_MAX)
 		mob.state.ai = ai;
 
+	if (mob.xs < 1)
+		mob.xs = 0;
```
Lemongrass3110: Maybe issue an additional warning on those?

src/map/script.c
```diff
@@ -9861,7 +9861,7 @@ BUILDIN_FUNC(monster)
 
 	if (script_hasdata(st, 9)) {
 		size = script_getnum(st, 9);
-		if (size > 3) {
+		if (size > SZ_BIG) {
 			ShowWarning("buildin_monster: Attempted to spawn non-existing size %d for monster class %d\n", size, class_);
```
Lemongrass3110: We should actually report the source npc here too, shouldn't we?
Actually on all of the checks in this function.

src/map/npc.c
```diff
@@ -3724,7 +3724,7 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
 		return strchr(start,'\n');// skip and continue
```
Lemongrass3110: A few lines above the format is in the comments. It also needs to be adjusted to the new one.
Can't comment on the correct line... >_>

Lemongrass3110: Monster event label can actually get a buffer overflow here, since the array is only 77 not 127.
The event name is also not validated(this might actually be the reason noone ever noticed anything until now).

vstumpf: Event names should only be validated at run time, just like any other script. You would have to change the load order to make all the monster scripts go last to validate the event names.

Lemongrass3110: There is a function for checking the event name syntax. I was referring to that one. Of course we could add checking of the event labels after everything was fully booted up too.

doc/script_commands.txt
```diff
@@ -155,6 +155,7 @@ is given as "--ja--", the 'japanese name' field from the monster database is
 used, (which, in rAthena, actually contains an English name) if it's "--en--", 
 it's the 'english name' from the monster database (which contains an uppercase 
 name used to summon the monster with a GM command).
+Monster level will change the level of the monsters spawn (optionnal).
```
aleos89: `<monster level> can be used to change the spawned monster's level from the database value (optional).`

doc/script_commands.txt
```diff
@@ -174,9 +175,21 @@ should start with "On". As with all events, if the NPC is an on-touch NPC, the
 player who triggers the script must be within 'trigger' range for the event to
 work.
 
-There are two optional fields for monster size and AI.  Size can be 0 (medium),
-1 (small), or 2 (big).  AI can be 0 (default), 1 (attack/friendly), 2 (sphere),
-3 (flora), or 4 (zanzou).
+There are two optionnals fields for monster size and AI.
```
aleos89: optionnals -> optional

src/map/npc.c
Atemo: ```xs``` and ```ys``` short int not initialized https://github.com/rathena/rathena/blob/1bf5e69416c30fb13d4eadc27fced2609dd20589/src/map/npc.c#L3691
and ```mob.xs = (signed short)xs;```.

Any idea to distinguish none input (always negative) from negative input?

src/map/script.c
Atemo: We don't need, the source already return ```SCRIPT_CMD_FAILURE```
https://github.com/rathena/rathena/blob/1bf5e69416c30fb13d4eadc27fced2609dd20589/src/map/script.c#L9866
https://github.com/rathena/rathena/blob/1bf5e69416c30fb13d4eadc27fced2609dd20589/src/map/script.c#L507

Lemongrass3110: True! :)

Issue #1190, opened 2016-04-19
make server error


jgfeliz:
I have the following error when I try to compile centos 7

`make[1]: Entering directory `/home/shinigami/rAthena/src/map'
            CC      itemdb.c
itemdb.c: In function 'itemdb_read_itemgroup_sub':
itemdb.c:618:4: error: incompatible type for argument 3 of 'itemdb_group_free'
    itemdb_group_free(db_ui2key(group_id), &data, NULL);
    ^
itemdb.c:546:12: note: expected 'va_list' but argument is of type 'void *'
 static int itemdb_group_free(DBKey key, DBData *data, va_list ap);
             ^
make[1]: *** [obj/itemdb.o] Error 1
make[1]: Leaving directory `/home/shinigami/rAthena/src/map'
make: *** [map] Error 2
`

Lemongrass3110:
This must have been broken since November then.
This line was changed by me here: https://github.com/rathena/rathena/commit/86d6113801db3c18d78d437f54ad623ccf5230df#diff-a77bb0f27e703467040fcc353196222fR616

For the time being you could try to add a cast to va_list for the NULL parameter.

jgfeliz:
I can not understand , what would be the solution?

jgfeliz:
again
![2016-04-20 23_12_35-new notification](https://cloud.githubusercontent.com/assets/18557926/14696938/8b03428e-074d-11e6-90fe-6e22dcae6da7.jpg)

Playtester:
Apparently centos doesn't like if you send NULL instead of a va_list (and it's really bad to do this because there's no memory allocated for it then).

Try to undo that part and see if that helps I'd say:

Replace:
```+			itemdb_group_free(db_ui2key(group_id), &data, NULL);```

With:
```
-			va_list ap;
-			itemdb_group_free(db_ui2key(group_id), &data, ap);
```

hazimjauhari90:
this not happen only in centos. also happen too in debian.. perhaps happen in linux OS i think..i've got this error too in debian.

![ra error](https://cloud.githubusercontent.com/assets/16514357/14729747/302e1182-0873-11e6-8c8b-af22f46eeaa1.png)
.

theoriginalspre:
Spre is back!

Anywho, This compiled beautifully without using va_list on ubuntu, which is a debian derivitive. I dont have centos to test need to get vmware reinstalled, but from what I gather changing NULL to 0 should work, should being the operative word

[...]

PR #1859, opened 2017-01-08
raspberry simple adaptation

johnfercher:
Hi, I was trying to run the rAthena server in my Raspberry PI 2, but I was getting one error in the file **itemdb.c** during the compile. Searching for an answer I found the #1190 issue, which has the same error. The user @Playtester suggested this change, but, the master repository still haven't the correction. 

I compile and used the server under Windows 10 and the latest version of Raspbian, none error happened. This modification is simple and very useful.


lighta:
This change will work.
I just check the function. Altought to be honest is far from being pretty, and it's annually going to the wrong side. Allow me to explain in a big wall of text =) (it's been a while).

So the change currently work cause the function called,
static int itemdb_group_free(DBKey key, DBData *data, va_list ap).
Does't use at all the va_list. So sending an uninitialized va_list work. (that why it was coded with NULL / 0).
Now arm / raspberry complain cause va_list have different implementation and the NULL one is only good for if it was a linked_list.. cf : https://bytes.com/topic/c/answers/721825-va_list
Now declaring the variable, va_list ap; will work but we won't be able to use it in any way cause there was no va_start(), va_end(), done. So it's not usable list =(, And with that change it make it less obvious that it's not usable. So that why I'm saying it's a step in the wrong direction.

Now back on original issue, we calling a function with this as signature :
static int itemdb_group_free(DBKey key, DBData *data, va_list ap).
This signature is imposed by other db_function to be able to loop trough them etc.. So we change touch that function signature. But we know for that call, and even the common use the ap is useless.. hmm ?
Well imo. let just create a new function signature for our need :
```
static int itemdb_group_free2(DBKey key, DBData *data)
{
//insert the previous definition here, since we're not using va_list anyway
}

//helper with correct signature, wich then call the reduced fonction
static int itemdb_group_free(DBKey key, DBData *data,va_list ap)
{
 return itemdb_group_free2(key,data);
}
```
And finally on src/map/itemdb.c:607.
`-            itemdb_group_free(db_ui2key(group_id), &data, ap);
+            itemdb_group_free2(db_ui2key(group_id), &data);`

Like this, well we don't create a unusable va_list, we explicitly show that our fonction doesn't need it for now and we keep the signature with va_list for those who need.
For the simple cost of adding a new definition in the text section of our prog, and some context switch  from itemdb_group_free to itemdb_group_free2. But this could probably be inlined

src/map/itemdb.c
```diff
@@ -1774,6 +1775,10 @@ static int itemdb_final_sub(DBKey key, DBData *data, va_list ap)
 	return 0;
 }
 
+/** NOTE:
+* In some SOs, like Raspbian, we aren't allowed to pass 0 in va_list.
```
secretdataz: Is this a typo? Did you mean OS?

src/map/itemdb.c
```diff
@@ -1792,6 +1797,27 @@ static int itemdb_group_free(DBKey key, DBData *data, va_list ap) {
 	return 0;
 }
 
+/** (Raspberry PI)
+* Adaptation of itemdb_group_free. This function enables to run rAthena on Raspbian SO.
```
secretdataz: Is this a typo? Did you mean OS?

johnfercher:
I added the function which you suggested and it's in fact more pretty. Was added comments in both functions about the modification. I tested again under Raspberry PI and Windows 10, none problems happened.

I'm sorry about the bad identation, I don't know why the Windows do that.

johnfercher:
Changed SO to OS and reverted compile configs to VS2017.

vcproj-15/char-server.vcxproj
```diff
@@ -227,4 +227,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
secretdataz: Missing newline

vcproj-15/login-server.vcxproj
```diff
@@ -211,4 +211,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
secretdataz: Missing newline

vcproj-15/map-server.vcxproj
```diff
@@ -366,4 +366,4 @@
     <Copy SourceFiles="..\db\import-tmpl\statpoint.txt" DestinationFolder="..\db\import\" ContinueOnError="true" Condition="!Exists('..\db\import\statpoint.txt')" />
     <Copy SourceFiles="..\db\import-tmpl\status_disabled.txt" DestinationFolder="..\db\import\" ContinueOnError="true" Condition="!Exists('..\db\import\status_disabled.txt')" />
   </Target>
-</Project>
+</Project>
```
secretdataz: Missing newline

vcproj-15/mapcache.vcxproj
```diff
@@ -150,4 +150,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
secretdataz: Missing newline

johnfercher:
Impossible, I copied the original files and pasted on project files.

lighta:
I merge this manually into the repo.
Did a little squash on your various commit to not bloat to much the log history.
Also remove the duplicate code, and add inline to avoid context switch.

secretdataz:
@lighta There's an option to automatically squash and merge :/

PR #1869, opened 2017-01-12
String escapes

Jeybla:


Lemongrass3110:
Nice catch. But I think this should probably be done in inter_savereg and probably both(key&value) should be escaped. Key should obviously always be escaped and value only if it is a string variable.

src/char/inter.c
```diff
@@ -524,45 +524,53 @@ void mapif_accinfo_ack(bool success, int map_fd, int u_fd, int u_aid, int accoun
  **/
 void inter_savereg(uint32 account_id, uint32 char_id, const char *key, unsigned int index, intptr_t val, bool is_string)
 {
-	if( key[0] == '#' && key[1] == '#' ) { // global account reg
+	char esc_val[254];
+	char esc_key[32];
+
+
```
Lemongrass3110: Remove one newline here please. It actually looks like you wanted to put it after 534(where it would fit better).

src/char/inter.c
```diff
@@ -524,45 +524,53 @@ void mapif_accinfo_ack(bool success, int map_fd, int u_fd, int u_aid, int accoun
...
-	if( key[0] == '#' && key[1] == '#' ) { // global account reg
+	char esc_val[254];
+	char esc_key[32];
+
+	
+	Sql_EscapeString(sql_handle, esc_key, key);
+	if( is_string && val ) {
+		Sql_EscapeString(sql_handle, esc_val, (char*)val);
+	}
+	if( esc_key[0] == '#' && esc_key[1] == '#' ) { // global account reg
```
Lemongrass3110: Please keep using the unescaped key for the prefix checks and for sending them to the login server and the specific error output for this usecase.

Your changes should actually be copied over to mmo_save_global_accreg in login/account.c for the account variables too instead of sending the escaped value directly.

Jeybla:
Thanks for your accurate review, Lemon.

I didn't actually see the unescaped strings in mmo_save_global_accreg.
I'll have a closer look for other unescaped strings.

Jeybla:
Should be save to merge. Sorry for late testing D:

Issue #1908, opened 2017-01-24
Windows Mapcache users missing batch?


Akkarinage:
* **rAthena Hash**: 09cbf2d
* **Client Date**: Any
* **Server Mode**: Any
* **Description of Issue**: 
There's no easy way to run the mapcache on windows without opening a new cmd window and navigating to the exe.

PR #1909, opened 2017-01-24
Batch file for Windows Mapcache users

Akkarinage:
Signed-off-by: Akkarinage <akkarin@rathena.org>

* **Addressed Issue(s)**: #1908 

* **Server Mode**: Both

* **Description of Pull Request**: Adds a new batch file for windows users

### DIFF
--- a/mapcache.bat
+++ b/mapcache.bat
@@ -1,2 +1,2 @@
 @ECHO OFF
-CALL serv.bat mapcache.exe Map Cache
+CALL serv.bat mapcache.exe Map-Cache
--- a/mapcache.bat
+++ b/mapcache.bat
@@ -1,0 +1,2 @@
+@ECHO OFF
+CALL serv.bat mapcache.exe Map Cache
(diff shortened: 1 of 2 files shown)


anacondaqq:
And? What is it doing? I see that you added call request for mapcache.exe to serv.bat script.
But for what is it? What benefits? What will it give? What expected the result? I really trying to understand the logic of it, but I can't. Can you help me out, please? Thx

P.S. not against changes, and it's not critic, I just trying to realise for what is it, and how to use it (i mean serv.bat & mapcache change)

Akkarinage:
mapcache.bat works the same way as any of the other batch files for running servers. Adding the batch file makes it easier for users to run the mapcache. The extra line in serv.bat stops the mapcache from restarting after 15 seconds. Most users seem to struggle to open cmd and navigate to mapcache.exe and then complain that it vanishes due to there being no "pause" built in.

It's just something small that will help users.

PR #1829, opened 2016-12-26
Initial Release: Translation/Localization Feature

cydh:
Initial release NPC Translations.
- Further info can be read in [README.md](https://github.com/cydh/rathena/blob/feature/translation/lang/README.md)
- Credits to https://github.com/HerculesWS/Hercules/commit/330e31cc71ece055908acb1eb967b4009ebc9c46, @aleos89

---
## Known issue
- The message inside `select` with `:` as separator, won't be parsed/translated

PR #1932, opened 2017-01-30
translation of mob_db_re and item_db_re to pt_br

johnfercher:
This commit will not affect any server mode, the use of it is totally optional.

There are at least two emulators of Ragnarok in portuguese [Cronus](https://github.com/Cronus-Emulator/Cronus) and [brAthena](https://github.com/brAthena/brAthena), the thing is, they haven't a community like rAthena, therefore, are outdated in several ways. Thinking in offer other option (up-to-date) to Brazilian servers, I translated the files **mob_db_re.sql** and **item_db_re.sql**. The files only updates the fields related to names, so, this contribution doesn't affect the rest of things.

I used the sql-files from brAthena to translated which was possible; in the files, the lines with '#' are the items/mobs which doesn't have translation yet, thus, it's simple to translate the things that still remain in english.


Lemongrass3110:
Hey john!
Thanks for your contribution.
Maybe it would make more sense to include those names into the .po files from #1829.
What do you think?

aleos89:
As @Lemongrass3110 commented before. Including this with PR #1829 would make more sense rather than creating separate imports for each language.

PR #1980, opened 2017-02-23
Allow leaderless party

macabu:
* **Addressed Issue(s)**: Parties should be able to exist without a leader
* **Server Mode**: Both
  * This just treats leader as another member but flags it and does the removal accordingly. Things like changing party option (exp share and item) already checked if no leader was present.  
  * I tested it and everything seems to be working client, server and db -wise.

PS: Don't know how to only include the last commit soz

### DIFF
--- a/src/char/int_party.c
+++ b/src/char/int_party.c
@@ -176,6 +176,13 @@ int inter_party_tosql(struct party *p, int flag, int index)
 			schema_config.char_db, party_id, p->member[index].account_id, p->member[index].char_id) )
 			Sql_ShowDebug(sql_handle);
 	}
+	
+	if( flag & PS_DELLEADER )
+	{// Remove leader status
+		if( SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `leader_id`='0', `leader_char`='0' WHERE `party_id`='%d'",
+			schema_config.party_db, party_id) )
+			Sql_ShowDebug(sql_handle);	
+	}
 
 	if( charserv_config.save_log )
 		ShowInfo("Party Saved (%d - %s)\n", party_id, p->name);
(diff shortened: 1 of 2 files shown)


Lemongrass3110:
You used your master branch again this is basically your problem because this makes it more difficult to merge upstream changes.
Create a branch in your fork and keep your master branch untouched and only use it to merge changes from upstream master into it.
There is a guide in rathena's readme how to do so.
If you need any help feel free to send any of us a message on discord.

macabu:
@Lemongrass3110 
I actually didn't add a configuration on purpose, because I wanted to discuss this.  
How would you like it to behave if someone had the setting enabled and then disables the setting and restarts the server? A cleanup on leaderless parties? Same for @reloadbattleconf?    
Regarding the instances, official behavior doesn't allow for parties without a leader to create an instance, which rAthena already does too.

aleos89:
How would you go about assigning a new leader to the party? Would you just go with the lowest index or randomly pick a person from the party? What about the leader leaving a party while the party has an instance?

Daegaladh:
In euRO servers when the leader leaves, the party become useless and there's no way to set another leader, you just have to leave and make a new one.

Jeybla:
> In euRO servers when the leader leaves, the party become useless and there's no way to set another leader, you just have to leave and make a new one.

Hostly this doesn't sound like an improvement to me. I'd pause this until official servers have a decent solution for those things. Also it's much more work to maintain multiple settings like you already mentioned `A cleanup on leaderless parties? Same for @reloadbattleconf?`

cydh:
> How would you go about assigning a new leader to the party? Would you just go with the lowest index or randomly pick a person from the party?

no new assignment

> What about the leader leaving a party while the party has an instance?

it works for anticipate the shitty party leader. u can continue the instance if the party leader is leaving. almost (or all?) instance need party leader to generate the timer, the rest are good even without leader.

u guys, need drop ur own opinion for official feature

PR #2006, opened 2017-03-15
Update main.sql

ex0ample:
* **Addressed Issue(s)**: 
edit database create script
* **Server Mode**: 
Both
edit database create script table 'db_roulette'


aleos89:
Might be worth while to adjust `clan` and `clan_alliance` with this as well.

Issue #2043, opened 2017-04-07
Mob exp returns negative value using getmonsterinfo and strmobinfo


technoken:
* **rAthena Hash**:  4b9ade33ee4f89bc65b3ec272eb5c5e6a5966838
* **Server Mode**:  Pre-renewal
* **Modifications that may affect results**: 
* **Description of Issue**: 
When exp rates are set higher, the exp shown when using getmonsterinfo and strmobinfo gave negative values.

I think this actually happens because when the exp reached over 2147483647. I don't know much about it's calculation and can't give any further info why it gives negative result.

```
// Rate at which exp. is given. (Note 2)
base_exp_rate: 25000

// Rate at which job exp. is given. (Note 2)
job_exp_rate: 25000
```

I use this script
```
-	script	test	-1,{
OnNPCKillEvent:
	.@mobid = killedrid;
	.@strmobinfo6 = strmobinfo(6,.@mobid);
	.@strmobinfo7 = strmobinfo(7,.@mobid);
	.@getmonBaseExp = getmonsterinfo(.@mobid,MOB_BASEEXP);
	.@getmonJobExp = getmonsterinfo(.@mobid,MOB_JOBEXP);

	dispbottom "MOB ID: "+.@mobid+"";
	dispbottom "Exp Reward: "+.@strmobinfo6+"";
	dispbottom "Job Exp Reward: "+.@strmobinfo7+"";
	dispbottom "Exp Reward: "+.@getmonBaseExp+"";
	dispbottom "Job Exp Reward: "+.@getmonJobExp+"";
}
```
then try to kill an MVP monster. Pharaoh, Lord of the Dead, etc..
It will show the negative results.

vstumpf:
It's because Athena uses 4 byte signed integers. 4 bytes = 32 bits, one bit is reserved for sign bit, leaving 31 bits. Therefore, the range of numbers available is [-2147483648, 2147483647].

The mob_db struct saves the exp as 4 byte unsigned integers.  Since there's no sign bit, the range is [0, 4294967295].

technoken:
Thanks @Lemongrass3110 ! Anyway does this fix cap the exp to 2147483647 or no?

vstumpf:
No, it only caps the value you get from the script commands. Something can give 4 billion exp, but if you use the command, it will show 2147483647

Lemongrass3110:
Fixed in e72c736

PR #2051, opened 2017-04-09
Capped a few fields for mob info script commands

Lemongrass3110:
* **Addressed Issue(s)**: 
#2043

* **Server Mode**: 
Both

The script engine uses a signed 32 bit integer type and therefore it is impossible to push any value higher than INT32_MAX into the script engine and expecting valid results.
Of course this fix will also be wrong, because the data types still do not match, but I think it looks better than getting a negative number.


Jeybla:
Why not cap it in script_pushint for everything? Maybe zero or `-1` would be a valid solution, too?

`struct script_data* push_val2(struct script_stack* stack, enum c_op type, int64 val, struct reg_db *ref)`

Just wondered about `int64 val` in its definition.

PR #2061, opened 2017-04-13
When we try to run the application we got the signal Abort trap 6 and…

rpdigos:
… the application is interrupted by the OS.

This issue is related with the fact that when using strncpy the source and destination string should not overlap, as we can see in the man page:

man strncpy()
...
The source and destination strings should not overlap, as the behavior is undefined.

We've replaced it by the function memmove:

man memmove()
...
The two strings may overlap; the copy is always done in a non-destructive manner.

Follow ec1fe15d7d480e4fa6ff3f56986bbfd49a4cde2a

* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/src/map/map.c
+++ b/src/map/map.c
@@ -3989,7 +3989,7 @@ int inter_config_read(char *cfgName)
 #ifdef RENEWAL
 			// Copy the original name
 			// Do not use safestrncpy here - enforces zero termination before copying and will break it [Lemongrass]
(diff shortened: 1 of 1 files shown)


rpdigos:
In the comment above I mean strncpy() instead of stpncpy() sorry for the typo

Jeybla:
Related src: http://www.cplusplus.com/reference/cstring/strncpy/
>  destination and source shall not overlap (see memmove for a safer alternative when overlapping).

Thanks for providing the pull request!

rpdigos:
Hello,

The OS is MacOS X 10.10 and the server was compiled using gcc version 5.2.0

Issue #1710, opened 2016-11-12
Bug: Item Combo - Cards and Equip in same index


esu1214:
for example 21012.
The equip that contains a same postion combo_db card will never work.

Edited Information by Jey:
Mode: Both
Hash: a75133eb4d8976c1c890d5420665e7a76349e226
Example Combo: `4335:21012`

esu1214:
The newest version of rA master.
you may have a test on 4335:21012.

Playtester:
I haven't checked if it's really bugged, but looking at the code it might be because of this:
```				if( pc_is_same_equip_index((enum equip_index)k, sd->equip_index, index) )
					continue;```
That combo mentioned by esu for example is a sword combined with a card that needs to be put into the sword. Because the sword is the same equip index as the card, it might be skipped?

@esu1214 - Please don't say "newest", that not very helpful over time. Also with mode it means if you are compiling on renewal or pre-renewal.

**Comment by esu1214** [comment]
The newest version of rA master.
you may have a test on 4335:21012.

**Comment by esu1214** [comment]
The newest version of rA master.
you may have a test on 4335:21012.

**Comment by Jeybla** [comment]
4335:21012 is a renewal combo, so mode should be renewal.

This issue reminds me of another bug in the item combo db. Every item combo in the import db overwrites a item combo in the standard db. So if there are 300 item combos in your import db, the first 300 standard item combos are disabled. Maybe this bug has been fixed long time ago. Just want to mention it here.

**Comment by esu1214** [comment]
I have anylyzed the function "pc_is_same_equip_index".
it's not the reason why the same equip position be skipped.
The reason is that it only compare equipped equip_index to trigger the item_combo but not contains its selfe when equip a equipment.

**Comment by Jeybla** [comment]
I can confirm this bug.

Fun fact:
Changing the combo from `4335:21012,[...]` to `21012:4335,[...]` fixes it.

In `pc_checkcombo` every recorded item is saved in `nb_itemCombo`. Non cards are just checked by their index:
```C
						for (z = 0; z < nb_itemCombo-1; z++)
							if(combo_idx[z].idx == index) //we already have that index recorded
								do_continue=true;
```
And Cards are checked in more detail:
```C
							for (c1 = 0; c1 < nb_itemCombo-1; c1++){
								if(combo_idx[c1].idx == index){
									for (c2 = 0; c2 < sd->inventory_data[index]->slot; c2++){
[...]

PR #2080, opened 2017-04-19
Fixes Item Combos in same equip index.

Jeybla:
* **Addressed Issue(s)**: #1710
* **Server Mode**: Both

I'd prefer to refactor some things in `pc_checkcombo`, but this is a quick fix for #1710 .


esu1214:
Good job.

PR #2081, opened 2017-04-20
Adjusted skill_nocast_db settings.

### LINKED ISSUE #1496: Manhole & Bloodlust without PvP
I'm really not sure if this is official, so if someone could check that... would be appreciated.
The thing is that Manhole & Bloodlust can be used on players **without the need of PvP mode on**.

So players can abuse it to trap other MVP hunters or keep them away **from even entering** the WoE Castles. Even though these maps have no PvP nor GvG activated?!

Is this really normal?

Jeybla:
For the following skills:
* SC_MANHOLE - PvP only
* SC_DIMENSIONDOOR - town and field, doesn't work in PvP
* SC_CHAOSPANIC - No restriction
* SC_BLOODYLUST - Enabled in town

Fixes #1496.

Thanks to @Everade and @aleos89.

### DIFF
--- a/db/re/skill_nocast_db.txt
+++ b/db/re/skill_nocast_db.txt
@@ -30,11 +30,12 @@
 2423,1	//WM_SIRCLEOFNATURE
 2430,1	//WM_SATURDAY_NIGHT_FEVER
 2455,1	//SO_ARRULLO
+2299,1	//SC_MANHOLE
 
 //----------------------------------------------------------------------------
 // PVP
 //----------------------------------------------------------------------------
-
+2300,2	//SC_DIMENSIONDOOR
 
 //----------------------------------------------------------------------------
 // GVG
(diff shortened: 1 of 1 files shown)


db/re/skill_nocast_db.txt
```diff
@@ -5,7 +5,7 @@
 // SkillID,Flag
 //
 // Legend for 'Flag' field (bitmask):
-// 1    - cannot be used in normal maps
+// 1    - cannot be used in normal maps (enabled on pvp, gvg, gvg_dungeon, gvg_castle, gvg_te, gvg_te_castle, battleground)
```
aleos89: If we adjust the comment here, we should probably make the change in the pre-re and import files.

Jeybla: Just removed the comment. Not really worth the future maintenance/merge conflicts. Was just nice to know for me.

Issue #2058, opened 2017-04-13
Homunculus Eleanor Combo Insta Kill All Mobs


feltenc:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/3cce333fc69766a04823eb3bd8bc7bc86061bab6 

* **Client Date**: 2017-01-25RagexeRE

* **Server Mode**: Renewall

* **Description of Issue**: Eleanor Homunculus is causing instakill in any monster, be it boss or even the prontera dummys. We tested with the prontera dummies and it killed them in less than 5 seconds, we also tested with some mvps like Ifrit and some other mvp's.

  * Result: Game Broken
  * Expected Result: Normal hits/damage
  * How to Reproduce: Summon a Eleanor with Biochemist/Creator and use the Eleanor Combo on bosses or dummies.
  * Official Information: None

* **Modifications that may affect results**: None

Replay Link: https://mega.nz/#!1IJBVA7b!gHHty6ttgUGQn9k1EskQ6OVhDVHjemUCJZrWYUuFz0A

admkakaroto:
In the map_server also some error messages appear when changing the mode of the homunculus.

http://imgur.com/TQTGhRQ


I also confirm the bug using Eternal Quick Combo.

Jeybla:
I can confirm this bug. Damage is over 9m!

I'll have a closer look into it.

Jeybla:
Seems to be a wrong skill formula for `Eternal Quick Combo`.

I can confirm this bug report. But I don't get any error message.
> In the map_server also some error messages appear when changing the mode of the homunculus.

Jeybla:
Thanks to @aleos89 for testing the skill on aegis! The damage formula is correct.

We'll just disable the skill on boss type monster.

feltenc:
Nice! congratulations for your job guys, you are all amazing!

PR #2082, opened 2017-04-20
Updated Homunculus S skill failure messages

aleos89:
* **Addressed Issue(s)**: #2058

* **Server Mode**: Pre-renewal and Renewal.

  * Added missing skill failure messages for Homunculus S.
  * Eternal Quick Combo can no longer be used on Status Immune monsters.
  * Disabled the Style Change message for Eleanor. Aegis does not display a message when switching fighting style.
Thanks to @feltenc and @Jeybla!


src/map/skill.c
```diff
@@ -5844,7 +5886,6 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 		{
 			TBL_HOM *hd = BL_CAST(BL_HOM,src);
```
Jeybla: Variable hd is now unused. This line can be removed.

PR #2098, opened 2017-04-27
Updated Giant Fly Wing behavior

aleos89:

* **Server Mode**: Pre-Renewal and Renewal

  * Party members should be recalled to their own positions in a 3x3 area around the party leader.
  * Added two missing client messages.
  * Adjusted script command warpparty to support range values.

Partial credits to: HerculesWS/Hercules@7064be0

### DIFF
--- a/npc/other/CashShop_Functions.txt
+++ b/npc/other/CashShop_Functions.txt
@@ -52,10 +52,8 @@ function	script	F_CashStore	{
 // - No arguments.
 function	script	F_CashPartyCall	{
 	warp "Random",0,0;
-	if(getpartyleader(getcharid(1),2) == getcharid(0)) {
-		getmapxy .@map$, .@x, .@y, UNITTYPE_PC;
-		warpparty "Leader", 0, 0, getcharid(1), .@map$, 3, 3;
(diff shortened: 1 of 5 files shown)


src/map/script.c
```diff
@@ -5838,6 +5840,17 @@ BUILDIN_FUNC(warpparty)
 		break;
 		case 3: // Leader
 		case 4: // m,x,y
+			if (rx && ry) {
```
Atemo: ```if (rx || ry) {``` to allow a warp in line

npc/other/CashShop_Functions.txt
```diff
@@ -48,13 +48,13 @@ function	script	F_CashStore	{
 // Giant Fly Wing
 //============================================================ 
 // - Warp party leader to random spot on the map.
-// - Summon Party members on party leader map to that location.
+// - Summon Party members on party leader map to a 3x3 location around the leader.
 // - No arguments.
 function	script	F_CashPartyCall	{
 	warp "Random",0,0;
```
Atemo: remove ```warp "Random",0,0;```

Issue #2111, opened 2017-04-30
Certain Hidden Stat Bonuses Not Being Applied


Canoe:
* **rAthena Hash**: 9e6b02ae590d22d1fe4bdc730e0c0f62eb80ca33
* **Client Date**: 20151104
* **Server Mode**: Pre-Renewal, but also affects Renewal

* **Description of Issue**: 
Seems to be caused by a75133eb4d8976c1c890d5420665e7a76349e226. 

These functions were given a new variable, bool display. However, none of them are called where display is false.

status_calc_batk
status_calc_matk
status_calc_hit
status_calc_critical
status_calc_flee
status_calc_flee2

This means for any of these that have bonuses inside the display check, those bonuses are not being applied.

This is affecting things like..
Mindbreaker's Matk buff
Box of Resentment
Abrasive

PR #2114, opened 2017-05-02
Independent display values for the status window.

Jeybla:
* Adds the battle config show_full_status to optionally
* show all hidden status changes in the status window.

* **Addressed Issue(s)**: #2111

This is just a suggestion.

Adds status display variables like `def_disp`. Status-Calc-functions like `status_calc_def` are calculating the actual value (`def`) and also a `hidden` value. The Display Variable `def_disp` gets calculated by  `def_disp = def - hidden` . This makes to calculation of the status window completely independent from the actual status calculation. And gives the possibility to show all status changes in the status window with `show_full_status = yes`.

Added status display variables:
```
atk_disp, atk2_disp, batk_disp, watk_disp, 
watk2_disp, matk_min_disp, matk_max_disp, 
hit_disp, flee_disp, cri_disp, flee2_disp,
def_disp, mdef_disp, def2_disp, mdef2_disp
```

### DIFF
--- a/conf/battle/player.conf
+++ b/conf/battle/player.conf
@@ -172,9 +172,9 @@ min_npc_vendchat_distance: 3
 // Default is 25. 100 = 100% Increase.
 rental_mount_speed_boost: 25
 
-// Display Katar critical bonus in the status window? (Note 1)
+// Display all status changes in the status window? (Note 1)
 // Default (official): no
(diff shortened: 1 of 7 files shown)


Lemongrass3110:
Closed by @aleos89 in ef3ad99

PR #2115, opened 2017-05-03
Fix few issues in 2016-12-28 & newer clients

hazimjauhari90:
Updated ZC_ACCEPT_ENTER2 for 2016-12-28 and newer clients

Fixed a issue with 2016-12-28 and newer clients not properly showing
player's vending list to other players viewing them.

Thank you to @Rytech2 and 3CeAM
https://sourceforge.net/p/v1-3ceam/code/817


src/map/clif.c
```diff
@@ -2508,6 +2510,8 @@ void clif_add_random_options(unsigned char* buf, struct item *it) {
 /// 02d4 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.W <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK3)
 /// 0990 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.L <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK_V5)
 /// 0a0c <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.L <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK_V6)
+/// 0a37 <index>.W <amount>.W <name id>.W <identified>.B <damaged>.B <refine>.B <card1>.W <card2>.W <card3>.W <card4>.W <equip location>.L <item type>.B <result>.B <expire time>.L <bindOnEquipType>.W (ZC_ITEM_PICKUP_ACK_V7)
+/// Note: Length for a37 is 3 bigger then a0c but I don't know what the added parts are for. // [Rytech]
```
Lemongrass3110: The note says 3, but the code is increased by 6 (47->53).
Intended?

src/map/clif.c
```diff
@@ -683,8 +683,10 @@ void clif_authok(struct map_session_data *sd)
 	const int cmd = 0x73;
 #elif PACKETVER < 20141022
 	const int cmd = 0x2eb;
-#else
+#elif PACKETVER < 20161228
```
Lemongrass3110: better turn the upper one into a or condition:
`#elif PACKETVER < 20141022 || PACKETVER >= 20161228`
We have it like that for other packets as well. Character deletion for example.

src/map/clif.c
hazimjauhari90: well.. based on this, yeap increase by 6.
https://sourceforge.net/p/v1-3ceam/code/817

src/map/clif.c
hazimjauhari90: done changed it. please review it back..

src/map/clif.c
dastgir: I guess rytech meant 3 WORD are added.

Lemongrass3110: Please add a PACKETVER condition for 0xa37 in clif_additem so it will really be sent as the new packet. The packet version where the new packet was added should be 20160921. But the size I have is 59 instead of 53. Could you test if it also works with that? If it works please also adjust the comment accordingly.

hazimjauhari90: just want to double conform. is it in void clif_additem or void clif_vendinglist?

Lemongrass3110: Sorry you are totally right. I mixed those two up. Nevertheless the PACKETVER check needs to be added and the packet_len needs to be stored in the hardcoded table around(https://github.com/rathena/rathena/blob/master/src/map/clif.c#L19545) so it get's sent with the correct length(59). Currently it's 56 that explains the comment - my bad!

hazimjauhari90: well.. i think i already made it already..

src/map/clif.c
```diff
@@ -7104,8 +7106,10 @@ void clif_vendinglist(struct map_session_data* sd, int id, struct s_vending* ven
 
 #if PACKETVER < 20150226
 	const int item_length = 22;
-#else
+#elif PACKETVER < 20161228
 	const int item_length = 47;
+#else
+	const int item_length = 59;
```
hazimjauhari90: @Lemongrass3110 already added packet length 59 in the void clif_vendinglist..

src/map/clif.c
```diff
@@ -7196,8 +7200,10 @@ void clif_openvending(struct map_session_data* sd, int id, struct s_vending* ven
...
```
hazimjauhari90: @Lemongrass3110 does this necessary? if not, i will remove it...

src/map/clif.c
```diff
@@ -7106,7 +7106,7 @@ void clif_vendinglist(struct map_session_data* sd, int id, struct s_vending* ven
 
 #if PACKETVER < 20150226
 	const int item_length = 22;
-#elif PACKETVER < 20161228
+#elif PACKETVER < 20160921
```
hazimjauhari90: @Lemongrass3110  already added packet length 59for PACKETVER condition 20160921 in the void clif_vendinglist..

src/map/clif.c
```diff
@@ -7196,8 +7200,10 @@ void clif_openvending(struct map_session_data* sd, int id, struct s_vending* ven
 
 #if PACKETVER < 20150226
 	const int item_length = 22;
-#else
+#elif PACKETVER < 20160921
 	const int item_length = 47;
+#else
+	const int item_length = 59;
```
hazimjauhari90: @Lemongrass3110 does this necessary? if not, i will remove it...

src/map/clif.c
```diff
@@ -7104,8 +7106,10 @@ void clif_vendinglist(struct map_session_data* sd, int id, struct s_vending* ven
...
```
hazimjauhari90: @Lemongrass3110  already added packet length 59 for PACKETVER condition 20160921 in the void clif_vendinglist..

src/map/clif.c
hazimjauhari90: @Lemongrass3110  does this necessary? if not, i will remove it...

src/map/clif.c
hazimjauhari90: just got time to test it on server.. well.. these are what happens if using packet length 59.. still items not properly shown in the vending. by using 2017-03-15cRagexeRE client..

screen1asiaro021
screen1asiaro022

hazimjauhari90: while this is when using packet length 53... it fixed the issue of items not properly shown in vending.

screen1asiaro023

Rytech2:
0xA0C and 0xA37 are packets for adding a single item into one's inventory. 0xA0C has a length of 56 and 0xA37 has a length of 59 which is why I made the commit that its 3 bigger and question what its for.

As for the length of the vending packet, the same packet is used but its length is increased from 47 to 53. No idea what the added 6 bytes are for but its fine leaving them blank for now until we know what their used for.

src/map/clif.c
hazimjauhari90: nope.. i think u are right & i'm the person who mixed up those two.. after 5 days, i finally understood & get what u mean... just added 20160921 PACKETVER condition for 0xa37 in clif_additem..

Jeybla:
Why did you close it @hazimjauhari90 ?

Lemongrass3110:
We pmed today on discord - he is still testing things, probably closed it due to that.

hazimjauhari90:
@Jeybla i closed it becaus i accidentally clicked on sync button at github desktop. and after that everything mixed up, i mean this PR is contaiminated with lots of non related commits.

@Lemongrass3110.. i just put a test with using 2017-03-15cRagexeRE..  based on our conversation earlier, u told me clif_additem is called when an item gets added to your inventory(when you collect a dropped item or receive it from an npc or something like that) even @item should trigger it

i made 2 tests.. summon 20 poring monster (@monster poring 20) & kill them all using killmonster command. so, i collected the dropped poring items.. seems all added to my inventory. another test that i made, by using @item command.. i tested on item ids 13201, 13029, 13039, 13054, 13058.. all these 5 items also appear & added to my inventory.. without missing any of them. then, i try to relog.. all those items obtain from these 2 tests are still appear in my inventory..

PR #1685, opened 2016-11-02
Status Change (SC) Config

cydh:
Just rebased ver of https://github.com/rathena/rathena/tree/sc_config

PR #2126, opened 2017-05-08
[WIP] Status Change display in status window.

Jeybla:
Cherry picked and fixed changes from #1481. 
Also added a show_status_sc configuration. Which allows to display all (also hidden) Status Changes.

(Currently testing this PR)

* **Addressed Issue(s)**: #554

* **Server Mode**: Both

### DIFF
--- a/conf/battle/player.conf
+++ b/conf/battle/player.conf
@@ -172,6 +172,10 @@ min_npc_vendchat_distance: 3
 // Default is 25. 100 = 100% Increase.
 rental_mount_speed_boost: 25
 
+// Display all status changes in the status window? (Note 1)
+// Default (official): no
+show_status_sc: no
+
 //===================================
 // VIP system
 //===================================
--- a/src/map/battle.c
+++ b/src/map/battle.c
@@ -15,6 +15,7 @@
 #include "map.h"
 #include "path.h"
 #include "pc.h"
+#include "status.h"
 #include "homunculus.h"
 #include "mercenary.h"
 #include "elemental.h"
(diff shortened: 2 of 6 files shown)


src/map/battle.c
```diff
@@ -4551,29 +4552,27 @@ struct Damage battle_calc_defense_reduction(struct Damage wd, struct block_list
...
-			def1 = (def1 * (100 + si)) / 100;
+
+			def1 = def1 * (100 + si) / 100;
+		}
+
+		i += sd->ignore_def_by_class[tstatus->class_] + sd->ignore_def_by_class[CLASS_ALL];
+		if (i) {
+			i = min(i, 100); //cap it to 100 for 0 def min
+			def1 -= def1 * i / 100;
+#ifndef RENEWAL
```
Jeybla: Should be `#ifdeaf` instead.

src/map/battle.c
```diff
@@ -4582,29 +4581,9 @@ struct Damage battle_calc_defense_reduction(struct Damage wd, struct block_list
 
 		i = min(i,100); //cap it to 100 for 0 def min
 		def1 = (def1*(100-i))/100;
+#ifndef RENEWAL
```
Jeybla: Should be `#ifdeaf` instead.

src/map/battle.c
```diff
@@ -4623,10 +4602,8 @@ struct Damage battle_calc_defense_reduction(struct Damage wd, struct block_list
 				def2 -= (target_count - (battle_config.vit_penalty_count - 1))*battle_config.vit_penalty_num;
 			}
 		}
+#ifndef RENEWAL
 		if (skill_id == AM_ACIDTERROR)
-#ifdef RENEWAL
-			def2 = 0; //Ignore only status defense. [FatalEror]
```
Jeybla: This is missing now. Is it intended @aleos89 ?

Jeybla:
Note: #2383

conf/battle/player.conf
```diff
@@ -172,6 +172,10 @@ min_npc_vendchat_distance: 3
 // Default is 25. 100 = 100% Increase.
 rental_mount_speed_boost: 25
 
+// Display all status changes in the status window? (Note 1)
+// Default (official): no
+show_status_sc: no
```
Jeybla: Do we want to have this config in `player.conf`? I think `status.conf` may suit better?

mrjnumber1:
i have a feeling that these status values not appearing in the display window is because they adjust the user's atkpercentinfo/matkpercentinfo/defpercent/mdefpercent (and so on). because curse, provoke, and other statuses act the same way.
i dropped some of that info on the rathena forum in this topic https://rathena.org/board/topic/112343-various-status-effects-off-from-officials-and-info-from-aegis/
really hope this helps!!

Jeybla:
@mrjnumber1 thanks for the info! I already thought about pausing this pr and continuing with #1685 first. Maybe that'd be the cleaner way to go?

Jeybla:
There are many things conflicting with this pr. I suggest to pause this pr until we migrated to C++ and applied the SC Config pr (or something similar in C++). I'll keep it open anyways.

Gardosen:
is this still relevant or could this be closed? this pr is from 2017 and nothing has happened since then.

mrjnumber1:
I believe it'll be obsoleted upon the (possibly) eventual status change db in yaml

eppc0330:
Why don't we  update this PR?

PR #2130, opened 2017-05-08
follow up to 9716233

hazimjauhari90:
added missing ZC_ACCEPT_ENTER2 for 2016-12-28 & newer clients

* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -681,7 +681,7 @@ void clif_authok(struct map_session_data *sd)
 {
 #if PACKETVER < 20080102
 	const int cmd = 0x73;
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
I left that out intentionally, because I do not think that 20161228 is the correct date here.

hazimjauhari90:
I know.. but, perhaps without any clientver condition, 2017 clients cannot login into char server as the setting still using ZC_ACCEPT_ENTER3 while newer clients are using old ZC_ACCEPT_ENTER2. Perhaps 2016_12_28 will act like dummy client for the condition packet ZC_ACCEPT_ENTER2.

PR #2132, opened 2017-05-09
Feature/warp2

Atemo:

* **Server Mode**: both

* **Description of Pull Request**: Unlike 'warp', 'warp2' NPC will also be triggered by hidden player.


src/map/npc.c
```diff
@@ -4391,7 +4398,7 @@ int npc_parsesrcfile(const char* filepath, bool runOnInit)
 			}
 		}
 
-		if( strcasecmp(w2,"warp") == 0 && count > 3 )
+		if( strncasecmp("warp", w2, 4) == 0 && count > 3 )
```
Lemongrass3110: this would also allow "warp5000" as definition, because it will only validate the first 4 characters.
`( strcasecmp(w2,"warp") == 0 || strcasecmp(w2,"warp2") == 0 ) && count > 3`

kukuasir1:
when I use warp2 in the instance, if `<from mapname>` and `<to mapname>` are different map, player will be warp to `<from mapname>`

Lemongrass3110:
Thanks! Fixed in ddee418, but this was not related to this pull request.

Issue #2154, opened 2017-05-17
Instance idle timer


Jeybla:
* **rAthena Hash**: 800f49181fe54caf4e3d1f7a54997d6294ab4894
* **Client Date**: 20151029
* **Server Mode**: both

* **Description of Issue**: 
  * Result: Instance idle timer starts even though there still is another player in the instance.
  * Expected Result: The instance should stay open if there are still users in it.
  * How to Reproduce: Join any instance with two characters. **Both chars should stay on a different map within the instance**. Disconnect one character. The idle timer starts.

PR #2155, opened 2017-05-17
Fixes Instance idle timer.

Jeybla:
* **Addressed Issue(s)**: #2154
* **Server Mode**: both
```C
if(map[im->map[i]->m].users > 1) // We check this before the actual map.users are updated, hence the 1
```
This is a problematic line in `instance_delusers`. If there are multiple maps with 1 user on it, the idle timer gets started though.


Jeybla:
Thanks @Lemongrass3110 .
I actually want to test this pr before pushing it.

Issue #412, opened 2015-04-29
bBaseAtk with negative value issue


reigneil:
it seems that this bBaseAtk if you insert negative value wont work. i.e
equipping item id 4339 "Mineral Card" wont reduce your Atk. can anyone confirmed this?

OT btw what is bAtk and bAtk2., is this the version of weapon attack and status attack?

Playtester:
Probably a renewal-only issue because there are different types of ATK in renewal. BaseATK can't go below 0, and if it's always 0 on renewal, that would explain why negative values wouldn't work.

reigneil:
if it is the case then why it not set to zero or reduce only? i.e weapon attack = 150 then i use bBaseAtk -50 it must be set to 100 weapon attack right... or if i have weapon attack = 150 then i use bBaseAtk -200 it must be set to 0 right or do i miss something?

Playtester:
You are probably right, but I'm not familiar enough with the renewal damage code, it's quite complex and different from pre-renewal. There are 4 attack types rather than just 2.
http://irowiki.org/wiki/ATK

But in pre-renewal it was as you said. Like if you have 10 ATK and equip a -20 ATK item it goes to 0 ATK.

jezznar:
* **rAthena Hash**: c212b2ed053d60855ce584a0368d98cff94c5059

* **Client Date**: 2015-11-04

* **Server Mode**: both

* **Description of Issue**: Issue is still visible in the latest git commit. https://gaming.youtube.com/watch?v=xvEGvxol27U&feature=share
  * Result: bBaseAtk is not working when negative values are used. No updates in status window. No reduction in actual damage output. Doesn't work even if it is moved from script field to equip_script field in the database. Works correctly when positive values are used
  * Expected Result: 

- [ ] bBaseAtk should work when negative values are used
- [ ] Status window should be updated
- [ ] Actualamage output should be reduced

  * How to Reproduce: Use an item with negative bBaseAtk bonus.
  * Official Information: N/A

* **Modifications that may affect results**:  N/A

PR #2191, opened 2017-06-07
Fix/#412. Allow negative bBaseAtk scripts

jezznar:
* **Addressed Issue(s)**: #412 

* **Server Mode**: Pre and Re

* **Description of Pull Request**: Fixed problem about map-server not being able to parse negative bBaseAtk values. Changed variable types of bAtk, watk, watk2 and eatk to SHORT instead of USHORT. Changed cap value of status->bAtk to SHRT_MIN to SHRT_MAX instead of 0 to USHRT_MAX.

Test Cases:
1. Status window update Base attack
* Pre: Adjustment should only affect Base attack
    - [x] 64bit
    - [x] 32bit
* Pre: Adjustment should only affect Bonus
    - [x] 64bit
    - [x] 32bit

2. Atk damage adjustment is evident when attacking monsters
    - [x] 64bit: Pre and Re
    - [x] 32bit: Pre and Re

### DIFF
--- a/src/map/status.h
+++ b/src/map/status.h
@@ -2074,8 +2074,7 @@ struct status_data {
 		hp, sp,  // see status_cpy before adding members before hp and sp
 		max_hp, max_sp;
 	short
-		str, agi, vit, int_, dex, luk;
-	unsigned short
+		str, agi, vit, int_, dex, luk,
 		batk,
 #ifdef RENEWAL
 		watk,
--- a/src/map/status.h
+++ b/src/map/status.h
@@ -2075,12 +2075,14 @@ struct status_data {
 		max_hp, max_sp;
 	short
 		str, agi, vit, int_, dex, luk,
-		batk,
+		batk
 #ifdef RENEWAL
 		watk,
 		watk2,
-		eatk,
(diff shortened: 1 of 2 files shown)


src/map/status.h
```diff
@@ -2074,14 +2074,15 @@ struct status_data {
 		hp, sp,  // see status_cpy before adding members before hp and sp
 		max_hp, max_sp;
 	short
-		str, agi, vit, int_, dex, luk;
-	unsigned short
-		batk,
+		str, agi, vit, int_, dex, luk,
+		batk;
```
aleos89: Shouldn't these still be unsigned?

jezznar: I didn't really touch other stats, but let me test if it'll have any adverse effect when used as negative bonuses

jezznar: @aleos89 I tried it and as expected it works fine in RE. but when in PRE it doesn't work well. The computation for each mode is different in approach.

jezznar: Btw, these comments are just valid for bAtk. Basic stats work done add ushort

PR #2194, opened 2017-06-11
ItemDB Improvement

crazystorm2017:
Fixed armors being read as a weapon in refinement.
Fixed some item being refineable.
Fixed some item scripts.

Server Mode: 
Renewal


aleos89:
Could you resolve the conflicts on your end so this can be pushed? Thanks!

PR #2207, opened 2017-06-20
Migrate core source files to C++. Add yaml-cpp project as dependency

secretdataz:
Addressed Issue(s): None
Server Mode: Both
Description of Pull Request:
This pull request enables the rAthena project as a whole to use C++ source code in the project.
core.c, login.c, char.c, and map.c's extension have been changed to .cpp to indicate that they are C++ source codes.
This pull request also adds jbeder/yaml-cpp as a 3rdparty dependency.

Special thanks to @aleos and @Lemongrass3110 for their assist!


src/common/yamlwrapper.cpp
```diff
@@ -0,0 +1,122 @@
...
+yamliterator* yamlwrapper::iterator() {
+	return new yamliterator(this->root);
+}
+
+yamlwrapper* yaml_load_file(const char* file_name) {
+	return new yamlwrapper(file_name);
+}
+
+extern "C++" YAML::Node yaml_get_node(YAML::Node& node, std::string& key) {
+	size_t pos = key.find('.');
```
Lemongrass3110: Missing nullpointer check for both node and key.
If key is null or empty no 'find' call is necessary.

src/common/yamlwrapper.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+char* yaml_get_c_string(yamlwrapper* wrapper, const char* key) {
+	std::string cpp_str = yaml_get_node(wrapper->root, std::string(key)).as<std::string>();
+	const char* c_str = cpp_str.c_str();
+	size_t str_size = std::strlen(c_str) + 1;
+	char* buf = (char*)aCalloc(1, str_size);
+	strcpy(buf, c_str);
+	return buf;
+}
+
```
Lemongrass3110: Maybe create a template function for retrieving the values(also as extern C++) and call it from the C functions.
In that template function implement nullpointer checks and possibly default values? Or what is returned by the as function if the node is null?

src/common/yamlwrapper.cpp
```diff
@@ -0,0 +1,122 @@
...
+	this->sequence = sequence;
+	this->index = 0;
+}
+
+yamliterator* yamlwrapper::iterator() {
+	return new yamliterator(this->root);
+}
+
+yamlwrapper* yaml_load_file(const char* file_name) {
+	return new yamlwrapper(file_name);
```
Lemongrass3110: I'd actually call `YAML::LoadFile(file)` directly and make sure a node is retrieved. And only return a new instance if I get a node. That way you can remove the constructor for std::string as well.

src/common/yamlwrapper.h
```diff
@@ -0,0 +1,79 @@
...
+*
+*	You should have received a copy of the GNU General Public License
+*	along with this program. If not, see <http://www.gnu.org/licenses/>.
+*
+*	Author(s)
+*		Jittapan Pluemsumran <secret@rathena.org>
+*/
+
+#pragma once
+#pragma warning(disable:4099) // Damn it, Microsoft! [Secret]
```
Lemongrass3110: If it's only for microsoft please add ifdef for it.

src/common/yamlwrapper.h
```diff
@@ -0,0 +1,79 @@
...
+yamliterator* yaml_get_iterator(yamlwrapper* wrapper);
+
+bool yaml_iterator_is_valid(yamliterator* it);
+yamlwrapper* yaml_iterator_first(yamliterator* it);
+yamlwrapper* yaml_iterator_next(yamliterator* it);
+bool yaml_iterator_has_next(yamliterator* it);
+void yaml_iterator_destroy(yamliterator* it);
+
+#ifdef __cplusplus
+}
```
Lemongrass3110: Add `/* extern "C" */` like you did in the cpp file.

src/common/yamlwrapper.h
```diff
@@ -0,0 +1,79 @@
...
+extern "C" {
+#endif
+
+#include "cbasetypes.h"
+
+typedef struct yamlwrapper yamlwrapper;
+typedef struct yamliterator yamliterator;
+
+#ifdef __cplusplus
+class yamlwrapper {
```
Lemongrass3110: Does the typedef above not cause this to be translated to `class struct yamlwrapper`?
Also the typedefs should not be necessary if __cplusplus is defined or not?
This part is really confusing as hell x_x

src/common/yamlwrapper.cpp
secretdataz: After some research, I found that std::string in non-pointer form can't be null.

PR #2208, opened 2017-06-20
Updated sleep2:

Atemo:

  * Currently sleep2 is supposed to return 0 or 1 if the unit is still attached. Actually mapserv throws a warning and the script end if the timer is still running and unit offline. With this update the script end if there is no unit attached for sleep2.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6383,7 +6383,7 @@ sleep and sleep2 will pause the script for the given amount of milliseconds.
 Awake is used to cancel a sleep. When awake is called on a NPC it will run as
 if the sleep timer ran out, and thus making the script continue. Sleep and sleep2
 basically do the same, but the main difference is that sleep will not keep the rid,
-while sleep2 does. Also sleep2 will stop the script if there is none unit attached.
+while sleep2 does. Also sleep2 will stop the script if there is no unit attached.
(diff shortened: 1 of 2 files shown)


doc/script_commands.txt
```diff
@@ -6383,7 +6383,7 @@ sleep and sleep2 will pause the script for the given amount of milliseconds.
 Awake is used to cancel a sleep. When awake is called on a NPC it will run as
 if the sleep timer ran out, and thus making the script continue. Sleep and sleep2
 basically do the same, but the main difference is that sleep will not keep the rid,
-while sleep2 does.
+while sleep2 does. Also sleep2 will stop the script if there is none unit attached.
```
secretdataz: There is "no" unit attached.

src/map/script.c
```diff
@@ -18410,7 +18410,7 @@ BUILDIN_FUNC(sleep)
 }
 
 /// Pauses the execution of the script, keeping the unit attached
-/// Returns if the unit is still attached
+/// Stop the script if none unit attached
 ///
 /// sleep2(<mili secconds>) -> <bool>
```
Lemongrass3110: Should be `sleep2(<milli seconds>)` and remove the `-> <bool>` part.

Issue #2173, opened 2017-05-28
mapwarp to savepoint is not working


jezznar:
* **rAthena Hash**:  4d126047e8e7664a56057eaef61a916029b34af1

* **Client Date**:  2015-11-05

* **Server Mode**:  re (prolly in pre also)

* **Description of Issue**: 
  * Result: mapwarp to savepoint is not working
`mapwarp .EventMap
EventMap$,"SavePoint",0,0;`
![image](https://cloud.githubusercontent.com/assets/13676977/26529609/8e33a26a-43f5-11e7-9beb-9cb4d8f9a4dd.png)

  * Expected Rewarp should be able to warp players to their savepointsavepoints
  * How to Reproduce: just use the mapwarp as stated above
  * Official Information: N/A
  

* **Modifications that may affect results**:  N/A

Atemo:
mapwarp to savepoint is ~~not working~~ __not implemented__
```*mapwarp "<from map>","<to map>",<x>,<y>{,<type>,<ID>};

This command will collect all characters located on the From map and warp them 
wholesale to the same point on the To map, or randomly distribute them there if 
the coordinates are zero. "Random" is understood as a special To map name and 
will mean randomly shuffling everyone on the same map.

Optionally, a type and ID can be specified. Available types are:

 0 - Everyone
 1 - Guild
 2 - Party

Example:

// Will warp all members of guild with ID 63 on map prontera to map alberta.
	mapwarp "prontera","alberta",150,150,1,63;
```
But yes we may add this feature.

jezznar:
Thanks @aleos89!

PR #2209, opened 2017-06-20
Cleaned up the warp functions

aleos89:
* **Addressed Issue(s)**: #2173

  * Created a central warp function to check for Random, SavePoint, and Save arguments.
  * Script commands areawarp, warppartner, warpwaitingpc, and areawarp now support Random, SavePoint and Save as destinations.
Thanks to @jezznar!

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -5635,12 +5635,44 @@ BUILDIN_FUNC(rand)
 	return SCRIPT_CMD_SUCCESS;
 }
 
-/*==========================================
- * Warp sd to str,x,y or Random or SavePoint/Save
- *------------------------------------------*/
+/**
+ * Warp character based on given parameters
+ * @param sd: Player data
+ * @param map: Map name/Warp type
+ * @param x: X location
+ * @param y: Y location
+ * @return 0 on success and failure otherwise
+ */
+static int buildin_warp_sub(struct map_session_data *sd, const char *map, int x, int y) {
+	int ret = 0;
+
+	nullpo_retr(1, sd);
+
+	if (strcmp(map, "Random") == 0)
+		ret = pc_randomwarp(sd, CLR_TELEPORT);
+	else if (strcmp(map, "SavePoint") == 0 || strcmp(map, "Save") == 0)
+		ret = pc_setpos(sd, sd->status.save_point.map, sd->status.save_point.x, sd->status.save_point.y, CLR_TELEPORT);
+	else {
+		int16 index = 0;
+
+		if (!(index = mapindex_name2id(map)))
+			return 1;
+
+		ret = pc_setpos(sd, index, x, y, CLR_OUTSIGHT);
+	}
+
+	if (ret)
+		ShowError("buildin_warp_sub: Moving player '%s' to \"%s\",%d,%d failed.\n", sd->status.name, map, x, y);
+
+	return ret;
+}
+
+/**
+ * Warp a player
+ * warp "map name",x,y{,char id};
+ */
 BUILDIN_FUNC(warp)
 {
-	int ret;
 	int x,y;
 	const char* str;
 	struct map_session_data* sd;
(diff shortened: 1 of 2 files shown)


src/map/script.c
```diff
@@ -5635,12 +5635,44 @@ BUILDIN_FUNC(rand)
...
-/*==========================================
- * Warp sd to str,x,y or Random or SavePoint/Save
- *------------------------------------------*/
+/**
+ * Warp character based on given parameters
+ * @param sd: Player data
+ * @param map: Map name/Warp type
+ * @param x: X location
+ * @param y: Y location
+ * @return 0 on success and failure otherwise
```
Lemongrass3110: After fixing #2210 this function should also return enum e_setpos.

src/map/script.c
```diff
@@ -5635,12 +5635,44 @@ BUILDIN_FUNC(rand)
...
+ */
+static int buildin_warp_sub(struct map_session_data *sd, const char *map, int x, int y) {
+	int ret = 0;
```
Lemongrass3110: Value initialization is useless in this case.

src/map/script.c
```diff
@@ -5699,13 +5728,19 @@ static int buildin_areawarp_sub(struct block_list *bl,va_list ap)
 			j++;
 		} while( map_getcell(m,tx,ty,CELL_CHKNOPASS) && j < max );
 
-		pc_setpos((TBL_PC *)bl,index,tx,ty,CLR_OUTSIGHT);
+		if (buildin_warp_sub((TBL_PC *)bl, str, tx, ty))
```
Lemongrass3110: As mentioned above it would be nice to move this part of code directly into the newly created sub function.

src/map/script.c
```diff
@@ -5652,36 +5684,33 @@ BUILDIN_FUNC(warp)
...
 	index = va_arg(ap,unsigned int);
 	x2 = va_arg(ap,int);
 	y2 = va_arg(ap,int);
 	x3 = va_arg(ap,int);
 	y3 = va_arg(ap,int);
+	str = va_arg(ap,char *);
 
-	if(index == 0)
-		pc_randomwarp((TBL_PC *)bl,CLR_TELEPORT);
+	if (index == 0)
```
Lemongrass3110: This part of code should be removed completely and leave it to the new sub function.

src/map/script.c
```diff
@@ -13117,58 +13147,56 @@ BUILDIN_FUNC(failedremovecards) {
...
+	y = script_getnum(st, 5);
+
+	if (script_hasdata(st, 7)){
+		check_val = script_getnum(st, 6);
+		check_ID = script_getnum(st, 7);
 	}
 
-	if((m=map_mapname2mapid(mapname))< 0)
-		return SCRIPT_CMD_SUCCESS;
+	if ((m = map_mapname2mapid(mapname)) < 0)
```
Lemongrass3110: Mapname is validated here again. SavePoint will not be able to pass.

src/map/script.c
```diff
@@ -13117,58 +13147,56 @@ BUILDIN_FUNC(failedremovecards) {
...
-					if(p->data[i].sd && p->data[i].sd->bl.m == m){
-						pc_setpos(p->data[i].sd,index,x,y,CLR_TELEPORT);
-					}
+			if (p) {
+				for (i = 0; i < MAX_PARTY; i++) {
+					if (p->data[i].sd && p->data[i].sd->bl.m == m)
+						buildin_warp_sub(p->data[i].sd, str, x, y);
 				}
 			}
 			break;
```
Lemongrass3110: Maybe we could add clan support here as well?

src/map/script.c
```diff
@@ -13117,58 +13147,56 @@ BUILDIN_FUNC(failedremovecards) {
...
-BUILDIN_FUNC(mapwarp)	// Added by RoVeRT
+/**
+ * Warp a given map
+ * mapwarp "<from map>","<to map>",<x>,<y>{,<type>,<ID>};
+ * @author [Reddozen], [RoVeRT]; improved by [Lance]
+ */
+BUILDIN_FUNC(mapwarp)
 {
-	int x,y,m,check_val=0,check_ID=0,i=0;
+	int x, y, m, check_val = 0, check_ID = 0, i = 0;
```
Lemongrass3110: While we touch it: It would be nice to match the variable names with the function description. type vs check_val and ID vs check_id. Also it would be nice if we would turn the type into an enum once again and export it into the script engine.

src/map/script.c
```diff
@@ -13356,11 +13388,15 @@ BUILDIN_FUNC(warppartner)
 
 	mapindex = mapindex_name2id(str);
```
Lemongrass3110: Name validation once again - no passing through for SavePoint.

src/map/script.c
```diff
@@ -13356,11 +13388,15 @@ BUILDIN_FUNC(warppartner)
...
 	if (mapindex) {
-		pc_setpos(p_sd,mapindex,x,y,CLR_OUTSIGHT);
-		script_pushint(st,1);
+		if (buildin_warp_sub(p_sd, str, x, y))
+			script_pushint(st, 0);
```
Lemongrass3110: Turn it into true/false maybe?

src/map/script.c
```diff
@@ -13356,11 +13388,15 @@ BUILDIN_FUNC(warppartner)
 
 	mapindex = mapindex_name2id(str);
 	if (mapindex) {
-		pc_setpos(p_sd,mapindex,x,y,CLR_OUTSIGHT);
-		script_pushint(st,1);
+		if (buildin_warp_sub(p_sd, str, x, y))
```
Lemongrass3110: Invert the check and remove the outer else this way the two pushint 0 can be merged.

src/map/script.c
```diff
@@ -5635,12 +5635,44 @@ BUILDIN_FUNC(rand)
...
+ * @param sd: Player data
+ * @param map: Map name/Warp type
+ * @param x: X location
+ * @param y: Y location
+ * @return 0 on success and failure otherwise
+ */
+static int buildin_warp_sub(struct map_session_data *sd, const char *map, int x, int y) {
+	int ret = 0;
+
+	nullpo_retr(1, sd);
```
Lemongrass3110: Nullpo check for mapname?

lighta:
I know I'm annoying with this, but does this support multi-map server ?
I remenber there was an issue with some warp function with it. it will be good to try with 2 map-serv see if that central function solve it as well.

Zarbony:
My Teleport Skill (Fly Wing, Teleport...)
Warps me Randomly to various fields and Dungeons, but not on Randomarea an the Map.

But i came from Prontera by using teleport lv1 or Fly wing to rickridge and then to magma dun and so on but not randomly in prontera :X

Issue #1734, opened 2016-11-20
Status changes on NPCs


Lemongrass3110:
The clan leader NPCs all have the status change SC_CLANINFO with their respective clan id to show the clan icon just like on players below them.

For players we currently have visible status changes that need to be sent out(`clif.c::clif_efst_status_change_sub`) to other players stored in the map_session_data in the following fields:
```
struct sc_display_entry **sc_display;
unsigned char sc_display_count;
```

This is required for NPCs now too.

cydh:
oh u talked about this back then, it's fine by me. add those 2 vars to `struct npc_data`, but don't forge the `eri *pc_sc_display_ers;`

aleos89:
Implemented in 45e91e2.

PR #2221, opened 2017-06-27
Added clan emblems to the NPCs

Lemongrass3110:
* **Addressed Issue(s)**: #1734

* **Server Mode**: Both

Partial merge of/based on idathena/trunk@1f84c5d
Thanks to @exneval

Added a missing NPC(Clan Helper)
Unified npc creation(npc_create_npc)
Changed the StatusDisplayType table from bool to a bitmask of BL types
Added support for storing and sending these kinds of status changes to NPCs

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -1484,7 +1484,7 @@ int clif_spawn(struct block_list *bl)
 				clif_spiritcharm(sd);
 			if (sd->status.robe)
 				clif_refreshlook(bl,bl->id,LOOK_ROBE,sd->status.robe,AREA);
-			clif_efst_status_change_sub(sd, bl, AREA);
+			clif_efst_status_change_sub(bl, bl, AREA);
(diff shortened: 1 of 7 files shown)


src/map/clif.c
```diff
@@ -6054,30 +6056,44 @@ void clif_status_change(struct block_list *bl, int type, int flag, int tick, int
  * @param bl: Objects walking into view
```
aleos89: Don't forget to change sd comment to tbl in the line above.

src/map/npc.c
```diff
@@ -2390,6 +2400,21 @@ bool npc_viewisid(const char * viewid)
 	return true;
 }
```
aleos89:
```c
/**
 * Create a bare NPC object.
 * @param m: Map ID
 * @param x: X location
 * @param y: Y location
 * @return npc_data
 */
```

PR #2281, opened 2017-07-23
Fix swapped accsessories

Temtaime:
Enum was swapped, you can view it by calling @itemlist with one acc. equiped.
Also i didn't have a modern client to test EQP_SHADOW_ACC_R to correctness

### DIFF
--- a/src/common/mmo.h
+++ b/src/common/mmo.h
@@ -205,8 +205,8 @@ enum equip_pos {
 	EQP_ARMOR            = 0x000010, // 16
 	EQP_SHOES            = 0x000040, // 64
 	EQP_GARMENT          = 0x000004, // 4
-	EQP_ACC_L            = 0x000008, // 8
-	EQP_ACC_R            = 0x000080, // 128
+	EQP_ACC_R            = 0x000008, // 8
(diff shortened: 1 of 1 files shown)


aleos89:
Are you sure you're not confusing yourself with the orientation of the equip window and how you're viewing it? The character is facing yourself so that means the left-hand and right-hand are mirrored to yourself.

Temtaime:
There's a left and right hand with a shield and weapon respectively. And they're named with "L-hand" and "R-hand".
And yes, i'm sure. I faced this bug developping a RO client using constants from rathena.

aleos89:
Ok, I did some testing around and you're right. But along with this change, there are two items that need to be flipped in the item database to match this change. They need to be adjusted to:
```
28310,Sarah's_Left_Earring,Sarah's Left Earring,4,10,,100,,,,0,0xFFFFFFFF,63,2,8,,145,0,,{ skill "AL_HEAL",1; },{},{}
28311,Sarah's_Right_Earring,Sarah's Right Earring,4,10,,100,,,,0,0xFFFFFFFF,63,2,128,,145,0,,{ skill "AL_TELEPORT",1; },{},{}
```
If you could make those changes in your PR we can push this. :)

PR #2322, opened 2017-08-08
Inter server C++ conversion

secretdataz:
* **Server Mode**: Both

This pull request converts inter.c to C++.
inter_server.conf which contain storage information is also converted to YAML format (to inter_server.yml) and has import folder support added.


src/char/inter.h
```diff
@@ -4,20 +4,25 @@
...
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 #include "../common/conf.h"
 #include "../common/mmo.h"
 #include "../common/sql.h"
 
+#ifdef __cplusplus // C codes can't see this
```
Lemongrass3110: Can't those 3 ifdefs be merged together?

karlseven:
Why is it is says invalid storage ID after we update to latest revision?

Issue #1950, opened 2017-02-07
Homunculus Thinks Target Is Unreachable


dragonis1701:
* **rAthena Hash**: Current
* **Client Date**: 2013-08-07
* **Server Mode**: Pre-renewal, probably happens in renewal, as well.
* **Description of Issue**: I made #1907 but I was wrong in my assumptions in that thread. This is very easy to reproduce with Lif. What happens is, when a Homunculus goes to attack a target, it randomly chooses one of eight cells surrounding the target. That part works perfectly fine. The issue arises when the master of the Homunculus is on the cell the Homunculus has chosen to walk to. When the Homunculus chooses a cell the master is on, the Homunculus believes the cell is unreachable and then stops moving. If the player walks, thus moving off the target cell, the Homunculus will start to move again because the cell became reachable once the master was no longer on the cell.

If you want to reproduce this, summon a lif homunculus, attack a target with no weapon equipped, then see what happens with Lif. If she walks up and attacks the target, she did not choose the cell the master is on. If she stands still, she chose the cell the master is on. You can then walk a few cells to make Lif fall out of range of the master so she must break target and follow the master, then she will check for another cell around the monster. This bug only has a 1/8 chance of occuring, so it may take a few attempts.

* **Modifications that may affect results**: I have nocellstack disabled but it occurs with it is enabled, as well.

aleos89:
Do you happen to know what the official behavior is? I would assume the Lif should continue to find a new cell around the target still (while not trying to stack with master).

dragonis1701:
Unfortunately, no, but I assume it would be as you said.

Lemongrass3110:
@Playtester any info on that?

PR #2325, opened 2017-08-08
Updated homunculus walk behavior

aleos89:
* **Addressed Issue(s)**: #1950

  * Updated homunculus walk behavior to closer mimic official behavior.
  * Homunculus will now stop in a random 3x3 area around the master.
  * Homunculus will no longer get stuck when looking for an empty cell when supporting the master during attacks.
Thanks to @dragonis1701!

### DIFF
--- a/src/map/clif.c
+++ b/src/map/clif.c
@@ -14629,7 +14629,7 @@ void clif_parse_HomMoveToMaster(int fd, struct map_session_data *sd){
 	else
 		return;
 
-	unit_calc_pos(bl, sd->bl.x, sd->bl.y, sd->ud.dir);
+	unit_calc_pos(bl, sd->bl.x, sd->bl.y);
 	ud = unit_bl2ud(bl);
 	unit_walktoxy(bl, ud->to_x, ud->to_y, 4);
 }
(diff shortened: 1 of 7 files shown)


dragonis1701:
This bug still occurs. It's easiest to see when you have Lif as your Homunculus and then go to wolves, pay_fild02. When you attack a wolf, Lif will not always move in to attack it. There is also when the wolves "assist" Lif will not move to attack the second wolf and will only attack the second wolf if Lif does not have to move into range. So, it sets the target but does not actually move.

aleos89:
Hm, I can go back and do some more adjustments then. As you stated, when I tried to reproduce it after the changes I couldn't get my Lif stuck. I'll see what I can do.

aleos89:
As for the chances of the homunculus getting stuck, the only thing I can think of is the client isn't sending a request to move the homunculus again.

I was able to reproduce it by making myself stand between a standing monster and my homunculus. I'm not sure there's anything else I can do at this point to fix that, unless @Playtester has a suggestion. He knows more about the walk code than I do.

Lemongrass3110:
It tells me the video is not available. Was it taken down already?

Could you maybe try to set up a script for us to test with?
Some script like setting cells to non-walkable and spawn an immobile mob so we have a high chance of reproducing your issue?

I will try to set up some walk packet debugging code in the meantime.

PR #2363, opened 2017-08-22
Quest_db overhaul

Atemo:

* **Server Mode**: Both

* **Description of Pull Request**: This PR converts quest_db.txt to yml format.


anacondaqq:
This is very very very very very very bad format of data right now for quest.yml.
Not readable, confusing, very uninformative, and ugly.
Current quest_db.txt looks much better and more clear.
Recommendations will post later in comments. 
Because this PR looks like "we change it not because it needed, but because we can do it, and nobody care if after 1 year will come other guy who love for example libconfg format and will return it back to libconf, or even csv". 

Do not get me wrong, but that what i was afraid, after seeing achievements_db.yml, refine_.yml
You are planning convert all dbs to yml, but many of such dbs ARE VERY BAD for yml format. and libconfig will be much better then yml. Plus it's extra 30-40 issues at issue tracker via each of such new conversion between formats, and a lot of administrators pain with highly customized servers. 

😭 😭 😡

Atemo:
>Not readable, confusing, very uninformative, and ugly.

Ha I thought this format was fine. We can see the list of quest ID like in libconfig. The list of field (ID, Quest Name) is closer to aegis format. There is also additionnal comment about the field unlike the old quest_db.

>Because this PR looks like "we change it not because it needed, but because we can do it, and nobody care if after 1 year will come other guy who love for example libconfig format and will return it back to libconf, or even csv".

We (actually I) propose to change quest_db in yaml format because currently it's a pain in the ass to add additionnal condition (or more mob to hunt for your custom server).
For example in episode 16.1 some hunting quest are completed when killing mob race X and level Y. But yeah we can add (again) more commas in our current quest_db..

Lemongrass3110:
@anacondaqq we plan to phase out all libconfig databases to drop libconfig from our source. I have not looked at the format yet, but if you have a better proposal for the new structure you are free to share it with us. But one thing for sure: libconfig will be dropped from rAthena.

anacondaqq:
No problem, I will propose a little bit different structure in the current quest_db.yml, because what I saw, not good to be clear names, and structures (sorry, please try to get me correctly, because me and many others will work with it on their own servers), and if you decide to convert db formats to yml, ok, no problem, but lets do it beautifully, and a little bit more readable. Suggestions will post not soon, but I guess quest_db.yml will not be released in that time (i do not ask to freeze development for that, or anything else (who the f**k am I to tell devs what they need to do, right?), just will review it a little bit later.

Jeybla:
@anacondaqq could you explain in detail what you actually dislike about the yaml format?

comparison:
```
10109,0,1154,10,0,0,0,0,0,0,0,0,0,0,0,0,0,"Sphinx dungeon - Pasana"
```
vs
```yml
- Quest_Infos: [10109, "Sphinx dungeon - Pasana"]
  Hunt: [1154,10]
```
I'd personally prefer the yaml format. Because you don't need to think about the current column and don't have all those empty cols zeroed.

anacondaqq:
@Atemo Please do not take offense at me, if I seem rude to you or have told you something boorish (or something is not correct).

@Jeybla 
I don't said that i do not like yml, i said i do not like how it explained in quest_yml. 

Look, main purpose of converting current format (csv) to yml to make a file readable without jumping or knowing about settings.

You showed the very great example of a code, but let me explain in detail where are problems.
The problem in confusion part of how some "things" named, they do not help to understand what is going on, they just more confused, because a lot of quests can have the same ID as a mob for example, and the mob can look like a quest_id and vice versa.

Examples
I think such format is bad:
```
# 2010 Headgear Quests
  - Quest_Infos: [1100, "Solo in the Sphinx Dungeon!"]
    Hunt: [1178,20]
  - Quest_Infos: [1101, "Soloing Sphinx Dungeon!"]
    Hunt: [1164,20]
  - Quest_Infos: [1102, "Soloing Clock Tower!"]
    Hunt: [1194,40]
  - Quest_Infos: [1103, "Soloing Clock Tower!"]
    Hunt: [1213,30]
  - Quest_Infos: [1104, "Solo at Louyang!"]
    Hunt: [1519,20]
  - Quest_Infos: [1105, "Solo at Louyang!"]
    Hunt: [1513,50]
  - Quest_Infos: [1106, "Solo at Amatsu Dungeon!"]
    Hunt: [1375,20]
  - Quest_Infos: [1107, "Solo at Amatsu Dungeon!"]
    Hunt: [1403,40]
  - Quest_Infos: [1108, "Solo at Louyang!"]
    Hunt: [1631,20]
```
Why? because there are no readable part, you trying to minize a space in format, which created to make things readable. For what? And why?

Why not try to do something like that:
```yml
# Type1
  - quest_name: "Solo in the Sphinx Dungeon!"
    quest_id: 1100
    monster_id: 1178
    monster_amount: 20
  - quest_name: "Soloing Sphinx Dungeon!"
    quest_id: 1100
    monster_id: 1164
    monster_amount: 20

# Or type2
  - id: 1100
    name: "Solo in the Sphinx Dungeon!"
    hunt_mobid: 1178
    hunt_amount: 20
  - id: 1100
    name: "Soloing Sphinx Dungeon!"
    hunt_mobid: 1164
    hunt_amount: 20

```

The idea in keeping each quest stand-alone and separate. And avoid filds with numbers or name what number must be inside a field.

Jeybla:
I understand what you're saying. I think Atemo paired the quest ID and -name into a single line, because every quest needs both.
For the `Hunt` approach I must say that Atemos approach is better for multiple monsters.

secretdataz:
Here's my take
```yaml
- Id: 1234
  Name: "I hunt, I collect, I wait!"
  Hunt:
    - Id: 1178
      Amount: 20
   - Id: 1179
      Amount: 20
  ExtraDrops: # Some quests add extra drops to certain target monsters
    - MobId: 1178
      Drops:
        - ItemId: 987
          Rate: 100 # 1%
        - ItemId: 988
          Rate: 200 # 2%
  Timers:
    - Limit: 3600
    - AnotherTimerFeaturekROUse: 7200
  # Expand here instead of adding comma and more zeroes
```

Lemongrass3110:
@secretdataz make it MobId in Hunt too? That's one of the points @anacondaqq mentioned, because you do not know if it is a quest, mob or item id...

For Timers I would suggest something like this:
```
TimeLimit:
  Variable: 3600 # seconds from the start of the quest
  Absolute: <time> # maybe HH:MM as string?
```

secretdataz:
@Lemongrass3110 Yeah I forgot about that Id naming. We could use that absolute timer format like you suggested.

Issue #1788, opened 2016-12-12
Double Attack/Triple Attack/Fear Breeze update


secretdataz:
As of a recent kRO update, Double Attack, Triple Attack, and Fear Breeze can now critically strike.
Source: http://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=6726&curpage=1

feltenc:
Looking forward to this implementation!

Badarosk0:
Waiting anxiously. Ahahaha

Felleonel:
Good change, waiting for this.

Questune09:
is this already available??? how can i update it??

Jeybla:
@Questune09 please use the forum for support. This issue is in progress, but you can merge the related pull request if you like to. 
(Any other unrelated questions will be removed from now on)

mazvi:
Need update :)

BrOgBr:
There's already some time left in the kRO.
We need it!

Lemongrass3110:
@BrOgBr you are free to create a pull request.

savagemao:
does it work?

Miraakol:
maybe this one can help https://github.com/idathena/trunk/commit/49f428a635a98189e17c8de43e94bc807d97c1f7#diff-c861cb68eadb0408c3c20bb633904e69

MiYanni:
@secretdataz @aleos89 I looked through the [commit](https://github.com/rathena/rathena/commit/08d160aaac8cdd286bbafb32532abe42b8a42376) that was marked as completing this issue. However, there is no code there to update this for *Raging Trifecta Blow* (aka *Triple Attack*). Is that intentional to not have criticals for that skill, or was it already working prior to that commit? If it isn't working currently, I'd vote to reopen the issue.

attackjom:
That skill can't critical in official server

MiYanni:
@attackjom Ah, the issue description is to blame, then. I looked at the post linked and yes, it doesn't list Triple Attack. My bad. 😅 It says (using Google Translate):
> - Changed so that critical hits are applied when some continuous strike skills that are activated as passive are activated.
>　→ Skill List: Double Attack, Chain Action, Shadow Warrior, Six Fist Kwon, Fear Breeze

The only one I didn't recognize was Six Fist Kwon. Searching using the Korean text, I arrive here: https://ro.gnjoy.com/guide/runemidgarts/skillview.asp?lineseq=6&skillid=371
[...]

PR #2365, opened 2017-08-23
According to #1788 double atk, triple atk and fear breeze can now crit

Milchdealer:
* **Addressed Issue(s)**: #1788 

* **Server Mode**: RE

* **Description of Pull Request**: Basically just adding the skills to the crit check


Badarosk0:
I did the test and still did not get any results in the double critic.

aleos89:
I don't think this is the correct way to implement this. I believe Aegis added a new damage type for this case. I've tried playing around with it before but the crit effect wasn't working properly due to some other issues in our code.

Badarosk0:
How sad. I also tried to make some adjustments, but I was not successful ...

PR #2368, opened 2017-08-24
Added script command sc_end_class

aleos89:

  * Added a new script command `sc_end_class` to remove status changes from any learned skills.
  * Adjusted the custom Reset NPC to make use of this behavior.
Thanks to @sader1992!

### DIFF
--- a/src/map/script.c
+++ b/src/map/script.c
@@ -11321,9 +11321,9 @@ BUILDIN_FUNC(sc_end)
 
 /**
  * Ends all status effects from any learned skill on the attached player.
- * sc_end_class {<GID>};
+ * sc_end_class {<char_id>};
  */
-BUILDIN_DEF(sc_end_class)
+BUILDIN_FUNC(sc_end_class)
 {
 	struct map_session_data *sd;
 	uint16 skill_id;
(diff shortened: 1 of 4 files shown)


sader1992:
nice :X a lot easier
I got wroing

```
pc.c: In function ‘pc_resetskill’:
pc.c:7353:10: warning: declaration of ‘skill_id’ shadows a previous local [-Wshadow]
   uint16 skill_id = skill_idx2id(i);
          ^
pc.c:7295:9: note: shadowed declaration is here
  int i, skill_id, skill_point=0;
         ^
```
can it be a different command script (so it can be used on other scripts)
like example cs_class_end; 
and it can be used without reseting the skills if we want to make the npc disable all the class buffs(if it's easy)

sader1992:
Work perfect but this `BUILDIN_DEF(sc_end_class)` need to be `BUILDIN_FUNC(sc_end_class)`
no warning or errors

src/map/script.c
```diff
@@ -11319,6 +11319,29 @@ BUILDIN_FUNC(sc_end)
 	return SCRIPT_CMD_SUCCESS;
 }
 
+/**
+ * Ends all status effects from any learned skill on the attached player.
+ * sc_end_class {<GID>};
```
Lemongrass3110: CID or char id

src/map/script.c
```diff
@@ -11653,7 +11676,7 @@ BUILDIN_FUNC(resetskill)
 	TBL_PC *sd;
 	if (!script_charid2sd(2,sd))
 		return SCRIPT_CMD_FAILURE;
-	pc_resetskill(sd,1);
+	pc_resetskill(sd, 1);
```
Lemongrass3110: .

Issue #2352, opened 2017-08-18
Bug Itemshop when try buying with bounded item


chadn4u:
* **rAthena Hash**: 8310253

* **Client Date**: 20151104a

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: cost item are not reduced but the item sold can still be obtained
  * Expected Result: cost item should be reduced
  * How to Reproduce: 
1. Create npc with itemshop and use any item as cost item(for this case im using Poring Coin(7539))
2. And then make two items based on the cost item, 1 for non bound item and 1 for bound item(for this case im using @atcommands to make items: `@item 7539 1` `@itembound 7539 1 1` )
3. Try to buy something by using the items in the npc itemshop you created earlier.

  * Official Information:
  

* **Modifications that may affect results**:  None.

dheerajsyz:
can post your shop script here?

chadn4u:
`prontera,156,179,4	itemshop	Porco Shop	436,7539,501:2`

aleos89:
I'm not aware of any bound restrictions when deleting items. I'll try to run a test later today.

PR #2373, opened 2017-08-25
Cleaned up Item Shop checks

aleos89:
* **Addressed Issue(s)**: #2352

  * Removed an extra bound check when selling items.
  * Added a check to return failure when unable to delete an item.
  * Added checks when counting sellable items in an Item Shop.
    * Items that can't be sold, favorite items, rental items, and bound items will now be ignored unless the GM has required permissions.
Thanks to @chadn4u!

### DIFF
--- a/src/map/npc.c
+++ b/src/map/npc.c
@@ -1412,19 +1412,24 @@ static enum e_CASHSHOP_ACK npc_cashshop_process_payment(struct npc_data *nd, int
 			{
 				struct item_data *id = itemdb_exists(nd->u.shop.itemshop_nameid);
 
-				if (cost[0] < (price - points)) {
-					char output[CHAT_SIZE_MAX];
+				if (id) { // Item Data is checked at script parsing but in case of item_db reload, check again.
+					if (cost[0] < (price - points)) {
+						char output[CHAT_SIZE_MAX];
 
-					memset(output, '\0', sizeof(output));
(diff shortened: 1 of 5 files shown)


src/map/npc.c
```diff
@@ -4795,3 +4802,28 @@ void do_init_npc(void){
...
 	// End of initialization
 }
+
+/**
+* Check if the player can sell the current item
+* @param sd map_session_data of the player
+* @param item struct of the checking item.
+* @return bool 'true' is sellable, 'false' otherwise
+*/
+bool npc_can_sell_item(struct map_session_data * sd, struct item * item) {
```
Lemongrass3110: Just from the description this sounds more like a pc_ function than a npc_ function.

src/map/npc.c
```diff
@@ -1412,19 +1412,24 @@ static enum e_CASHSHOP_ACK npc_cashshop_process_payment(struct npc_data *nd, int
 			{
 				struct item_data *id = itemdb_exists(nd->u.shop.itemshop_nameid);
 
-				if (cost[0] < (price - points)) {
-					char output[CHAT_SIZE_MAX];
+				if (id) { // Item Data is checked at script parsing but in case of item_db reload, check again.
```
Lemongrass3110: Maybe invert that if to keep the changes small and make the flow more readable.

Issue #2372, opened 2017-08-25
RequiredEquipment does not work


M4karov:
* **rAthena Hash**: 9db3c386ead6ea58efc44cdb14883b510e8562a3

* **Client Date**: 20170614

* **Server Mode**: Pre-Renewal / Renewal

* **Description of Issue**: RequiredEquipment does not work even if you do not have the item in the inventory, it is still possible to use the skill.
  * Result: 
  * Expected Result: 
  * How to Reproduce: I tested using the NC_HOVERING skill in both pre-renewal and renewal mode.
  * Official Information: [Link](http://irowiki.org/wiki/Hover)
  

* **Modifications that may affect results**: None

PR #2374, opened 2017-08-25
Fixes required equipment check for skills

aleos89:
* **Addressed Issue(s)**: #2372

  * Required equipment check was allowing the skill to cast without the proper equipment.
Thanks to @M4karov!


M4karov:
I thank you for the excellent work. The correction was very agile.

Jeybla:
Are there any skills which require only one of a list of items?
`// Wearing an applicable item.`
This sounds like it was intentionally written to require only one of the items.

M4karov:
Yes, but when it comes to equipment as a requirement, it did not work even though it was just one item.

M4karov:
When using the ability he does not recognize the name of the equipment needed to continue the use. How should I correct this?

image

aleos89:
The reason it shows that is because the `clif_skill_fail` format has changed a bit and needs to be updated for newer clients.

src/map/skill.c
```diff
@@ -15623,20 +15623,49 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
...
+				case NC_SELFDESTRUCTION:
+				case NC_SHAPESHIFT:
+				case NC_EMERGENCYCOOL:
+				case NC_MAGNETICFIELD:
+				case NC_NEUTRALBARRIER:
+				case NC_STEALTHFIELD:
+					if (pc_search_inventory(sd,reqeqit) < 0) {
+						count--;
+						if (!count) {
+							clif_skill_fail(sd,skill_id,USESKILL_FAIL_NEED_EQUIPMENT,reqeqit<<16);
```
Lemongrass3110: Probaby this should use:
`clif_skill_fail(sd,skill_id,USESKILL_FAIL_NEED_ITEM,reqeqit<<16|1);`

src/map/skill.c
```diff
@@ -15623,20 +15623,49 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
...
-					return false;
-				}
-			} else
-				break; // Wearing an applicable item.
+				break; // Skill has no required item(s); get out of here
+			switch(skill_id) { // Specific skills require multiple items while default will handle singular cases
+				case NC_PILEBUNKER:
+					if (!pc_checkequip2(sd,reqeqit,EQI_ACC_L,EQI_MAX)) {
+						count--;
+						if (!count) {
```
Lemongrass3110: Since up at the top you have !reqeqit this can cause problems if there is a zero in between valid items.
Nevertheless you could get rid of the counting and use something like this:
`if( ( i + 1 ) < require.eqItem_count && !require.eqItem[i+1] )`
With this you can peek the next entry and display the error message in that case.
Would have the same problem with zeros in between though.

aleos89: During the parsing process of `skill_require_db` we check whether these item ID are valid or not. So we shouldn't have any problems at this point.

PR #2375, opened 2017-08-28
Create vs-make-server.bat

mazvi:
An easy way to compile a server with visual studio on windows like make server on linux. 
Based on ideas: @anacondaqq 
https://rathena.org/board/topic/111269-guide-lightweight-compiler-for-windows-for-replacement-heavy-visual-studio-ide-compile-rathena-faster-save-your-bandwidth-and-disk-space/

Readme.md will be update if this pull accepted

* **Addressed Issue(s)**: 

* **Server Mode**: 


Akkarinage:
This should use the Windows PATH variable to look for VS installations. I also don't see much point in this since it's designed to save ~30 seconds waiting for VS to load and a user to mouse-click twice.

I'm pretty sure the IDE shows additional output for warnings/errors than the MSBuild tool does if executed from command line by itself. I could be wrong though.

mazvi:
Yes, true PATH in bat file should be changed depending on installation location. Therefore there should be a change of readme later, to support this file.

For error messages I do not think I found it, and the server can run.

PR #2381, opened 2017-08-29
Adjusted script command progressbar_npc behavior

aleos89:

  * Allow players to stay attached.
  * Use Work In Progress value to restrict while progress bar is active.


attackjom:
This has a important in ep17.1 script++

but, I think it will be good if we have optional to detach or attach to character.

JohnnyPlayy:
It is also very important in 17.2

src/map/script.cpp
```diff
@@ -21014,6 +21014,7 @@ BUILDIN_FUNC(progressbar)
  * progressbar_npc "<color>",<seconds>{,<"NPC Name">};
  */
 BUILDIN_FUNC(progressbar_npc){
+	map_session_data *sd;
```
Atemo, line 21017:
```suggestion
	map_session_data *sd = map_id2sd(st->rid);
```

src/map/script.cpp
```diff
@@ -21042,8 +21043,10 @@ BUILDIN_FUNC(progressbar_npc){
 			return SCRIPT_CMD_FAILURE;
 		}
 
-		// detach the player
-		script_detach_rid(st);
+		if (script_rid2sd(sd)) { // Player attached - keep them from doing other things
```
Atemo, line 21046:
```suggestion
		if (sd) { // Player attached - keep them from doing other things
```

src/map/script.cpp
```diff
@@ -21055,6 +21058,11 @@ BUILDIN_FUNC(progressbar_npc){
 	// Second call(by timer after sleeping time is over)
 	} else {
 		// Continue the script
+		if (script_rid2sd(sd)) { // Player attached - remove restrictions
```
Atemo, line 21061:
```suggestion
		if (sd) { // Player attached - remove restrictions
```

Issue #1796, opened 2016-12-13
Autolooting prevents item_flag broadcast


whupdo:
I set 512,16 in item_flag.txt
If you use @ alootid +512
This is the time to get 512 items
Will not broadcast normally

Git hash : https://github.com/rathena/rathena/commit/93339212bcb9ee58907870e42ab0d93217aef3d6

aleos89:
Are you in a party? What is the item share set to? According to this comment:
```
// Somehow, if party's pickup distribution is 'Even Share', no announcement
```
there won't be a broadcast if it's set to Even Share.

whupdo:
sorry my English is not good
I mean I set 512,16
It will be broadcast normally when auto picking is not set
If there is a set @ alootid +512 will not broadcast

Akkarinage:
Confirmed.
Using git hash c6c8d0bc8884f683f7d66c31e741e0a3031c5e12 in RE mode, 20151104 client.

Adding Jellopy (909) to import/item_flag.txt displays announcement when picking a Jellopy off the floor.

After adding to autoloot list `@alootid +909` I collect the Jellopy from a monster but there is no announcement. I am not in a party.

Probable offending code: https://github.com/rathena/rathena/blob/master/src/map/mob.c#L2091

PR #2392, opened 2017-08-30
Corrected broadcast item flag with autoloot

aleos89:
* **Addressed Issue(s)**: #1796

  * Moved the broadcast flag check before party share loot check.
Thanks to @whupdo!


aleos89:
I tested this in-game before creating the PR.

PR #2400, opened 2017-09-03
Adds pk_mode_mes displayed on joining a pk zone.

Jeybla:

* **Server Mode**: both

When pk_mode and pk_mode_mes is enabled, the player gets notified joining a pk-zone map.
Suggestion by @RadianFord


src/map/clif.c
```diff
@@ -10004,6 +10004,23 @@ static bool clif_process_message(struct map_session_data* sd, bool whisperFormat
...
+ * (Only enabled in pk_mode)
+ * @param sd
+ * @return void
+ *------------------------------------------*/
+inline void clif_pk_mode_message(struct map_session_data * sd)
+{
+	if (battle_config.pk_mode && battle_config.pk_mode_mes &&
+	    sd && map[sd->bl.m].flag.pvp) {
+		// 407: You've entered the PK Zone.
+		clif_disp_overhead_(&sd->bl, msg_txt(sd,407), BC_SELF);
```
Jeybla: I'd like to use a different display mode, because this one spams the chat window. Is there a display_overhead option which will not be displayed in the chat window?

RadianFord: How come im getting error on this part? `clif_disp_overhead_(&sd->bl, msg_txt(sd,407), BC_SELF);`

Jeybla: Can you specify the error?

RadianFord:
```
clif.c:10150:3: warning: implicit declaration of function ‘clif_disp_overhead_’ [-Wimplicit-function-declaration]
   clif_disp_overhead_(&sd->bl, msg_txt(sd,500), BC_SELF);
```

and

```
/tmp/ccCDv4ka.ltrans18.ltrans.o: In function `clif_parse_LoadEndAck.part.76.ltopriv.592':
<artificial>:(.text+0xb13b): undefined reference to `clif_disp_overhead_'
collect2: error: ld returned 1 exit status
Makefile:69: recipe for target 'map-server' failed
make[1]: *** [map-server] Error 1
make[1]: Leaving directory '/home/wickedro/wickedz/src/map'
Makefile:51: recipe for target 'map' failed
make: *** [map] Error 2
```

Jeybla: Shouldn't be a implicit declaration. The definition (https://github.com/rathena/rathena/blob/feature/pk_mode_mes/src/map/clif.c#L9566) is in the same file some lines above the call (https://github.com/rathena/rathena/blob/feature/pk_mode_mes/src/map/clif.c#L10018). Also the declaration is in the header (https://github.com/rathena/rathena/blob/feature/pk_mode_mes/src/map/clif.h#L834) Are you sure you merged it correctly?

Jeybla: I'm sorry I already tested it. It' definitely an issue on your side. Also I'm using `clif_disp_overhead_` not `clif_disp_overhead` and your call is in a different line (`clif.c:10150`), pretty sure you're not testing on this branch. Please try to checkout this clean branch. If that works you should sanitize your merge.

conf/msg_conf/map_msg.conf
```diff
@@ -423,7 +423,7 @@
 404: War of Emperium SE is currently in progress.
 405: War of Emperium SE has been ended.
 406: War of Emperium SE is currently not in progress.
-//407 free
+407: You've entered the PK Zone.
```
aleos89: `You've entered a PK Zone.

src/map/clif.c
```diff
@@ -10004,6 +10004,23 @@ static bool clif_process_message(struct map_session_data* sd, bool whisperFormat
 	return true;
 }
 
+/* clif_pk_mode_message
+ * Displays a message if the player joins a PK Zone
+ * (Only enabled in pk_mode)
+ * @param sd
+ * @return void
+ *------------------------------------------*/
```
aleos89:
```c
/**
 * Displays a message if the player enters a PK Zone (during pk_mode)
 * @param sd: Player data
 */
```

Jeybla:
I still wonder if there is a option to see a `display_overhead` on self without adding it to the chat. I'm sure I used such a function in a script before :/

Atemo:
@Jeybla `showscript` https://github.com/rathena/rathena/blob/master/doc/script_commands.txt#L3846

Lemongrass3110:
Just an idea, but would it make sense to respect `pk_min_level` as well?

Jeybla:
@Lemongrass3110 didn't thought about that. Good point. But perhaps a little annoying if the player levels up in that area and gets killed without even noticing he's in a PK zone? Not sure what would be better. Depends on the server in my opinion.

PR #2418, opened 2017-09-13
update item group file location

sader1992:
* **Addressed Issue(s)**: 

* **Server Mode**: 
both

just update the file location

### DIFF
--- a/doc/item_group.txt
+++ b/doc/item_group.txt
@@ -40,6 +40,8 @@ The table below explains which fields are accessed in each.
 
 GroupID: See the "Item Group ID" section in 'src/map/itemdb.h'.
          Supports IG_* constants.
+	and "Item Group ID" section in 'src/map/script_constants.h'.
+	 Supports export_constant(IG_* constants.
 
 ---------------------------------------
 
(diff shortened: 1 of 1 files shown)


sader1992:
i remember that there was #define MAX_ITEMGROUP option in the source files but i did't find it :x

Jeybla:
@sader1992 7f4a7ba26f8cfd8f929ddc2402f774c7b7294c82 Itemgroup has been changed to DBMap in that commit

doc/item_group.txt
```diff
@@ -40,6 +40,8 @@ The table below explains which fields are accessed in each.
 
 GroupID: See the "Item Group ID" section in 'src/map/itemdb.h'.
          Supports IG_* constants.
+	and "Item Group ID" section in 'src/map/script_constants.h'.
+	 Supports export_constant(IG_* constants.
```
aleos89:
```
         See the "item groups" section in 'src/map/script_constants.h".
         Supports export_constant(IG_*) constants.
```

sader1992: do you mean remove the first line ? (See the "Item Group ID" section in 'src/map/itemdb.h'.)
if not i think for the second line it's better to start it with (and) so they know that there is two files must be edited to add new IG

Issue #2440, opened 2017-09-22
Sorcerer's Elemental Cure should heal for casters HP/SP%


Barigas:
I know, wording here is a bit unclear, but on officials Sorcerer's Elemental Cure removes 10% of caster's hp and sp, and gives this amount to the summoned spirit. So, if caster has 20000 hp and 1500sp, it should give 2000hp and 150sp to the elemental.
Here is not mine video from iRO, look at 7:40 timing at Agua sp, it becomes full in 2-3 casts.
https://www.youtube.com/watch?v=KCXxPMcSIXQ&feature=youtu.be&t=7m40s
To fix this, it is enough to change:
Line 10332 in skill.c 
status_heal(&ed->bl,e_hp,e_sp,3);
change to
status_heal(&ed->bl,s_hp,s_sp,3);
or somehow like this.
Revision: any

aleos89:
For future reports, could you try to fill in the template that's provided when you open your issue. Thanks.

PR #2443, opened 2017-09-22
Corrected Elemental Cure healing behavior

aleos89:
* **Addressed Issue(s)**: #2440

  * The HP/SP heal should be based on the caster's current HP/SP, not the Elemental's max HP/SP.
Thanks to @Barigas!

### DIFF
--- a/src/map/skill.c
+++ b/src/map/skill.c
@@ -10319,17 +10319,20 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case SO_EL_CURE:
 		if( sd ) {
 			struct elemental_data *ed = sd->ed;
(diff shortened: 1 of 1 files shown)


src/map/skill.c
```diff
@@ -10319,17 +10319,20 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case SO_EL_CURE:
 		if( sd ) {
 			struct elemental_data *ed = sd->ed;
-			int s_hp = sd->battle_status.hp * 10 / 100, s_sp = sd->battle_status.sp * 10 / 100;
```
Jeybla: Weird, that it was actually calculated two times oO

PR #2461, opened 2017-09-29
Fix to compile in C89 mode

rpdigos:

  * Fix to compile in C89 mode as the loop initial declarations are only allowed in C99 mode


rpdigos:
The error message from gcc:
  error: 'for' loop initial declarations are only allowed in C99 mode
  note: use option -std=c99 or -std=gnu99 to compile your code

Jeybla:
Why should we fix issues for C89 mode when we require a C++11 compiler?

rpdigos:
Hello Jeybla,
Because it was compiling before and the costs nothing to merge the changes since it's already done.

Issue #2415, opened 2017-09-11
[custom] RDMOPT_DEC_SPELL_CAST_TIME View Bug


uddevil:
* **rAthena Hash**: 3cc29d04cc04817f3ffcf6c67cc8c27e6be53c61

* **Client Date**: 2015-11-04a RagexeRE

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: 
- Difference item description when using RDMOPT_DEC_SPELL_CAST_TIME Random Option
- This is when viewed from my own inventory : 
![cacad bin_2017-09-11_22-30-55](https://user-images.githubusercontent.com/10105610/30283239-ba5bb144-9741-11e7-9aa8-074c1c3057d6.png)
- This is when viewed from other people inventory (using Shift+Click on item)
![cacad bin_2017-09-11_22-29-02](https://user-images.githubusercontent.com/10105610/30283263-c933b0f4-9741-11e7-9858-0c0a8763fe78.png)

  * Expected Result: 
- Shouldn't it have the same script description? (Reduce variable cast time +-5%)

  * How to Reproduce: 
1. db/const.txt :
`RDMOPTG_CCD_PlaceHolder2	102`

2. db/import/mob_drop.txt :
`1002,1680,10000,RDMOPTG_CCD_PlaceHolder2 //Poring test with 100% drop channce`

3. db/import/item_randomopt_group.txt:
`RDMOPTG_CCD_PlaceHolder2,500,RDMOPT_DEC_SPELL_CAST_TIME,-5,0,RDMOPT_CLASS_MDAMAGE_BOSS,10,0`

4. kill poring until you have that 1680 (Crimson Rod)
5. view in your inventory, it has the real description (like in 2nd screenshot above)
6. give it to other people, ask them to shift click it
7. click the link in chat window
8. the description is different now (0%)

  * Official Information:
  
There's no official info, since its custom

* **Modifications that may affect results**:

uddevil:
![navicat_2017-09-15_09-28-52](https://user-images.githubusercontent.com/10105610/30463839-57c26348-99f8-11e7-9ec6-bce134fde210.png)
i don't know if its relevant, but is it because of the db is in unsigned smallint? I just realized that if i put negative number on the script, the db will be saved as 0

secretdataz:
Item link showing negative values as 0 seems to be a client issue which we can fix. Although the database signedness issue has been fixed in ea28ff4. Thanks

PR #2465, opened 2017-10-01
Fixed random options with negative values being stored as 0 in database.

secretdataz:
* **Addressed Issue(s)**: #2415
* **Server Mode**: Both
This pull request removes the `unsigned` modifier for random option columns in rAthena's MySQL script. Upgrade scripts are also provided.


Jeybla:
Looks fine. Just for clarifications: When dies these values become negative?

secretdataz:
@Jeybla It can be seen in the referenced issue.

Jeybla:
Ah, thanks. Didn't see that.

secretdataz:
@Jeybla Since the client could display the -5 value for the item holder just fine, I concluded that both the client and the server should support signed values and the zero value is just an item link bug.

PR #2469, opened 2017-10-02
Fixes spawn infos after mobdb reload.

### LINKED ISSUE #2360: @whereis missing after @reloadmobdb
* **rAthena Hash**: 6f985080a8b071e5869679e430274a4b166971c2

* **Client Date**: 2015-11-04aRagexeRE

* **Server Mode**: Renewal

* **Description of Issue**: 
`@whereis` not showing spawn location after `@reloadmobdb`
  * Result: 
`@whereis` not showing spawn location after `@reloadmobdb`
  * Expected Result: 
`@reloadmobdb` should not affect shown spawn on `@whereis`
  * How to Reproduce: 
1. do command `@reloadmobdb`
2. command `@whereis` after that will show "Monster does not spawn normally' on all monster.
  * Official Information:
  

### LINKED ISSUE #2460: Taekwon Mission Monster Change Issue after @reloadmobdb ( Issue #2360 )
* **rAthena Hash**: c71f10508325fdd8a97a1b07ba28b845950c51b0 and up...

* **Client Date**: 2015-09-16

* **Description of Issue**: Taekwon using Taekwon Mission with only one monster: Scorpion which is the first available monster in db ( **cause of @reloadmobdb** )
  * Result: only one monster used for Taekwon Mission: Scorpion, even with change mission;
  * Expected Result: random monster from db, even with 1% chance of change monster with 0 kills;
  * How to Reproduce: create Taekwon, use skill ``Taekwon Mission``, make @reloadmobdb - you will get this problem.
  * Official Information
  

  Using @reloadmobdb command;
#2360

Jeybla:
* **Addressed Issue(s)**: #2360 and #2460

* **Server Mode**: Both

* Fixes whereis after mobdb reload. Fixes #2360
* Fixes Taekwon Mission after mobdb reload. Closes #2460
* Moves the monster spawn infos from the mob_db struct to an own
  data structure. Sadly it's not directly accessible from the mob_db
  struct anymore, but this will be possible in the future, when we're
  able to easily add getter methods to structs. It's now accessible via
  mob_get_spawn() and to just check if a spawn point is available just use
  mob_has_spawn().
* Also this adds a typedef for the MobID. I recognized most functions
  are either int or unsigned short without a common norm. I suggest to
  change all MobID usage into the typedef in a latter commit.
* Thanks to @uddevil, @mgksrt8 and @aleos89!

Tests done: (need to be re-tested)
- [x] `@whereis` on multiple monsters after some `reloadmobdbs` and `reloadscripts` (fixed in this pr)
- [x] `@whereis` cumulated monster spawns (new cumulate function)
- [x] `Taekwon Mission` after mobdb reload (fixed in this pr)
- [x] `@spawn`ed some monsters (due to searchname changes)
- [x] `@mi` tested monster info on some monsters (due to searchname changes)
- [x] respawn of fixed spawn positions (this pr shouldn't have any impact on it)


ecdarreola:
Hi @Jeybla , can you update this pr? Tested this before cpp migration and working fine.

Jeybla:
@ecdarreola sure, I'll update it.
Since we can fully use C++ features on the map server now, I'm able to improve it a bit.

src/map/mob.cpp
```diff
@@ -66,6 +70,16 @@ struct mob_db *mob_dummy = NULL;	//Dummy mob to be returned when a non-existant
...
 
+// Monster Spawn informations
+class MobSpawns {
+public:
+	std::vector<struct spawn_info> m_spawns;
+	MobSpawns(struct spawn_info new_spawn);
+	void add_spawn(struct spawn_info new_spawn);
+};
+// unordered map mob_id -> MobSpawns
+std::unordered_map<MobID, MobSpawns> mob_spawn_data;
```
Jeybla: Use vector directly as second map value, remove the class `MobSpawns`

src/map/atcommand.cpp
```diff
@@ -7828,24 +7829,26 @@ ACMD_FUNC(whodrops)
 
 ACMD_FUNC(whereis)
 {
-	struct mob_db *mob_array[MAX_SEARCH];
-	int count;
-	int i, j, k;
+	MobID mob_ids[MAX_SEARCH] = {0};
+	int count = 0;
+	const int MAX_SPAWN_DISP = 10;
```
Jeybla: Should we remove that? It's currently not used and just displays all spawns.

aleos89: Yeah I think the MAX_SEARCH value is what the true cap is.

lighta:
I was prefering your map version. Way less code, it was probably easier to understand / maintain.
The 3rd element of map is a comparator, which could be used to sort your map. 
So you can sort them by qty like this.

Jeybla:
@lighta I thought the same. But from what I've read, it is bad practice to sort maps by their value. I could use a set, but the update would look equally bad and it would be just an unnecessary overhead like the map version. Sometimes it's preferable to use very simple containers

PR #2500, opened 2017-10-10
Migrated skill.c to C++

aleos89:
* **Addressed Issue(s)**: N/A

  * Needed in preparation for future projects.


src/map/skill.cpp
```diff
@@ -5876,8 +5880,7 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 		if( src->type == BL_ELEM ) {
 			struct elemental_data *ele = BL_CAST(BL_ELEM,src);
 			struct status_change *tsc_ele = status_get_sc(&ele->bl);
-			sc_type type = status_skill2sc(skill_id), type2;
-			type2 = type-1;
+			sc_type type = status_skill2sc(skill_id), type2 = static_cast<sc_type>(type - 1);
```
lighta: I'll put this on 2 line

src/map/skill.cpp
```diff
@@ -3038,14 +3042,14 @@ void skill_attack_blow(struct block_list *src, struct block_list *dsrc, struct b
...
-			if (skill_blown(dsrc,target,blewcount,dir,BLOWN_NO_KNOCKBACK_MAP|BLOWN_MD_KNOCKBACK_IMMUNE|BLOWN_TARGET_NO_KNOCKBACK|BLOWN_TARGET_BASILICA) < blewcount)
+			if (skill_blown(dsrc,target,blewcount,dir,(enum e_skill_blown)(BLOWN_NO_KNOCKBACK_MAP|BLOWN_MD_KNOCKBACK_IMMUNE|BLOWN_TARGET_NO_KNOCKBACK|BLOWN_TARGET_BASILICA)) < blewcount)
 				skill_addtimerskill(src, tick + status_get_amotion(src), target->id, 0, 0, LG_OVERBRAND_PLUSATK, skill_lv, BF_WEAPON, flag|SD_ANIMATION);
 			break;
 		case SR_KNUCKLEARROW:
 			// Ignore knockback damage bonus if in WOE (player cannot be knocked in WOE)
 			// Boss & Immune Knockback stay in place and don't get bonus damage
 			// Give knockback damage bonus only hits the wall. (bugreport:9096)
-			if (skill_blown(dsrc, target, blewcount, dir_ka, BLOWN_IGNORE_NO_KNOCKBACK|BLOWN_NO_KNOCKBACK_MAP|BLOWN_MD_KNOCKBACK_IMMUNE|BLOWN_TARGET_NO_KNOCKBACK|BLOWN_TARGET_BASILICA) < blewcount)
+			if (skill_blown(dsrc, target, blewcount, dir_ka, (enum e_skill_blown)(BLOWN_IGNORE_NO_KNOCKBACK|BLOWN_NO_KNOCKBACK_MAP|BLOWN_MD_KNOCKBACK_IMMUNE|BLOWN_TARGET_NO_KNOCKBACK|BLOWN_TARGET_BASILICA)) < blewcount)
```
lighta: you don't need "enum" keyword. and static_cast is faster. (cause at compilation instead runtime).
=> static_cast<e_skill_blown>(blaballbal)

src/map/map-server.vcxproj
```diff
@@ -248,7 +248,9 @@
       <CompileAs>CompileAsCpp</CompileAs>
     </ClCompile>
     <ClCompile Include="searchstore.c" />
-    <ClCompile Include="skill.c" />
+    <ClCompile Include="skill.cpp">
+      <CompileAs>CompileAsCpp</CompileAs>
```
lighta: you don't need this since is named .cpp now.

src/map/skill.cpp
aleos89: Static cast won't work in this case because it's a bitmask enum. We have to convert it to an `enum class` to properly support this enum.

lighta: I may not have understand well what you mean.
BitmaskEnum or not the cast will work. here a sample with my explanation:
http://cpp.sh/3hbhf
I'm not that found over enum class. It's actually the otherway arround, if you do an enum class you can't play with bitmask like that cause they will complain. So you usually have to cast them at the end to use them... (if you do mask with them that it).
The only nice thing that I like with enum class is that the enumeration value do not collide with another enum, like it's the current case without.

PR #2502, opened 2017-10-12
Added mailmap file to coalesce together commits.

kisuka:
Added mailmap file to link SVN commits to current github users who are still contributing to some degree.

Also links commits together where the user had a bunch of different emails used during commits.
ie: Lighta : p


Jeybla:
Two questions about that:
1. Shouldn't we add a explicit name for each email owner?
2. Can we automate the creation/modification of that file?

GodLesZ:
Thx for the ping, @Lemongrass3110.
You may use `godlesz_at_hotmail.de` as e-mail for me.

Lemongrass3110:
Thanks for the mail! :-)
Haven't heard from you in a long time. Nice to see you are still around.

PR #2515, opened 2017-10-16
Jobmaster: Third Class Non Trans Option

mazvi:
// .ThirdClassNonTrans
//	= Disable it if you want player MUST BE rebirth class first to get job 3
//	= Enable it if you want player with job 2 have choice get job 3 or job Rebirth

This feature is for server need long story line so people must became rebirth first if want to get job 3

* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/npc/custom/jobmaster.txt
+++ b/npc/custom/jobmaster.txt
@@ -18,7 +18,7 @@
 //= 1.6 Added option to get job related equipment on change. [Braniff]
 //= 1.7 Readability changes. Also added BabyExpanded and BabySummoner classes. [Jey]
 //= 1.8 Added option to disable Baby Novice Only but Baby Class can be Enabled [mazvi]
-//= 1.9 Added option ThirdClassNonTrans please read Setting additional Note [mazvi]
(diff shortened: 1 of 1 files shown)


Akkarinage:
This file has received a alot of attention over the last few weeks. Considering its a custom script and has nothing to do with the official job scripts, perhaps attention would be better utilised elsewhere.

mazvi:
i just want to help. because there are also private servers who want to use long stories in the journey job. but if it is felt this is not necessary I apologize. because it's too excited.

So what sould i do? Detete this pull? Iam sory for my fault..

Jeybla:
https://github.com/rathena/rathena/commit/2676d05394008288a882c8966a4d6da3865d0eaa#commitcomment-24824085

Already told here that I don't see this option needed in our repository. If anyone else is asking for it, we may add it later, but I think you're current the only one who's looking for this option. So for everyone else this is just confusing.
Anyway, thanks for your contributions :)

mazvi:
Don't we need this? Because job 4 can't be with regular job 3, it has to be job 3 trans :D

Issue #2497, opened 2017-10-10
SO_VACUUM_EXTREME behavior not correct


esu1214:
rAthena Hash:
7351b88
Client Date: 20170125
Server Mode: renewal
Description of Issue:
Players will be sucked to the center of UNT_VACUUM_EXTREME if nearby in Aegis server.
But players only stop walk instead in rA.

aleos89:
Confirmed also that players with no knock back items (IE: Strong Shield) are still sucked into the center.

PR #2526, opened 2017-10-17
Corrected Vacuum Extreme behavior

aleos89:
* **Addressed Issue(s)**: #2497

  * Vacuum Extreme will now pull players into the center.
Thanks to @esu1214!


cydh:
Talk about the issue, since he said "nearby" I just think there are some conditionals that make player will be sucked.
So here's my test, player will be sucked if
1. Player that has walk's destination is in Vacuum area. (n-1 x n-1)
2. The Vacuum area is placed near the player

and the way u removed the uni knockback check will suck every players/objects in no knockback map such GVG castles

PR #2535, opened 2017-10-21
Enabled import for translations by default

Lemongrass3110:

* **Server Mode**: Both

Right now the translations are the only configurations that are not imported by default.
Added possibility to override MAP_MAX_MSG with defines_pre.h


Daegaladh:
Please, add the possibility to override LANG_ENABLE (msg_conf.h) with defines_pre.h

Lemongrass3110:
Will check it later today :) there are a lot of defines that should be checked if they are defined already

PR #2541, opened 2017-10-22
Mapflag Nomemo: Novice Island

mazvi:
Added Novice Island RE Maps. [mazvi]

* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/npc/re/mapflag/nomemo.txt
+++ b/npc/re/mapflag/nomemo.txt
@@ -9,6 +9,7 @@
 //= 1.2 Added dali, dali02 [Nova]
 //= 1.3 Added Nightmare Glastheim [Ridley8819]
 //= 1.4 Added GVG TE Maps. [Cydh]
+//= 1.5 Added Novice Island RE Maps. [mazvi]
 //============================================================
 
 // Episode 13 ====================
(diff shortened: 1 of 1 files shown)


Atemo:
Actually there is no npc to warp to iz_int / int_land maps, only news novices should be on the maps. Novices can't use /memo so adding **nomemo** doesn't seem useful.

We could eventually add **nowarpto** mapflag to prevent `@warp` command (note **nowarpto** disable /memo too) but even if others class are warped in there I don't see potential exploit.

mazvi:
okay thanks but why on PRE novice training ground there was nomemo flag

cydh:
>Actually there is no npc to warp to iz_int / int_land maps, only news novices should be on the maps. Novices can't use /memo so adding nomemo doesn't seem useful.

this is why Aegis was (or still?) abused

mazvi:
Yes like on official idRO aegis.. mamy rumor other job can back to training ground... i dont know how other job can back to traning ground. i think that is hoax..

Lemongrass3110:
Well I guess we could simply merge this since it does not really hurt. Additionally @aleos89 will clean those up soon anyway.

Atemo:
May you change **nomemo** to **nowarpto** mapflag (in the proper file) @mazvi ?

mazvi:
Closed move to: https://github.com/rathena/rathena/pull/2566

PR #2542, opened 2017-10-24
Update warper npc

sader1992:
* **Addressed Issue(s)**: Update

* **Server Mode**: both

* **Description of Pull Request**:  updating the warper npc

- Adding maps
1. Towns:
- [x] lasagna

2. Fields:
- [x] lasa_fild01
- [x] lasa_fild02
- [x] moc_fild 4,5,6,8,9,10,14,15

3. Dungeons:
- [x] lasa_dun01
- [x] lasa_dun02
- [x] lasa_dun03

4. Castles:
- [x] Kafragarten 1,2,3,4,5
- [x] Gloria 1,2,3,4,5

5. Guild_Dungeons:
- [x] Kafragarten > teg_dun01
- [x] Gloria > teg_dun02

6. Instances:
- [x] Eclage Interior
- [x] Faceworms Nest
- [x] Ghost Palace
- [x] Horror Toy Factory
- [x] Sarah and Fenrir Instance
- [x] Sara Memory
- [x] Geffen Magic Tournament
- [x] Devil Tower

7. Special:
- [x] Dimensional Rift (dali)
- [x] Para Market

8. Options:
- [x] Add setting to disable Satan Morroc maps and enable the previous morroc maps
- [x] Add setting to enable only the first Field map in the list
- [x] Add setting to enable only the first Dungeon map in the list

9. **TODO:** (maybe for another pull request)
Making the script generate the options OnInit ?
Make the Restrict function remove the option instead of showing it with a message
can add maps to Restrict function OnInit

### DIFF
--- a/npc/custom/warper.txt
+++ b/npc/custom/warper.txt
@@ -12,7 +12,7 @@
 //= 1.4 Added new Guild Dungeons.
 //= 1.4a Slight edits.
 //= 1.4b Added Wolfchev's Laboratory warp.
-//= 1.5 Added Lasagna ,Para Market ,WOE TE ,Instances and setting for old morroc maps [sader1992].
+//= 1.5 Added Lasagna ,Para Market ,WOE TE ,Instances and settings [sader1992].
 //============================================================
 
 -	script	Warper	-1,{
@@ -80,7 +80,16 @@ function Disp {
 function Pick {
 	set .@warp_block,@warp_block;
 	set @warp_block,0;
-	set .@select, select(@menu$);
+	if((@f && .OnlyFirstFld) || (@d && .OnlyFirstDun)){
+		set .@select,1;
+		if(.@warp_block){
+			while(.@warp_block & (1<<.@select)){
+				.@select += 1;
+			}
+		}
+	}else{
+		set .@select, select(@menu$);
+	}
 	if (getarg(0) == "") {
 		set .@i, .@select;
 		set .@map$, getarg(.@i);
@@ -95,6 +104,7 @@ function Pick {
 	set .@x, @c[.@i*2];
 	set .@y, @c[.@i*2+1];
 	deletearray @c[0],getarraysize(@c);
+	@f = false; @d = false;
 	Go(.@map$,.@x,.@y);
 }
 function Restrict {
@@ -172,6 +182,7 @@ T37: Go("yuno",157,51);
 // --------------------------------------------------
 	Fields:
 // --------------------------------------------------
+@f = true;
 menu	"Amatsu Fields",F1, "Ayothaya Fields",F2, "Bifrost Fields", F3,
     	"Brasilis Fields",F4, "Comodo Fields",F5, "Dewata Fields",F6,
     	"Eclage Fields",F7, "Einbroch Fields",F8, "El Dicastes Fields",F9,
@@ -266,6 +277,7 @@ F29: Restrict("Pre-RE",5,10);
 // --------------------------------------------------
 	Dungeons:
 // --------------------------------------------------
+@d = true;
 menu	"Abyss Lakes",D1, "Amatsu Dungeon",D2, "Anthell",D3,
     	"Ayothaya Dungeon",D4, "Beach Dungeon",D5, "Bifrost Tower",D41,
     	"Bio Labs",D6, "Brasilis Dungeon",D7, "Byalan Dungeon",D8, "Clock Tower",D9,
@@ -510,6 +522,8 @@ S10: Go("turbo_room",99,114);
 
 OnInit:
 	.Satan_Morroc = true;	//	false will enable moc_fild 4,5,6,8,9,10,14,15 while disable moc_fild 20,21,22 Default is true.
+	.OnlyFirstFld = false;	//	true will teleport to the first level of the Fields  Default is false.
+	.OnlyFirstDun = false;	//	true will teleport to the first level of the Dungeons  Default is false.
 }
 
 // --------------------------------------------------
--- a/npc/custom/warper.txt
+++ b/npc/custom/warper.txt
@@ -12,7 +12,7 @@
 //= 1.4 Added new Guild Dungeons.
 //= 1.4a Slight edits.
 //= 1.4b Added Wolfchev's Laboratory warp.
-//= 1.5 Added Lasagna ,WOE TE ,Instances and setting for old morroc maps [sader1992].
+//= 1.5 Added Lasagna ,Para Market ,WOE TE ,Instances and setting for old morroc maps [sader1992].
 //============================================================
 
 -	script	Warper	-1,{
@@ -492,7 +492,7 @@ I19: Restrict("RE");
 // --------------------------------------------------
 menu	"Auction Hall",S1, "Battlegrounds",S2, "Casino",S3, "Dimensional Rift",S4,
 		"Eden Group Headquarters",S5, "Gonryun Arena",S6, "Izlude Arena",S7, 
-		"Monster Race Arena",S8, "Turbo Track",S9;
+		"Monster Race Arena",S8, "Para Market",S9, "Turbo Track",S10;
 
 S1: Go("auction_01",22,68);
 S2: Go("bat_room",154,150);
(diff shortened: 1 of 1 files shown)


npc/custom/warper.txt
```diff
@@ -69,7 +70,19 @@ function Go {
 	set lastwarp$, getarg(0);
 	set lastwarpx, getarg(1,0);
 	set lastwarpy, getarg(2,0);
-	warp getarg(0),getarg(1,0),getarg(2,0);
+	for(.@i=0;.@i<getarraysize(.random_coordinates$);.@i++){
```
secretdataz: If the array’s size is static for every iteration, store the size in a variable instead.

npc/custom/warper.txt
```diff
@@ -69,7 +70,19 @@ function Go {
...
+		if(lastwarp$ == .random_coordinates$[.@i] && getarg(3,0) != 1){
+			set lastwarpx, 0;
+			set lastwarpy, 0;
+		}
+	}	
+	for(.@i=0;.@i<getarraysize(.blocked_maps$);.@i++){
```
secretdataz: If the array’s size is static for every iteration, store the size in a variable instead.

npc/custom/warper.txt
Lemongrass3110: This whole array is useless. Simply override the towns you want to randomize in the OnInit with zeroes. This way you do not have to compare it on each warp.

npc/custom/warper.txt
```diff
@@ -18,6 +18,7 @@
 //= 1.4 Added new Guild Dungeons.
 //= 1.4a Slight edits.
 //= 1.4b Added Wolfchev's Laboratory warp.
+//= 1.5 Added Lasagna and other Instances and settings
```
Lemongrass3110: Add your name in brackets.

npc/custom/warper.txt
```diff
@@ -3,7 +3,7 @@
 //===== By: ================================================== 
 //= Euphy
 //===== Current Version: =====================================
-//= 1.4b
+//= 1.5
```
Lemongrass3110: Would be nice to convert it to the new default header and getting rid of this and some other useless fields.

npc/custom/warper.txt
Lemongrass3110: Same here, block them in the OnInit code, maybe using negative coordinates?

npc/custom/warper.txt
```diff
@@ -424,47 +451,72 @@ G6: Go("schg_dun01",200,124);
...
+S5: Restrict("RE");
+	Go("moc_para01",31,14,1);
+S6: Go("gon_test",48,10,1);
+S7: Go("arena_room",100,88,1);
+S8: Go("p_track01",62,41,1);
+S9: Go("turbo_room",99,114,1);
+
+
+OnInit:
+	.Satan_Morroc = 1;	//	0 will enable moc_fild 4,5,6,8,9,10,14,15 wjile disable moc_fild 20,21,22 Default is 1
```
Lemongrass3110: while. and please try to use "true" and "false" instead of 1/0.

npc/custom/warper.txt
```diff
@@ -518,3 +570,4 @@ ecl_in01,51,60,4	duplicate(Warper)	Warper#ecl	811
 malangdo,134,117,6	duplicate(Warper)	Warper#mal	811
 malaya,231,204,4	duplicate(Warper)	Warper#ma	811
 mora,57,152,4	duplicate(Warper)	Warper#mora	811
+lasagna,196,187,4	duplicate(Warper)	Warper#las	811
```
Lemongrass3110: put it in alphabetical order

npc/custom/warper.txt
sader1992: the idea of this array is for the MVP maps while some servers block them and other make them warp-able to random coordinates i did try to see if there is way to check if a MVP-map didn't find that , and it maybe letting the server owner can choose the map him self would be better idea for some people would like to make the warp normal for orc maps and maybe disable some mini-bosses map like Valkyrie and Ghostring

Lemongrass3110: Yeah that's okay. I'm just saying that you dont need to do this logic on each warp.
Just do this parsing at the end of the OnInit event and you are fine. You parse the array once per reload and it should work the same way just being faster.

npc/custom/warper.txt
sader1992: here i am thinking if i can move this check to the Pick function so the blocked map wont be showing as an option for the players

npc/custom/warper.txt
sader1992: i think i did get what you mean , making the check load once on server reload and not every-time the player click on the npc right ?

Lemongrass3110: exactly. :)

npc/custom/warper.txt
```diff
@@ -424,47 +451,72 @@ G6: Go("schg_dun01",200,124);
...
+S6: Go("gon_test",48,10,1);
+S7: Go("arena_room",100,88,1);
+S8: Go("p_track01",62,41,1);
+S9: Go("turbo_room",99,114,1);
+
+
+OnInit:
+	.Satan_Morroc = 1;	//	0 will enable moc_fild 4,5,6,8,9,10,14,15 wjile disable moc_fild 20,21,22 Default is 1
+	//Warp the player to random Coordinates in the map .
+	setarray .random_coordinates$,"";	//	Example: setarray .random$,"prontera","Alberta",Aldebaran";
```
Atemo: I'm against this setting. Using 0 for warps coordinates is more productive.
`Go("payon",0,0);` instead of `Go("payon",179,100);` for random warp on payon.

npc/custom/warper.txt
```diff
@@ -424,47 +451,72 @@ G6: Go("schg_dun01",200,124);
...
+	//Block map from the warper .
+	setarray .blocked_maps$,"";	//	Example: setarray .blocked_maps$,"prontera","Alberta",Aldebaran";
```
Atemo: I'm against this setting. I don't understand the point of adding a special warper with maps blocked.
The user should just remove the warps entries from the script, the player shouldn't see theses maps in the menu.

Atemo:
Thanks for the update!

npc/custom/warper.txt
sader1992: all of the server owners who ask me to make (PRE-RE)server ask for Episode base server so the first configuration is removing Satan Morroc Episode and get back the old Morroc maps monsters and quests , so they want the players to farm for Valkyrie items etc before they install the Episode 12: Satan Morroc ,and so no one would have advantage farming Satan Morroc while the server still new and in low population.

Atemo: It seems legit, I agree.

sader1992:
i did not find a way to override the options in OnInit since the options loaded when the player talk to the npc , so i did remove the options for now and leave it for maybe a rewrite in the future

mazvi:
Can you added feature dungeon 1st level only (place after entrance) and field outside town only?

sader1992:
nice idea i will do that

mazvi:
I think the field settings, very complex.. for example the nearest field prontera there are 4 of them:

prt_fild08
prt_fild05
prt_fild06
prt_fild01

but is very complex configuration..

mazvi:
can you add feature price like healer, there was price configuration...

Town Price
Field Price
Dungeon Price

and more...

npc/custom/warper.txt
```diff
@@ -149,42 +155,45 @@ T14: Go("gonryun",160,120);
...
+T33: Go("splendide",201,147);
+T34: Go("thor_camp",246,68);
+T35: Go("umbala",97,153);
+T36: Go("veins",216,123);
+T37: Go("yuno",157,51);
 
 // --------------------------------------------------
 	Fields:
 // --------------------------------------------------
+@f = true;
```
Jeybla: I know it's the easiest solution. But most of the time you should not use char-variables in an NPC if you only need the variable in the same scope. Another solution to that would be for example another parameter for `Pick()` or another function like `PickDungeon()` (Which just calls Pick() with the dungeon parameter).

Anyway since the whole NPC is built on similar behavior (`setarray @c`, `@warp_block`), it should be fine. I just wanted to mention it for your future projects.

Issue #2127, opened 2017-05-08
Linux Compile Error


rofasindo:
Compile error, I'm using Linux Centos 6 32Bit with latest git rAthena today 2017/05/08 (svn 19694), error _./configure_ and cannot command _make server_

> checking whether byte ordering is bigendian... yes
configure: error: bigendian is not supported... stopping

> [root@vultr server]# make server
make: *** No rule to make target `server'.  Stop.

![error compile 1](https://cloud.githubusercontent.com/assets/24239587/25788896/adff951a-33d6-11e7-841e-5530b678210b.jpg)
![error compile 2](https://cloud.githubusercontent.com/assets/24239587/25788897/ae2c5884-33d6-11e7-96fa-3f13a4e35ce9.jpg)

Jeybla:
Please retry it by using `./configure --disable-64bit `

rofasindo:
@Jeybla it same problem on here

> checking whether byte ordering is bigendian... yes
configure: error: bigendian is not supported... stopping

Jeybla:
Mh.. This eventually is caused by 4a004ad4b83d2b46f56b4e57a8a6f3e487372ea6 .
Related: `https://github.com/bitcoin/bitcoin/issues/5177`

@secretdataz may has more info about this issue. You could try to checkout a3282c17305c7f0311d056a2eb6e4c85029b8ae9 .

secretdataz:
Adding `-fpermissive` to the `CXXFLAG` define in src/common/Makefile.in could be a workaround.

rofasindo:
adding like this?
> CXXFLAG =-std=c++11-fpermissive

still not work, while I have try to _**./configure --disable-64bit** or **./configure**_ and same error on 
> configure: error: bigendian is not supported... stopping

secretdataz:
CXXFLAG =-std=c++11 -fpermissive

rofasindo:
@secretdataz still same, configure not working.

secretdataz:
@rofasindo Can you try installing g++ package?

rofasindo:
@secretdataz thank you, just missing g++ package. Now fix, I can configure and make server. thanks again.

yerunkar11:
I am facing the same on centos, this time i have an gcc-c++ package installed.

NandoPrates:
i too same error in centos 6 but have Package gcc-c++-4.4.7-18.el6.x86_64 already installed and latest version

NandoPrates:
[...]

PR #2546, opened 2017-10-26
Add check in configure to ensure c++ compiler

lighta:
Hard fix for #2127


mazvi:
new configure not compatible with centos 7 64bit with maria db https://rathena.org/board/topic/104271-compiling-rathena-for-mariadb-101-in-centos-7-x64-configure-error-mysql-not-found-or-incompatible/

PR #2580, opened 2017-11-05
Update mob_avail.txt

Balferian:
* **Server Mode**: Pre/Re
Added view equipment **"robe"** to **mob_avail.txt**

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -4311,7 +4311,7 @@ static bool mob_readdb_mobavail(char* str[], int columns, int current)
 	mob_db_data[mob_id]->vd.class_ = sprite_id;
 
 	//Player sprites
-	if(pcdb_checkid(sprite_id) && columns==12) {
+	if(pcdb_checkid(sprite_id) && columns==13) {
 		mob_db_data[mob_id]->vd.sex=atoi(str[2]);
 		mob_db_data[mob_id]->vd.hair_style=atoi(str[3]);
 		mob_db_data[mob_id]->vd.hair_color=atoi(str[4]);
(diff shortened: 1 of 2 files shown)


src/map/mob.cpp
```diff
@@ -4331,7 +4332,7 @@ static bool mob_readdb_mobavail(char* str[], int columns, int current)
 #endif
 	}
 	else if(columns==3)
-		mob_db_data[mob_id]->vd.head_bottom=atoi(str[2]); // mob equipment [Valaris]
+		mob_db_data[mob_id]->vd.robe = atoi(str[2]); // mob equipment [Valaris]
```
Lemongrass3110: Why not pet equipment anymore, without updating documentation?

src/map/mob.cpp
```diff
@@ -4311,17 +4311,18 @@ static bool mob_readdb_mobavail(char* str[], int columns, int current)
 	mob_db_data[mob_id]->vd.class_ = sprite_id;
 
 	//Player sprites
-	if(pcdb_checkid(sprite_id) && columns==12) {
+	if(pcdb_checkid(sprite_id) && columns==13) {
```
Lemongrass3110: Might be better to change this to > 12 and add column 13 as optional inside with a default value.

Atemo:
What about a convertion in yaml format?

hannicaldummy:
I've made all these changes and my map-server keeps saying that the maximum colums at mob_avail is 12, anything? :/

aleos89:
I'm going to go ahead and close this. We have future plans of converting this to YAML which should allow for easier expansions.

Issue #2579, opened 2017-11-05
src/map/mob.hpp : No error when setting MAX_MOB_DB too high


uddevil:
* **rAthena Hash**: e7e9cf6d7333345d0e11d9f0bf9f1badb6ff3cd9

* **Client Date**: 2015-11-04a RagexeRE

* **Server Mode**: renewal

* **Description of Issue**: 
  * Result: 
    - map-server compilation finished without any error when setting `MAX_MOB_DB` too high (i'm using 81000)
    - then, when `map-server.exe` executed, it stops after loading `import/mob_drop.txt` without any error (i don't know if its really stopped, but i've waited about 5 minutes and still no response)
  * Expected Result: 
    - it should warn or error when `MAX_MOB_DB` is set too high
    - or at least there's error when `map-server.exe` executed
  * How to Reproduce: 
    - set `MAX_MOB_DB` on `src/map/mob.hpp` to 81000
    - recompile the server (I'm using vs 2015 CE)
    - execute `map-server.exe`
  * Official Information:
    - None as its custom
  

* **Modifications that may affect results**: 
    - So far i'm using `MAX_MOB_DB` 51000 and there's no problem.

psychonaut04:
Can you try this:

[src/map/mob.cpp](https://github.com/psychonaut04/rathena/commit/70fd19b93bddcca530f3f7c3a575fa8324f4dfa6)
[src/map/mob.hpp](https://github.com/psychonaut04/rathena/commit/b52d74cbf4936f3091b379313259bcb6cb3c6299
)
So it should be 51000 and anything more than that: error?

uddevil:
@psychonaut04 
It worked, and the compilation failed with error when i set the MAX_MOB_DB with 81000.
I don't really have / implement that many mobs though.

But that makes me wondering, what is the real maximum MAX_MOB_DB ? is it USHRT_MAX ?

psychonaut04:
I'm really not sure about that, can be something to try, but it can have a exact size.

psychonaut04:
If your USHRT_MAX is 51000(and if you can't put anything more than 51000) it can be that, so i can try to get the maximum value instead of 51000(cause it would be able change for another user). But i'm not sure from where that max comes.

Lemongrass3110:
As of b2ab972 MAX_MOB_DB is only used for ID checks and should not be touched anymore.
If you still think you need to adjust it, we will assume you know what you are doing and ignore any issues that result from it.
[...]

PR #2581, opened 2017-11-06
Fix to issue #2579: No error when setting MAX_MOB_DB too high

psychonaut04:
* **Addressed Issue(s)**: 

#2579

* **Server Mode**: 

Renewal(from issue), please check if it can apply to both.

struct mob_db *mob_db will return mob_dummy if **MAX_MOB_DB** is **51000** or more, plus compile will return error.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -64,7 +64,7 @@
 struct mob_db *mob_db_data[MAX_MOB_DB+1];
 struct mob_db *mob_dummy = NULL;	//Dummy mob to be returned when a non-existant one is requested.
 
-struct mob_db *mob_db(int mob_id) { if (mob_id < 0 || MAX_MOB_DB > 51000 || mob_id > MAX_MOB_DB || mob_db_data[mob_id] == NULL) return mob_dummy; return mob_db_data[mob_id]; }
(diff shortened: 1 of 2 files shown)


Lemongrass3110:
Having the ifdef check makes the other case unreachable code, since you will never be able to compile it like that.

psychonaut04:
Like, could only change one file? I got that before but still changed the other case just to be sure, but don't need i guess.

lighta:
The code look ok, but I don't really get, why does this have to be limited to 51k ?

Jeybla:
The mobid is mostly declared as `unsigned short` (max. 65535) or `short` (max. 32767) in the src. For the current code base I'd suggest to raise the error already at 32767. @psychonaut04 why did you decide to raise the error at 51000? @lighta it should be better to have an error message instead of simply stopping execution.

I think we'd need to refactor that some time anyway.

lighta:
well if it unsigned short, make the error when set over UINT16_MAX instead that would make more sens imo. (or INT16_MAX) if we worry about the wrap.

psychonaut04:
Oh can change it so, i got no clue about the max value, just did it using the value that the user talked about in the issue.

Jeybla:
@Lemongrass3110 is about to increase the maximum mobdb entries in #2604

Lemongrass3110:
Thanks for your contribution!

Sadly this is not needed anymore as @Jeybla already mentioned. As of b2ab972 MAX_MOB_DB is only used for ID checks and should not be touched anymore.

Looking forward to your next contribution though!

Issue #2478, opened 2017-10-05
Migrate login server completely to C++


Jeybla:
* **rAthena Hash**: 872cb42a480b2f52731b7bc8d7ff889db00b5cc0

As most of you already know, we are currently slowly shifting into complete C++ support. Step by step we're migrating map-server parts into C++ and it's foreseeable to migrate the map-server completely to C++ soon.

But before we actually use C++ features map-server wide I'd like to do some less important tests by migrating the login-server first. I don't want us to run into complicated issues with the map server, when we're able to test it on a less important and less big project like the login server.

Things to do (test):
- [x] Convert files to C++ format. (a587b91)
- [ ] Migrate login server to C++ completely
- [ ] Refactor some code in C++ and document benefits and issues
- [ ] Leave it for a while in the repository and wait for some user issues
- [ ] Eventually write a small plugin system for testing purposes (?)

After these test, we'll have a smother and more confident switch for the main (map-server) project.

sader1992:
plugins  ![smile](https://user-images.githubusercontent.com/5981261/31258801-117031a4-aa42-11e7-9710-ddcf303c5753.jpg)

Jeybla:
To clarify this: We are still far away from having plugins. And we should not go for a quick and bad solution. Please don't expect plugins in near future from us.

lighta:
@Jeybla, dunno if you already have check but I did convert all those module back then. you can found it there : https://github.com/lighta/rathena/tree/refactor/rA-cpp-linux-test
It's probably a bit outdated, altough I doubt char and login serv have change that much.

Jeybla:
Oh great, didn't know that! Looks really good.

The current plan was to slowly convert the map-server step by step. Examples:
https://github.com/rathena/rathena/pull/2435
https://github.com/rathena/rathena/pull/2466

But we are currently not able to use C++ features project wide (for example classes in header files). I just wondered when we are actually able to do that and which issues we may encounter. That's why I wanted to simulate it on the login-server. I'm not sure if we'll get the same "simulation effect" if we're using your code.

Maybe it'd be better to convert everything in a big coding session (one weekend with 2-4 developers involved). It also may be good to train our selfs to get on a equal level in C++ before we'll do that.

One thing is sure: We should concentrate first on making rathena compile-able and (nearly) bug-free in C++, before we refactor or reimplement things. We need to work with our community, because they will already hate us for having a massive amount of merge conflicts :D

[...]

PR #2583, opened 2017-11-07
Login-serv initial c++ version.

lighta:
ref : #2478

* **Addressed Issue(s)**: 

* **Server Mode**: 


src/login/account.cpp
```diff
@@ -517,25 +517,25 @@ static bool mmo_auth_fromsql(AccountDB_SQL* db, struct mmo_account* acc, uint32
...
-	Sql_GetData(sql_handle,  7, &data, NULL); acc->unban_time = atol(data);
-	Sql_GetData(sql_handle,  8, &data, NULL); acc->expiration_time = atol(data);
-	Sql_GetData(sql_handle,  9, &data, NULL); acc->logincount = (unsigned int) strtoul(data, NULL, 10);
-	Sql_GetData(sql_handle, 10, &data, NULL); safestrncpy(acc->lastlogin, data==NULL?"":data, sizeof(acc->lastlogin));
-	Sql_GetData(sql_handle, 11, &data, NULL); safestrncpy(acc->last_ip, data, sizeof(acc->last_ip));
-	Sql_GetData(sql_handle, 12, &data, NULL); safestrncpy(acc->birthdate, data==NULL?"":data, sizeof(acc->birthdate));
-	Sql_GetData(sql_handle, 13, &data, NULL); acc->char_slots = (uint8) atoi(data);
-	Sql_GetData(sql_handle, 14, &data, NULL); safestrncpy(acc->pincode, data, sizeof(acc->pincode));
-	Sql_GetData(sql_handle, 15, &data, NULL); acc->pincode_change = atol(data);
+	Sql_GetData(sql_handle,  0, &data, &out_len); acc->account_id = atoi(data);
```
Lemongrass3110: This results in 17 useless operations, since the value is not used.

src/login/logincnslif.hpp
```diff
@@ -7,13 +7,27 @@
  * @author rAthena Dev Team
  */
 
+#pragma once
 #ifndef CONSOLEIF_H
 #define	CONSOLEIF_H
 
 #ifdef	__cplusplus
```
Lemongrass3110: Shouldnt this simpl ybe removed?

src/login/CMakeLists.txt
```diff
@@ -7,8 +7,8 @@ set( LOGIN_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}  CACHE INTERNAL "" )
 #
 if( BUILD_SERVERS )
 message( STATUS "Creating target login-server" )
-file(GLOB LOGIN_HEADERS ${LOGIN_SOURCE_DIR}/*.h)
-file(GLOB LOGIN_SOURCES_C ${LOGIN_SOURCE_DIR}/*.c)
+file(GLOB LOGIN_HEADERS ${LOGIN_SOURCE_DIR}/*.hpp)
+#file(GLOB LOGIN_SOURCES_C ${LOGIN_SOURCE_DIR}/*.c)
```
Lemongrass3110: I think we could remove it completely. Guess we will not need it again.

src/login/account.cpp
aleos89: Agreed. This wasn't done in other cases, especially for the char-server. Is there a reason it needs to be done here?

lighta: only for string, but I didn't change those yet.
I'musing the std::string(const char*, size_t) constructor.

Lemongrass3110: Then I'd love to revert it as long as we dont use std::string. Might be overkill to do this in a preparation pull though.

PR #2586, opened 2017-11-07
Resolved a possible memory leak

aleos89:
* **Addressed Issue(s)**: N/A

  * When an invalid skill ID is given to a monster, the allocated skill object was not being freed.
Thanks to @Tokeiburu!


aleos89:
I'll go ahead and close this since we will be cleaning up the mob system.

PR #2607, opened 2017-11-15
Refactored achievements to utilize C++ features

aleos89:
* **Addressed Issue(s)**: N/A

  * Cleaned up the YAML parser.
  * Moved achievements from DBMap to an unordered_map.
  * Moved achievement targets from DBMap to a vector.
  * Changed all struct arrays into vectors.
  * Changed all char arrays to strings.
  * Removed achievement_dummy as it's no longer needed.


src/map/achievement.cpp
```diff
@@ -510,21 +498,17 @@ int *achievement_level(struct map_session_data *sd, bool flag)
...
  * Update achievement objectives.
- * @see DBApply
+ * @param sd: Player to update
+ * @param ad: Achievement data to compare for completion
+ * @param group: Achievement group to update
+ * @param update_count: Objective values player has
+ * @return 1 on success and false on failure
  */
-static int achievement_update_objectives(DBKey key, DBData *data, va_list ap)
+static int achievement_update_objectives(struct map_session_data *sd, struct s_achievement_db *ad, enum e_achievement_group group, int update_count[MAX_ACHIEVEMENT_OBJECTIVES])
```
lighta: Transfert that array via it's adress / ref instead ? (Could be wrap with std::array if we want to keep a fixed size)

src/map/achievement.cpp
```diff
@@ -595,15 +579,15 @@ static int achievement_update_objectives(DBKey key, DBData *data, va_list ap)
...
+			for (i = 0; i < ad->targets.size(); i++) {
 				if (objective_count[i] < ad->targets[i].count)
 					objective_count[i] += update_count[0];
 			}
 
 			changed = true;
 
-			ARR_FIND(0, ad->target_count, k, objective_count[k] < ad->targets[k].count);
-			if (k == ad->target_count)
+			ARR_FIND(0, ad->targets.size(), k, objective_count[k] < ad->targets[k].count);
```
lighta: std::find_if ?

src/map/achievement.cpp
```diff
@@ -971,208 +952,230 @@ struct av_condition *parse_condition(const char *p, const char *file, int line)
...
  * Reads and parses an entry from the achievement_db.
- * @param wrapper: The YAML wrapper containing the entry.
+ * @param node: YAML node containing the entry.
  * @param n: The sequential index of the current entry.
  * @param source: The source YAML file.
- * @return The parsed achievement entry or NULL in case of error.
+ * @return True on successful parse or false otherwise
  */
-struct achievement_db *achievement_read_db_sub(yamlwrapper *wrapper, int n, const char *source)
+bool achievement_read_db_sub(const YAML::Node &node, int n, const std::string &source)
```
lighta: The whole function should be try/catch I think

src/map/achievement.cpp
```diff
@@ -1197,83 +1200,31 @@ void achievement_script_free(struct av_condition *condition)
...
 		return;
-	memset(&achievement_dummy, 0, sizeof(achievement_dummy));
-	achievement_db = idb_alloc(DB_OPT_BASE);
-	achievementmobs_db = idb_alloc(DB_OPT_BASE);
 	achievement_read_db();
 }
 
 void do_final_achievement(void)
 {
 	if (!battle_config.feature_achievement)
```
lighta: I have a feeling that check could cause some leak. (Well no more with your change but I just feel it's error prone)

src/map/intif.cpp
```diff
@@ -2232,7 +2233,7 @@ int intif_achievement_reward(struct map_session_data *sd, struct achievement_db
 	WFIFOW(inter_fd, 10) = adb->rewards.nameid;
 	WFIFOL(inter_fd, 12) = adb->rewards.amount;
 	safestrncpy(WFIFOCP(inter_fd, 16), sd->status.name, NAME_LENGTH);
-	safestrncpy(WFIFOCP(inter_fd, 16+NAME_LENGTH), adb->name, ACHIEVEMENT_NAME_LENGTH);
+	safestrncpy(WFIFOCP(inter_fd, 16+NAME_LENGTH), adb->name.c_str(), ACHIEVEMENT_NAME_LENGTH);
```
lighta: It's fine but ee probably need some raString with some maxlenght as basetype

lighta:
what is needed on this now ?

PR #2609, opened 2017-11-19
Respect the default platform toolset

Lemongrass3110:

* **Server Mode**: Both

With this you will not have to upgrade your solution everytime you clone a fresh copy of rAthena and the problem with having to stash between branch switches should also be gone.
If you want to compile it for another toolset you simply have to set the target toolset by yourself.


anacondaqq:
This is very important and interesting PR, due to a lot of SDKs released every 6 months from Microsoft. Which makes impossible to work in a team with someone who is just installed/reinstalled operation system. The PR solve the problem, and not not need to keep a Zoo of SDKs (8.1, 150**, and so on).

Lemongrass3110:
This is what was getting on our nerves as well. 👍
Thanks for reviewing it so fast!

Issue #2528, opened 2017-10-18
Suggestion: Floating Tax Vending


mazvi:
* **rAthena Hash**: b4106eb

* **Client Date**: 20170517RE

* **Server Mode**: RENEWAL

* **Description of Issue**: 
  * Result: The tax vending on rAthena is STATIC can set by 2% for example
  * Expected Result: The tax vending is floating like 20170517 Client side view please check screenshot
  * How to Reproduce: 
  * Official Information: Check screenshot
  

* **Modifications that may affect results**: 
  
  

![screenchaos806](https://user-images.githubusercontent.com/8316103/31725000-79cee514-b44d-11e7-9432-a48ce07c24da.jpg)
![screenchaos803](https://user-images.githubusercontent.com/8316103/31724562-4f41d28a-b44c-11e7-8fa6-e2f778786db7.jpg)
![screenchaos804](https://user-images.githubusercontent.com/8316103/31724563-4f9327ca-b44c-11e7-83a1-583d510c6bbd.jpg)
![screenchaos805](https://user-images.githubusercontent.com/8316103/31724564-4fda6b3a-b44c-11e7-8913-e148a6d9e0df.jpg)
![screenchaos807](https://user-images.githubusercontent.com/8316103/31725012-829618a2-b44d-11e7-8561-b07eadb6c3b4.jpg)
![screenchaos808](https://user-images.githubusercontent.com/8316103/31725014-82d92318-b44d-11e7-9a28-c67a647003fd.jpg)

anacondaqq:
```c
if(price >= 900000000) tax = 10;
else if(price >= 100000000) tax = 8;
else if(price >= 10000000) tax = 6;
else if(price >= 1000000) tax = 4;
else if(price >= 100000) tax = 2;
else if(price >= 10000) tax = 0;

```

interesting,..

aleos89:
Here are my tests on kRO:
![100000](https://user-images.githubusercontent.com/3521753/33042504-1a9eb48e-ce10-11e7-9259-414eb319900e.PNG)
![1000000](https://user-images.githubusercontent.com/3521753/33042518-1e02590a-ce10-11e7-9126-c6fbbe3f86cf.PNG)
![10000000](https://user-images.githubusercontent.com/3521753/33042522-1fe320d8-ce10-11e7-8cfd-fefe59a39642.PNG)
[...]

PR #2620, opened 2017-11-20
Adjusting Vending tax to match official

aleos89:
* **Addressed Issue(s)**: #2528

  * There are no taxes on items being sold for less than 100 million. Anything above incurs a 5% tax.
  * Refactored the tax code to use a single function.
Thanks to @mazvi!

### DIFF
--- a/conf/battle/items.conf
+++ b/conf/battle/items.conf
@@ -20,8 +20,8 @@ vending_over_max: yes
 // When a tax is applied, the item's full price is charged to the buyer, but
 // the vender will not get the whole price paid (they get 100% - this tax).
 // Officially there is no tax for anything less than 100 million zeny, but anything greater incurs a 5% tax.
-// Value of 0 will follow official behavior, anything else will apply a global tax.
-vending_tax: 0
+// -1 will follow official behavior, 0 will disable taxes while anything else will apply a global tax.
+vending_tax: -1
 
 // Show the buyer's name when successfully vended an item
 buyer_name: yes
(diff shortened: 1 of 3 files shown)


conf/battle/items.conf
```diff
@@ -19,7 +19,9 @@ vending_over_max: yes
 // Tax to apply to all vending transactions (eg: 10000 = 100%, 50 = 0.50%)
 // When a tax is applied, the item's full price is charged to the buyer, but
 // the vender will not get the whole price paid (they get 100% - this tax).
-vending_tax: 200
+// Officially there is no tax for anything less than 100 million zeny, but anything greater incurs a 5% tax.
+// Value of 0 will follow official behavior, anything else will apply a global tax.
```
Jeybla: Unfortunately this excludes the 0% tax option :/

src/map/vending.cpp
```diff
@@ -200,8 +216,7 @@ void vending_purchasereq(struct map_session_data* sd, int aid, int uid, const ui
 
 	pc_payzeny(sd, (int)z, LOG_TYPE_VENDING, vsd);
 	achievement_update_objective(sd, AG_SPEND_ZENY, 1, (int)z);
-	if( battle_config.vending_tax )
-		z -= z * (battle_config.vending_tax/10000.);
+	vending_calc_tax(sd, z);
```
Jeybla: For clarification it's better to use `z = vending_calc_tax(sd, z);` instead of just returning void.

lighta: if he do that, it's better to remove zeny pass as reference to avoid any possible confusion,
param2= zenyin
return = zenytaxed.

PR #2633, opened 2017-11-23
Instance Enter Function

sader1992:
* **Server Mode**:  both

it's just an idea to make this function in the Global_Functions files since it used in every instance script
Of course it's not completed yet i need to delete the enter code and call for this function in the instances files
but before that i would need your approval about this idea maybe it's not ok for who make the instances files usually ( @Atemo )
Because this will result that all the instances have the same error line not special ones
it can adding the instance npc in the function but the message would be the same

this function will return only if it's IE_OK so after calling the function will be the codes for quests and variables
i can may put example for the code in one of the instances


sader1992:
too much files i will make a clean PR

PR #2634, opened 2017-11-24
Instance Entry Functoin

sader1992:
* **Server Mode**: both

it's just an idea to make this function in the Global_Functions files since it used in every instance script
Of course it's not completed yet i need to delete the enter code and call for this function in the instances files
but before that i would need your approval about this idea maybe it's not ok for who make the instances files usually ( @Atemo )
Because this will result that all the instances have the same error statement not special ones
it can adding the instance npc in the function but the message would be the same

this function will return only if it's IE_OK so after calling the function will be the codes for quests and variables
i can may put example for the code in one of the instances

Gain:
remove duplication .
switch the instance entry with one line .

Loss:
unique statement for every instance .


sader1992:
did add Endless Tower for the sake of argument

lighta:
I suppose this is good to reduce code duplication.
I'd rename it "F_GenericInstanceEnter", tough, as from my understanding this would be a default handler for instance_entering, but could be specialized for other need.

npc/other/Global_Functions.txt
```diff
@@ -749,3 +749,36 @@ function	script	F_GetPlatinumSkills	{
...
+//	Example: 
+//	callfunc("F_InstanceEnter","Endless Tower");
+//	callfunc("F_InstanceEnter","Endless Tower","e_tower");
+//	callfunc("F_InstanceEnter","Endless Tower","e_tower","Tower Protection Stone");
+//////////////////////////////////////////////////////////////////////////////////
+function	script	F_InstanceEnter	{
+.@instance_name$ = getarg(0);
+.@map_name$ = getarg(1);
+.@npc_name$ = getarg(2);
+	switch(instance_enter(.@instance_name$)){
```
lighta: what happen here when arg_count = 0 ? can we add a check to prevent this ?

Atemo:
Although I agree there is some redundancy in **instance_enter** messages, after seeing the change I think we shouldn't do theses change for readibility.
Waiting for more opinion.

sader1992:
i had to agree after editing i feel like it's better to have the unique message for each instance instead
however i feel like it's better to leave the function in the Global_Functions file for other usage ?

cydh:
I have same words like Atemo. also and because we follow the official dialog, the function breaks it.

PR #2637, opened 2017-11-25
Fix #2591 spinlock issue.

### LINKED ISSUE #2591: EnterSpinLock can cause deadlock with main thread
* **rAthena Hash**: 6decf6c

* **Client Date**: 2015-11-04

* **Description of Issue**: Pretty niche use case here, but I'm working on setting up a second worker thread and came across a deadlock. EnterSpinLock() uses the thread ID to check which thread owns a lock; this is fine, except that the main thread defaults to thread ID 0. EnterSpinLock() will assume this means the lock is free, allow another thread to take ownership and increment lock->nest to 2, and when the main thread exits won't decrement the nest count (because another thread now owns the lock). After the second thread exits, nest is decremented to 1 (instead of 0), the lock isn't freed, and no other thread can enter the spin lock, causing a dead lock. I think that's the most confusing paragraph I've ever written.
  * Result: Main thread fails to take ownership of the spin lock correctly, causes a deadlock if another thread enters the spin lock before the main thread exits. 
  * Expected Result: Main thread takes ownership of the spin lock and other threads are forced to wait.
  * How to Reproduce: Start a second thread, enter spin lock on the main thread, and before the main thread exits, enter the same spin lock on the second thread. Exit from the main thread, then the second thread. The main thread will deadlock if it tries to enter the spin lock again. 
  * Official Information: N/A 
  

The second worker thread is a modification of course, but if rAthena is updated to take advantage of threading in the future (for example putting log queries on a separate thread), this would likely cause issues. Figured I'd put this here so it can be resolved instead of causing possible issues/confusion later when it gets used.

To fix it, in rathread_init() change the main thread to thread ID 1, and adjust l_threads[].myID to i+1. rathread_self() needs to be adjusted as well to use the correct l_threads[] index (g_rathread_ID-1). As far as I'm aware this should be all that's necessary.

As an aside, this only happens if HAS_TLS is defined, though given the fact that attempting to compile without it causes compilation errors, I'm assuming it's never not defined.

lighta:
Add basic_test, for few integration test.

* **Addressed Issue(s)**: 

* **Server Mode**: 


src/test/basic_test.vcxproj.filters
```diff
@@ -0,0 +1,36 @@
...
+      <Extensions>h;hh;hpp;hxx;hm;inl;inc;xsd</Extensions>
+    </Filter>
+    <Filter Include="Fichiers de ressources">
+      <UniqueIdentifier>{67DA6AB6-F800-4c08-8B7A-83BB121AAD01}</UniqueIdentifier>
+      <Extensions>rc;ico;cur;bmp;dlg;rc2;rct;bin;rgs;gif;jpg;jpeg;jpe;resx;tiff;tif;png;wav;mfcribbon-ms</Extensions>
+    </Filter>
+  </ItemGroup>
+  <ItemGroup>
+    <ClCompile Include="basictest.cpp">
+      <Filter>Fichiers sources</Filter>
```
aleos89: Please use English - `Source Files`

src/test/basic_test.vcxproj.filters
```diff
@@ -0,0 +1,36 @@
...
+    </ClCompile>
+    <ClCompile Include="test_thread.cpp">
+      <Filter>Fichiers sources</Filter>
```
aleos89: Please use English - `Source Files`

src/test/basic_test.vcxproj.filters
```diff
@@ -0,0 +1,36 @@
...
+    </ClCompile>
+    <ClCompile Include="test_spinlock.cpp">
+      <Filter>Fichiers sources</Filter>
```
aleos89: Please use English - `Source Files`

src/test/basic_test.vcxproj.filters
```diff
@@ -0,0 +1,36 @@
...
+    </ClCompile>
+  </ItemGroup>
+  <ItemGroup>
+    <ClInclude Include="test_spinlock.hpp">
+      <Filter>Fichiers d%27en-tête</Filter>
```
aleos89: Please use English - `Header Files`

src/test/basic_test.vcxproj.filters
```diff
@@ -0,0 +1,36 @@
...
+    </ClInclude>
+    <ClInclude Include="test_thread.hpp">
+      <Filter>Fichiers d%27en-tête</Filter>
```
aleos89: Please use English - `Header Files`

src/test/basictest.cpp
```diff
@@ -0,0 +1,29 @@
+// basictest.cpp : Définit le point d'entrée pour l'application console.
```
aleos89: Please use English.

PR #2639, opened 2017-11-25
packet.h clarification

sader1992:
* **Server Mode**: both

i see a lot of people edit this without known , so this line just to tell them do not edit it .

### DIFF
--- a/src/config/packets.h
+++ b/src/config/packets.h
@@ -9,9 +9,9 @@
  **/
 
 #ifndef PACKETVER
-	///Do not edit this line. To set your client version, please do this instead.
-	///In Windows: add this line in your src\custom\defines_pre.h file: #define PACKETVER YYYYMMDD
-	///In Linux: The same as above, or run the following command: ./configure --enable-packetver=YYYYMMDD
+	/// Do NOT edit this line! To set your client version, please do this instead:
+	/// In Windows: Add this line in your src\custom\defines_pre.h file: #define PACKETVER YYYYMMDD
+	/// In Linux: The same as above or run the following command: ./configure --enable-packetver=YYYYMMDD
 	#define PACKETVER 20151104
 #endif
 
--- a/src/config/packets.h
+++ b/src/config/packets.h
@@ -9,9 +9,9 @@
  **/
 
 #ifndef PACKETVER
-	///Do not edit this line. To set your client version, please do one of the folowing instead.
-	///for Windows and Linux: add this line in your src\custom\defines_pre.h file: #define PACKETVER YYYYMMDD
-	///for Linux Only: run the following command: ./configure --enable-packetver=YYYYMMDD
+	///Do not edit this line. To set your client version, please do this instead.
+	///In Windows: add this line in your src\custom\defines_pre.h file: #define PACKETVER YYYYMMDD
+	///In Linux: The same as above, or run the following command: ./configure --enable-packetver=YYYYMMDD
 	#define PACKETVER 20151104
 #endif
 
--- a/src/config/packets.h
+++ b/src/config/packets.h
@@ -9,9 +9,9 @@
  **/
 
 #ifndef PACKETVER
-	///Do not edit this line. To set your client version, please do this instead.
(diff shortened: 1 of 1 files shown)


lighta:
hmm, the comment could be misleading if you want to lower the packet_version.
What is the common problematic ?
Perhaps if we want to details we should put a note to see doc/packet_config.txt or something of that sort ?

sader1992:
@lighta 
when people edit the packets.h to change the `PACKETVER` to 2017 clients RagexeRE
they see like this
`#ifndef PACKETVER
	#define PACKETVER 20151104
#endif

#ifndef PACKETVER_RE
	/// From this point on only kRO RE clients are supported
	#if PACKETVER > 20151104
		#define PACKETVER_RE
	#endif
#endif`
they have to edit `#define PACKETVER 20151104`
but they also edit `#if PACKETVER > 20151104`
and this will lead for this error
screenrathena014
so i would move this option from here to prevent misleading when they change the `PACKETVER`
but since my knowledge in C/C++ is limited i did make a statement to those who want to change it
i did think if i made the statement like
"Don't edit if you use RagexeRE"
but then they will wander what they have to do if they use Ragexe

Daegaladh:
I think you should add a comment in #define PACKETVER too, something like:

> 
> Do not edit this line. To set your client version, please do this instead.
> In Linux: run the following command: ./configure --enable-packetver=YYYYMMDD
> In Windows: add this line in your src\custom\defines_pre.h file: #define PACKETVER YYYYMMDD

lighta:
@sader1992 so basically, the comment should be on the #if with condition.
(i.e #if PACKETVER > 20151104). here only. The 1st part is perfectly valid to edit is just that they getting confuse and chose extra condition.

Now we could put your comment warning but there a little issue with that, is that we have tons of macro like this, so we may end up doing tons of comment.
To be fair editing a condition is never a good idea if you don't know why it was there...

Now we could hide that PACKETVER_RE enabling precondition. Since one of the issue is that this is in config/ area so people are allowed to edit stuff here.
What we could do is moved down those PACKETVER minimun version a little down to the file after the 
/**
* No settings past this point
**/

src/config/packets.h
```diff
@@ -9,16 +9,12 @@
  **/
 
 #ifndef PACKETVER
+	///Do not edit this line. To set your client version, please do this instead.
+	///In Linux: run the following command: ./configure --enable-packetver=YYYYMMDD
+	///In Windows: add this line in your src\custom\defines_pre.h file: #define PACKETVER YYYYMMDD
```
aleos89: The `defines_pre.h` can also be used for Linux, not just Windows.

The documentation is good, we just don't want it to be misleading that it's only applicable to Windows builds.

src/config/packets.h
```diff
@@ -9,16 +9,12 @@
...
 	#define PACKETVER 20151104
 #endif
 
-#ifndef PACKETVER_RE
-	/// From this point on only kRO RE clients are supported
-	#if PACKETVER > 20151104
-		#define PACKETVER_RE
-	#endif
-#endif
-
```
aleos89: Is there a reason this was moved down?

sader1992: some people get confused by that `#define PACKETVER 20151104` is the same of `#if PACKETVER > 20151104` so they edit both
while we can't control the server owner option to edit the version in here when he edit both to put his version while he use RagexeRE +2016 he will get the error in the screenshot above
moving it down will decrease the error rate by the servers owner
however if people start to use the other methods to set there version , no one would get the error and get confused BY `#define PACKETVER` AND `#if PACKETVER`

src/config/packets.h
Daegaladh: What about...

> ///In Windows: add this line in your src\custom\defines_pre.h file: #define PACKETVER YYYYMMDD
> ///In Linux: The same as above, or run the following command: ./configure --enable-packetver=YYYYMMDD

sader1992:
after being a while
i do believe it's better to Encourage people to edit the version in the other files rather then changing the check place

PR #2647, opened 2017-11-26
Fix #2425 Character variable lengths and char-server

### LINKED ISSUE #2425: Character variable lengths and char-server
* **rAthena Hash**: 8222208
* **Client Date**: Any
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: When saving long character variable strings, the char-server can easily crash.
  * Expected Result: No crash.
  * How to Reproduce: I've used the following script to reproduce the issue:
```
prontera,155,176,3	script	Pront Test	77,{
	test$ = "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345";
	test2$ = "anything really, doesn't matter at this point";
	end;
}
```
  * Other Information: There are no size checks for these variables in intif_saveregistry, it just truncates the size to 
```
WFIFOB(inter_fd, plen) = (unsigned char)len;
```
and past 254, the problems start (the max size of the char array is 254 in the char-server). The same issue also happens with variables having a name length above 32. What I've personally done is
```
		len = strlen(varname)+1;

		if (len > 32) {
			ShowError("intif_saveregistry: Variable name length is too long (aid: %d, cid: %d): %s\n", sd->status.account_id, sd->status.char_id, varname);
			continue;
		}

and ...

			if( p->value ) {
				len = strlen(p->value)+1;

				if (len > 254) {
					ShowError("intif_saveregistry: Variable content length is too long (aid: %d, cid: %d): %s, content: %s\n", sd->status.account_id, sd->status.char_id, varname, p->value);
					len = 254;
					p->value[len - 1] = '\0';
				}
```
But it would probably be best to do those checks when the values are set, not when they're saved.

lighta:
* **Addressed Issue(s)**: #2425 

* **Server Mode**: Both

Medium term fox for crash on char-serv caused by too long character variable.
Add few error msg and safety.
Add doc/sample/npc_test_longvar.txt as a basic test.

### DIFF
--- a/doc/sample/npc_test_longvar.txt
+++ b/doc/sample/npc_test_longvar.txt
@@ -1,0 +1,17 @@
+//===== rAthena Script =======================================
+//= Sample: Character variable lengths Test
+//===== By: ==================================================
+//= rAthena Dev Team
+//===== Last Updated: ========================================
+//= 20171125
+//===== Description: ========================================= 
+//= An validation test about long character variable stored on char-serv
+//= from https://github.com/rathena/rathena/issues/2425
+//============================================================
+
+prontera,155,176,3	script	Pront Test	77,{
+	mes "Testing to store long variable on char-serv";
+	test$ = "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345";
+	test2$ = "anything really, doesn't matter at this point";
+	end;
+}
--- a/src/map/intif.cpp
+++ b/src/map/intif.cpp
@@ -422,11 +422,14 @@ int intif_saveregistry(struct map_session_data *sd)
 		src->update = false;
 
 		len = strlen(varname)+1;
-
+    if (len > 32) { //this is sql colum size, must be retrive from config
+			ShowError("intif_saveregistry: Variable name length is too long (aid: %d, cid: %d): '%s' sz=%d\n", sd->status.account_id, sd->status.char_id, varname, len);
+			continue;
+		}
 		WFIFOB(inter_fd, plen) = (unsigned char)len; // won't be higher; the column size is 32
 		plen += 1;
 
-		safestrncpy(WFIFOCP(inter_fd,plen), varname, len);
+		safestrncpy(WFIFOCP(inter_fd,plen), varname, len); //the key
 		plen += len;
 
 		WFIFOL(inter_fd, plen) = script_getvaridx(key.i64);
@@ -435,13 +438,18 @@ int intif_saveregistry(struct map_session_data *sd)
 		if( src->type ) {
 			struct script_reg_str *p = (struct script_reg_str *)src;
 
-			WFIFOB(inter_fd, plen) = p->value ? 2 : 3;
+			WFIFOB(inter_fd, plen) = p->value ? 2 : 3; //var type
 			plen += 1;
 
 			if( p->value ) {
 				len = strlen(p->value)+1;
+        if ( len > 254 ) { // error can't be higher; the column size is 254. (nb the transmission limit with be fixed with protobuf revamp)
+          ShowDebug( "intif_saveregistry: Variable value length is too long (aid: %d, cid: %d): '%s' sz=%d to be saved with current system and will be truncated\n",sd->status.account_id, sd->status.char_id,p->value,len);
+          len = 254;
+          p->value[len - 1] = '\0'; //this is backward for old char-serv but new one doesn't need this
+        }
 
-				WFIFOB(inter_fd, plen) = (unsigned char)len; // won't be higher; the column size is 254
+				WFIFOB(inter_fd, plen) = len; 
 				plen += 1;
 
 				safestrncpy(WFIFOCP(inter_fd,plen), p->value, len);
--- a/src/map/intif.cpp
+++ b/src/map/intif.cpp
@@ -420,9 +420,9 @@ int intif_saveregistry(struct map_session_data *sd)
 			continue;
 
 		src->update = false;
-
-		len = strlen(varname)+1;
-    if (len > 32) { //this is sql colum size, must be retrive from config
+		bool lValid = script_check_RegistryVariableLenght(0,varname,&len);
+		++len;
+		if (!lValid) { //this is sql colum size, must be retrive from config
 			ShowError("intif_saveregistry: Variable name length is too long (aid: %d, cid: %d): '%s' sz=%d\n", sd->status.account_id, sd->status.char_id, varname, len);
 			continue;
 		}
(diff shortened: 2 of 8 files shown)


doc/sample/npc_test_longvar.txt
```diff
@@ -0,0 +1,17 @@
...
+//= 20171125
+//===== Description: ========================================= 
+//= An validation test about long character variable stored on char-serv
+//= from https://github.com/rathena/rathena/issues/2425
+//============================================================
+
+prontera,155,176,3	script	Pront Test	77,{
+	mes "Testing to store long variable on char-serv";
+	test$ = "0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345";
+	test2$ = "anything really, doesn't matter at this point";
```
Lemongrass3110: Please also add a testcase with a variable name longer than 32 characters.

src/map/pc.cpp
```diff
@@ -9235,6 +9235,18 @@ int pc_setregistry(struct map_session_data *sd, int64 reg, int val)
 	return 1;
 }
 
+//maybe move this as intif or script ?
+bool pc_check_RegistryVariableLenght(struct map_session_data *sd, int pType, const char *val, size_t* vlen)
```
Lemongrass3110: Typo in "length"

src/map/pc.cpp
```diff
@@ -9235,6 +9235,18 @@ int pc_setregistry(struct map_session_data *sd, int64 reg, int val)
 	return 1;
 }
 
+//maybe move this as intif or script ?
```
Lemongrass3110: Should be usable/used by all modules that have script variables.

lighta: yeah I think I'd just put it in intif as it's the one "blocking/sending" the data.
Plus I don't wanna bloated pc.c more then it's already is.

src/map/pc.cpp
lighta: which typo. I wrote it lije I wrote lighta =) jk, will fix.

cydh: `lighta != lengtha`

src/map/script.cpp
```diff
@@ -3121,6 +3136,12 @@ void script_array_update(struct reg_db *src, int64 num, bool empty)
 int set_reg(struct script_state* st, struct map_session_data* sd, int64 num, const char* name, const void* value, struct reg_db *ref)
 {
 	char prefix = name[0];
+	size_t vlen = 0;
+	if ( !script_check_RegistryVariableLength(0,name,&vlen) )
+	{
+		ShowError("set_reg: Variable name length is too long (aid: %d, cid: %d): '%s' sz=%d\n", sd->status.account_id, sd->status.char_id, name, vlen);
```
Lemongrass3110: Will crash for non-player variables.

src/map/script.cpp
```diff
@@ -748,6 +748,21 @@ static unsigned int calc_hash(const char* p)
...
+{
+	size_t len = strlen(val);
+
+	if (vlen)
+		*vlen = len;
+	switch (pType) {
+		case 0:
+			return (len < 33); // key check
+		case 1:
+			return (len < 255); // value check
```
Lemongrass3110: Still think those two sizes should be moved to mmo.h and value should be limited to UINT8_MAX, since that's what actually limits the value, because we send the size of the string as unsigned 8 bit integer.

src/map/intif.cpp
```diff
@@ -435,13 +440,19 @@ int intif_saveregistry(struct map_session_data *sd)
...
 			plen += 1;
 
 			if( p->value ) {
-				len = strlen(p->value)+1;
+				lValid = script_check_RegistryVariableLength(1,p->value,&len);
+				++len;
+				if ( !lValid ) { // error can't be higher; the column size is 254. (nb the transmission limit with be fixed with protobuf revamp)
+					ShowDebug( "intif_saveregistry: Variable value length is too long (aid: %d, cid: %d): '%s' sz=%d to be saved with current system and will be truncated\n",sd->status.account_id, sd->status.char_id,p->value,len);
+					len = 254;
+					p->value[len - 1] = '\0'; //this is backward for old char-serv but new one doesn't need this
```
Lemongrass3110: This is not only backward compatability, we still do not support anything higher.

src/map/script.cpp
lighta: why would it crash it just checking the len of a string.

src/map/script.cpp
lighta: I think not. I mean yes from the prototocol inter-serv point of view he limit is uint8_max, but it won't be limited on that specific case then. All our packet are constructed without check on variable bound.
Now we could add those but it's an overhead and since we receive only our packet atm would that really make sens ?
I mean if at least the packet was created and sent in an async manner that wouldn't impact the server but atm =/.
I feel that since I will change the inter-serv com soon enough it doesn't worth the trouble.

From a system point of view, the limit is really the len of sql.acc_reg_num and sql.acc_reg_str key and value.
Now I could add the attribute len table in a .h, which is probably a bit better, but really from a tiny bit. The current size is localized in same spot atm with that function. A macro won't bring more, as in most time you would want to check against the value of that macro a string (thus the function).
Only somewhat interesting case would be to create array with fixed sized with that macro. (stack array). But if you do this to soon you prevent yourself to detect len error, then within the same process passing an adress or std::move would be better then to have an array so in the end ~

Now current case or macro we still have 2 different place that define the sql table attribute length. Which I feel is not ideal. 
We could at server startup grab that info and set it to a static to check against. (but it's map-serv that would have to ask char-serv the size of his db bound...)  
Or we could have that conf file and make the sql query be based upon this.

src/map/script.cpp
secretdataz: sd is used here without proper checking.

aleos89: Is `set_reg` ever called with a null session?

Lemongrass3110: judging by the code a few lines below without checking the source references i'd say: definitely.

lighta: ohh I get it, you meant in the error msg, gotcha, yeah that annoying, I tough you where commenting in 
"if ( !script_check_RegistryVariableLength(0,name,&vlen) )"
K then should I remove the (aid: %d, cid: %d) part, or do you want that info if the sd is avalaible ?

src/map/intif.cpp
lighta: setting the '\0' is the backward. we don't need this really as the string_len is passed as previous argument and that I reconstruct the string with that arg.
Only usefull for strlen or thing like this.

PR #2651, opened 2017-11-28
Introduce utils2 to gradually replace utils.

lighta:
Replace current levenshtein with much faster implementation

* **Addressed Issue(s)**: 

* **Server Mode**: 


Lemongrass3110:
https://ci.appveyor.com/project/rAthenaAPI/rathena/build/refactor/cpp_core2-2240/job/ev3ylr1i5rbxqm5q#L256

src/common/common.vcxproj
```diff
@@ -224,4 +226,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
Lemongrass3110: Missing newline.

src/common/common.vcxproj
```diff
@@ -75,6 +76,7 @@
     <ClCompile Include="thread.c" />
     <ClCompile Include="timer.c" />
     <ClCompile Include="utils.c" />
+    <ClCompile Include="utils2.cpp" />
```
Lemongrass3110: Maybe go with utility instead of utils2?

src/common/common.vcxproj.filters
```diff
@@ -172,5 +175,8 @@
     <ClCompile Include="yamlwrapper.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="utils2.cpp">
+      <Filter>Source Files</Filter>
+    </ClCompile>
   </ItemGroup>
 </Project>
```
Lemongrass3110: Missing newline, not related to this pull though.

src/common/utils2.cpp
```diff
@@ -0,0 +1,69 @@
...
+cScopeTimer::cScopeTimer() 
+	: aPimpl(new sPimpl())
+{}
+
+/**
+ * Calculates the Levenshtein distance of two strings.
+ * @author http://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Levenshtein_distance#C.2B.2B
+ * comparison test was done here http://cpp.sh/2o7w
+ */
+int levenshtein(const std::string &s1, const std::string &s2)
```
Lemongrass3110: Telling me to not use "auto" everywhere and reading your code afterwards is a bit strange O.o

src/common/utils2.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+	// To change the type this function manipulates and returns, change
+	// the return type and the types of the two variables below.
+	int s1len = static_cast<int>(s1.size());
```
Lemongrass3110: Why not use size_t instead here?

lighta:
should have taken into account vs stupidity.. will fix tomorrow.

src/common/utils2.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <iostream>
+#include <numeric> //iota
+
+
+using Clock = std::chrono::steady_clock;
+using std::chrono::time_point;
+using std::chrono::duration_cast;
+using std::chrono::microseconds;
+
+struct cScopeTimer::sPimpl {
```
Lemongrass3110: What's this used for?
Could we keep it out as long as we do not need it yet?

src/common/utils2.hpp
```diff
@@ -0,0 +1,13 @@
...
+
+//a small class to perform time mesurement
+class cScopeTimer {
+	struct sPimpl; //this is to avoid long compilation time
+	std::unique_ptr<sPimpl> aPimpl;
+    
+	cScopeTimer();
+};
+
+int levenshtein( const std::string &s1, const std::string &s2 );
```
Lemongrass3110: Missing newline

src/common/utils2.cpp
Jeybla: http://en.cppreference.com/w/cpp/language/pimpl

Looks like it's quite useful in large libraries. I don't think we really need it in our small project.

src/common/utils2.cpp
lighta: I wasn't sure of the return value. the precedent funciton was using int, maybe the distance was negative. (yes I know negative dist uh ?) but you never know.
I was being cautious here.

src/common/utils2.cpp
lighta: tough it might be usefull. why you guy not liking it ?
what the matter with it ?

Jeybla: I wouldn't say I don't like it. It just feels hacky/dirty. Maybe it could cause issues in future versions. Maybe it could become a part of C++ itself with a single statement.
But most importantly: Why should we have it if we don't need it?

lighta: hacky ? uh =(, (I just didn't want to expose it to avoid the need of othrer class to know full std::chrono and stuff implementation.
Why do we need it?
1) Well in debug mode using it would be perfectly valid. To know how long that suff take etc... The call could just be wrapped with some macro to disable. (in release you could too but it's less common as you want the full speed..)
2) I view it as a tool. Since I used firewall / pimpl template. the compilation time won't be altered, all that we really losing is on the .exe since. Not that important for me atm.
3) Basic_test or some other will need it, plus I think it's good to have an example of how to play with those timer.
For example I didn't know it would have fail with this implementation on vs. (I tryed with gcc as the comment say)
If next version of c++ have it one line that good, but I'm not sure will be able to use it right away. (gcc/clang/vs..)

src/common/utilities.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+#include <memory>
+#include <string>
+
+//a small class to perform time mesurement
```
Lemongrass3110: measurement

src/common/utilities.hpp
```diff
@@ -0,0 +1,13 @@
...
+class cScopeTimer {
+	struct sPimpl; //this is to avoid long compilation time
+	std::unique_ptr<sPimpl> aPimpl;
+    
+	cScopeTimer();
+};
+
+int levenshtein( const std::string &s1, const std::string &s2 );
```
Lemongrass3110: missing newline

PR #2658, opened 2017-12-01
Added the custom and test scripts to the buildbot

Lemongrass3110:

* **Server Mode**: Both

All scripts inside the custom and test npc directory are now parsed and executed automatically by the buildbot.
The only exceptions are the custom battleground scripts that would cause duplication warnings.

Thanks to @Jeybla for the Unix script!


sader1992:
so they will be loaded without adding them to the scripts_custom.conf ?

Lemongrass3110:
@sader1992 this is only intended for the buildbot, so that we see if one of the scripts we have in our repo gets broken by changes we do. Right now the custom directory is not checked because all those scripts are commented in the config.

tools/ci/npc.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+out=../../npc/scripts_custom.conf
+
+printf "\n" >> $out
```
Jeybla: Just to mention it:
You're always appending to the file. If this script is executed more than once, the file would contain duplicates. (But afaik the build bots are completely cleared for every test)

Issue #2660, opened 2017-12-01
Can't enable title on title UI.


hnomkeng:
* **rAthena Hash**: 02044c0
* **Client Date**: 2017-06-14b
* **Server Mode**: Pre-renewal
* **GCC**: 5.3.1
* **Description of Issue**: Can't enable title on title UI.
Vdo : https://youtu.be/fTlabDAS-WE

sader1992:
open your database char table
in your char add a title id let's say 1034
relogin into the server and see your char does it have title 1034 = [VIP] ?

hnomkeng:
@sader1992  add on char table is show.

hnomkeng:
I can disable title on GUI. I can't enable title on GUI.

sader1992:
do you get any message about unknown packet in the terminal ?

hnomkeng:
@sader1992 No have error or warning msg on console.

sader1992:
i did test it now i can confirm this it's the same on 2015-11
i am not sure it's from rAathena but it seems to

hnomkeng:
@sader1992  after hash c5ceda0 not working ?

sader1992:
in between

sader1992:
i think maybe the issue here > 1c66035761d256474bde7df703eb406255514d8c 
this is the only update that edit the title system (in between the hashs )
not sure it's better to wait for rAthena developers

PR #2661, opened 2017-12-01
fix title select in ui window

sader1992:
* **Addressed Issue(s)**: 
thanks to @hnomkeng https://github.com/rathena/rathena/issues/2660

* **Server Mode**: both

not really good at c/c++ so please check the code
this for me fixed the issue , but not sure if it's the right solution


sader1992:
not the right solution

PR #2662, opened 2017-12-01
fix title select in ui window

sader1992:
* **Addressed Issue(s)**: 
thanks to @hnomkeng https://github.com/rathena/rathena/issues/2660

* **Server Mode**: both

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -19020,7 +19020,7 @@ void clif_change_title_ack(struct map_session_data *sd, unsigned char result, un
  */
 void clif_parse_change_title(int fd, struct map_session_data *sd)
 {
-	int title_id,i;
+	int title_id;
 
 	nullpo_retv(sd);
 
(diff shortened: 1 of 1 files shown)


src/map/clif.cpp
```diff
@@ -19032,7 +19032,7 @@ void clif_parse_change_title(int fd, struct map_session_data *sd)
 	}else if( title_id <= 0 ){
 		sd->status.title_id = 0;
 	}else{
-		if (std::find(sd->titles.begin(), sd->titles.end(), title_id) != sd->titles.end()) {
+		if (std::find(sd->titles.begin(), sd->titles.end(), i) != sd->titles.end()) {
```
Jeybla: `i` would be uninitialized in that line.

I think the actual issue is a logical mistake. The find checks if the player already has the title. If not (`find==end()`) execute `clif_change_title_ack`. So we'd need to change `!=` to `==`

aleos89: Jeybla is right. In this case you define `i` uninitialized and with `std::find` the third parameter is what you're searching for in the vector.

sader1992: `std::find (vector.begin(), vector.end(), item) != vector.end()` if this is true then there is item in the vector
what i think is the problem here `title_id = RFIFOL(fd, 2);`
the title id format is changed and it look like i != the title id anymore
not sure tho

sader1992: while i look more in it i did knew that 
`clif_change_title_ack(sd, 1, title_id);`
		`	return;`
is what making the problem not (not)passing check
so that should as you did say == not !=
while the `i` edit did fix the problem but that cuz it making it don't pass the check
so the right is `==` not `!=` 
sorry i got confused a little bit by `clif_change_title_ack(sd, 1, title_id);`

PR #2063, opened 2017-04-14
Terminate script execution when a command fails

Lemongrass3110:
* **Addressed Issue(s)**: None

* **Server Mode**: Both

* **Description of Pull Request**: 
Since script commands only fail in reasonable and noteworthy cases we should cancel script execution and report any data we can provide the user with as soon as a script command fails during runtime.

PR #2673, opened 2017-12-04
Update sc_end_class

sader1992:

* **Server Mode**: both

sc_end_class now can be end the effect of specific job 
if no job id was given it will end the skill effect for the job of the invoking character
sc_end_class {<char_id>{,<job_id>}};

Example:
the test npc
[sc_end_class.txt](https://github.com/rathena/rathena/files/1532031/sc_end_class.txt)

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5389,7 +5389,7 @@ and theirs val1, val2, val3, and val4 usage in source.
 perform a complete removal of all statuses (although permanent ones will re-apply).
 
 'sc_end_class' works like 'sc_end' but will remove all status effects from any learned
-skill on the invoking character. If <job_id> is provided it will end the effect for the that job.
+skill on the invoking character. If <job_id> is provided it will end the effect for that job.
 
 Examples:
 	// This will poison the invoking character for 10 minutes at 50% chance.
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5409,17 +5409,17 @@ Examples:
 	// This will end the Freezing status for the invoking character.
 	sc_end SC_FREEZE;
 	
-	//will end the effect of any learned skill for the invoking character.
+	// This will end the effect of any learned skill for the invoking character.
 	sc_end_class;
 	
-	//will end the effect of any learned skill for the invoking character.
+	// This will end the effect of any learned skill for the character with the <char_id> 150000.
 	// val1: <char_id>
 	sc_end_class(150000);
 	
-	//will end the effect of any Arch Bishop skill for the invoking character.
+	// This will end the effect of any Arch Bishop skill for the invoking character.
 	// val1: <char_id> (-1 return the <char_id> of the invoking character.)
 	// val2: <job_id> of Arch Bishop
-	sc_end_class(-1,4057);
+	sc_end_class(-1,Job_Arch_Bishop);
 
 Note: to use SC_NOCHAT you should alter Manner
 	set Manner, -5;	// Will mute a user for 5 minutes
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5389,7 +5389,7 @@ and theirs val1, val2, val3, and val4 usage in source.
 perform a complete removal of all statuses (although permanent ones will re-apply).
 
 'sc_end_class' works like 'sc_end' but will remove all status effects from any learned
-skill on the invoking character. if <job_id> was given it will end the effect for that job.
+skill on the invoking character. If <job_id> is provided it will end the effect for the that job.
 
 Examples:
 	// This will poison the invoking character for 10 minutes at 50% chance.
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5390,7 +5390,6 @@ perform a complete removal of all statuses (although permanent ones will re-appl
 
 'sc_end_class' works like 'sc_end' but will remove all status effects from any learned
 skill on the invoking character. if <job_id> was given it will end the effect for that job.
-if the <char_id> was -1 it will return the <char_id> for the invoking character
 
 Examples:
 	// This will poison the invoking character for 10 minutes at 50% chance.
@@ -5417,9 +5416,9 @@ Examples:
 	sc_end_class(150000);
 	
 	// This will end the effect of any Arch Bishop skill for the invoking character.
-	// val1: <char_id> (-1 return the <char_id> of the invoking character.)
+	// val1: <char_id>
 	// val2: <job_id> of Arch Bishop
-	sc_end_class(-1,Job_Arch_Bishop);
+	sc_end_class(getcharid(0),Job_Arch_Bishop);
 
 Note: to use SC_NOCHAT you should alter Manner
 	set Manner, -5;	// Will mute a user for 5 minutes
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -5350,7 +5350,7 @@ Used in reset NPC's (duh!)
 *sc_start2 <effect type>,<ticks>,<value 1>,<value 2>{,<rate>,<flag>{,<GID>}};
 *sc_start4 <effect type>,<ticks>,<value 1>,<value 2>,<value 3>,<value 4>{,<rate>,<flag>{,<GID>}};
 *sc_end <effect type>{,<GID>};
-*sc_end_class {<char_id>};
+*sc_end_class {<job_id>{,<char_id>}};
 
 These commands will bestow a status effect on a character.
 
@@ -5389,7 +5389,7 @@ and theirs val1, val2, val3, and val4 usage in source.
 perform a complete removal of all statuses (although permanent ones will re-apply).
 
 'sc_end_class' works like 'sc_end' but will remove all status effects from any learned
-skill on the invoking character.
+skill on the invoking character. if job_id was given it will end the effect for that job.
 
 Examples:
 	// This will poison the invoking character for 10 minutes at 50% chance.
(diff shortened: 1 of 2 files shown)


secretdataz:
Now you can't do end statuses from `whatever job another char has`'s skills.

sader1992:
yes it was my plan too to do that
i did try something like 
`	if (script_hasdata(st, 2))`
`		if (script_hasdata(st, 2) != 0)`
`			class_ = script_getnum(st, 2);`
with my limited knowledge it didn't work ~o-o~
and also 0 is the novice job
i wanted it to be 
if sc_end_class(0,<job_id>); it will end the skill effect of the job that <char_id> have

src/map/script.cpp
```diff
@@ -11337,18 +11337,28 @@ BUILDIN_FUNC(sc_end)
...
+ * if <job_id> was given it will end all the effect for that job
+ * if <job_id> is 0 it will return the job of the invoking character
+ * sc_end_class {<job_id>{,<char_id>}};
  */
 BUILDIN_FUNC(sc_end_class)
 {
 	struct map_session_data *sd;
 	uint16 skill_id;
-	int i;
+	int class_ = 0;
```
Lemongrass3110: Remove init here

src/map/script.cpp
```diff
@@ -11337,18 +11337,28 @@ BUILDIN_FUNC(sc_end)
...
+	struct block_list *bl = map_id2bl(st->rid);
 
-	if (!script_charid2sd(2, sd))
-		return SCRIPT_CMD_FAILURE;
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, -1);
```
Lemongrass3110: Please keep this as SCRIPT_CMD_FAILURE and remove the value push.
This is intended for future merge of #2063

src/map/script.cpp
```diff
@@ -11337,18 +11337,28 @@ BUILDIN_FUNC(sc_end)
...
+		return SCRIPT_CMD_SUCCESS;
+	}
+
+	if (script_hasdata(st, 2))
+		class_ = script_getnum(st, 2);
+	if (class_ == 0)
```
Lemongrass3110: just turn this into an else case

src/map/script.cpp
```diff
@@ -11337,18 +11337,28 @@ BUILDIN_FUNC(sc_end)
...
 
-	if (!script_charid2sd(2, sd))
-		return SCRIPT_CMD_FAILURE;
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, -1);
+		return SCRIPT_CMD_SUCCESS;
+	}
+
+	if (script_hasdata(st, 2))
+		class_ = script_getnum(st, 2);
```
Lemongrass3110: probably adding a class validation into this if would be nice, otherwise you might fail in the skill tree lookup

doc/script_commands.txt
```diff
@@ -5389,7 +5389,8 @@ and theirs val1, val2, val3, and val4 usage in source.
 perform a complete removal of all statuses (although permanent ones will re-apply).
 
 'sc_end_class' works like 'sc_end' but will remove all status effects from any learned
-skill on the invoking character.
+skill on the invoking character. if job_id was given it will end the effect for that job.
+if the job was 0 it will return the <job_id> for the invoking character
```
Lemongrass3110: I'd rather have it like if you specify a class it will take that value for granted(maybe a small validation, but nothing more).
This would also be better since novice is also job class 0, if I do not remember wrong here.

src/map/script.cpp
sader1992: here i didn't understand what to do

sader1992:
so i put it like if the char id is 0 it will return the invoking character
or maybe something else

src/map/script.cpp
aleos89: He means remove the `= 0` part. Just define the variable: `int class_;`

src/map/script.cpp
sader1992: i will search about it << still don't know a lot of things

sader1992: just to make sure
you want something like
`if (class_ != pc_jobid2mapid(class_))`
`			return SCRIPT_CMD_FAILURE;`
so the map server wont crash if the class id was not a class id right ?

Lemongrass3110: pcdb_checkid

src/map/script.cpp
```diff
@@ -11337,18 +11337,34 @@ BUILDIN_FUNC(sc_end)
 
 /**
  * Ends all status effects from any learned skill on the attached player.
- * sc_end_class {<char_id>};
+ * if <char_id> = -1 it will return the attached player
+ * if <job_id> was givin it will end the effect of that class for the attached player
```
Lemongrass3110: given

doc/script_commands.txt
```diff
@@ -5407,6 +5408,18 @@ Examples:
 
 	// This will end the Freezing status for the invoking character.
 	sc_end SC_FREEZE;
+	
+	//will end the effect of any learned skill for the invoking character.
```
Lemongrass3110: Follow the formatting/phrasing of the example above:
`//<whitespace>This <your part of the sentence>`

doc/script_commands.txt
```diff
@@ -5407,6 +5408,18 @@ Examples:
...
+	sc_end_class;
+	
+	//will end the effect of any learned skill for the invoking character.
```
Lemongrass3110: this is not a sample of an "invoking" character, since you specify a char id of any/another player.
So probably call it "for the specified character" or like you did below say "for the character with the char id 150000.

src/map/script.cpp
```diff
@@ -11337,18 +11337,34 @@ BUILDIN_FUNC(sc_end)
...
+	}
+	else if (!script_charid2sd(2, sd))
+			return SCRIPT_CMD_FAILURE;
+
+	if (script_hasdata(st, 3))
+		class_ = script_getnum(st, 3);
+	else
+		class_ = sd->status.class_;
+
+	if (!pcdb_checkid(class_)) {
```
sader1992: i did notice that this check wont work if the script give a string job id aka Job_Arch_Bishop
i did test `sc_end_class(-1,Job_Arch_Bisho);` so this must return error but it's not because the check ask for number
and strangely there is no crash or error , it just like the line in the script not there

secretdataz: `Job_Arch_Bisho` is not a valid constant and will be resolved as an unset permanent character variable with value of 0 instead if there is a player attached to the script.

Lemongrass3110: Note: 0 = Novice, so still valid and not an error that we could detect somehow... :(

lighta: simple isn't it some function to check parameter viability ? isValidParam()

src/map/script.cpp
```diff
@@ -11359,6 +11359,11 @@ BUILDIN_FUNC(sc_end_class)
 	else
 		class_ = sd->status.class_;
 
+	if (!pcdb_checkid(class_)) {
+		ShowError("sc_end_class: bad job id given ,[%s] is not a job id.\n", script_getstr(st, 3));
```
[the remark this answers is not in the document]
aleos89: `ShowError("buildin_sc_end_class: Invalid job ID '%d' given.\n", script_getnum(st, 3));`

doc/script_commands.txt
```diff
@@ -5389,7 +5389,8 @@ and theirs val1, val2, val3, and val4 usage in source.
 perform a complete removal of all statuses (although permanent ones will re-apply).
 
 'sc_end_class' works like 'sc_end' but will remove all status effects from any learned
-skill on the invoking character.
+skill on the invoking character. if <job_id> was given it will end the effect for that job.
+if the <char_id> was -1 it will return the <char_id> for the invoking character
```
Atemo: `sc_end_class` + `job_id` will be used in few case. I believe adding a shortcut for attached char ID is meaningless when `getcharid(3)` does the job.

src/map/script.cpp
sader1992: can't we do something like
		`if (pc_jobid2mapid(script_getstr(st, 3)) == -1) {`
			`ShowError("sc_end_class: bad job id given ,[%s] is not a job id.\n", script_getstr(st, 3));`
			`return SCRIPT_CMD_FAILURE;`
		`}`
however i did try that getstr not work `int pc_jobid2mapid(unsigned short b_class)`
and i don't know when the `script_hasdata(st, 3)` will be 0 in the script

aleos89: From your documentation, you're not accepting string values for this parameter. As such, don't confuse the `Job_*` constant values as strings because they are typed out. The only time you'd use `script_getstr` is when you have quotation marks around your parameter.

So in this case you'd use `script_getnum` because the `Job_*` constants are integers.

doc/script_commands.txt
sader1992: true when i did that i was thinking about getpartymember i forgot that even if can be -1 it will eventually needed to use $@partymembercid[]
i will remove the -1 option

src/map/script.cpp
```diff
@@ -23894,7 +23905,7 @@ struct script_function buildin_func[] = {
 	BUILDIN_DEF2(sc_start,"sc_start2","iiii???"),
 	BUILDIN_DEF2(sc_start,"sc_start4","iiiiii???"),
 	BUILDIN_DEF(sc_end,"i?"),
-	BUILDIN_DEF(sc_end_class,"?"),
+	BUILDIN_DEF(sc_end_class,"??"),
```
sader1992: is this need to be `BUILDIN_DEF(sc_end_class,"ii"),` instead of `??` ?

doc/script_commands.txt
```diff
@@ -5389,7 +5389,7 @@ and theirs val1, val2, val3, and val4 usage in source.
 perform a complete removal of all statuses (although permanent ones will re-apply).
 
 'sc_end_class' works like 'sc_end' but will remove all status effects from any learned
-skill on the invoking character.
+skill on the invoking character. if <job_id> was given it will end the effect for that job.
```
aleos89: `If <job_id> is provided it will end the effect for the that job.`

src/map/script.cpp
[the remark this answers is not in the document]
aleos89: Don't forget to adjust this error line. 👍

doc/script_commands.txt
aleos89: Sorry. I just realized I made a typo myself. I corrected the comment above to what it should be. Also, don't forget this: https://github.com/rathena/rathena/pull/2673#discussion_r154879679

PR #2698, opened 2017-12-10
Adventurer Pack (item_id: 16420) bonus script fix

Vennren:
* **Addressed Issue(s)**: The adventurer pack gives stat food items that are not working and not having a continuous reference to items in the same ID range. ID range is consistent now, stat food gives the actual buffs.

* **Server Mode**: Renewal

* **Description of Pull Request**: Ready made item_db changes that will this problem from occurring again.


cydh:
we're going to check for the given Item IDs to confirm it. except you can give us some proof that the Item IDs are correct (link, screen shot, packet sniffing result, or else). thank you

Vennren:
For item ID 16420 there is no further information available on [divine pride](https://www.divine-pride.net/database/item/16420/adventurer-pack), however for an identical item (adventurer pack, id: 12992) there is more information from [divine pride/irowiki](http://db.irowiki.org/db/item-info/12992/).

Either item ID for the consumables can be used as long the effect is properly in place, which it isn't right now. I have currently no other source of information available to further reinforce this pull request, other than going by the descriptions given by the above sources and the original item DB referral IDs for the consumable stat food items and the lack of consistency in their ID range.

Lemongrass3110:
@cydh he only exchanged 12206 with 12449 (the other were just sorted into ascending order).
And he added the SC to the existing items. The only thing we need to confirm is if one of those items does have a different EFST like the normal one, so that they could get stacked. But I do not really think it will be like that.

PR #2702, opened 2017-12-12
Added November and current December clients

Lemongrass3110:

* **Server Mode**: Both

This pull requests adds support for all clients up until 2017-12-06
Credits to @4144

### DIFF
--- a/src/map/clif_obfuscation.hpp
+++ b/src/map/clif_obfuscation.hpp
@@ -396,8 +396,10 @@
 		packet_keys(0x03221B0E,0x73AB7A5F,0x0DD80FD4);
 	#elif PACKETVER == 20171129 // 2017-11-29aRagexeRE
 		packet_keys(0x581B77A7,0x62F74D7A,0x4F102173);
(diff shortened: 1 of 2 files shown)


admkakaroto:
any information about the problem of the party system bug?

secretdataz:
What about the PACKETVER_RE define for 2017-12-06 since it is first ever Ragexe available since mid 2016?

Lemongrass3110:
I did not see that when I copied it over, but we currently do not support the mid 2016 Ragexe clients as well.
If we plan on adding support for Ragnarok Zero clients and kRO main clients or even support for other servers we should come up with a design for storing the packet versions/user configurations.

xEasycore:
It would be great if rA adds support to Ragnarok Zero

idamonli:
The ZERO related update that I tried to transplant HERC/4144 failed
https://github.com/4144/Hercules/commits?author=4144

Lemongrass3110:
This pull request was merged already. ZERO clients are not on our agenda for now.

Issue #2703, opened 2017-12-13
new script like pet <pet_id>;


mjonrest:
* **rAthena Hash**: 

* **Client Date**: 

* **Server Mode**: 

* **Description of Issue**: 
  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**: 
  
  

can we have this kind of script, so we could catch many mob by using 1 taming item

by example:

unripe apple can catch poring or drops, or poporing

Lemongrass3110:
```
*makepet <pet id>;

This command will create a pet egg and put it in the invoking character's
inventory. The kind of pet is specified by pet ID numbers listed in
'db/pet_db.txt'. The egg is created exactly as if the character just successfully
caught a pet in the normal way.

	// This will make you a poring:
	makepet 1002;

Notice that you absolutely have to create pet eggs with this command. If you try
to give a pet egg with 'getitem', pet data will not be created by the char
server and the egg will disappear when anyone tries to hatch it.
```

Lemongrass3110:
Sorry I misread this issue.

sader1992:
pet 0; well tame all the monsters in pet db
i don't know if this is a bug
i can't find the *pet in the script.cpp

aleos89:
```c
	//catch_target_class == 0 is used for universal lures (except bosses for now). [Skotlex]
	if (sd->catch_target_class == 0 && !status_has_mode(&md->status,MD_STATUS_IMMUNE))
		sd->catch_target_class = md->mob_id;
```
Intended behavior.

sader1992:
@aleos89 thanks for the info!

mjonrest:
if i use pet 0; it will catch all mob in pet_db.. but how if we need to catch several mobs only.. can we add more command just like pet <pet_id>; command?

aleos89:
@mjonrest : Honestly I haven't tried that. You could try specifying multiple mobs ie: `pet 1001; pet 1002;` and see how that works.

mjonrest:
#2705 fixed

sader1992:
@aleos89 the problem with using `{ pet 1002; pet 1113; pet 1004; }`
that only process the last command ,so in this example it will only catch 1004 and if you try it on 1002 and 1113 it will always fail

Everade:
Huh?
`{ itemskill 297,1; }`
This worked for over a decade xD

PR #2705, opened 2017-12-14
Update *pet Command Script

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/2703

* **Server Mode**: both

when pet command used like this
pet -1;
it will catch any monster that required the item used to catch
so the control would be from the pet_db.txt

i did think that it may be confusing to use -1 ,however i do think it would be better with -1


sader1992:
need more test sorry ❌

sader1992:
now it work as intended

src/map/pet.cpp
```diff
@@ -660,6 +660,10 @@ int pet_catch_process2(struct map_session_data* sd, int target_id)
 	//catch_target_class == 0 is used for universal lures (except bosses for now). [Skotlex]
 	if (sd->catch_target_class == 0 && !status_has_mode(&md->status,MD_STATUS_IMMUNE))
 		sd->catch_target_class = md->mob_id;
+	
+	//catch_target_class == -1 for catching any mob required the item used
+	if (sd->catch_target_class == -1 && sd->itemid == pet_db[i].itemID)
+			sd->catch_target_class = md->mob_id;
```
cydh: just remove 1 extra tab here

aleos89:
Can you update your branch to match current master (9315781) and adjust your PR appropriately? Thanks!

doc/script_commands.txt
```diff
@@ -9019,16 +9019,21 @@ This command is used in all the item scripts for taming items. Running this
 command will make the pet catching cursor appear on the client connected to the
 invoking character, usable on the monsters with the specified pet ID number. It
 will still work outside an item script.
+if the <pet id> is 0 it will return the monster target id as long as it's in
+the pet_db.txt and not MD_STATUS_IMMUNE monster mode.
+all bosses use MD_STATUS_IMMUNE monster mode so it wont work on bosses.
+see 'doc/mob_db_mode_list.txt' for more info about the monster mode.
+if <pet id> is -1 it will catch the monster that require the item used to catch.
```
aleos89:
```
If the <pet id> is PET_CATCH_UNIVERSAL it will return the monster target ID
as long as it is in the pet_db.txt and does not have the MD_STATUS_IMMUNE
monster mode.
If the <pet id> is PET_CATCH_UNIVERSAL_ITEM it will return the monster target
ID as long as it is in the pet_db.txt and the targeted monster requires the lure
item used.
See 'doc/mob_db_mode_list.txt' for more information about monster modes.
```

sader1992:
i will

sader1992:
i will make a new PR with all the changes i kinda messed up the branch in my local XD

src/map/pet.cpp
```diff
@@ -660,6 +660,10 @@ int pet_catch_process2(struct map_session_data* sd, int target_id)
 	//catch_target_class == PET_CATCH_UNIVERSAL is used for universal lures (except bosses for now). [Skotlex]
 	if (sd->catch_target_class == PET_CATCH_UNIVERSAL && !status_has_mode(&md->status,MD_STATUS_IMMUNE))
 		sd->catch_target_class = md->mob_id;
+	
+	//catch_target_class == PET_CATCH_UNIVERSAL_ITEM is used for catching any monster required the lure item used
+	if (sd->catch_target_class == PET_CATCH_UNIVERSAL_ITEM && sd->itemid == pet_db[i].itemID)
```
Lemongrass3110: Turn this into an else case and we are ready to merge! 👍

doc/script_commands.txt
```diff
@@ -9031,16 +9031,21 @@ This command is used in all the item scripts for taming items. Running this
 command will make the pet catching cursor appear on the client connected to the
 invoking character, usable on the monsters with the specified pet ID number. It
 will still work outside an item script.
+If the <pet id> is PET_CATCH_UNIVERSAL it will return the monster target ID
+as long as it is in the pet_db.txt and does not have the MD_STATUS_IMMUNE monster mode.
```
Lemongrass3110: I'd just say pet database instead of pet_db.txt here. Who knows how long it will be a .txt file...

doc/script_commands.txt
```diff
@@ -9031,16 +9031,21 @@ This command is used in all the item scripts for taming items. Running this
...
+If the <pet id> is PET_CATCH_UNIVERSAL_ITEM it will return the monster target
+ID as long as it is in the pet_db.txt and the targeted monster requires the lure item used.
+See 'doc/mob_db_mode_list.txt' for more information about monster modes.
```
Lemongrass3110: @aleos89 
For both of these above:
"it will return" sounds odd to me, because this command does not return any value.
It would be better to refer to the catching process.
For example:
If the <pet id> is PET_CATCH_UNIVERSAL_ITEM the item will enable to catch the monster target as long as it is in the pet database and the targeted monster requires the lure item used.

doc/script_commands.txt
```diff
@@ -9031,16 +9031,21 @@ This command is used in all the item scripts for taming items. Running this
 command will make the pet catching cursor appear on the client connected to the
 invoking character, usable on the monsters with the specified pet ID number. It
 will still work outside an item script.
+If the <pet id> is PET_CATCH_UNIVERSAL it will return the monster target ID
```
Lemongrass3110: I know this is not your fault, but could you add the alternative names for those commands while we are cleaning up the documentation?
bpet's real name is birthpet and pet's real name is catchpet.
Thanks a lot!

src/map/pet.cpp
sader1992: you mean `else if` right ?

doc/script_commands.txt
sader1992: when i did use `catchpet` instead of pet in the item_db i get error
```
[Error]:
script error on db/import/item_db.txt line 70
    parse_line: expect command, missing function name or calling undeclared function
*   70 : { 'c'atchpet PET_CATCH_UNIVERSAL_ITEM; }
```

sader1992: wont we need to add them in the script.cpp too ?
```
BUILDIN_DEF(catchpet,"i"),
BUILDIN_DEF(birthpet,""),
```

src/map/pet.cpp
aleos89: Correct!

doc/script_commands.txt
aleos89: Yeah those are missing.

PR #2715, opened 2017-12-18
correct mapcache.bat

sader1992:
* **Addressed Issue(s)**: 

* **Server Mode**: both

image

when compile the map cache is map-cache.exe not mapcache.exe
(Microsoft Visual Studio) i don't know if it's the same on cmake
 but it should be


aleos89:
I think the best solution for this would be to rename the Windows project files to to `mapcache` and adjust the rAthena solution to match as well so that we don't have name discrepancies between Windows and Linux.

Issue #2716, opened 2017-12-19
sc_start bug when using -1 on timer


keitenai:
* **rAthena Hash**: 15e509f
* **Client Date**:  20160203

* **Server Mode**: Any

* **Description of Issue**: 
      sc_start gives negative stats to character when using -1 on timer.
      The character's stats will return to normal after warping.

  * How to Reproduce:
      By using the item script with infinite duration:
      '**sc_start SC_EXPBOOST,-1,50**' or '**sc_start SC_WINDWEAPON,-1,1**' or '**sc_start SC_WALKSPEED,-1,150**'
      or '**sc_start SC_FOOD_STR_CASH,-1,5**' and more.
      
* **Modifications that may affect results**:  none

Lemongrass3110:
Just confirmed this. But I'm somewhat thinking that this might be client side issue.

keitenai:
everything works fine with the -1 duration if i add "sleep2 10;" before "sc_start"

keitenai:
I don't know when was this issue got fixed but upon testing this again, the **issue is no longer present**.

This issue is related to:
https://github.com/rathena/rathena/issues/131

PR #2717, opened 2017-12-19
fix negative on timer in sc_start

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/2716

* **Server Mode**: both

when using negative value for the timer the buff stuck with no time
~the PR is for if the timer value is less or equal 0 it will return the time from the skill_db~

after thinking about it there is effects dose not have timer in the skill db (like foods effect) so i am making it return script fail

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -11222,7 +11222,7 @@ BUILDIN_FUNC(sc_start)
 	else
 		bl = map_id2bl(st->rid);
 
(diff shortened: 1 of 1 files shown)


sader1992:
after more thinking it look like a feature not a bug but it's not in the doc folder

Lemongrass3110:
-1 is a valid tick which stands for infinite duration.

PR #2726, opened 2017-12-20
Switched mob db searches to find

Lemongrass3110:

* **Server Mode**: Both

As suggested by @Jeybla
This way no (caught) exceptions will be reported in visual studio anymore

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -75,11 +75,10 @@
 
 //Dynamic mob database
 std::map<uint16, struct mob_db> mob_db_data;
-
-struct mob_db *mob_db( int mob_id ){ 
-	try{
+struct mob_db *mob_db( int mob_id ){
+	if( mob_db_data.find( mob_id ) != mob_db_data.end() ){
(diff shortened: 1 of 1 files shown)


Jeybla:
`As suggested by @Jeybla`
I also suggested to remove these functions and handle missing object in the calling functions, just to avoid nullpointer dereferences :P
Anyway if we still want to handle with pointers it actually may be better to directly store smart pointers in
the mob_db_data. @Lemongrass3110 already said this could save us some memory usage.

src/map/mob.cpp
```diff
@@ -75,11 +75,10 @@
...
 std::map<uint16, struct mob_db> mob_db_data;
-
-struct mob_db *mob_db( int mob_id ){ 
-	try{
+struct mob_db *mob_db( int mob_id ){
+	if( mob_db_data.find( mob_id ) != mob_db_data.end() ){
 		return &mob_db_data.at(mob_id);
-	}catch( std::out_of_range ){
+	}else{
 		return NULL;
```
lighta: nullptr please.

PR #2727, opened 2017-12-20
Turned pet db into a dynamic map

Lemongrass3110:

* **Server Mode**: Both

With this the limit of monsters that can be turned into pets is removed.
Each monster can now be a pet if you put it into the pet db.
With this it will only consume RAM for the declared pets and not for a fixed amount of 300 entries.


src/map/pet.cpp
```diff
@@ -1646,44 +1637,46 @@ void read_petdb()
...
-			pet_db[j].capture=atoi(str[13]);
-			pet_db[j].speed=atoi(str[14]);
-			pet_db[j].s_perfor=(char)atoi(str[15]);
-			pet_db[j].talk_convert_class=atoi(str[16]);
-			pet_db[j].attack_rate=atoi(str[17]);
-			pet_db[j].defence_attack_rate=atoi(str[18]);
-			pet_db[j].change_target_rate=atoi(str[19]);
-			pet_db[j].pet_script = NULL;
-			pet_db[j].pet_loyal_script = NULL;
+			struct s_pet_db* pet = pet_db(nameid);
```
Jeybla: We could directly use `&pet_db_data[nameid]; // adds the entry if missing` here, don't we?

Lemongrass3110: true since we do not have any failure cases below this anymore

src/map/pet.cpp
```diff
@@ -403,10 +406,11 @@ static int pet_return_egg(struct map_session_data *sd, struct pet_data *pd)
  * @param sd : player requesting
  * @param pet : pet requesting
```
lighta: return ?

src/map/pet.cpp
```diff
@@ -655,21 +659,21 @@ int pet_catch_process2(struct map_session_data* sd, int target_id)
 
 	//FIXME: delete taming item here, if this was an item-invoked capture and the item was flagged as delay-consume [ultramage]
 
-	i = search_petDB_index(md->mob_id,PET_CLASS);
+	pet = pet_db(md->mob_id);
```
lighta: if(!pet) return ?;

src/map/pet.cpp
```diff
@@ -1646,44 +1637,42 @@ void read_petdb()
...
+			pet->intimate=atoi(str[11]);
+			pet->die=atoi(str[12]);
+			pet->capture=atoi(str[13]);
+			pet->speed=atoi(str[14]);
+			pet->s_perfor=(char)atoi(str[15]);
+			pet->talk_convert_class=atoi(str[16]);
+			pet->attack_rate=atoi(str[17]);
+			pet->defence_attack_rate=atoi(str[18]);
+			pet->change_target_rate=atoi(str[19]);
+			pet->pet_script = NULL;
```
lighta: can we use nullptr instead NULL ?, nullptr is safer in few case and there is no need to stick with NULL since it's not a C compiler anymore.

src/map/pet.cpp
Lemongrass3110: true!

src/map/pet.cpp
```diff
@@ -403,10 +406,11 @@ static int pet_return_egg(struct map_session_data *sd, struct pet_data *pd)
...
  */
-int pet_data_init(struct map_session_data *sd, struct s_pet *pet)
+bool pet_data_init(struct map_session_data *sd, struct s_pet *pet)
 {
 	struct pet_data *pd;
-	int i = 0, interval = 0;
+	struct s_pet_db *pet_db_ptr;
+	int interval = 0;
 
 	nullpo_retr(1, sd);
```
Lemongrass3110: wrong datatype - should return a bool value.

src/map/pet.cpp
Lemongrass3110: yes of course since those are all .cpp files.

src/map/pet.cpp
Lemongrass3110: Seeing it now again, it's actually wrong since we have comments here that other mobs will display angry emoticons. So we still need to continue the code here.

src/map/pet.cpp
```diff
@@ -662,13 +662,16 @@ int pet_catch_process2(struct map_session_data* sd, int target_id)
...
-	//catch_target_class == PET_CATCH_UNIVERSAL_ITEM is used for catching any monster required the lure item used
-	else if (pet && sd->catch_target_class == PET_CATCH_UNIVERSAL_ITEM && sd->itemid == pet->itemID)
-		sd->catch_target_class = md->mob_id;
+	// If the target is a valid pet, we have a few exceptions
+	if( pet ){
+		//catch_target_class == PET_CATCH_UNIVERSAL is used for universal lures (except bosses for now). [Skotlex]
+		if (sd->catch_target_class == PET_CATCH_UNIVERSAL && !status_has_mode(&md->status,MD_STATUS_IMMUNE)){
+			sd->catch_target_class = md->mob_id;
+		//catch_target_class == PET_CATCH_UNIVERSAL_ITEM is used for catching any monster required the lure item used
+		}else if (pet && sd->catch_target_class == PET_CATCH_UNIVERSAL_ITEM && sd->itemid == pet->itemID){
```
aleos89: I added `pet` here since it wasn't needed in the prior universal check but since you made a large encapsulation then this `pet` check isn't needed.

Issue #2700, opened 2017-12-12
Quest DB Map Crash


ignizh:
* **rAthena Hash**:  d08bc1ce2d7c652f9a5d0c399ed724f248c51dea

* **Client Date**: 20170607

* **Server Mode**:  RE

* **Description of Issue**: Crash of Map server at entering instances with (maybe) duplicated entries of quest_db
  * Result: Here is the core dump, don't know if is this caused by duplicated entries of quest_db or is something else, but I had those quests duplicated, after deleting the duplicated ones, the crash is still there, so, maybe is related with checkquest command.

> Core was generated by `./map-server'.
> Program terminated with signal SIGSEGV, Segmentation fault.
> #0  pc_show_questinfo_sub (sd=0x6c0c950, qi_display=0x0, qi=0x7faf698624fc, show=true) at pc.cpp:12236
> 12236                   if ((*qi_display) != 1) {
> (gdb) bt full
> #0  pc_show_questinfo_sub (sd=0x6c0c950, qi_display=0x0, qi=0x7faf698624fc, show=true) at pc.cpp:12236
> No locals.
> #1  0x0000000000505e27 in pc_show_questinfo (sd=0x6c0c950) at pc.cpp:12322
>         qi = 0x7faf698624fc
>         i = 0
>         mystate = <optimized out>
>         j = <optimized out>
>         failed = <optimized out>
> #2  0x000000000054c91b in buildin_setquest (st=0x7faf7a7767a4) at script.cpp:19269
>         sd = 0x6c0c950
>         quest_id = 15008
>         __FUNCTION__ = "buildin_setquest"
> #3  0x000000000053537d in run_func (st=st@entry=0x7faf7a7767a4) at script.cpp:4049
>         data = <optimized out>
>         i = <optimized out>
>         start_sp = <optimized out>
>         end_sp = <optimized out>
>         func = 537
> #4  0x0000000000539740 in run_script_main (st=0x7faf7a7767a4) at script.cpp:4326
>         cmdcount = 655300
>         gotocount = 2048
>         sd = <optimized out>
>         stack = 0x7faf769e5304
> #5  0x00000000004dbc05 in npc_event_sub (sd=0x6c0c950, ev=0x7faf6b94c9dc,
>     eventname=0x7faf6b989f7c "#delaybios2::OnPCLoginEvent") at npc.cpp:876
> No locals.
[...]

PR #2728, opened 2017-12-20
Fixed a crash caused by setquest in login events

Lemongrass3110:
* **Addressed Issue(s)**: #2700 

* **Server Mode**: Both

Fixed a bug where using setquest in a login event and a player spawning next to a NPC with setquestinfo causing a crash, because the players quest display data was not initialized yet.

Thanks to @ignizh

Reproducible through the following script and logging in at lasagna,173,157 for example.

```
-	script	Issue2700	-1,{
OnPCLoginEvent:
	setquest 1000;
	end;
}
```

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -12256,6 +12256,8 @@ void pc_show_questinfo(struct map_session_data *sd) {
 		return;
 	if (!map[sd->bl.m].qi_count || !map[sd->bl.m].qi_data)
 		return;
+	if (map[sd->bl.m].qi_count != sd->qi_count)
+		return; // init was not called yet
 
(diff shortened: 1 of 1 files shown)


lighta:
So hmm just to understand with this it prevent the crash ok but does the command is nullify then ?

Lemongrass3110:
No in loadendack the init will be called and the packet will be sent out!

Jeybla:
> No in loadendack the init will be called and the packet will be sent out!

Nice. Did you test that?

Lemongrass3110:
Yup I first had a state.new check there, which caused the call at the very bottom of loadendack to cancel out as well :-)

PR #2734, opened 2017-12-22
Dynamic Duel.

lighta:
Remove max number of duel at once.
update duel module. ready for multi-thread, and safer management.

Remove various statically allocated tab.
Remove duplicate code for npc_movenpc.
Fix few warnings.

* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/src/map/atcommand.cpp
+++ b/src/map/atcommand.cpp
@@ -5394,7 +5394,7 @@ ACMD_FUNC(skilloff)
  *------------------------------------------*/
 ACMD_FUNC(npcmove)
 {
-	short x = 0, y = 0, m;
+	short x = 0, y = 0;
 	struct npc_data *nd = 0;
 	char npc_name[NPC_NAME_LENGTH];
 
--- a/src/map/atcommand.cpp
+++ b/src/map/atcommand.cpp
@@ -5412,18 +5412,12 @@ ACMD_FUNC(npcmove)
 		return -1;
 	}
 
-	if ((m=nd->bl.m) < 0 || nd->bl.prev == NULL)
-	{
+	if ( npc_movenpc( nd, x, y ) ) 
+	{ //actually failed to move
 		clif_displaymessage(fd, msg_txt(sd,1154)); // NPC is not on this map.
 		return -1;	//Not on a map.
-	}
-
-	x = cap_value(x, 0, map[m].xs-1);
-	y = cap_value(y, 0, map[m].ys-1);
-	map_foreachinallrange(clif_outsight, &nd->bl, AREA_SIZE, BL_PC, &nd->bl);
-	map_moveblock(&nd->bl, x, y, gettick());
-	map_foreachinallrange(clif_insight, &nd->bl, AREA_SIZE, BL_PC, &nd->bl);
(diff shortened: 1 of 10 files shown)


src/map/duel.hpp
```diff
@@ -12,18 +12,23 @@ struct duel {
...
 };
 
-#define MAX_DUEL 1024 //max number of duels on server
-extern struct duel duel_list[MAX_DUEL]; //list of current duel
-extern int duel_count; //current number of duel on server
+//#define MAX_DUEL 1024 //max number of duels on server
+//extern struct duel duel_list[MAX_DUEL]; //list of current duel
+//extern int duel_count; //current number of duel on server
+
+duel& duel_GetDuelId( size_t did );
```
anacondaqq: duel_getid() will be better, is not it?

src/map/duel.hpp
```diff
@@ -12,18 +12,23 @@ struct duel {
...
+bool duel_exist( size_t did );
+size_t duel_Getcount();
```
anacondaqq: duel_playercount() is not it better? or duel_count()?

src/map/atcommand.cpp
```diff
@@ -8307,14 +8301,12 @@ ACMD_FUNC(invite)
...
-	if(did == 0) {
+	if(did == 0 || !duel_exist(did) ) {
 		clif_displaymessage(fd, msg_txt(sd,350)); // "Duel: @invite without @duel."
 		return 0;
 	}
 
-	if(duel_list[did].max_players_limit > 0 &&
-		duel_list[did].members_count >= duel_list[did].max_players_limit) {
-
+	if( !duel_check_player_limit( duel_GetDuelId(did) ) ){
```
anacondaqq: totally against of mixing UpperCaseAndLower_CaseTogether. Lets stick to only one lower with "_". 😃

src/map/duel.hpp
lighta: Duel_count. Sorry if misleading but this show the num of duel active not the number of player in duel

anacondaqq: then: duel_totalduels

or duel_countfights
or duel_countduels
or duel_countallduels
or duel_totalactive
or duel_countactive
or duel_countactiveduels
or duel_totalcount
or duel_counttotal
or duel_totalactiveduels

xD just suggestions 📟

src/map/duel.hpp
lighta: getbyid then, we don't return the id of the duel but the specific duel from an id. (it's a little confusing since there is the duel module and object)

src/map/duel.hpp
lighta: duel_counttotal() and maybe I'll add duel_countactive() for those that are being accepted.
(the 1st one just return when someone has created one but no fight is going on yet)

laziem:
Hi,
After I update rAthena, everyt time I accepting duel, I got this message `Duel: The limit of players has been reached`. And the duel can't be started.

uddevil:
@laziem i'm having the same issue. I think its because of typo of the condition check ( #2778 )

Issue #2749, opened 2017-12-31
There is no way to programatically determine whether an item is tradable in rAthena's script engine


secretdataz:
* **rAthena Hash**: 5602bb0

* **Description of Issue**: Currently, we can only give players bound, rental item. There is no way to check if an item is tradable via the script engine. This also leads to exploits in item manipulating scripts when it comes to dealing with rental, bound items.

anacondaqq:
To be clear there are many ways to check it. But no direct way via single function for doing it.  Thank you for PR. 😃

secretdataz:
Hmm yeah, I reviewed our script commands and found that you can use `getinventorylist` to retrieve the expire and bound variable of the entire inventory. That's overkill if you wanna check if an item is tradable, though.

Lemongrass3110:
Partially fixed in bfb6972
Commands for inventory are still missing(since expire and bound is not sufficient to know if it can be traded).

aleos89:
Finalized with 7cc59c670822695833e9ac812cd5b46c9146eff2.

PR #2750, opened 2017-12-31
Implemented getequiptradability script command

secretdataz:
* **Addressed Issue(s)**: #2749
* **Server Mode**: Both
This PR provides a new script command `getequiptradability` script command as an optional sanity check against equipped bound, rental items.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -2937,7 +2937,7 @@ If no item ID/name given, all possible items in player's inventory will be merge
 
 ---------------------------------------
 
-*gettradability(<equipment slot>{,<char id>});
+*getequiptradability(<equipment slot>{,<char id>});
 
 Returns true if the item in <equipment slot> is tradable.
 Returns false otherwise.
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -2936,6 +2936,13 @@ Merge all stackable items that separated by GUID flags
 If no item ID/name given, all possible items in player's inventory will be merged.
 
 ---------------------------------------
+
+*gettradability(<equipment slot>{,<char id>});
+
+Returns true if the item in <equipment slot> is tradable.
+Returns false otherwise.
+
+---------------------------------------
 //
 2,1.- End of item-related commands.
 //
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -23703,19 +23703,21 @@ BUILDIN_FUNC(getequiptradability) {
 	num = script_getnum(st, 2);
 
 	if (!script_charid2sd(3, sd)) {
-		script_pushint(st, 0);
 		return SCRIPT_CMD_FAILURE;
 	}
 
 	if (equip_index_check(num))
 		i = pc_checkequip(sd, equip_bitmask[num]);
+	else
+		return SCRIPT_CMD_FAILURE;
+
 	if (i >= 0) {
 		bool tradable = (sd->inventory.u.items_inventory[i].expire_time == 0 &&
 			(!sd->inventory.u.items_inventory[i].bound || pc_can_give_bounded_items(sd)));
 		script_pushint(st, tradable);
 	}
 	else
-		script_pushint(st, 1);
+		script_pushint(st, false);
 
 	return SCRIPT_CMD_SUCCESS;
 }
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -23697,7 +23697,7 @@ BUILDIN_FUNC(round) {
 }
 
 BUILDIN_FUNC(getequiptradability) {
-	int i = -1, num;
+	int i, num;
 	TBL_PC *sd;
 
 	num = script_getnum(st, 2);
@@ -23708,8 +23708,10 @@ BUILDIN_FUNC(getequiptradability) {
 
 	if (equip_index_check(num))
 		i = pc_checkequip(sd, equip_bitmask[num]);
-	else
+	else{
+		ShowError("buildin_getequiptradability: Unknown equip index '%d'\n",num);
 		return SCRIPT_CMD_FAILURE;
+	}
 
 	if (i >= 0) {
 		bool tradable = (sd->inventory.u.items_inventory[i].expire_time == 0 &&
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -23709,8 +23709,11 @@ BUILDIN_FUNC(getequiptradability) {
 
 	if (equip_index_check(num))
 		i = pc_checkequip(sd, equip_bitmask[num]);
-	if (i >= 0)
-		script_pushint(st, !(sd->inventory.u.items_inventory[i].expire_time || (sd->inventory.u.items_inventory[i].bound && !pc_can_give_bounded_items(sd))));
+	if (i >= 0) {
+		bool tradable = (sd->inventory.u.items_inventory[i].expire_time == 0 &&
+			(!sd->inventory.u.items_inventory[i].bound || pc_can_give_bounded_items(sd)));
+		script_pushint(st, tradable);
+	}
 	else
 		script_pushint(st, 1);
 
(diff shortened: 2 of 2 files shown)


doc/script_commands.txt
```diff
@@ -2935,6 +2935,13 @@ Merge all stackable items that separated by GUID flags
 (either by flag 4 item_flag.txt or GUID  in item_group).
 If no item ID/name given, all possible items in player's inventory will be merged.
 
+---------------------------------------
+
+*gettradability(<equipment slot>{,<char id>});
```
anacondaqq: Hi, don't be offended, just suggestion because will use the command many times and the name is very bad for not native english speakers. Because if I wish to check if equip is tradable I starting to think about totally different name like

`equip_istradable(EQI_SLOT, char_id)`
or
`equip_cantrade(EQI_SLOT,char_id)`
or
`equip_tradable(EQI_SLOT,char_id)`

src/map/script.cpp
```diff
@@ -23696,6 +23696,27 @@ BUILDIN_FUNC(round) {
...
 }
 
+BUILDIN_FUNC(getequiptradability) {
+	int i = -1, num;
+	TBL_PC *sd;
+
+	num = script_getnum(st, 2);
+
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, 0);
```
anacondaqq: never do this with line below, because in that combo you will have map-server crash when something is wrong.

Better to have something like:
```
if( !script_charid2sd(3, sd) )
    return SCRIPT_CMD_FAILURE;
```

src/map/script.cpp
```diff
@@ -23696,6 +23696,27 @@ BUILDIN_FUNC(round) {
...
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (equip_index_check(num))
+		i = pc_checkequip(sd, equip_bitmask[num]);
+	if (i >= 0)
+		script_pushint(st, !(sd->inventory.u.items_inventory[i].expire_time || (sd->inventory.u.items_inventory[i].bound && !pc_can_give_bounded_items(sd))));
```
anacondaqq: do not confuse people with this line. Just suggestion by extraction this expression to several small one in one block. Or better pre-define macro function with all important logic for such things.

sader1992:
i would also suggest to add this to `getinventorylist` too example `@inventorylist_trade[]` so we can check for the non equipment items

src/map/script.cpp
```diff
@@ -23696,6 +23696,27 @@ BUILDIN_FUNC(round) {
...
+	else
+		script_pushint(st, 1);
+
+	return SCRIPT_CMD_SUCCESS;
```
anacondaqq: before the line better to add some var if condition expression will have unexcepted behavior which will return to script value. like:
`script_pushint(st, status)`
Where status defined as 0 at the beginning and changed in if condition block to a proper value like 1 if everything success than at the end we will have only two values 0 and 1, 0 if everything is bad, 1 if is good.

doc/script_commands.txt
secretdataz: I'm following the format of old commands. Thanks for the suggestion ♥

src/map/script.cpp
secretdataz: Older commands also do this 'combo'. (see countitem) Does it actually crash the server?

secretdataz:
@sader1992 For getinventorylist,
```
.@tradable = !@inventorylist_expire[.@idx] && !@inventorylist_bound[.@idx];
```
already does the trick.

src/map/script.cpp
anacondaqq: Yes, it will. This is wrong behavior (tested). When no rid and no a "someone" to whom push int value started very strange and difficult to debug bugs and crashes of course. Do not ask me how reproduce it, because it was ~30 days+ ago. But this behavior is 100% incorrect and not need to return any value if SCRIPT_RETURN_FAILURE given. Because this constant is parsed by script engine and as I understand its a part of future PR which will present more detailed "very bad behavior" in the script what happened to map-server.

sader1992: you could `script_rid2sd(sd);`

aleos89: I agree with @anacondaqq on this one. Since we have #2063 in the works, we are going to be removing any return values on valid script failures.

src/map/script.cpp
```diff
@@ -23696,6 +23696,30 @@ BUILDIN_FUNC(round) {
...
+
+	if (equip_index_check(num))
+		i = pc_checkequip(sd, equip_bitmask[num]);
+	if (i >= 0) {
+		bool tradable = (sd->inventory.u.items_inventory[i].expire_time == 0 &&
+			(!sd->inventory.u.items_inventory[i].bound || pc_can_give_bounded_items(sd)));
+		script_pushint(st, tradable);
+	}
+	else
+		script_pushint(st, 1);
```
aleos89: Shouldn't this be returning false instead of 1(true)? The index value would be bad (returned -1) so this should fail as well.

src/map/script.cpp
```diff
@@ -23696,6 +23696,30 @@ BUILDIN_FUNC(round) {
...
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, 0);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (equip_index_check(num))
+		i = pc_checkequip(sd, equip_bitmask[num]);
+	if (i >= 0) {
+		bool tradable = (sd->inventory.u.items_inventory[i].expire_time == 0 &&
+			(!sd->inventory.u.items_inventory[i].bound || pc_can_give_bounded_items(sd)));
```
Lemongrass3110: I'm somehow missing something like `itemdb_cantrade_sub` here?

src/map/script.cpp
```diff
@@ -23696,6 +23696,30 @@ BUILDIN_FUNC(round) {
...
+	TBL_PC *sd;
+
+	num = script_getnum(st, 2);
+
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, 0);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (equip_index_check(num))
```
Lemongrass3110: Should return script command failure if the index check fails as well.

src/map/script.cpp
Lemongrass3110: Yeah I agree with aleos. This might not be a script command failure case, but it should return not tradable at least.

src/map/script.cpp
```diff
@@ -23703,19 +23703,23 @@ BUILDIN_FUNC(getequiptradability) {
...
 		bool tradable = (sd->inventory.u.items_inventory[i].expire_time == 0 &&
-			(!sd->inventory.u.items_inventory[i].bound || pc_can_give_bounded_items(sd)));
+			(!sd->inventory.u.items_inventory[i].bound || pc_can_give_bounded_items(sd)) &&
+			itemdb_cantrade(&sd->inventory.u.items_inventory[i], pc_get_group_level(sd), pc_get_group_level(sd))
+			);
 		script_pushint(st, tradable);
 	}
 	else
-		script_pushint(st, 1);
+		script_pushint(st, false);
```
anacondaqq: script_pushINT() but returning bool. Why?

Lemongrass3110: because our script engine does not know datatypes except integers and strings.

src/map/script.cpp
```diff
@@ -23696,6 +23696,34 @@ BUILDIN_FUNC(round) {
...
+	num = script_getnum(st, 2);
+
+	if (!script_charid2sd(3, sd)) {
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (equip_index_check(num))
+		i = pc_checkequip(sd, equip_bitmask[num]);
+	else
+		return SCRIPT_CMD_FAILURE;
```
Lemongrass3110: ShowError( "buildin_getequiptradability: Unknown equip index '%d'\n", num );

PR #2782, opened 2018-01-07
Fixed a few name typos with Niflheim

Lemongrass3110:

* **Server Mode**: Both

There are a few typos with the name Niflheim all over the place.
I simply searched and replaced them with the correct spelling with only one F.


Daegaladh:
There are some "niffleheim", "niffleim", "niffelheim" and "nifleheim" too.

Lemongrass3110:
I only found niffleheim in the item db, where it is actually spelled incorrectly by iRO:
https://www.divine-pride.net/database/item/20398
I am not really sure what to do about that...
Thanks for the others though! 👍

Everade:
**Niffleheim** is actually correct.
It is traditionally spelled **Niflheim** though.

Lemongrass3110:
Even though I like Marvel their interpretation of the norse mythology is even further away from the real deal than the stories in Ragnarok Online, so we are going to stick with Niflheim, since that also matches the mapname and a lot of other references in NPCs and client side data.

Freyja-Folkvangr:
Are these ???? marks ok?
screen shot 2018-01-17 at 12 15 47

PR #2795, opened 2018-01-10
Workaround to solve issue with news cards in enchants NPCs

Atemo:

* **Server Mode**: Both

Thanks to @Lemongrass3110 to point this issue.


Everade:
@Atemo 
`rAthena Clean\npc\re\merchants\enchan_ko.txt`

That's the script i have initial reported to Lemongrass.
This is where i found out that cards are being deleted.

I think you forgot that one.

PR #2796, opened 2018-01-10
[WIP] Implemented array sorting utility script command

secretdataz:
* **Server Mode**: Irrelevant
Added `sortarray` script command which sorts an integer array with Quicksort + Hoare's partitioning.
There are still rooms for improvement. Suggestions are appreciated.

To-do:
[ ] Add support for string arrays

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -9,6 +9,8 @@
 
 #include "script.hpp"
 
+#include <stack>
+
 #include <math.h>
 #include <stdlib.h> // atoi, strtol, strtoll, exit
 #include <setjmp.h>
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -23728,6 +23730,148 @@ BUILDIN_FUNC(getequiptradability) {
...
+		ShowError("buildin_sortarray: Not a variable!\n");
+		script_reportdata(array_);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	name = reference_getname(array_);
+	ref = reference_getref(array_);
+
+	if (not_server_variable(name[0])) {
```
Lemongrass3110: Merge inner and outer if

src/map/script.cpp
```diff
@@ -23728,6 +23730,148 @@ BUILDIN_FUNC(getequiptradability) {
...
+ * @param end Ending index
+ */
+int partition_array(script_state *st, map_session_data *sd, script_data *array_, int start, int end) {
+	auto ref = reference_getref(array_);
+	auto id = reference_getid(array_);
+	auto name = reference_getname(array_);
+	int i = start, j = end;
+	int pivotidx = rnd() % (end - start);
+	void* pivot = get_val2(st, reference_uid(id, start + pivotidx), ref);
+	bool is_string = is_string_variable(name);
```
Lemongrass3110: Maybe use a  function pointer instead? One that compares two void pointers and return -1,0,1?
Maybe even a delegate would be enough?

src/map/script.cpp
```diff
@@ -23728,6 +23730,148 @@ BUILDIN_FUNC(getequiptradability) {
...
+ * @param idx1 index of first element
+ * @param idx2 index of second element 
+ */
+void swap_array_element(script_state *st, map_session_data *sd, script_data* array_, int idx1, int idx2) {
+	void *v1, *v2;
+	auto id = reference_getid(array_);
+	auto name = reference_getname(array_);
+	auto ref = reference_getref(array_);
+
+	v1 = get_val2(st, reference_uid(id, idx1), ref);
```
Lemongrass3110: would be nice to store the two uids and reuse them

mimishi:
Suggestion if it is possible to sort highest to lowest?

Issue #2804, opened 2018-01-14
Wrong message on skills


Daegaladh:
* **rAthena Hash**: a6a2c4301bcf295b1769411a9123fd6a39ba613f

* **Client Date**: 2015-05-13

* **Server Mode**: Both

* **Description of Issue**: 
  * Result: When using Aura Blade without a weapon a white message appears "[Revolver] weapon class must be equipped."
  * Expected Result: Red message "The skill cannot be used with this weapon."
  * How to Reproduce: Just use Aura Blade without a weapon.
  * Official Information:
  

* **Modifications that may affect results**:

Artuvazro:
Confirmed on clean installation:

https://i.imgur.com/dTkPhHQ.jpg

sader1992:
![image](https://user-images.githubusercontent.com/5981261/34920905-219a52b8-f983-11e7-95c1-6a1b06a2949e.png)

however i don't know what the Expected Result

Daegaladh:
Futher testing revealed it happens with all skills that include revolver as required weapon in the skill_require_db.

PR #2808, opened 2018-01-15
Fixed required weapon messages

Daegaladh:
* **Addressed Issue(s)**: #2804

* **Server Mode**: both

* **Description of Pull Request**: Fixes the required weapon messages, showing the specific ones only when that only weapon is required.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -15672,18 +15672,26 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 			case RA_AIMEDBOLT:
 				break;
 			default:
-				if (require.weapon&(1<<W_REVOLVER))
-					clif_msg(sd, SKILL_NEED_REVOLVER);
-				else if (require.weapon&(1<<W_RIFLE))
-					clif_msg(sd, SKILL_NEED_RIFLE);
-				else if (require.weapon&(1<<W_GATLING))
-					clif_msg(sd, SKILL_NEED_GATLING);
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -15672,18 +15672,26 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
...
-					clif_msg(sd, SKILL_NEED_RIFLE);
-				else if (require.weapon&(1<<W_GATLING))
-					clif_msg(sd, SKILL_NEED_GATLING);
-				else if (require.weapon&(1<<W_SHOTGUN))
-					clif_msg(sd, SKILL_NEED_SHOTGUN);
-				else if (require.weapon&(1<<W_GRENADE))
-					clif_msg(sd, SKILL_NEED_GRENADE);
-				else
-					clif_skill_fail(sd, skill_id, USESKILL_FAIL_THIS_WEAPON, 0);
+				switch((unsigned int)log2(require.weapon)) {
```
Daegaladh: Not sure if this is the best way of reversing the bitwise, but at least it works.

Lemongrass3110: I guess in this case it would have been best to exchange the bitwise and with an equality operator.

Since log2 is C99 and we are only featuring VS2013 and above we should be fine with your approach as well though 👍

aleos89: The only way I can think of is to do like @Lemongrass3110 suggested and have specific equality checks. Maybe just leave a comment and say what the `log2` function is used here for if we keep it this way. :)

PR #2810, opened 2018-01-17
Heal Cleanup

syracuase:
* **Addressed Issue(s)**: -
* **Server Mode**:  Renewal
((Base Level + INT)/5) * 30

could be further simplified to

(Base_Level+INT)*6

* **Addressed Issue(s)**: 

* **Server Mode**: 


Lemongrass3110:
The compiler will optimize that anyway. In this case we kept it more for readability. :-)

aleos89:
As pointed out by @Stolao as well, this will change the way bonuses are given every 5 INT. Thanks for your contribution!

PR #2829, opened 2018-01-25
SendMail Command Script

sader1992:

* **Server Mode**: both

so this just an idea for now
the codes not completed yet
but i want to know what you think about it first

my plan is making a mail command can send zeny and items in it

this code work for zeny but not yet the items
i want to find the best way to make the command more user friendly for the items 
i did think about 
`SendMail(strcharinfo(0), "title", "body", 5, <zeny>,<itemid>,<amount>,<itemid>,<amount>);`till 5 items
but it wont look good

so i am thinking of making 3 or more commands like this
```
SendMail.item(<itemid>,<amount>);
SendMail.item2(<item2 values>);
SendMail.zeny(<zeny>);
SendMail.send(strcharinfo(0), "title", "body", 5);
```
i don't know if this would work 
those command work like we give the attachment for the mail then we send it

### DIFF
--- a/src/map/mail.cpp
+++ b/src/map/mail.cpp
@@ -429,6 +429,10 @@ void npc_mail_send(struct npc_data *nd, const char *dest_name, const char *title
 	msg.zeny = zeny;
 	//msg.item[1].nameid = 502;
 	//msg.item[1].amount = 1;
+	for (int i = 0; i < MAIL_MAX_ITEM; i++) {
+		msg.item[i].nameid = 0;
+		msg.item[i].amount = 0;
+	}
 	msg.id = 0; // id will be assigned by charserver
 	msg.send_id = nd->bl.id;
 	msg.dest_id = 0; // will attempt to resolve name
--- a/src/map/mail.cpp
+++ b/src/map/mail.cpp
@@ -12,6 +12,7 @@
 #include "itemdb.hpp"
 #include "clif.hpp"
 #include "pc.hpp"
+#include "npc.hpp"
 #include "intif.hpp"
 #include "date.hpp" // date_get_dayofyear
 #include "log.hpp"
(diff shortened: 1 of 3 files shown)


src/map/mail.cpp
```diff
@@ -418,3 +419,39 @@ void mail_refresh_remaining_amount( struct map_session_data* sd ){
...
 }
+
+void npc_mail_send(struct npc_data *nd, const char *dest_name, const char *title, const char *body_msg, int body_len, int zeny) {
+	struct mail_message msg;
+
+	if (body_len > MAIL_BODY_LENGTH)
+		body_len = MAIL_BODY_LENGTH;
+
+	msg.zeny = zeny;
+	//msg.item[1].nameid = 502;
```
anacondaqq: typo?

src/map/mail.cpp
```diff
@@ -418,3 +419,39 @@ void mail_refresh_remaining_amount( struct map_session_data* sd ){
 		sc_start2( &sd->bl, &sd->bl, SC_DAILYSENDMAILCNT, 100, doy, 0, -1 );
 	}
 }
+
+void npc_mail_send(struct npc_data *nd, const char *dest_name, const char *title, const char *body_msg, int body_len, int zeny) {
```
anacondaqq: 1. Why VOID? Why not to try to use ENUMs with named constants which will signal that something wrong?
2. suggestion: const char body_msg to just msg, and body_len to msg_len

src/map/mail.cpp
```diff
@@ -418,3 +419,39 @@ void mail_refresh_remaining_amount( struct map_session_data* sd ){
...
+
+	msg.zeny = zeny;
+	//msg.item[1].nameid = 502;
+	//msg.item[1].amount = 1;
+	for (int i = 0; i < MAIL_MAX_ITEM; i++) {
+		msg.item[i].nameid = 0;
+		msg.item[i].amount = 0;
+	}
+	msg.id = 0; // id will be assigned by charserver
+	msg.send_id = nd->bl.id;
```
anacondaqq: add nullpointer check for nd->bl.id before usage.

src/map/mail.cpp
```diff
@@ -418,3 +419,39 @@ void mail_refresh_remaining_amount( struct map_session_data* sd ){
...
+	msg.dest_id = 0; // will attempt to resolve name
+	safestrncpy(msg.send_name, nd->name, NAME_LENGTH);
```
anacondaqq: nd->name nullpointer check before usage prefered.

src/map/mail.cpp
```diff
@@ -418,3 +419,39 @@ void mail_refresh_remaining_amount( struct map_session_data* sd ){
...
+	safestrncpy(msg.dest_name, (char*)dest_name, NAME_LENGTH);
+	safestrncpy(msg.title, (char*)title, MAIL_TITLE_LENGTH);
+	msg.type = MAIL_INBOX_NORMAL;
```
anacondaqq: how about adding this to parameters list?
The whole function will be agiler if will be added the type options to the parameters...
So people can decide what type of mail they wish to send 🍅

src/map/mail.cpp
```diff
@@ -418,3 +419,39 @@ void mail_refresh_remaining_amount( struct map_session_data* sd ){
...
+
+	if (msg.title[0] == '\0') {
+		return; // Message has no length and somehow client verification was skipped.
+	}
+
+	if (body_len)
+		safestrncpy(msg.body, (char*)body_msg, body_len + 1);
+	else
+		memset(msg.body, 0x00, MAIL_BODY_LENGTH);
```
anacondaqq: is not better replace it with `\0`? (I can be wrong)

src/map/script.cpp
```diff
@@ -24376,6 +24392,8 @@ struct script_function buildin_func[] = {
 	BUILDIN_DEF2(round, "ceil", "i"),
 	BUILDIN_DEF2(round, "floor", "i"),
 	BUILDIN_DEF(getequiptradability, "i?"),
+	
+	BUILDIN_DEF(sendmail, "?????"),
```
anacondaqq: why?
id, id, str, str, str, num, num
Why ??????  then? It can be a little bit confused.

Just suggestion:
iisssii

Daegaladh:
I suggest to name it just mail, since can be confusing with the sendmail commands in other languages that send actual email.

Lemongrass3110:
As discussed with @sader1992 in private chat.

PR #2840, opened 2018-01-27
(Clear) command script

sader1992:
* **Addressed Issue(s)**: NONE

* **Server Mode**: Both

so `clear;` command will clear the dialog and continues the script

## Why this command script would be usefull:
so here i made 2 examples on how we can use `clear` it may have a lot of usage too
### Example 1:
when we need to switch options and rewrite the dialog normally we do `next` before or inside the case
so this example is on the switch
```
prontera,155,193,3	script	clear test1	826,{
	mes"How Can I Help You ?";
	switch(select("clear:next")){
		case 1:
			clear;
			mes"As you wish";
			break;
		case 2:
			next;
			mes"OK";
			break;
	}
	end;
}
```
### Example 2:
we can use it in other ways `get creative`
so here is an example that came to me while i am adding the PR
```
prontera,155,193,3	script	clear test2	826,{
	mes"hi";
	sleep2 3000;
	clear;
	mes"i am npc";
	sleep2 3000;
	clear;
	mes"and i am testing new script command";
	sleep2 3000;
	clear;
	mes"see you later my friend!";
	close;
}
```

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -1150,6 +1150,20 @@ will terminate.
 
 ---------------------------------------
 
+*clear;
+
+This command will clear the dialog text and continues the script without player interaction.
+
+	mes"This how clear command work .";
+	sleep2 3000;
+	clear;	//this will clear what in the dialog and move to the next one.
+	mes"i will show you again .";
+	sleep2 3000;
+	clear;
+	mes"see you latter my friend";
+
+---------------------------------------
+
 *close;
 
 This command will create a 'close' button in the message window for the invoking
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -1152,15 +1152,17 @@ will terminate.
 
 *clear;
 
-This command will clear the dialog text and continues the script without player interaction.
+This command will clear the dialog text and continue the script without player interaction.
 
-	mes"This how clear command work .";
+Example:
+	mes "This is how the 'clear' script command works.";
 	sleep2 3000;
-	clear;	//this will clear what in the dialog and move to the next one.
-	mes"i will show you again .";
+	clear; // This will clear the dialog and continue to the next one.
+	mes "I will show you again.";
 	sleep2 3000;
 	clear;
-	mes"see you latter my friend";
+	mes "Bye!";
+	close;
 
 ---------------------------------------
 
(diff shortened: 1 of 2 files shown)


hurtsky:
pretty nice, like `next;` but no need click `next` 👍

src/map/script.cpp
```diff
@@ -5087,6 +5087,21 @@ BUILDIN_FUNC(next)
...
 
+/// Clear The dialog text and continues the script without Next button.
+///
+/// clear;
+BUILDIN_FUNC(clear)
+{
+	TBL_PC* sd;
+
+	if (!script_rid2sd(sd))
+		return SCRIPT_CMD_SUCCESS;
```
Lemongrass3110: please make it return script_cmd_failure in this case.

src/map/script.cpp
```diff
@@ -5087,6 +5087,21 @@ BUILDIN_FUNC(next)
...
+
+	st->state = RUN;
```
Lemongrass3110: Is changing the state really necessary?
It should be running anyway. Otherwise it would not reach the statement.

doc/script_commands.txt
```diff
@@ -1150,6 +1150,20 @@ will terminate.
...
+
+This command will clear the dialog text and continues the script without player interaction.
+
+	mes"This how clear command work .";
+	sleep2 3000;
+	clear;	//this will clear what in the dialog and move to the next one.
+	mes"i will show you again .";
+	sleep2 3000;
+	clear;
+	mes"see you latter my friend";
```
Lemongrass3110: Missing a blank between mes and the parameter in all mes calls.

Lemongrass3110: @aleos89 could you do an English cleanup?

aleos89:
```
*clear;

This command will clear the dialog text and continue the script without player interaction.

Example:
	mes "This is how the 'clear' script command works.";
	sleep2 3000;
	clear; // This will clear the dialog and continue to the next one.
	mes "I will show you again.";
	sleep2 3000;
	clear;
	mes "Bye!";
	close;

```

src/map/script.cpp
```diff
@@ -5087,6 +5087,21 @@ BUILDIN_FUNC(next)
 	return SCRIPT_CMD_SUCCESS;
 }
 
+/// Clear The dialog text and continues the script without Next button.
```
aleos89: `Clears the dialog and continues the script without a next button.`

PR #2849, opened 2018-01-29
Mob.cpp update for Mob_Avail

TheDerpySupport:
* **Addressed Issue(s)**: 
With Bodystyles being official in rAthena now this allows Mob_Avail to use them.

* **Server Mode**: 
Both

Allows Mob_Avail to support Bodystyles


aleos89:
This is something I would like to look into after #2780 is merged so that `mob_avail` could be YAML-fied as well. This also coincides with #2580.

We appreciate the PR and don't want this closure to discourage from future support! Thanks!

Issue #2853, opened 2018-01-31
AL_HOLYWATER casting on water without bottles


Daegaladh:
* **rAthena Hash**: 31988abd1919a42e7ccedaf597fa13d4a7308865

* **Client Date**: Not relevant

* **Server Mode**: Both

* **Description of Issue**: AL_HOLYWATER should not cast or consume SP on water when not having empty bottles.
  * Result: It casts and consume SP despite showing the skill fail message.
  * Expected Result: AL_HOLYWATER should not cast or consume SP when not having empty bottles, both in water (and water skills) and outside water.
  * How to Reproduce: 
  * Official Information: See "Expected Result", tested on iRO.
  

* **Modifications that may affect results**:

aleos89:
Holy Water uses this state check from `skill_require_db`:
```
31,0,0,10,0,0,0,99,0,0,water,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0								//AL_HOLYWATER
```

Lemongrass3110:
Hmm I think I remember @Playtester was saying something like this once?

Playtester:
I think you are referring to "Further testing" post I made here: https://github.com/rathena/rathena/issues/1119

Daegaladh:
Apparently euRO never shows "skill has failed" messages, even if you have the chat option active.
Tested again on iRO, it shows "skill has failed" when requirements aren't met, the rest remains the same.

Daegaladh:
Ok after more testing i've confirmed WZ_WATERBALL is working correctly, and AL_HOLYWATER just needs a little fix.

mrjnumber1:
MG_FIREWALL has similar behavior. the player has its own internal firewall count variable. once they've got 3 used already, the skill will instantly fail to cast. on rathena, it tries to cast, and then checks that if the player has maxcount after cast completion, rather than before starting.

Daegaladh:
@mrjnumber1 You're right, I've just tested it on iRO Classic

PR #2856, opened 2018-01-31
Fixed AL_HOLYWATER and WZ_WATERBALL to match Aegis behavior

Daegaladh:
* **Addressed Issue(s)**: #2853, #1119

* **Server Mode**: Both

Fixed AL_HOLYWATER and WZ_WATERBALL to match Aegis behavior (should not show failed skill messages).
Also fixed failed skill messages showing on skills casted on players outside pvp.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -14968,6 +14968,7 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 		case AL_HOLYWATER:
 			if(pc_search_inventory(sd,ITEMID_EMPTY_BOTTLE) < 0)
 				return false;
+			break;
 		case MO_CALLSPIRITS:
 			if(sc && sc->data[SC_RAISINGDRAGON])
 				skill_lv += sc->data[SC_RAISINGDRAGON]->val1;
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -7568,8 +7568,6 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 					skill_delunit(su);
 				clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
 			}
-			else
-				clif_skill_fail(sd,skill_id,USESKILL_FAIL_LEVEL,0);
 		}
 		break;
 
@@ -11090,7 +11088,7 @@ static int8 skill_castend_id_check(struct block_list *src, struct block_list *ta
 			case AB_CLEARANCE:
 				return USESKILL_FAIL_TOTARGET;
 			default:
-				return USESKILL_FAIL_LEVEL;
+				return USESKILL_FAIL_MAX;
 		}
 	}
 
@@ -14967,6 +14965,9 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 				return false;
 			}
 			break;
+		case AL_HOLYWATER:
+			if(pc_search_inventory(sd,ITEMID_EMPTY_BOTTLE) < 0)
+				return false;
 		case MO_CALLSPIRITS:
 			if(sc && sc->data[SC_RAISINGDRAGON])
 				skill_lv += sc->data[SC_RAISINGDRAGON]->val1;
@@ -15571,7 +15572,6 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 				break;
 			if (map_getcell(sd->bl.m,sd->bl.x,sd->bl.y,CELL_CHKWATER) && !map_getcell(sd->bl.m,sd->bl.x,sd->bl.y,CELL_CHKLANDPROTECTOR))
 				break;
-			clif_skill_fail(sd,skill_id,USESKILL_FAIL_LEVEL,0);
 			return false;
 		case ST_RIDINGDRAGON:
 			if( !pc_isridingdragon(sd) ) {


src/map/skill.cpp
```diff
@@ -11090,7 +11088,7 @@ static int8 skill_castend_id_check(struct block_list *src, struct block_list *ta
 			case AB_CLEARANCE:
 				return USESKILL_FAIL_TOTARGET;
 			default:
-				return USESKILL_FAIL_LEVEL;
+				return USESKILL_FAIL_MAX;
```
Daegaladh: Not sure if this changes other intended behaviors too, but in Aegis damage skills casted on players on non-pvp maps should not show any message.

src/map/skill.cpp
```diff
@@ -14967,6 +14965,9 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 				return false;
 			}
 			break;
+		case AL_HOLYWATER:
+			if(pc_search_inventory(sd,ITEMID_EMPTY_BOTTLE) < 0)
```
Lemongrass3110: I'm just thinking if it makes sense to put this into skill_require_db. But I cannot remember anymore where we had a similar requirement. I think it was one of the Mechanic skills... Just look for something like this code snippet:
`struct skill_condition require = skill_get_requirement(sd, skill_id, skill_lv);`

src/map/skill.cpp
```diff
@@ -7568,8 +7568,6 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 					skill_delunit(su);
 				clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
 			}
-			else
-				clif_skill_fail(sd,skill_id,USESKILL_FAIL_LEVEL,0);
```
Lemongrass3110: Are you sure that this is not producing an error message?
For example if you have a full inventory?

src/map/skill.cpp
Lemongrass3110: Also it looks weird without a break in this switch case.

src/map/skill.cpp
```diff
@@ -15571,7 +15572,6 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 				break;
 			if (map_getcell(sd->bl.m,sd->bl.x,sd->bl.y,CELL_CHKWATER) && !map_getcell(sd->bl.m,sd->bl.x,sd->bl.y,CELL_CHKLANDPROTECTOR))
 				break;
-			clif_skill_fail(sd,skill_id,USESKILL_FAIL_LEVEL,0);
```
Lemongrass3110: Undocumented change?

src/map/skill.cpp
Daegaladh: Yeah, I'm sure. And you can't use skills over 90% of weight, so that could never happen.

src/map/skill.cpp
Daegaladh: But adding it to the skill_require_db would make it return a "skill failed" message right?
Btw when testing this pc_search_inventory, I noticed it returns -1 when have no empty bottles, and 15 always when have 1 or more, weird.
It doesn't need a break, since it just checks if you don't have bottles does nothing, otherwise continues to the ST_WATER case.

src/map/skill.cpp
Daegaladh: In aegis this never shows "skill failed" messages.

src/map/skill.cpp
Lemongrass3110: I'm not talking about 90% weight limit, but 100 item slots.

src/map/skill.cpp
Lemongrass3110: It returns the inventory index(-1 if not found, 0 to MAX_INVENTORY=100 if found).
You can then access the player's inventory with this index and check the count.

I'm just wondering because the case was not there and I'm a bit unsure if the skill level of HOLYWATER should really be increased by SC_RAISINGDRAGON.
The water check should be done elsewhere I guess, since the HOLYWATER case has not been there before and it was still checked if you are standing on water.

src/map/skill.cpp
Lemongrass3110: It's strange I've never heard of the requirement, that CALL SPIRIT requires to be standing on water?

src/map/skill.cpp
Daegaladh: No, that belongs to the skill MO_CALLSPIRITS and isn't related to my changes XD
The case I added just forces to do nothing when the player doesn't have any empty bottles.
Without this, the skill would cast and show a skill failed message, and that's not what happens in Aegis, it should do just nothing.

src/map/skill.cpp
Daegaladh: ? I think you're mixing things, this isn't related to MO_CALLSPIRITS, but AL_HOLYWATER and WZ_WATERBALL

src/map/skill.cpp
Lemongrass3110: Since you did not add a break it will run through these cases as well. That's what I'm saying!
See:
"Also it looks weird without a break in this switch case."

src/map/skill.cpp
Artuvazro: imagen

src/map/skill.cpp
Daegaladh: Ohh, right!

Lemongrass3110: # is used for issues and pull requests :)
But now it's good! 👍

src/map/skill.cpp
Lemongrass3110: Is the holy water dropped on the floor or not?

Daegaladh: Nope, is not generated.

src/map/skill.cpp
Lemongrass3110: You removed this code portion from MO_CALLSPRITIS and not WZ_WATERBALL.
That's why I do not understand this. :(

Daegaladh: No, this is case ST_WATER ._.

src/map/skill.cpp
```diff
@@ -15571,7 +15572,6 @@ bool skill_check_condition_castbegin(struct map_session_data* sd, uint16 skill_i
 				break;
```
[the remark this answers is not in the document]
Daegaladh: No, that's github visualization fault, look at the line numbers, that break doesn't belong to that case...

Daegaladh:
Apparently euRO never shows "skill has failed" messages, even if you have the chat option active so, I'll close this PR and make a new one with the correct behavior, now tested on iRO.

src/map/skill.cpp
Lemongrass3110: Ah damn github tricked me :(
Sorry totally my fault here!

PR #2860, opened 2018-01-31
Fixed AL_HOLYWATER casting and consuming sp when having no bottles

Daegaladh:
* **Addressed Issue(s)**: #2853

* **Server Mode**: Both

* **Description of Pull Request**: AL_HOLYWATER should not cast in water when you have no bottles, just shows a skill failed message.

It's done this way because skill_require_db would cause the skill to cast and show the empty bottle required message instead.


mrjnumber1:
MG_FIREWALL should work a similar way. the player has its own internal firewall count variable. once they've got 3 used already, the skill will instantly fail to cast.

PR #2873, opened 2018-02-06
Re-gain Bioethics on Skill Reset

syracuase:
If a PC reset using neuralizer, there is currently no way that they can regain the bioethics skill. Modified first and second check to fallback to teaching the skill if AM_BIOETHICS is not available.


Atemo:
Neuralizer doesn't reset quest skill unless you are setting `quest_skill_reset` to yes, at your own risk.
https://github.com/rathena/rathena/blob/master/conf/battle/player.conf#L44

Issue #2875, opened 2018-02-07
Rebellion Skill 'Heat Barrel'


rye305:
rAthena Hash: 4171cd1

Client Date: 25 jan 2017

Server Mode: Renewal

Result: Heat Barrel consumes all coins but only considers that one coin is used no matter what the amount of coins consumed was. Thus, giving minimal ATK and ASPD bonus.

Expected Result: Heat Barrel should consumes all coins and scale according to the amount of coins consumed. Namely, 16% ATK per coin for a maximum of 160%

Source: https://irowiki.org/wiki/Hit_Barrel

This bug along with the Crimson Marker bug (https://github.com/rathena/rathena/issues/2863) and the Bound Trap bug (https://github.com/rathena/rathena/issues/2874) Is crippling the rebel class on my server and causing a high number of complaints.

admkakaroto:
is missing ASPD bonus:
```
	[SKID.RL_HEAT_BARREL] = {
		"Heat Barrel (Acceleration Bullet)",
		"Max Level: 5",
		"^CC3399Requirement: Rich's Coin 1^000000",
		"Skill Form: ^339900Buff^000000",
		"Description: ^777777Consumes all coins to",
		"increase ASPD and ATK, reduces fixed cast",
		"time at the cost of HIT. Skill's level",
		"affect ASPD, ATK and fixed cast time bonus",
		"and HIT reducement. If the caster",
		"switches equipment during the duration,",
		"the skill's effect will dissapear. This skill",
		"can't be used with Madness Canceller or",
		"Platinum Alter. Skill duration is 60 sec.^000000",
		"[Lv 1]: ^777777ASPD +1, HIT -30^000000",
		"[Lv 2]: ^777777ASPD +2, HIT -35^000000",
		"[Lv 3]: ^777777ASPD +3, HIT -40^000000",
		"[Lv 4]: ^777777ASPD +4, HIT -45^000000",
		"[Lv 5]: ^777777ASPD +5, HIT -50^000000"
	},
```


Font: Zackdreaver GitHUB!

rye305:
according to https://github.com/rathena/rathena/issues/2857 the ASPD part is working as intended. everything else isnt tho and this is one of the core skills of rebel

admkakaroto:
test:

after:
https://imgur.com/IWI1n05

before:
https://imgur.com/vfw1705

no change, I use the latest version of rathena.

aleos89:
[...]

PR #2882, opened 2018-02-08
Corrected Heat Barrel bonuses

aleos89:
* **Addressed Issue(s)**: #2875

* **Server Mode**: Renewal

  * Attack was not factoring in coins.
  * Corrected fixed cast time to also factor in coins.
Thanks to @rye305!

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -10948,10 +10948,16 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			val2 = src->id;
 			break;
 		case SC_HEAT_BARREL:
-			//kRO Update 2016-05-25
-			val2 = val1 * 5; // -fixed casttime
-			val3 = 6 + val1 * 2; // ATK
-			val4 = 25 + val1 * 5; // -hit
+			{
+				uint8 n = 10;
+				if (sd)
+					n = (uint8)sd->spiritball_old;
+
+				//kRO Update 2016-05-25
+				val2 = n * 5; // -fixed casttime
+				val3 = (6 + val1 * 2) * n; // ATK
+				val4 = 25 + val1 * 5; // -hit
+			}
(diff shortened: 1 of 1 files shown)


rye305:
working fine my man!

LordWhiplash:
I was looking at Rebellion skills and noticed that in Heat Barrel, the description says that the ASPD is modified by a fixed number, rather than a percentage 

http://kafra.kr/#!/en/KRO/skillinfo/2568

aleos89:
rAthena currently grants Heat Barrel at a fixed value, not as a percentage.

LordWhiplash:
I've just applied the #4428 and checked on master, the place and code of the bonus is exactly equals mine, and in my tests, the aspd is in %.

Just to confirm i'm not missinterpreting what fixed means... it means that ASPD bonus will ever be +5, independent of user stats, right?

image
(1 agi, no Heat Barrel)

image
(1 agi, Heat Barrel level 5)

image
(130 agi, no Heat Barrel)

image
(130 agi, Heat Barrel level 5)

aleos89:
I tested on kRO and confirmed it is a fixed amount.

erosama28:
i want to confirm the formula above increases the attack at the end of the formula. AFAIK it factors only weapon attack much like racial modifier correct .

erosama28:
`	//Skill damage modifiers that stack linearly
	if(sc && skill_id != PA_SACRIFICE) {
		if(sc->data[SC_OVERTHRUST])
			skillratio += sc->data[SC_OVERTHRUST]->val3;
		if(sc->data[SC_MAXOVERTHRUST])
			skillratio += sc->data[SC_MAXOVERTHRUST]->val2;
		if (sc->data[SC_HEAT_BARREL])` i'm trying to code this on my own server and based on all video i reviewed heat barrel should add the modifier linearly.

Issue #2900, opened 2018-02-12
pet_data::petDB points to dangling pointer after reloadmobdb usage


secretdataz:
* **rAthena Hash**: b48f0e8
* **Client Date**: Irrelevant
* **Server Mode**: Both
* **Description of Issue**: 
  * Result: pet_data::petDB points to dangling pointers.
  * Expected Result: pet_data::petDB points to the correct s_pet_db entry.
  * How to Reproduce:

1. Hatch a pet
2. Use `@reloadmobdb`
3. Return it to egg
4. Segfault may be triggered or the pet may yield an incorrect egg

* **Modifications that may affect results**: None

PR #2901, opened 2018-02-12
Converted pet_data::petDB into a getter function

secretdataz:
* **Addressed Issue(s)**: #2900 
* **Server Mode**: Both
pd->petDB pointer is not updated upon monster database reloading. This PR fixes the issue by removing the pointer storing behavior and put a getter function to retrieve the pointer instead.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -2882,7 +2882,7 @@ void status_calc_pet_(struct pet_data *pd, enum e_status_calc_opt opt)
 		memcpy(&pd->status, &pd->db->status, sizeof(struct status_data));
 		pd->status.mode = MD_CANMOVE; // Pets discard all modes, except walking
 		pd->status.class_ = CLASS_NORMAL;
-		pd->status.speed = pd->petDB->speed;
+		pd->status.speed = pd->get_pet_db()->speed;
 
 		if(battle_config.pet_attack_support || battle_config.pet_damage_support) {
 			// Attack support requires the pet to be able to attack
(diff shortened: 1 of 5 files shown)


src/map/status.cpp
```diff
@@ -3724,9 +3724,11 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
 
 	if( sd->pd ) { // Pet Bonus
 		struct pet_data *pd = sd->pd;
-		if( pd && pd->petDB && pd->petDB->pet_loyal_script && pd->pet.intimate >= battle_config.pet_equip_min_friendly )
-			run_script(pd->petDB->pet_loyal_script,0,sd->bl.id,0);
-		if( pd && pd->pet.intimate > 0 && (!battle_config.pet_equip_required || pd->pet.equip > 0) && pd->state.skillbonus == 1 && pd->bonus )
+		s_pet_db *pet_db_ptr = pd->get_pet_db();
+
+		if( pet_db_ptr != nullptr && pd->get_pet_db() && pd->get_pet_db()->pet_loyal_script && pd->pet.intimate >= battle_config.pet_equip_min_friendly )
```
Lemongrass3110: Getter is called way too often here. you stored it in the pointer already!

src/map/status.cpp
```diff
@@ -3724,9 +3724,11 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
...
+			run_script(pd->get_pet_db()->pet_loyal_script,0,sd->bl.id,0);
+		if( pet_db_ptr != nullptr && pd->pet.intimate > 0 && (!battle_config.pet_equip_required || pd->pet.equip > 0) && pd->state.skillbonus == 1 && pd->bonus )
```
Lemongrass3110: Is pet_db_ptr really needed here?

PR #2902, opened 2018-02-12
Made the second parameter for bonus script command optional

secretdataz:
* **Server Mode**: Irrelevant
Made the second parameter for bonus script command optional. Would be nice to make some item bonuses that doesn't require any parameter cleaner.


Lemongrass3110:
@teededung make sure you have `BUILDIN_DEF(bonus,"i?"),` like this.

teededung:
@Lemongrass3110 Yes, I'm sure. I have recompiled 3 times

RadianFord:
I got the same issue with @teededung

Issue #2858, opened 2018-01-31
Items generated through skills with your inventory full (not weight) should not drop to the ground


Artuvazro:
* **rAthena Hash**:  31988abd1919a42e7ccedaf597fa13d4a7308865

* **Client Date**: not relevant.

* **Server Mode**: both.

* **Description of Issue**: 
  * Result: When you have 100 items and use a skill that creates a new item such as Holy Water, the created item drops to the ground. 
  * Expected Result: The item shouldn't be created at all. 
  * How to Reproduce: Use any skill that generates a new item. 
  * Official Information: Tested in aegis. 
  

Screenshot from aegis:
![imagen](https://user-images.githubusercontent.com/8776528/35634991-5ebadbd6-06ad-11e8-82b0-024d378c1e86.png)

* **Modifications that may affect results**:  none.

Lemongrass3110:
Just because #2856 was closed because of this:
Which AEGIS server were you trying on?

Artuvazro:
euRO. Will test in iRO when I have time and update this.

Artuvazro:
Aegis screenshot:
![unknown](https://user-images.githubusercontent.com/8776528/36273682-997589f4-1285-11e8-845e-a2e2e8167b19.png)

PR #2914, opened 2018-02-15
Fixed item creation with full inventory

Lemongrass3110:
* **Addressed Issue(s)**: #2858

* **Server Mode**: Both

Prevented items from being created if adding them to the inventory failed for whatever reason.
Introduced a new config that allows to keep the old behavior.


Daegaladh:
The battle_config.skill_drop_items_full check should go in the same if as the pc_additem, and you also forgot the "you can't have this item because you will exceed the weight limit." message (not sure if there is already a clif function for that :/ ).

PR #2915, opened 2018-02-16
Changing DBMap into STL::Map in C++ for guild storage

inhyositsu:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

### DIFF
--- a/src/map/storage.cpp
+++ b/src/map/storage.cpp
@@ -3,6 +3,8 @@
 
 #include "storage.hpp"
 
+#include <map>
+
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


src/map/storage.cpp
```diff
@@ -558,13 +532,18 @@ s_storage* create_guildstorage(int guild_id)
...
+	struct s_storage *gs;
 
-	if (guild_search(guild_id) == NULL)
-		return NULL;
+	if (guild_search(guild_id) == nullptr)
+		return nullptr;
 
-	if (gs == NULL)
-		gs = create_guildstorage(guild_id);
+	gs = guild2storage(guild_id);
```
inhyositsu: Is this correct here? A function call it by itself

PR #2916, opened 2018-02-16
Fixed athena-start newlines not set to unix type

Daegaladh:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

* **Description of Pull Request**: Set athena-start newlines to unix type (LF) as it should.


Daegaladh:
I was tired of using dos2unix everytime i move files from Windows to Linux. D:

PR #2929, opened 2018-02-23
Fixed no reiteration unit skills fail message back to USESKILL_FAIL_LEVEL

Daegaladh:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

* **Description of Pull Request**: Non-overlapping unit skills should show USESKILL_FAIL_LEVEL instead of USESKILL_FAIL_DUPLICATE_RANGEIN. Tested on both Aegis 15.2 and iRO.
This was changed by @cydh on feec60e8c853caeaccfb51745e72b9676dfe042e for some reason, maybe some skill does actually show that message?


Daegaladh:
I've tested the skills related to feec60e just to be sure, and the USESKILL_FAIL_DUPLICATE_RANGEIN never shows up.

PR #2931, opened 2018-02-24
Fixed Sheltering Bliss (CG_MOONLIT) max level

Daegaladh:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

* **Description of Pull Request**: It's max level is wrong since the first implementation of the skill, around 2006. Nobody noticed in 12 years lol.


mrjnumber1:
```SK_CG_MOONLIT::GetMaxLevel()
{
  return 5;
}```

i'm confused! this is how it is on pre-re at least..

Daegaladh:
Isn't that the same I'm saying? It should be 5 instead of 1.

mrjnumber1:
oh whoops, i misread the diff and thought you were changing the other way around. you're right!

PR #2943, opened 2018-02-28
Cleaned up the mapflag system

aleos89:
* **Addressed Issue(s)**: N/A

  * Created setter and getter functions.
  * Adjusted all calls to use these functions.
  * Cleaned up several functions to be more dynamic to reduce redundancy that was all over the place.
  * Renamed nosumstarmiracle to nosunmoonstarmiracle.
  * Adjusted skill_damage mapflag to use proper defined constants.
Thanks to @Lemongrass3110 for a lot of help!


mrjnumber1:
really like this change, ideally adding a new mapflag is as painless as possible and all these cases are handled dynamically though.

src/map/script.cpp
```diff
@@ -12292,331 +12256,118 @@ BUILDIN_FUNC(setmapflagnosave)
...
-	if (sd->pvp_timer == INVALID_TIMER) {
-		sd->pvp_timer = add_timer(gettick() + 200, pc_calc_pvprank_timer, sd->bl.id, 0);
-		sd->pvp_rank = 0;
-		sd->pvp_lastusers = 0;
-		sd->pvp_point = 5;
-		sd->pvp_won = 0;
-		sd->pvp_lost = 0;
+	mf = script_getnum(st, 3);
+
+	if( mf <= MF_MIN || mf > MF_MAX ){
```
secretdataz: Should be `mf < MF_MIN` or adjust `MF_MIN`'s value since it shares the same value with `MF_NOMEMO`.

src/map/map.cpp
```diff
@@ -4319,37 +4319,493 @@ void map_skill_damage_free(struct map_data *m) {
  * @param other Rate to Other target
  * @param caster Caster type
  **/
-void map_skill_damage_add(struct map_data *m, uint16 skill_id, int pc, int mob, int boss, int other, uint8 caster) {
+void map_skill_damage_add(struct map_data *m, uint16 skill_id, int rate[SKILLDMG_MAX], uint16 caster) {
```
secretdataz: Array size has no effect in this context. It will be converted into a pointer anyways.

src/map/script.cpp
secretdataz: Should be `mf < MF_MIN` or adjust `MF_MIN`'s value since it currently shares the same value with `MF_NOMEMO`, making the flag unavailable.

secretdataz: There are more occurrences of this but I forgot to comment ;w;

PR #2961, opened 2018-03-08
Costumes update

zackdreaver:
* **Addressed Issue(s)**: n/a

* **Server Mode**: Renewal

* **Description of Pull Request**: Updates official costumes up to February 2018

### DIFF
--- a/db/re/item_db.txt
+++ b/db/re/item_db.txt
@@ -11967,11 +11967,11 @@
 31233,C_Long_Twin_Purple,C Long Twin Yellow,4,0,,0,,0,,0,0xFFFFFFFF,63,2,4096,,1,0,1628,{},{},{}
 31234,C_Persica,C Persica,4,0,,0,,0,,0,0xFFFFFFFF,63,2,2048,,1,0,659,{},{},{}
 31235,C_Large_Ribbon_Muffler_Mid,C Large Ribbon Muffler Mid,4,0,,0,,0,,0,0xFFFFFFFF,63,2,2048,,1,0,1312,{},{},{}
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
I am still unsure about the view ids of 7 of these items:

- [ ] 20511
- [ ] 20512
- [ ] 31236
- [ ] 31240
- [ ] 31264
- [ ] 31311 (looks definitely wrong since bigger than 6k)
- [ ] 31323

Could you tell me how you got those?

zackdreaver:
20511 - From kRO itemInfo, should be correct
20512 - From jRO idnum2itemresnametable, the resname is same with ROBE_BAG_OF_ADVENTURER spriterobeid.lub
31236 - Should be 1267
31240 - Should be 321
31264 - From jRO idnum2itemresnametable, the only one that match with resnam is accessoryID 990
31311 - Yeah typo, that should be HG location, view id should be 666
31323 - Should be 484

Lemongrass3110:
Ok could you push a follow up and fix those that are wrong? :)

zackdreaver:
Should we use long version or short version for iRO names?
In this update i rather use C instead of Costume (eg: Costume Stripe Hat > C Stripe Hat)
though I can change it for the sake of scripts.

Lemongrass3110:
I think it is better we stick with C_ :)

Lemongrass3110:
Ah sorry misread that.
Well i'd go with how they'll probably have it. I've seen a few that have the full "Costume" as prefix now a few times already. Others have the "C" prefix. So I guess I'd just look how they add it and use that one.
Guess some of those items you added will possibly never get any iRO name anyway.

Issue #2866, opened 2018-02-03
renewal mob atk/matk is all wrong!


mrjnumber1:
* **rAthena Hash**: current
* **Server Mode**: Renewal ONLY
* **Description of Issue**: 
  * Result: Currently mob MATK and ATK is completely inaccurate values!
  * Official Information: 
having extracted more information from aegis zone, i've determined how off mob atk/matk has been for many years (seems hercules noticed as well, based on what they've done with mob/hom/merc matk).
now in aegis, mobs' atk1/atk2 parameters are in fact atk/matk base values. all of the values in our db need completely changed unfortunately..

said succinctly, it appears that mob MATK should be `blevel+int+70~130%atk2` ... right now it is using the same formula as players'
```
int CNPC::GetMagicAttPower()
{
  atk2 = m_characterInfo.atk2;
  minmatk = 7 * atk2 / 10;
  maxmatk = 13 * atk2 / 10;
  if ( !minmatk )
    minmatk = 1;
  if ( !maxmatk )
    maxmatk = 1;
  blevel = m_characterInfo.clevel;
  _int = m_characterInfo.Int;
  return blevel + _int + GetServerRandom(minmatk, maxmatk) + m_fvarshistory_ItemMATK_Amount.m_Value;
}
```

the MATK is harder to prove unfortunately.. it seems like merc matk is still `int_ + int_ / 5 * (int_ / 5)`, 
```
int CNPC::GetItemAttPower()
{
  int atk1 = m_characterInfo.atk1;
  int minbonus = 8 * atk1 / 10;
  int maxbonus = 12 * atk1 / 10;
  if ( !GetEffective(EFST_MAXIMIZE, 0) )
    minbonus = maxbonus;
  dmg = m_characterInfo.Str + m_characterInfo.clevel + GetServerRandom(minbonus, maxbonus);
  return ((dmg * m_fvarshistory_INC_ItemATK_Percent.m_Value) / 100.0
        + (dmg + m_fvarshistory_ItemATK_Amount.m_Value));
}
```
so atk is 80%-120% atk + lv+str

you can verify (the ATK values, at least...) by checking what gnjoy website lists.
http://ro.gnjoy.com/guide/runemidgarts/popup/monsterview.asp?monsterID=BAPHOMET
http://ro.gnjoy.com/guide/runemidgarts/popup/monsterview.asp?monsterID=BAPHOMET_

[...]

PR #2965, opened 2018-03-09
Mob MATK Update

syracuase:
* **Addressed Issue(s)**: [https://github.com/rathena/rathena/issues/2866](url)

* **Server Mode**: Renewal

* **Description of Pull Request**: In AEGIS atk2 is the variance. To get variance, i just minus atk2-atk1. For MATK values it is just 70-130% of atk2[variance]. Therefore in rathena it is 70-30%[atk2-atk1]. No need to revamp the whole mob_db as in my opinion, rAthena have a better atk1/atk2 system..

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -2581,16 +2581,14 @@ void status_calc_misc(struct block_list *bl, struct status_data *status, int lev
 	status->matk_min = status->matk_max = status_base_matk(bl, status, level);
 
 	///! TODO: Confirm these RENEWAL calculations. Currently is using previous calculation before 083cf5d9 (issue: #321) and until re/mob_db.txt is updated.
-	//switch (bl->type) {
-	//	case BL_MOB:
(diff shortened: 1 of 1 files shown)


mrjnumber1:
Strongly disagree with how this is done. Aside from the code not being set up properly.... It's nonsensical to store the mob matk in db as atk+matk (even though that'd make it not require a big commit)

It needs some cleaning up to adjust every issue.
It doesn't actually fix the issues with Mob MATK being wrong. It only (attempts to) give them proper bonus MATK.
Mob ATK is completely unaddressed as well!!

ecdarreola:
Tried this one but have a warning on console. 
image
Atleast I tried

syracuase:
@ecdarreola fixed.. was in a haste. 

@mrjnumber1  herc uses variance as atk2. this simply uses hercules' method without the need to convert all the mob's matk value.

slyx88:
May I know the type of attack of monster have? is it wATK or bATK?

aleos89:
Appreciate the effort in this, but we have an update for mob_db coming soon which will nullify this PR.

Issue #2087, opened 2017-04-23
power absorb vs force vanguard and skill oboro kagerou!


jamon18:
Server Mode: Renewal
Description of Issue:
force vanguard is affected with power absorb (bug)
fire /ice /wind /earth charm skill oboro kagerou dont affected with power absorb(bug)
http://irowiki.org/wiki/Power_Absorb
http://irowiki.org/wiki/Vanguard_Force

PR #2980, opened 2018-03-13
Corrects Spirit/Charm absorption behaviour on Ninja Spirit Charm

Chisee:
Fixes Sura Skill SR_ASSIMILATEPOWER not absorbing Spirit Charms.

* **Addressed Issue(s)**:  References part of  #2087 .

* **Server Mode**: Renewal

* **Description of Pull Request**: The conditional was missing a check on Kagero/Oboro's Spirit Charm, therefore not removing the charms nor applying SP regeneration on caster.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -9961,7 +9961,7 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case SR_ASSIMILATEPOWER:
 		if (flag&1) {
 			i = 0;
-			if (dstsd && dstsd->spiritball && (sd == dstsd || map_flag_vs(src->m)) && (dstsd->class_&MAPID_BASEMASK)!=MAPID_GUNSLINGER) {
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -9961,7 +9961,7 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case SR_ASSIMILATEPOWER:
 		if (flag&1) {
 			i = 0;
-			if (dstsd && dstsd->spiritball && (sd == dstsd || map_flag_vs(src->m)) && (dstsd->class_&MAPID_BASEMASK)!=MAPID_GUNSLINGER) {
+			if (dstsd && (dstsd->spiritball || dstsd->spiritcharm) && (sd == dstsd || map_flag_vs(src->m)) && (dstsd->class_&MAPID_BASEMASK)!=MAPID_GUNSLINGER) {
```
aleos89: Maybe it'd be better to just remove the `dstsd->spiritball && dstsd->spiritcharm` check here since we check again after this if-statement succeeds.

Chisee: Indeed! Took the oportunity to do the same on MO_ABSORBSPIRITS as the absorption of ninja charms was also unreachable.

PR #2982, opened 2018-03-14
 Added multi-hit critical damage display support

secretdataz:
* **Addressed Issue(s)**: #1788 

* **Server Mode**: Renewal (skill part), Both (clif part)
    * Double Attack and Fear Breeze can now critically strike.
    * Critical damage from those skills will now displayed as critical damage properly (Client date 20171207 or later).
    * The code is hacky and need some tests.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -4756,6 +4756,7 @@ static int clif_hallucination_damage()
 ///     10 = critical hit
 ///     11 = lucky dodge
 ///     12 = (touch skill?)
+///     13 = multi-hit critical
 int clif_damage(struct block_list* src, struct block_list* dst, unsigned int tick, int sdelay, int ddelay, int64 sdamage, int div, enum e_damage_type type, int64 sdamage2, bool spdamage)
 {
 	unsigned char buf[34];
@@ -4776,7 +4777,8 @@ int clif_damage(struct block_list* src, struct block_list* dst, unsigned int tic
 	nullpo_ret(src);
 	nullpo_ret(dst);
 
-	type = clif_calc_delay(type,div,damage+damage2,ddelay);
+	if (type != DMG_MULTI_HIT_CRITICAL)
+		type = clif_calc_delay(type,div,damage+damage2,ddelay);
 	sc = status_get_sc(dst);
 	if(sc && sc->count) {
 		if(sc->data[SC_HALLUCINATION]) {
@@ -4838,6 +4840,9 @@ int clif_damage(struct block_list* src, struct block_list* dst, unsigned int tic
 	if(src == dst) {
 		unit_setdir(src, unit_getdir(src));
 	}
+
+	// In case this assignment is bypassed by DMG_MULTI_HIT_CRITICAL
+	type = clif_calc_delay(type, div, damage + damage2, ddelay);
 	//Return adjusted can't walk delay for further processing.
 	return clif_calc_walkdelay(dst, ddelay, type, damage+damage2, div);
 }
--- a/db/pre-re/skill_db.txt
+++ b/db/pre-re/skill_db.txt
@@ -16,6 +16,7 @@
 //    0x20 - Skill ignores target's defense (misc type always ignores)
 //    0x40 - Skill ignores target's flee (magic type always ignores)
 //    0x80 - Skill ignores target's def cards
+//    0x100 - Skill can critical
 // 07 splash/effect range
 //      -1 - for screen-wide
 //       0 - no splash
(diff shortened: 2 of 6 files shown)


zackdreaver:
iirc, the client date should be around dec 2016, not 2017.

secretdataz:
@zackdreaver Oh, yeah. I got the year off by one year. Thanks!

BrOgBr:
Looks works xD

**Critical:**
zzzz

**Normal:**
zzzz2

RadianFord:
Tested this using Pre-renewal. and works perfect.

syracuase:
using your code, auto-counter does not display the crit animation anymore. several changes need to be put. maybe add:

if(skill_id == KN_AUTOCOUNTER)
return true;

at the is_attack_critical line? im not sure if it's just me though

iubantot:
im testing this PR, and i tried using snake hat + fear breeze result is the fear breeze is not effect is i think is canceled cause its only double attacks showing.

talesofragnarok:
i try this but as i notice critical working 100% but the double attack are gone

Badarosk0:
It seems that if you use double critical the fear of Ranger stops working. Does it?

idamonli:
@BrOgBr 
What skills are used?

idamonli:
Active attack skills don't seem to support them.

secretdataz:
According to my test results on kRO, triple attack’s change is likely to be reverted or not implemented in the first place.

BrOgBr:
@idamonli I didn't understand you. oO
In the SS, I was using **Double Attack** and it's working.

erenaryana:
excuse me. sorry but where do i get a client date 2017 as mentioned ? i wanted to test this on the offline rathena

cydh:
in skill.hpp, member of `struct s_skill_db`
please change the `uint8 nk;` to `uint16 nk;` at least.

secretdataz:
@cydh Good catch. Thanks

mazvi:
How about this update?

Atemo:
See the to-do list above ^

jezznar:
Tested this PR.
- [ ] Test interaction with endure'd enemies
- [ ] Test interaction with MVPs
extra:
- Test interaction in PVP: Doesn't work

src/map/clif.cpp
```diff
@@ -4847,6 +4849,9 @@ int clif_damage(struct block_list* src, struct block_list* dst, unsigned int tic
 	if(src == dst) {
 		unit_setdir(src, unit_getdir(src));
 	}
+
+	// In case this assignment is bypassed by DMG_MULTI_HIT_CRITICAL
+	type = clif_calc_delay(type, div, damage + damage2, ddelay);
```
Lemongrass3110: Shouldnt this be inside an if to avoid multiple calculations for ither types?

BrOgBr:
Against MVPs I've already tested it and confirmed that it was working xD

Atemo:
Yeha the system displays the crit on mvp in this PR. But previously crit + high aspd could 'hit lock' the mob, in vulgar term. The question is: the multi hit crit can 'hit lock' a mvp or a mob under endure on official? The answer will probably be subjective

redlightliu:
Update those source will let Fist Spell(SO_SPELLFIST) can crit. 
Fist Spell should not crit.

jezznar:
@secretdataz what do you mean about positional desync? I'm not quite sure about what you mean but now that you mentioned something about positions, I still get to attack a monster when it pushes me off. somehow, the attack became ranged? here's a video. https://www.youtube.com/watch?v=SgaFJAupveI

nutstartza:
god

Questune09:
is it normal that when you use a katar the critical wont show and if you use a dagger critical always showing?

luan122:
> is it normal that when you use a katar the critical wont show and if you use a dagger critical always showing?

is the damage different when you do critical and when you don't with katar?

darhylism:
is this good to go?

laziem:
crt
I just test this PR. Normal attack now always become critical.

Miraakol:
maybe this one can help https://github.com/idathena/trunk/commit/49f428a635a98189e17c8de43e94bc807d97c1f7#diff-c861cb68eadb0408c3c20bb633904e69

aleffmoura:
Its not working?

secretdataz:
Make sure you are using exe, hexed, or whatever you call your client dated 2016-12-07 or later.
I will not help you if you don't state that you are using client version 2016-12-07 or newer.

Issue #2968, opened 2018-03-10
About KG_KAGEMUSYA skill problem


redlightliu:
* **rAthena Hash**: bc02c4e

* **Client Date**: 20160203

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: When I use skill KG_KAGEMUSYA Lv5 to my charater.I get skill TF_DOUBLE Lv2.
  * Expected Result: This skill should be get TF_DOUBLE Lv2 -> Lv4 -> Lv6 -> Lv8 -> Lv10 by KG_KAGEMUSYA Lv1 -> Lv2 -> Lv3 -> Lv4 -> Lv5.
  * How to Reproduce: Use skill KG_KAGEMUSYA in you self.

  
http://www.inven.co.kr/board/ro/1939/17871
https://www.youtube.com/watch?time_continue=28&v=LKcXm_p8rCc

admkakaroto:
Where did you read that the skill level duble goes up according to the skill level KAGEMUSYA?

```
	[SKID.KG_KAGEMUSYA] = {
		"Shadow Warrior",
		"Max Level: 5",
		"^CC3399Requirement: Empty Shadow 3^000000",
		"Description: ^777777Imbues Double Attack skill",
		"to a target for a period of time.",
		"Consumes 1 SP per second while active.",
		"Consumes 1 Shadow Orb.^000000",
		"[LV 1]: ^777777Duration 60 seconds^000000",
		"[LV 2]: ^777777Duration 90 seconds^000000",
		"[LV 3]: ^777777Duration 120 seconds^000000",
		"[LV 4]: ^777777Duration 150 seconds^000000",
		"[LV 5]: ^777777Duration 180 seconds^000000"
},
```

https://irowiki.org/wiki/Shadow_Warrior

redlightliu:
http://www.inven.co.kr/board/ro/1939/17871

redlightliu:
And this film show the skill RA_FEARBREEZE and TF_DOUBLE can use together.
But rAthena only trigger the skill TF_DOUBLE. It will not trigger RA_FEARBREEZE.

PR #2988, opened 2018-03-15
Corrected Shadow Warrior double attack rate

aleos89:
* **Addressed Issue(s)**: #2968

* **Server Mode**: Renewal

  * The rate of double attack should match the rate of Thief's Double Attack even levels.
  * Confirmed that the double attack bonus does not work when weapons are not equipped.
Thanks to @redlightliu!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3322,12 +3322,16 @@ static struct Damage battle_calc_multi_attack(struct Damage wd, struct block_lis
 	if( sd && !skill_id ) {	// if no skill_id passed, check for double attack [helvetica]
 		short i;
 		if( ( ( skill_lv = pc_checkskill(sd,TF_DOUBLE) ) > 0 && sd->weapontype1 == W_DAGGER )
-			|| ( sd->bonus.double_rate > 0 && sd->weapontype1 != W_FIST ) //Will fail bare-handed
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -3322,12 +3322,16 @@ static struct Damage battle_calc_multi_attack(struct Damage wd, struct block_lis
...
+			|| ( sd->bonus.double_rate > 0 && sd->weapontype1 != W_FIST ) // Will fail bare-handed
+			|| ( sc && sc->data[SC_KAGEMUSYA] && sd->weapontype1 != W_FIST )) // Will fail bare-handed
 		{	//Success chance is not added, the higher one is used [Skotlex]
-                        int max_rate = max(5*skill_lv,sd->bonus.double_rate);
-                        if(sc && sc->data[SC_KAGEMUSYA]) max_rate= max(max_rate,sc->data[SC_KAGEMUSYA]->val1*3);
-                        
+			int max_rate = 0;
+
+			if (sc && sc->data[SC_KAGEMUSYA])
+				max_rate = sc->data[SC_KAGEMUSYA]->val1 * 10; // Same rate as even levels of TF_DOUBLE
```
Lemongrass3110: this looks like the comment is actually for the else case

PR #2990, opened 2018-03-16
Fixed turbo track fee and min level in solo mode

Daegaladh:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

* **Description of Pull Request**: Fixed fee and level requirements in turbo track solo mode.


aleos89:
```
  SetFeeZeny 1000
  SetReqLevel 10 99
```
The only difference I see is the level 99 cap. I'm not sure of the state of Turbo Track after renewal/3rd classes.

Daegaladh:
Yeah i managed to test it of official servers and it works for 99+ characters

Issue #814, opened 2015-12-06
kRO Max HP Limit


Jedzkie:
Hi! in kRO, the max HP limit is set by specific level (Renewal only), reference:

https://github.com/HerculesWS/Hercules/issues/243

cydh:
I saw there is kRO's update/mt notice that say so.

Akkarinage:
Does anyone have any links to info on this?

Atemo:
http://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=6322&curpage=51

PR #3003, opened 2018-03-23
Apply MHP limit based on bLv

flamefury:
* **Addressed Issue(s)**: #814
* **Server Mode**: Renewal
Changes max HP configuration into 3 values based on the level limits as defined in the kRO news post:
http://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=6322

With defaults matching the values there.


Lemongrass3110:
You forgot to commit battle.hpp

RadianFord:
Will this affect a server with a max level 500?

flamefury:
@RadianFord yes, but you can just modify the Lv175 limit to whatever you want it to be

Atemo:
Some indentations are still broken in pc and status.cpp (space instead of tab)

PR #3005, opened 2018-03-24
Iceymoon patch 2

iceymoon:
The first two quests after arriving in Izlude


iceymoon:
This was for a fork, please disregard.

PR #3007, opened 2018-03-26
Added support for homunculus autofeeding

Lemongrass3110:

* **Server Mode**: Both

This feature is available in 2017-09-20bRagexeRE or later

Credits to @Asheraf in HerculesWS/Hercules#1898

Fixed a missing entry in clif_shuffle for 2017-09-13


Lemongrass3110:
Sending out after call is not necessary, since the setting is sent on login even when the homunculus is vaporized.

PR #3008, opened 2018-03-27
Added a config for disabling novice/summoner character creation

secretdataz:

* **Server Mode**: Both

    * Added a configuration in char_athena.conf for disabling novice/doram creation.
    * TODO: Needs to separate renewal/pre-renewal logic


secretdataz:
@zackdreaver Do you know if Summoner's special trait of being brute instead of demi-human is removed or not?

zackdreaver:
From what I've read from some threads, warpportal forum and other site such inven.co.kr and ro-blog.livedoor.biz
there is no change to Doram, they're still a Brute race in Small size.

zackdreaver:
additional note: RC_Player works on them.

Issue #2972, opened 2018-03-11
Shadow Chaser skill rebalance: Auto Shadow Spell and Reproduce


zackdreaver:
* **rAthena Hash**: 40d9d5ab7399c75b21e2038e728044620cc4d42d
* **Client Date**: n/a
* **Server Mode**: Renewal

According : http://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=6836&curpage=1
Shadow Chaser's Auto Shadow Spell:
- Will be activated even if the physical attack misses
- Level of the spell cast by Auto Shadow Spell increased
from 1/1/2/2/3/3/4/4/5/5
to 3/3/4/4/5/5/6/6/7/7
- The effects of Auto Shadow spells are also applied to ranged physical attacks.

Shadow Chaser's Reproduce:
- Reduces SP penalty of using skills from 100% to 30% of SP

Stalker's Preserve:
- Preserve status will not be canceled when changing or moving to spesific maps.

PR #3014, opened 2018-03-28
Updated Shadow Chaser skills

aleos89:
* **Addressed Issue(s)**: #2972

  * Updated Auto Shadow Spell:
    * Will now trigger even on ranged physical attacks.
    * Increased spell cast level to 3/3/4/4/5/5/6/6/7/7.
  * Updated Reproduce:
    * SP penalty for casting spells is now 30% extra.
Thanks to @zackdreaver!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -7450,7 +7450,7 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
 	}
 	if (sd) {
 		uint16 r_skill = 0, sk_idx = 0;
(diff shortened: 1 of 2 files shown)


zackdreaver:
Just asking,
I saw that you add new line for reproduced skill's sp cost
``require.sp += require.sp * 30 / 100;``

Does that means, before this patch is applied,
the reproduced skill's sp cost is normal? without penalty?

aleos89:
Yeah, based on the changelog it would seem the extra +100% increase in SP usage was missing.

Issue #3024, opened 2018-04-02
update my battleground script templates


AnnieRuru:
.... I just remove all the rubbish information then ...

I noticed rAthena now has its own updated battleground script commands
this also means my battleground script templates needs an update

I am lazy to make rAthena fork, so I just update this in paste
http://upaste.me/45a9495597107216b
http://upaste.me/712949560abb0683d

remember to review 1st before update it

AnnieRuru:
**[type:bug]** ?
nonono, the old version is also working fine actually,
or else Euphy wont get this script inside the emulator

it should be **[type:enhancement]**

PR #3025, opened 2018-04-02
Cleaned up custom battleground scripts

secretdataz:
* **Addressed Issue(s)**: #3024 

* **Server Mode**: Both

Cleaned up AnnieRuru's BG scripts.

### DIFF
--- a/npc/custom/battleground/bg_emp.txt
+++ b/npc/custom/battleground/bg_emp.txt
@@ -3,7 +3,7 @@
 //===== By: ==================================================
 //= AnnieRuru
 //===== Current Version: =====================================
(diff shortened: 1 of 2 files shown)


Atemo:
Anything left to be done?

AnnieRuru:
I thought I look through the script over 5 times before I press the open issue button
yet the delwall thing still can slip through my eyes ~ give you a rep

now this is interesting .... delwall should throw an error if the wall has never been set
I guess this is another bug ?

`void map_iwall_remove(const char *wall_name)`
hmm ... maybe should change to `bool map_iwall_remove` and return true/false

PR #3029, opened 2018-04-02
Updated getunitdata / setunitdata

Atemo:

* **Server Mode**: both

Added the possibility to retrieve / change the current target id of a mob through getunitdata / setunitdata

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -7632,6 +7632,7 @@ Parameters (indexes) for monsters are:
 	UMOB_AMOTION
 	UMOB_ADELAY
 	UMOB_DMOTION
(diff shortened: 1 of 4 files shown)


AnnieRuru:
can I know what kind of situation require changing mob GID ?
I never encounter this kind of situation before ...

Lemongrass3110:
@AnnieRuru this is changing the target GID not the GID of the mob itself.
This means that it will attack the target with GID after this call.

AnnieRuru:
ah, thanks for the clarification
it seems rathena community really do use mob controller stuffs and has exceeded my knowledge on this subject
keep this up :heart:

Issue #3064, opened 2018-04-22
suggestion std::find command script


sader1992:
* **rAthena Hash**: a16fe436e34c52a981c6d1c0898ac605ae5af89d

* **Client Date**: 2015-11-04

* **Server Mode**: both

* **Description of suggestion**: 
so i was thinking if we can make script command for that in rathena
since most of the scripts need for loop for array comparison
Example:
```
===================================================
setarray .@myarray,2,50,15,17;
if(arraysearch(.@myarray ,5)){
	//false
}

if(arraysearch(.@myarray ,15)){
	//true
}

===================================================

setarray .@myarray$,"test1","test2","testtttt","testttt1";
if(arraysearch(.@myarray$ ,"test")){
	//false
}

if(arraysearch(.@myarray$ ,"testttt1")){
	//true
}
===================================================
```

so if the element is found the command script return true
else return false

(also auto freeloop it if the array is big [i don't know if this is necessary since it's in the src])

i know std::find for vectors but well it's just an idea

`( i did try to make it but no success xD )`

it would take time until all the scripters use it but it would be nice command script

Edit:
thinking about it more
it's better to be like this if we will loop inside the src
```
searcharray(<array>,<"value1">{,<"value2">}[etc up for some number])
return 0 if non is true
return 1 if all true
return 2 if one is true
```
this would be a shortcut for 
```
if(value == value1 || value == value2)
and
if(value == value1 && value == value2)
```
also this would help to not over use the command like
```
if(searcharray(.@array,"value") ||{&&} searcharray(.@array,"value2")
```
maybe this would be better

secretdataz:
Since we don't know the precondition of the data we are going to search, the best we could do is O(n) (linear search) plus the cost of string comparison if the array is a string array. It could be more efficient than implementing linear search in the script tho. I'll consider implementing it in my free time.
(or you can open a PR, linear search is easy)

Lemongrass3110:
Implemented in 151c847

PR #3071, opened 2018-04-24
inarray script commands

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/3064

* **Server Mode**: both

this script command is shortcut to array comparison
the original idea was checking if true or false
but i did change the command to return the number of the true cases instead

well it's just a suggestion
also i don't know if there is better way to write it
i am all up for any suggestion , optimization and name changes
(i am aware that it may or may not be accepted for rathena and it's ok for me either ways)

this command would be useful for large scripts more than the smaller ones
but here is example for how the script command should work

**Example 1:**
```
	setarray .@array,1,2,3,4,5,6;
	mes "is the number 3 inside .@array array ?";
	///both options will do the same thing
	switch(select("without the command:with the command")){
		clear;
		
		case 1:
			for(.@i=0;.@i<getarraysize(.@array);.@i++){
				if(.@array[.@i] == 3){
					mes "3 is an element of the array .@array";
					close;
				}
			}
			mes "3 is not an element of the array .@array";
				close;
			break;
			
		case 2:
			if(searcharray(.@array,3)){
				mes "3 is an element of the array .@array";
				close;
			}else{
				mes "3 is not an element of the array .@array";
				close;
			}
	end;
	}
```

**Example 2:**
```
	setarray .@array,1,2,3,4,5,6;
	mes "is the numbers 3 and 6 inside .@array array ?";
	///both options will do the same thing
	switch(select("without the command:with the command")){
		clear;
		
		case 1:
			for(.@i=0;.@i<getarraysize(.@array);.@i++){
				if(.@array[.@i] == 3 ||.@array[.@i] == 6){
					.@count ++;	//there is other ways to do that but for the sake of argument
				}
			}
			if(.@count == 2){
				mes "true";
				close;
			}
			mes "false";
				close;
			break;
			
		case 2:
			if(searcharray(.@array,3,6) == 2){
				mes "true";
				close;
			}else{
				mes "false";
				close;
			}
	end;
	}
```
**Example 3:**
```
	setarray .@array,1,2,3,4,5,6;
	mes "is one of the numbers 3 or 6 inside .@array array ?";
	///both options will do the same thing
	switch(select("without the command:with the command")){
		clear;
		
		case 1:
			for(.@i=0;.@i<getarraysize(.@array);.@i++){
				if(.@array[.@i] == 3 ||.@array[.@i] == 6){
					mes "true";
					close;
				}
			}
			mes "false";
				close;
			break;
			
		case 2:
			if(searcharray(.@array,3,6) == 1){
				mes "true";
				close;
			}else{
				mes "false";
				close;
			}
	end;
	}
```


anacondaqq:
Very simple, not effective, but how it can be effective on an unsorted array? Thank you for your efforts!

sader1992:
@anacondaqq it work with int or string arrays no sort is required ,the array i write just for example but it would work either ways  >>  example `setarray .@array,3,2,6,5,4,1;`

my plan was making it more efficient
but while i am writing it in the end i did feel it wont be a gain much else than script side
and while every large script i loop checking for the arrays , so i felt like it's better to push it and take more opinions and see if it would be good for rathena or not

either ways i think it would be eventually useful for scripts since (i didn't think a lot about it but >>) i think we can't make a script function (in the global functions) to return the same result without losing preference (except some like global variable) coz we can't really pass an array with `callfunc`

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+{
+	struct script_data *data;
+	const char* name;
+	int id , i, j, array_size , case_count = 0;
+	struct map_session_data* sd = NULL;
+	data = script_getdata(st, 2);
+	
+	if (!data_isreference(data))
+	{
+		ShowError("script:searcharray: not a variable\n");
```
Lemongrass3110: buildin_searcharray:

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+		script_reportdata(data);
+		script_pushnil(st);
```
Lemongrass3110: You do not need a pushnil if you end the script in the next line anyway.

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	id = reference_getid(data);
+	name = reference_getname(data);
+
+	if (not_server_variable(*name)) {
+		if (!script_rid2sd(sd))
```
Lemongrass3110: Merge those two ifs together please and change the return code to SCRIPT_CMD_FAILURE.

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+
+	if (array_size > SCRIPT_MAX_ARRAYSIZE)
+	{
+		ShowError("script:searcharray: The array is to large.\n");
+		script_reportdata(data);
+		script_pushnil(st);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
```
Lemongrass3110: Move the id initialization here:
`id = reference_getid(data);`

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+			return SCRIPT_CMD_SUCCESS;
+	}
+
+	array_size = script_array_highest_key(st, sd, name, reference_getref(data)) - 1;
+
+	if (array_size > SCRIPT_MAX_ARRAYSIZE)
+	{
+		ShowError("script:searcharray: The array is to large.\n");
+		script_reportdata(data);
+		script_pushnil(st);
```
Lemongrass3110: no pushnil required again.

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+	if (not_server_variable(*name)) {
+		if (!script_rid2sd(sd))
+			return SCRIPT_CMD_SUCCESS;
+	}
+
+	array_size = script_array_highest_key(st, sd, name, reference_getref(data)) - 1;
+
+	if (array_size > SCRIPT_MAX_ARRAYSIZE)
+	{
+		ShowError("script:searcharray: The array is to large.\n");
```
Lemongrass3110: buildin_searcharray again

hurtsky:
@anacondaqq in my opinion, this better and simpler, rather than doing loops check in script.

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+	}
+
+	if (is_string_variable(name)) 
+	{
+		const char* temp;
+		const char* value_list;
+		for (i = 0; i <= array_size; ++i) {
+			temp = (char*)get_val2(st, reference_uid(id, i), reference_getref(data));
+			j = 3;
+			while (script_hasdata(st, j))
```
Lemongrass3110: `for( j = 3; script_hasdata(st, j); j++ )`
Applies to both cases of the if/else.

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (is_string_variable(name)) 
+	{
+		const char* temp;
+		const char* value_list;
+		for (i = 0; i <= array_size; ++i) {
+			temp = (char*)get_val2(st, reference_uid(id, i), reference_getref(data));
```
Lemongrass3110: You could store `reference_getref(data)` into a variable as well. Will be a few ms faster again.

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+				{
+					case_count++;
+				}
+				j++;
+			}
+			script_removetop(st, -1, 0);
+		}
+	}
+	
+	push_val(st->stack, C_INT, case_count);
```
Lemongrass3110: Why not pushint?

src/map/script.cpp
```diff
@@ -6659,6 +6659,88 @@ BUILDIN_FUNC(getelementofarray)
...
+				j++;
+			}
+			script_removetop(st, -1, 0);
+		}
+	}
+	else 
+	{
+		int temp , value_list;
+		for (i = 0; i <= array_size; ++i) {
+			temp = (int)get_val2(st, reference_uid(id, i), reference_getref(data));
```
Lemongrass3110: Build is failung on linux because of this here.
You need to use `(int32)__64BPRTSIZE()`.
Just look for this macro in this file and you will find some samples.

sader1992:
@Lemongrass3110 maybe returning the index of the array be a better idea
or if it's ok we can make a type for usage
type 1 return the true matches and type 2 return the index of first value passed via the command script
this would be very helpful
or what you think ?

Jeybla:
I like that command; nice work sader.

In most cases it is important to know if any equal item is in the array, (e.g. useful for white-/blacklisting). But it would be really great to have both options, like you said: `type 1 return the true matches and type 0 return the index of first value`.

Maybe it'd be an good idea to split this into three commands?
- inarray (searches for any equal item in the array)
- searcharray (returns the index of the first equal item)
- countinarray (returns the count of equal items in an array)

Whereby the first two options could cancel the execution after the first hit to save calculation time.

Anyway this is imo one of the most important (and missed) script commands.

secretdataz:
@sader1992 Might be a better idea to define 2 commands then.
`searcharray(<array>,...);` and `countiteminarray(<array>,...);`
It could have a better name tho.

Lemongrass3110:
I could imagine that for strings it might be interesting to have an option to make it case insensitive as well.
But I agree with secret that splitting it might be cleaner.

doc/script_commands.txt
```diff
@@ -2260,6 +2260,31 @@ array, shifting all the elements beyond this towards the beginning.
 
 ---------------------------------------
 
+*searcharray <array name>[<first value>],<value>{,<value>...<value>};
+
+This command will check for the match cases between the array elements and
+the values, and return the number of matches.
+
+    setarray @array[0], 100, 200, 300, 400, 500, 600;
```
Lemongrass3110: Spaces instead of tab here.

src/map/script.cpp
```diff
@@ -6659,6 +6659,84 @@ BUILDIN_FUNC(getelementofarray)
...
+		script_reportdata(data);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	name = reference_getname(data);
+	ref = reference_getref(data);
+
+	if (not_server_variable(*name) && !script_rid2sd(sd))
+			return SCRIPT_CMD_FAILURE;
```
Lemongrass3110: Too many tabs.

aleos89:
Sorry, I forgot to lead up with Lemongrass' comment. If others agree, I think what Jeybla and Secret suggested is a good name. This script command should be called something like: `countinarray`.

Lemongrass3110:
@AnnieRuru it is neither of both. It does not return the index or check it, but actually chrcks the value of the array at each index. Also it does not search a value and return its index. I mistook it for this as well.
This is why I suggested using a different name.

anacondaqq:
xD I have started to write a comment with suggestions, but then read @Jeybla comment, and looks like we have some a man who reads other people's thoughts from very long distance here 🎩 , and I totally agree and support his comment about renaming commands to like he said, really will be much better to read scripts without knowing what exactly do command and touching documentation.

searcharray() is worst fi compare to `inarray` or `isvalueinarray` or `isinarray`.

doc/script_commands.txt
```diff
@@ -2260,6 +2260,31 @@ array, shifting all the elements beyond this towards the beginning.
 
 ---------------------------------------
 
+*countinarray <array name>{[<start index>]},<value1>{,<value2>...<valueN>};
+
+This command will check for matches between the array values and return the number of matches.
+While being optional, if [<start index>] is supplied, the search will begin from the given index value.
```
sader1992: i still need to receive the index and make `i` start from it

also i am thinking of making the other array command we talk about here in one PR
`countinarray` the command here
`inarray` accept 1 arg and return the index of the value while true and return -1 while false

src/map/script.cpp
```diff
@@ -6659,6 +6659,160 @@ BUILDIN_FUNC(getelementofarray)
...
+			if (temp == value)
+			{
+				script_pushint(st, i);
+				return SCRIPT_CMD_SUCCESS;
+			}
+			
+		}
+	}
+
+	script_pushint(st, -1);
```
sader1992: i don't know about the return -1 i feel like it must be something else
but while 0 is valid index i don't know what else we can return else than -1

secretdataz: It's a common thing to return -1 when no match is found.

doc/script_commands.txt
```diff
@@ -2260,10 +2260,35 @@ array, shifting all the elements beyond this towards the beginning.
...
+
+For more details, see the sample in 'doc/sample/inarray.txt'.
+
+---------------------------------------
+
+*countinarray <array name>,<value1>{,<value2>...<valueN>};
 
 This command will check for matches between the array values and return the number of matches.
-While being optional, if [<start index>] is supplied, the search will begin from the given index value.
+While being optional.
```
aleos89: Was this on purpose? You can check out `cleararray` which supports the index value to add to your script command.

src/map/script.cpp
```diff
@@ -6659,6 +6659,169 @@ BUILDIN_FUNC(getelementofarray)
...
+		ShowError("buildin_countinarray: The array is too large.\n");
+		script_reportdata(data);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	i = reference_getindex(data);
+	if (array_size < i)
+	{	//To prevent unintended behavior
+		ShowError("buildin_countinarray: The given index of the array is higher than the array size.\n");
```
sader1992: or maybe 
`ShowError("buildin_countinarray: index out of range.\n");`
or
`script_pushint(st, 0);` since this would be true

Lemongrass3110: Personally I would prefer an error message instead of silently ignoring the issue.
This way people will notice if they do something wrong and maybe if they read their map-server output(I know this does not happen very often) they will read their scripts again and check for the problem themselves instead of making another spoonfeed request on the forum or discord.

doc/script_commands.txt
```diff
@@ -2260,6 +2260,56 @@ array, shifting all the elements beyond this towards the beginning.
...
+	//return 0 because @array[0] == 100
+	
+	inarray(@array, 800);
+	//return -1 because 800 is not an element of the array @array
+
+For more details, see the sample in 'doc/sample/inarray.txt'.
+
+---------------------------------------
+
+*countinarray <array name>{[<start index>]},<value1>{,<value2>...<valueN>};
```
Atemo: What about to return the number of matches of 2 arrays for dynamical script instead of `<value1>{,<value2>...<valueN>}`?

doc/script_commands.txt
```diff
@@ -2260,6 +2260,56 @@ array, shifting all the elements beyond this towards the beginning.
...
+
+This command will check for matches between the array values and return the number of matches.
+While being optional, if [<start index>] is supplied, the search will begin from the given index value.
+
+	setarray @array[0], 100, 200, 300, 400, 500, 600;
+	// So we have made @array[]
```
Atemo: Use `.@var` and update the comment

doc/script_commands.txt
sader1992: thanks , i will .

sader1992:
`countinarray` command now accept 2 arrays instead of 1 array and value list as @Atemo suggested

src/map/script.cpp
```diff
@@ -6659,6 +6659,189 @@ BUILDIN_FUNC(getelementofarray)
...
+	if (is_string_variable(name))
+	{
+		const char* temp;
+		const char* value;
+		value = script_getstr(st, 3);
+		for (i = 0; i <= array_size; ++i)
+		{
+			temp = (char*)get_val2(st, reference_uid(id, i), ref);
+			script_removetop(st, -1, 0);
+			if (!strcmp(temp, value))
```
anacondaqq: temp never will be set.
Because at line above
```
script_removetop(st, -1, 0);
```

you're already removing data from stack.
You can try to test on any string array, and will see results.

Change to:
```cpp
	if (is_string_variable(name))
	{
		const char* temp;
		const char* value;
		value = script_getstr(st, 3);
		for (i = 0; i <= array_size; ++i)
		{
			temp = (char*)get_val2(st, reference_uid(id, i), ref);
			if (!strcmp(temp, value))
			{
				script_removetop(st, -1, 0);
				script_pushint(st, i);
				return SCRIPT_CMD_SUCCESS;
			}
			script_removetop(st, -1, 0);
			
		}
	}
```

the same do for interger, and for countinarray, you can test, if do not trust :D

sader1992: isn't doing this would remove only the last one from the stack ?

so i would need to change the -1 to what ever the arraysize when the for loop break 

i am not sure but i thing this code may send all the values in the array to the script
if you try to `mes` it it will write all the array elements -1
correct me if i am wrong please

sader1992:
i am wondering if there is something wrong in this comment or it's just on hold for now
(i may be missed something and you are waiting me to fix it xD ?)

doc/script_commands.txt
```diff
@@ -2260,6 +2260,62 @@ array, shifting all the elements beyond this towards the beginning.
 
 ---------------------------------------
 
+*inarray <array name>,<value>;
+
+This command return the index of the first value found in the array
+It will return -1 if the value not found.
```
aleos89: `This command returns the index of the first matching value found in the array.`
`It will return -1 if the value is not found.`

doc/script_commands.txt
```diff
@@ -2260,6 +2260,62 @@ array, shifting all the elements beyond this towards the beginning.
...
+
+	setarray .@array[0], 100, 200, 300, 400, 500, 600, 100;
+	
+	inarray(.@array[0], 200);
+	//return 1 because the number 200 is the index 1 for the array
```
aleos89: `//return 1 because 200 is in index 1`

doc/script_commands.txt
```diff
@@ -2260,6 +2260,62 @@ array, shifting all the elements beyond this towards the beginning.
...
+	//another way to say it that .@array[1] == 200
+	
+	.@index = inarray(.@array[0], 600);
+	//.@index is now 5 because .@array[5] == 600
+	
+	inarray(.@array[0], 100);
+	//while .@array[6] == 100 but the command will return the first index it found.
```
aleos89: `//while index 6 is also 100, the command will return the first instance it finds`

src/map/script.cpp
```diff
@@ -6659,6 +6659,189 @@ BUILDIN_FUNC(getelementofarray)
 	return SCRIPT_CMD_SUCCESS;
 }
 
+/// Return the index number of the value in an array.
```
aleos89: `/// Return the index number of the first matching value in an array.`

doc/sample/inarray.txt
```diff
@@ -0,0 +1,115 @@
...
+//===== Description: ========================================= 
+//= Demonstrates the 'inarray' and 'countinarray' commands.
+//============================================================
+
+prontera,160,190,3	script	inarray sample	847,{
+	switch(select("inarray:countinarray")){
+		case 1:	//inarray command test
+			mes "[inarray Test]";
+			setarray .@array1[0],100,200,300,400,500,600,700;
+			setarray .@array2$[0],"One Hundred","Two Hundreds","Three Hundreds","Four Hundreds","Five Hundreds","Six Hundreds","Seven Hundreds";
```
aleos89: Remove the `s` after hundred.

doc/sample/inarray.txt
```diff
@@ -0,0 +1,115 @@
...
+			
+			mes .@array2$[inarray(.@array1,100)];	//return One Hundred
+			mes .@array2$[inarray(.@array1,300)];	//return Three Hundreds
```
aleos89: Remove the `s` after hundred.

doc/sample/inarray.txt
```diff
@@ -0,0 +1,115 @@
...
+			
+			//mes .@array2$[inarray(.@array1,800)];	//this give error
```
aleos89: `//this will return with an error`

src/map/script.cpp
Lemongrass3110: @anacondaqq to clear things up:
By using get_val2 the variable name is pushed onto the top of the stack before the value is resolved.
The actual value behind the variable stays untouched.
By removing the top stack entry after the call all you do is removing the variable name again. So it does not matter when you do it - you just have to do it.

PR #3099, opened 2018-05-03
Instance name command

Atemo:

* **Server Mode**: both

Just a suggestion to retrieve the db name of instance ID after the implementation of
`instance_info("<instance name>",<info type>{,<instance_db map index>});`

It could be a way to load the npcs of unique instance from the same original map.


Jeybla:
I like the idea. Once I needed a similiar command to retrieve the id of a currently running instance.

`instance_info("<instance name>",<info type>{,<instance_db map index>});`
We could also change the header of instance_info in a more generic way, like
`instance_info(<info type>{,"<instance name or id>",<instance_db map index>});`

Whereby <instance name or id> works like your command:
1. Not given - running instance (if in progress)
2. String - search for the name
3. Integer - search for the instanceid

src/map/script.cpp
```diff
@@ -20309,27 +20309,62 @@ BUILDIN_FUNC(instance_check_clan)
...
+	else {
+		data = script_getdata(st, 3);
+		get_val(st, data);
+		if( data_isstring(data) ) {
+			name = script_getstr(st, 3);
+			db = instance_searchname_db(name);
+		}
+		else if( data_isint(data) || script_getnum(st, 3) ) {
+			id = script_getnum(st, 3);
+			if( id ) {
```
Lemongrass3110: Should we add a check for valid instance id here? > 0 && < MAX_INSTANCE_DATA

src/map/script.cpp
```diff
@@ -20309,27 +20309,62 @@ BUILDIN_FUNC(instance_check_clan)
...
+		}
+	}
+	else {
+		data = script_getdata(st, 3);
+		get_val(st, data);
+		if( data_isstring(data) ) {
+			name = script_getstr(st, 3);
+			db = instance_searchname_db(name);
+		}
+		else if( data_isint(data) || script_getnum(st, 3) ) {
```
Lemongrass3110: I would only treat this as "else" case or is there anything else possible than string or number?

src/map/script.cpp
```diff
@@ -20381,6 +20381,58 @@ BUILDIN_FUNC(instance_info)
...
+	int type = script_getnum(st, 2);
+	int id = 0;
+
+	if (script_hasdata(st, 3) == NULL)
+		id = script_instancegetid(st);
+	else
+		id = script_getnum(st, 3);
+
+	if (id) {
+		im = &instance_data[id];
```
cydh: it's a bit unsafe here. if `id` is defined in script, but no range check at least for `if (id < 1 || id >= MAX_INSTANCE_DATA)`

Atemo: Good catch!

src/map/script.hpp
```diff
@@ -699,6 +699,12 @@ enum instance_info_type {
 	IIT_MAP
 };
 
+enum instance_live_info_type {
```
aleos89:
```suggestion
enum e_instance_live_info_type : uint8 {
```

PR #3109, opened 2018-05-11
Added support for changedress command

Lemongrass3110:

* **Server Mode**: Both

This adds support for /changedress and /nocosplay by adding an atcommand for it and redirecting the client command to it.

Thanks to @Everade for finding this.


src/map/clif.cpp
```diff
@@ -20243,6 +20243,21 @@ void clif_achievement_reward_ack(int fd, unsigned char result, int achievement_i
 	WFIFOSET(fd, packet_len(0xa26));
 }
 
+/*
+ * This packet is sent by /changedress or /nocosplay
+ *
+ * 0ae8
```
sctnightcore: Is it CZ or ZC?

Lemongrass3110: Its an incoming packet so CZ.

sctnightcore: ty sir .

PR #3162, opened 2018-05-27
Added extra check in some renewal instances to prevent unexpected behaviour

Atemo:
* **Addressed Issue(s)**: NA

* **Server Mode**: Re

* **Description of Pull Request**: Added extra check in some renewal instances to prevent unexpected behaviour

### DIFF
--- a/npc/re/instances/SaraMemory.txt
+++ b/npc/re/instances/SaraMemory.txt
@@ -1083,72 +1083,73 @@ OnDisable:
 OnEnable:
 	enablenpc instance_npcname("#MobsSara01", instance_id());
 	set .@map$, instance_mapname("1@sara");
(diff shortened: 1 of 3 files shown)


npc/re/instances/SaraMemory.txt
```diff
@@ -1083,6 +1083,7 @@ OnDisable:
 OnEnable:
 	enablenpc instance_npcname("#MobsSara01", instance_id());
 	set .@map$, instance_mapname("1@sara");
+	killmonster .@map$, instance_npcname("#MobsSara01")+"::OnMyMobDead";
```
Lemongrass3110: Seeing the lines below this:
Would it not be better to store the event string into a variable(`.@event$`) like in faceworm as well?
This applies to all others below as well.

Issue #2772, opened 2018-01-05
OGH buffs


Atemo:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/c04c2e7ef1830474d9d4cd9c5404a6374d987f19

* **Client Date**: irrevelant

* **Server Mode**: renewal

* **Description of Issue**: 
  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information:
  
Following https://github.com/rathena/rathena/pull/1554 we need more infos on some OGH effects
EFST_GLASTHEIM_ATK
EFST_GLASTHEIM_DEF
EFST_GLASTHEIM_HEAL
EFST_GLASTHEIM_HIDDEN
EFST_GLASTHEIM_STATE
EFST_GLASTHEIM_ITEMDEF
EFST_GLASTHEIM_HPSP

* **Modifications that may affect results**: none

mrjnumber1:
these are only in OGH beginner mode on kRO, so for what we have now (which is only normal mode), it's not needed.
that said there are probably some instance variables and duplicates we need to make to account for the beginner and hard modes ..

Atemo:
https://github.com/rathena/rathena/blob/master/npc/re/instances/OldGlastHeim.txt#L1553
Theses npcs give buffs as well in normal mode depending on boss remaining hp.

mrjnumber1:
okay.. i guess it was changed since i last saw it. sorry about that.

mrjnumber1:
from 15.2 episode zone
EFST_GLASTHEIM_ATK
+100% damage to corruption root and amdarais (same as Orc Lady for those),
+100% MDEF ignore against undead/demon

EFST_GLASTHEIM_DEF
-100% dmg from corruption root and amdarais (as increase soil card does)

EFST_GLASTHEIM_HEAL
+100% heal power (like recov staff)
+50% received healing (As in sprint mail)

EFST_GLASTHEIM_HIDDEN
-90% damage from the HIDDEN_MOB HIDDEN_MOB2 and HIDDEN_MOB_W - presumably, this means you take less damage from the boss's summon spells

EFST_GLASTHEIM_STATE
+20 str/agi/dex/vit/int/luk

EFST_GLASTHEIM_ITEMDEF
+200 def, +50 mdef (as in arclouse card!)

EFST_GLASTHEIM_HPSP
+10.000 HP
+1.000 SP

PR #3167, opened 2018-05-27
Added Old Glastheim Buffs

aleos89:
* **Addressed Issue(s)**: #2772

* **Server Mode**: Renewal

- [x] Added missing Old Glastheim buffs.
- [x] added the buffs in ogh instance script

Thanks to @Atemo and @mrjnumber1!


Questune09:
after trying this changes same as i reported before the npc's at the top of the first map untill last map is very long wait to talk with

can anyone try the instance to test it if im wrong?

and also i notice the monsters are not enough some are like 5pcs on a map and some are 10pcs only i think it should have a lot of monsters inside it i cant define it clearly because as i test on nova server.

last thing the respawn of abysmal and some monsters at the second map are in 1 locations notice it when you kill a bunch of monster on 1 location and wait until it respawn they respawn in same directions also the knight job monster on the second portal at the right side after killing 2 or 3 monsters only it appeared quickly

sorry for long explanation and bad english hopefully this will help also and i think its clear? XD

Atemo:
We 'only' added old glast heim buffs in this PR, please open an issue if you find a bug and provide NPC name / screenshots / coordinates with your description of the issue. Thanks !

Issue #3189, opened 2018-06-07
Negative MDEF should not increase damage from magic


flamefury:
* **rAthena Hash**: 046d2e6c440083d868d0f89a86d6463e586afac1
* **Client Date**: Doesn't matter
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: Having negative hard MDEF will increase the damage you take from enemy magic attacks.
  * Expected Result: Negative hard MDEF should be treated as 0 hard MDEF
  * How to Reproduce: Equip items that reduce hard MDEF and use Odin's Power + Assumptio to further magnify the effects.
  * Official Information: https://www.novaragnarok.com/forum/topic/7885-interaction-of-assumptio-odins-power-players-hard-mdef-and-monster-matk/?do=findComment&comment=39045
* **Modifications that may affect results**: None

flamefury:
Some more proof without Assumptio:
AB soft MDEF = 41, hard MDEF = -18
Elder Willow MATK according [Divine-Pride](https://www.divine-pride.net/database/monster/1033/): 43-52
Fire Bolt single hit damage: [Low range of 2](https://cdn.discordapp.com/attachments/346870797446545411/459387651771072512/unknown.png)) to [high range of 11](https://cdn.discordapp.com/attachments/346870797446545411/459387578253049866/unknown.png)

With 0 hard MDEF:
> Damage = RANGE(43, 52) * (1000 + 0) / (1000 + 0 * 10) - 41
> Damage = RANGE(43, 52) - 41
> Damage = RANGE(2, 11)

If it was actually using -18 hMDEF, it would've looked like this instead:
> Damage = RANGE(43, 52) * (1000 - 18) / (1000 - 180) - 41
> Damage = RANGE(43, 52) * 982 / 820 - 41
> Damage = RANGE(51, 62) - 41
> Damage = RANGE(10, 21)

In other words, I should never see numbers 2-9 and should see numbers 12-21, but I do frequently while I was waiting the minimum and maximum values to appear.

[kRO replay file](if you wish to check for equipment effects / shadow equipment.)

PR #3190, opened 2018-06-07
Negative MDEF should not increase damage taken from magic

flamefury:
* **Addressed Issue(s)**: #3189 
* **Server Mode**: Renewal
* **Description of Pull Request**: Set negative MDEF to 0 before damage reduction calculation.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -6286,8 +6286,8 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 			 * RE MDEF Reduction
 			 * Damage = Magic Attack * (1000+eMDEF)/(1000+eMDEF) - sMDEF
 			 */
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -6286,8 +6286,8 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
 			 * RE MDEF Reduction
 			 * Damage = Magic Attack * (1000+eMDEF)/(1000+eMDEF) - sMDEF
 			 */
-			if (mdef < -99)
-				mdef = -99; // Avoid divide by 0
+			if (mdef < 0)
+				mdef = 0; // Negative eMDEF is treated as 0 on official
```
Lemongrass3110: @mrjnumber1 can you confirm this?

Issue #3195, opened 2018-06-10
Wrong item script


Balferian:
* **rAthena Hash**:  f675895
* **Client Date**: -
* **Server Mode**: Renewal
* **Description of Issue**: wrong item script
`15074,Geffen_Mage_Robe,Geffen Magic Robe,4,10,,1000,,40,,1,0xFFFFFFFF,63,2,16,,99,1,0,{ .@r = getrefine(); bonus bUseSPrate,-10-((.@r>=7) ? 5: 0); **bonus bVariableCastrate,-15-((.@r>=9) ? 5: 0);** bonus bNoCastCancel2; },{},{}`

it must be: `bonus bVariableCastrate,15-((.@r>=9) ? 5: 0);`
https://www.divine-pride.net/database/item/15074

PR #3196, opened 2018-06-10
Corrects Variable Cast Rate for Geffen Magic Robe (Renewal)

danielbernard:
* **Addressed Issue(s)**: #3195 

* **Server Mode**: Renewal

  * Corrects Variable Cast Rate for Geffen Magic Robe
  * Geffen Magic Robe (Renewal) (15074) is currently set to reduce variable cast rate by 15%, with an additional 5% at refine level 9. This should be an increase of 15%, reduced to 10% at level 9.


Atemo:
Thanks you !

PR #3198, opened 2018-06-12
Added missing maps to map_cache and its mapflags

cydh:
* **Addressed Issue(s)**: -
* **Server Mode**: Both
  * Added 2009 and 2012 RWC maps to map_cache (renewal and pre-renewal)
  * Added GVG-ish mapflags


Atemo:
Additionnally I would like `pvp_noparty` to `Ignores party alliances on PvP-enabled maps`, `nowarp` to block warpparty and `nowarpto` to block `@` commands

cydh:
will be good to add

Atemo:
Oh well I commented according to [guild_vs1 gvg mapflag](https://github.com/rathena/rathena/blob/master/npc/mapflag/gvg.txt#L24) and [pvp_noparty mapflag](https://github.com/rathena/rathena/blob/master/npc/mapflag/pvp_noparty.txt#L16)
however it should be gvg_noparty on gvg maps ([battle src](https://github.com/rathena/rathena/blob/master/src/map/battle.cpp#L7891)). Blame me

PR #3200, opened 2018-06-12
Moved common, config and mapcache to C++

Lemongrass3110:

* **Server Mode**: Both

Moves all remaining C files to C++.
Dropped everything related to BETA_THREAD_TEST.
Dropped mt19937 3rdparty library since it is part of C++.
Dropped msinttypes 3rdparty library since it is part of Visual Studio SDK.

All that is left now is to drop libconfig in a follow up after this pull was merged.


cydh:
Why that threading is dropped?

Issue #3191, opened 2018-06-07
SC_FREEZING effects update


flamefury:
* **rAthena Hash**: 046d2e6c440083d868d0f89a86d6463e586afac1
* **Client Date**: Doesn't matter
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: Freezing status as is currently coded matches iRO Wiki's results. That is:
Movement Speed - 50%, Attack Speed - 15%, Physical Defence - 10%, Variable Cast Time + 15%.
Minimum duration of 6 seconds.
  * Expected Result: Effects should instead be:
Movement Speed to 70%, Attack Speed - 30%, Physical DEF - 30%, Fixed Cast Time + 50%.
Minimum duration of 10 seconds.
  * How to Reproduce: Become affected by Freezing (Frost Misty, Dragon Breath - Water, Ice Bound Trap).
  * Official Information: http://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=5352
Specifically, this section:
** 빙결 상태이상의 효과가 변경되었습니다.
--
  | 이동속도 70% 감소 / 공격 속도 30% 감소 / 고정 캐스팅 시간 50% 증가
  | 방어력이 30% 감소합니다.
  | 화속성 몬스터 및 화속성 갑옷을 입은 플레이어도 빙결 효과의 영향을 받도록 변경됩니다.
  | - 빙결 효과의 지속시간이 최소 10 초가 되도록 변경되었습니다.
  | - 빙결 효과의 기본 지속시간이 늘어났습니다.

Please note that this newspost is newer than the following one, which lists effects closer to the iRO Wiki written effects (though not an exact match):
http://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=5123

Note that monsters only receive a DEF drop of 10%: https://github.com/rathena/rathena/issues/3191#issuecomment-397362050

* **Modifications that may affect results**: None

aleos89:
Do you have anything newer? That source is from 2009 which is prior to Renewal release. It's possible it has changed from then as well.

danielbernard:
I tried to do some digging, and found that irowiki switched **from** the "newer" values mentioned above, to what we currently use.
[2011.12.25](https://irowiki.org/w/index.php?title=Status_Effects&oldid=34727#Freezing) → [2012.09.28](https://irowiki.org/w/index.php?title=Status_Effects&oldid=37091#Freezing)
[...]

PR #3207, opened 2018-06-15
Fix SC_Freezing effects

flamefury:
* **Addressed Issue(s)**:  #3191
* **Server Mode**: Renewal
* **Description of Pull Request**: Update Freezing effects based on the information found from testing and newspost.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -6444,7 +6444,10 @@ static defType status_calc_def(struct block_list *bl, struct status_change *sc, 
 	if (sc->data[SC_FLING])
 		def -= def * (sc->data[SC_FLING]->val2)/100;
 	if( sc->data[SC_FREEZING] )
-		def -= def * 10 / 100;
+		if (bl->type == BL_PC)
+			def -= def * 30 / 100;
+		else
+			def -= def * 10 / 100;
 	if( sc->data[SC_ANALYZE] )
 		def -= def * (14 * sc->data[SC_ANALYZE]->val1) / 100;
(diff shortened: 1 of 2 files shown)


src/map/status.cpp
```diff
@@ -6531,8 +6534,6 @@ static signed short status_calc_def2(struct block_list *bl, struct status_change
 			  + def2 * ( sc->data[SC_JOINTBEAT]->val2&BREAK_WAIST ? 25 : 0 ) / 100;
 	if(sc->data[SC_FLING])
 		def2 -= def2 * (sc->data[SC_FLING]->val3)/100;
-	if( sc->data[SC_FREEZING] )
-		def2 -= def2 * 10 / 100;
```
aleos89: Would be best to remove all `SCB_DEF2` flags defined in `initChangeTables`.

src/map/status.cpp
```diff
@@ -6444,7 +6444,10 @@ static defType status_calc_def(struct block_list *bl, struct status_change *sc,
 	if (sc->data[SC_FLING])
 		def -= def * (sc->data[SC_FLING]->val2)/100;
 	if( sc->data[SC_FREEZING] )
-		def -= def * 10 / 100;
+		if (bl->type == BL_PC)
+			def -= def * 30 / 100;
+		else
+			def -= def * 10 / 100;
```
aleos89: You could 1 line this: `def -= def * (bl->type == BL_PC ? 30 : 10) / 100;`

src/map/status.cpp
flamefury: Done!

Issue #3213, opened 2018-06-16
End Sektura should inflict Crystallize, not Frozen


flamefury:
* **rAthena Hash**: 967f28ed000592fdddc5be16119cae32965bf432
* **Client Date**: Doesn't matter
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: End Sektura has a chance to inflict Frozen.
  * Expected Result: End Sektura has a chance to inflict Crystallize.
  * How to Reproduce: Use End Sektura on a target until its effect triggers.
  * Official Information: https://www.divine-pride.net/database/item/1393
http://ro.gnjoy.com/guide/runemidgarts/itemview.asp?itemSeq=1&ID=1393

Specifically, this line:
근접물리공격 시 일정확률로 적과 자신에게 **냉동**을 건다.

The characters for Frozen (동빙) or Freezing (빙결) are nowhere in the description.

* **Modifications that may affect results**: None

PR #3216, opened 2018-06-16
Corrected End Sektura bonus

aleos89:
* **Addressed Issue(s)**: #3213

* **Server Mode**: Renewal

  * Should inflict Crystallization status, not Freeze.
  * Added Eff_Freezing item bonus effect.
Thanks to @flamefury!


flamefury:
Crystallization, not Freezing

Issue #3225, opened 2018-06-17
Being Crystalized should increase damage taken from Wind magic


flamefury:
* **rAthena Hash**: 105cc3ab7daa6b1b5e1278273aa1b41dd7c6ed5a
* **Client Date**: Doesn't matter
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: Wind magic does normal damage to a Crystallized target.
  * Expected Result: Wind magic does increased damage to a Crystallized target.
  * How to Reproduce: Crystallize a target, then use a Wind magic attack against them.
  * Official Information: Ripped from data.grf because I'm too lazy to get on kRO right now to check if my Sorc has enough skill points to learn Diamond Dust:

냉동 상태
이동, 공격, 스킬, 아이템 사용 불가
지속적인 HP, SP 감소
둔기, 도끼, 양손 도끼에 의한 피해 증가
_**풍속성 마법에 의한 피해 증가**_
단검, 검, 양손검, 화살에 의한 피해 감소

* **Modifications that may affect results**: None

flamefury:
https://streamable.com/wtgbf
Quick test, Sorc using Lightning Bolt vs Roda Frog.
Normal damage = 57
Crystallized damage = 85

Should be a 50% damage increase.

PR #3226, opened 2018-06-18
Corrected Crystallized state effect

aleos89:
* **Addressed Issue(s)**: #3225

  * While Crystallized Wind damage is increased 50%.
  * Corrected an arrow state check for Crystallization on weapon damage reduction.
  * Corrected elemental damage calculation for targets in pre-renewal mode.
Thanks to @flamefury!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -492,9 +492,12 @@ int64 battle_attr_fix(struct block_list *src, struct block_list *target, int64 d
 	if (tsc && tsc->count) { //increase dmg by target status
 		switch(atk_elem) {
 			case ELE_FIRE:
-				if (tsc->data[SC_SPIDERWEB]) {
-					//Double damage
+				if (tsc->data[SC_SPIDERWEB]) { //Double damage
+#ifdef RENEWAL
+					ratio += 100;
+#else
 					damage *= 2;
+#endif
 					//Remove a unit group or end whole status change
 					status_change_end(target, SC_SPIDERWEB, INVALID_TIMER);
 				}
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -503,32 +506,60 @@ int64 battle_attr_fix(struct block_list *src, struct block_list *target, int64 d
...
 				if (tsc->data[SC_VENOMIMPRESS])
+#ifdef RENEWAL
 					ratio += tsc->data[SC_VENOMIMPRESS]->val2;
+#else
+					damage += (int64)(damage * tsc->data[SC_VENOMIMPRESS]->val2 / 100);
+#endif
 				break;
 			case ELE_WIND:
-				if (tsc->data[SC_WATER_INSIGNIA])
+				if (tsc->data[SC_WATER_INSIGNIA] || tsc->data[SC_CRYSTALIZE])
```
flamefury: Ah, does it not stack with Water Insignia? Or does this get run twice?

aleos89: No, you're right. I was trying to keep code redundancy to a minimum and didn't think about the two stacking!

BrOgBr:
Can not attack, move, use most items or use most skills
Continuously lose 2% of HP and 1% of SP per second
Increase damage received from Maces, Axes and Two-handed Axes by 50%
Increase damage received from Wind property magic attacks by 50%
Reduce damage received from Daggers, Swords, Two-handed Swords, Arrows and gun ammunition by 50%

http://ragnarok.wikia.com/wiki/Crystallization

PR #3246, opened 2018-06-24
Removed the unused damage log on each player

Lemongrass3110:

* **Server Mode**: Both

This array was being filled and cleared, but never read anywhere.


Lemongrass3110:
Thanks to @cydh - this is used.

Issue #3256, opened 2018-06-27
Adding atcommand reloadcashdb


gengstergile:
* **rAthena Hash**: 
f196304f0

* **Client Date**: 
20151104

* **Server Mode**: 
Both

* **Description of Issue**: 
No ability to switch either to use cashdb sql only.

  * Result: -
  * Expected Result: -
  * How to Reproduce: -
  * Official Information:-
  

* **Modifications that may affect results**: 
separating cash shop from other itemdb and reloading only cashdb when updating cash shop to reduce loading time.
adding function to use txt itemdb while using sql cashdb

gengstergile:
yes i noticed the issue, for logged in user need to update the cash shop they need to relog for the changes to take effect, but this issue also exist with the reloaditemdb. adding this functionality can separate the need to use all db either sql or txt since cash shop need to be update much regularly.

sader1992:
i don't think its a need as `@reloaditemdb` work they way you want the command to work

also rAthena is moving to yaml so i think we better not care a lot about how these things work for now ,Because this may be changed soon , however this is only my opinion.

cydh:
1. about this issue & PR, yeah separate to reload cashdb from itemdb is reasonable
2. @Lemongrass3110 force to send or resend cash shop list in **kRO clients** will cause the client crash. iirc the client request to get cash shop list on player first login (until client reopened). (uhh sorry it was years ago, I noticed when play with openkore). So as long as after reloading then (I guess our code is safe) then we do some check on buy request, it's ok.
3. (i don't want to mention) implementing and worry about yml conversion will bring you nowhere.

Akkarinage:
Merged in fd9b475

PR #3257, opened 2018-06-27
Added a command to reload cash shop db

gengstergile:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/3256

* **Server Mode**: 
Both

separating cash shop from other itemdb and reloading only cashdb when updating cash shop to reduce loading time.
adding function to use txt itemdb while using sql cashdb

Added command to reload cash shop db only
 * use @reloadcashdb
Added cashdb switch function to use sql/db
 * can be modified at inter_athena.conf


src/map/atcommand.cpp
```diff
@@ -3829,6 +3829,9 @@ ACMD_FUNC(reload) {
 	if (strstr(command, "itemdb") || strncmp(message, "itemdb", 4) == 0) {
 		itemdb_reload();
 		clif_displaymessage(fd, msg_txt(sd,97)); // Item database has been reloaded.
+	} else if (strstr(command, "cashdb") || strncmp(message, "cashdb", 4) == 0) {
+		cashdb_reload();
+		clif_displaymessage(fd, "Cash shop database has been reloaded.");
```
thervingi: Why not use msg_txt instead of hard-coded string?

conf/inter_athena.conf
```diff
@@ -153,6 +153,9 @@ roulette_table: db_roulette
 // Use SQL item_db, mob_db and mob_skill_db for the map server? (yes/no)
 use_sql_db: no
 
+// Use SQL item_cash_db, item_cash_db2 for the map server? (yes/no)
+use_sql_cashdb: yes
```
thervingi: Maybe it's better if it is **no** by default? I mean, up until now it was one with the item_db so by default it would've been **no**. Besides, wouldn't this require rAthena users to import an additional SQL file to be able to run a fresh server?

src/map/atcommand.cpp
gengstergile: oh, yeah. because i wrote the modification for my server and just put it into the pull request, and forgot to make additional changes. in my server bcoz im too lazy to change if rathena update msg_txt and conflicting with the changes i made.

conf/inter_athena.conf
gengstergile: yes it is better to set no by default, sorry forgot to change that as well.

Zarbony:
Very usefull an to edit with npc an with SQL instead of db whish needs Note Time and ja Not Very Dynamic.
I Love this one and hope it will merged into Master:)

src/map/itemdb.cpp
```diff
@@ -1979,6 +1979,17 @@ void itemdb_reload(void) {
 	mapit_free(iter);
 }
 
+/**
+* Reload Cash DB
+*/
+void cashdb_reload(void) {
+	struct s_mapiterator* iter;
+	struct map_session_data* sd;
+
```
teededung: Remove `struct s_mapiterator* iter;` and `struct map_session_data* sd;` ? I got warning when compiling it.

PR #3272, opened 2018-07-02
Improved `char_del_option`

cydh:
* **Addressed Issue(s)**: -
* **Server Mode**: Both
  * Renamed to `char_deletion_code` and moved to login_athena.conf
  * The value is now using column name instead the number, available options are:
    * `email` (except a@a.com),
    * `birthday` (YYMMDD),
    * `pincode`,
    * and the last is `Sub-query` which allows you to defines your custom deletion code (as example you ask user for secret code while registering an account, or one-time-password for character deletion)
  * Extends E-mail length from 39 to 49
  * Reordering `0x2717` to make notation with variable length be on last


Daegaladh:
Love the idea, but some example of sub-query or further documentation would be fine ;)

cydh:
thank you
well actually sub-query can be *everything*, the `pincode` column itself is the example. I can't figure how write a doc there, must I refer to [MySQL Sub-query reference](https://dev.mysql.com/doc/refman/8.0/en/subqueries.html)? xd

Haikenz:
If I apply this diff I can not log in, the message appears:
`make_connection: connect failed (socket #8, error 111: Connection refused)!`

Everade:
@cydh 
@aleos89 
Any specific reason this hasn't been merged yet?

Issue #3277, opened 2018-07-03
Script (disable_items;) duplicating items


Badarosk0:
* **rAthena Hash**: 
https://github.com/rathena/rathena/commit/e4eddc3e617bd5702f5ce29f605ee66207fe66b4

* **Client Date**: 
2018

* **Server Mode**: 
Renewal

* **Description of Issue**: 

  * Result: Currently scripts using disable_itens; are not safe. Depending on the script's to cheat and remove items or duplicate. I've seen players doing this in privates. This is a sad thing.

  * Expected Result: Can not remove any item when the script exists in the disable_itens code; even using any skill with the NPC chat open.

  * How to Reproduce: Open an enchanting NPC or quest. Here's an option before you enchant. Use any skill to break the disable_itens; script. Ready now you can remove your item, add another one among others. I got NPCs that the staff provides here on rAthena and to remove an item before enchanting, you will get a new item then due to getitem2.
In my opinion it is a serious problem.

  * Official Information: You can not remove any item from the char when it is in NPCs, especially spells.

* **Modifications that may affect results**:

ecdarreola:
Try this setting in conf/battle/items.conf

Find line:
`// Allow changing of equipment while interacting with NPCs? (Note 1)
// Default: yes
item_enabled_npc: yes
`

and change to:
`// Allow changing of equipment while interacting with NPCs? (Note 1)
// Default: yes
item_enabled_npc: no
`

Should fix the changing part.

Badarosk0:
This solves the problem, but the script would be mocking when using disable_itens;

petrecci:
Fix, by lemon
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -12160,10 +12160,9 @@ void clif_parse_UseSkillToId(int fd, struct map_session_data *sd)
 			clif_msg(sd, WORK_IN_PROGRESS);
 			return;
 		}
-#else
+#endif
 		if (!sd->npc_item_flag || !(inf&INF_SELF_SKILL))
 			return;
-#endif

PR #3279, opened 2018-07-03
Hotfix for issue 3277

Lemongrass3110:
* **Addressed Issue(s)**: #3277

* **Server Mode**: Both

Fixes losing state of disable_items in certain scenarios.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -12160,10 +12160,9 @@ void clif_parse_UseSkillToId(int fd, struct map_session_data *sd)
 			clif_msg(sd, WORK_IN_PROGRESS);
 			return;
 		}
-#else
+#endif
 		if (!sd->npc_item_flag || !(inf&INF_SELF_SKILL))
 			return;
-#endif
 	}
 
 	if( (pc_cant_act2(sd) || sd->chatID) && skill_id != RK_REFRESH && !(skill_id == SR_GENTLETOUCH_CURE &&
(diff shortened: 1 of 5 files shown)


src/map/clif.cpp
```diff
@@ -12160,10 +12160,9 @@ void clif_parse_UseSkillToId(int fd, struct map_session_data *sd)
 			clif_msg(sd, WORK_IN_PROGRESS);
 			return;
 		}
-#else
+#endif
 		if (!sd->npc_item_flag || !(inf&INF_SELF_SKILL))
 			return;
```
cydh: thow this with clif wip message

src/map/status.cpp
```diff
@@ -4227,6 +4228,10 @@ int status_calc_pc_(struct map_session_data* sd, enum e_status_calc_opt opt)
 	if( (skill = pc_checkskill(sd,SU_SPRITEMABLE)) > 0 && !sd->sc.data[SC_SPRITEMABLE] )
 		sc_start(&sd->bl, &sd->bl, SC_SPRITEMABLE, 100, 1, -1);
 
+	if( previous_st ){
+		script_attach_state(previous_st);
```
ecdarreola: Causes error in latest commit.

Lemongrass3110: Why would it? I think you merged badly.

ecdarreola: My bad it was causing the other error. I'l post it in conversation. Sorry.

Issue #3301, opened 2018-07-09
Guild Expansion Skill


spinzaku:
* **rAthena Hash**: Latest (will check when i get home)

* **Client Date**: 2017-06-28

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Level 5 Guild Expansion Storage only has 100 item space @gstorage
  * Expected Result: It should be giving 500 item space
  * How to Reproduce: If you have Level 1 Guild Expansion and you do @gstorage it will work as intended, you will get 100 item space. But if you upgrade the skill to level 2,3,4,5 it will still be 100 item space. But if you created a new guild and instantly leveled the Guild Expansion to Level 5 (max) and you @gstorage for the first time, the item space will be 500.

![gstorage](https://user-images.githubusercontent.com/18629476/42428054-559af6d8-8364-11e8-8887-6c4a391ca3df.png)

BrOgBr:
Confirmed here.
It only worked as I've restarted the server.

Haikenz:
![sem titulo](https://user-images.githubusercontent.com/17690412/42507743-9eb2f470-841c-11e8-9bd3-a06df3722327.png)
Works fine to me.

aleos89:
The most likely cause is that the `autosave_time` (60 seconds by default) hasn't triggered yet to tell the char-server to save the data that's in memory.

PR #3306, opened 2018-07-10
Fixed Guild Storage Expansion skill saving

aleos89:
* **Addressed Issue(s)**: #3301

* **Server Mode**: Renewal

  * Force the Guild Storage Expansion skill to save on level up so the next time the guild storage is opened the proper increase is shown.
Thanks to @spinzaku!


xTakami:
Hello doing some tests the following happened to me, if a person upgrade the Guild Storage Expansion skill with the Guild Storage open it does not add the 100 spaces per level, for view the change of the 500 spaces it is necessary to restart the emulator.

spinzaku:
@xTakami yep that's what i reported. I think that's also what @aleos89 fixed. but i haven't tested yet

xTakami:
@spinzaku I applied the solution that public @aleos89  and the problem still occurs when the skill is raised to level 1 and then used @gstorage, if you keep the Guild Storage open while the skill is upgrade, stop adding the extra slot.

PR #3319, opened 2018-07-15
Added prt_q and prt_prison

mazvi:
Added prt_q and prt_prison

* **Addressed Issue(s)**: 

* **Server Mode**: 


Atemo:
https://github.com/rathena/rathena/blob/master/npc/re/mapflag/nomemo.txt#L188
`nomemo` mapflags are already enabled for ep16 maps

Daegaladh:
But nomemo doesn't avoid recalling family members :/

Atemo:
Oh! I see, thanks for clarifying.

Issue #3317, opened 2018-07-14
Guild Expansion Saving Level


admkakaroto:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/7a8ac464211d53bd11725d30d909a139db39e307
* **Client Date**: 2018-05-30
* **Server Mode**: RE
* **Description of Issue**: the problem is simple, if the player puts only 1 point or more, but do not put the 5 points and open the warehouse, the guild saves the amount of spaces and does not increase even more placing the rest of the points...
  * How to Reproduce:
1. Create a guild.
2. Level UP the guild.
3. Put 4 points expansion skill of guild storage, but it has to be less than 5, 4 or any value lower than that.
4. Now open the guild warehouse.
5. Close the warehouse.
6. Put remaining dots in the expansion ability.

**Players can not take items out of the warehouse.**

BrOgBr:
Already fixed. oO

admkakaroto:
The problem is caused by this commit.

xTakami:
The same thing is happening to me, thanks for generating the report @admkakaroto

PR #3321, opened 2018-07-17
Resolved Guild Storage Expansion not updating

aleos89:
* **Addressed Issue(s)**: #3317

* **Server Mode**: Renewal

  * Resolves the Guild Storage Expansion skill not refreshing the map server cached data.
  * Updated some inter server checks to grab the proper max size of guild storages.
Thanks to @admkakaroto!

### DIFF
--- a/src/map/storage.cpp
+++ b/src/map/storage.cpp
@@ -603,7 +603,12 @@ char storage_guild_storageopen(struct map_session_data* sd)
 		return GSTORAGE_ALREADY_OPEN;
 	}
 
-	if((gstor = guild2storage2(sd->status.guild_id)) == nullptr || (gstor && gstor->max_amount < guild_checkskill(sd->guild, GD_GUILD_STORAGE) * 100)) {
+	if((gstor = guild2storage2(sd->status.guild_id)) == nullptr
+#ifdef OFFICIAL_GUILD_STORAGE
+		|| (gstor && gstor->max_amount < guild_checkskill(sd->guild, GD_GUILD_STORAGE) * 100)
+#endif
+		)
+	{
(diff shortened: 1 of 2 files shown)


admkakaroto:
Thanks man!

admkakaroto:
Congratulations on the great work!

anacondaqq:
Good.

xTakami:
Thanks!

src/map/storage.cpp
```diff
@@ -574,7 +574,7 @@ void storage_guild_delete(int guild_id)
  */
 char storage_guild_storageopen(struct map_session_data* sd)
 {
-	struct s_storage *gstor;
+	struct s_storage *gstor = nullptr;
```
Lemongrass3110: useless init

src/map/storage.cpp
```diff
@@ -603,7 +603,12 @@ char storage_guild_storageopen(struct map_session_data* sd)
 		return GSTORAGE_ALREADY_OPEN;
 	}
 
-	if((gstor = guild2storage2(sd->status.guild_id)) == NULL) {
+	if((gstor = guild2storage2(sd->status.guild_id)) == nullptr
+#ifdef OFFICIAL_GUILD_STORAGE
+		|| (gstor && gstor->max_amount < guild_checkskill(sd->guild, GD_GUILD_STORAGE) * 100)
```
Lemongrass3110: Checking gstor for not null in this "or" condition is useless, since it will short-circuit if the check before returns null. Hence gstor can only be not null after this check.
Also I would not only check if max_amount is smaller, but if its not identical. In case the guild skills got reseted.

Issue #3316, opened 2018-07-14
NPC Santa change location


AzarthMZintos:
* **rAthena Hash**:  in the screenshot

* **Client Date**:  2018-05-30b

* **Server Mode**: renewal

* **Description of Issue**: 
  * Result: The navigation system tell me to go to that location instead of the old location of santa claus
  * Expected Result: The NPC should be move?!
  * How to Reproduce: Just use navigation system to toy dugeon
  * Official Information:I dont have any sorry, just the client it self
  

* **Modifications that may affect results**: 
  
  

<a href="https://ibb.co/nBzDcT"><img src="https://preview.ibb.co/b0gRxT/screenr_Athena002.jpg" alt="screenr_Athena002" border="0"></a>

PR #3329, opened 2018-07-18
Fixed npc location for Lutie quest

Lemongrass3110:
* **Addressed Issue(s)**: #3316

* **Server Mode**: Renewal

Split Lutie town npcs to pre-re and re.

Thanks to @AzarthMZintos


aleos89:
What if we threw this into `npc/cities/aldebaran.txt` with a blank NPC define and then just add a NPC duplicate for pre-re and renewal.

Issue #3336, opened 2018-07-20
Server crash after destroying the party


teededung:
hash: 1d88575f90b5f572cffba4a361f3028003008ffb
client date: 20170614

clif.cpp

The lines provoke map-server crash:

```
		if( (battle_config.bg_flee_penalty != 100 || battle_config.gvg_flee_penalty != 100) &&
			(map_flag_gvg(sd->state.pmap) || map_flag_gvg(sd->bl.m) || map_getmapflag(sd->state.pmap, MF_BATTLEGROUND) || map_getmapflag(sd->bl.m, MF_BATTLEGROUND)) )
			status_calc_bl(&sd->bl, SCB_FLEE); //Refresh flee penalty
```

when commented, no crash happen.

Extra info:

```

		/**
		 * Get a key-value pair and return the key value
		 * @param map: Map to search through
		 * @param key: Key wanted
		 * @param defaultValue: Value returned if key doesn't exist
		 * @return Key value on success or defaultValue on failure
		 */
		template <typename K, typename V> V map_get(std::map<K, V>& map, K key, V defaultValue) {
			auto it = map.find(key); //  <-------------------- this one

			if (it != map.end())
				return it->second;
			else
				return defaultValue;
		}

```

When key = MF_GVG on line above.

```

How to reproduce

1.) Create any char
2.) "`@warp bif_fild01,158,340`"
3.) create party (`@party fsdfsd`)
4.) create instance and login to the instance
5.) inside instance do /leave to leave party
6.) boom map-server crash

Which is provoked by line below:
```
		if( (battle_config.bg_flee_penalty != 100 || battle_config.gvg_flee_penalty != 100) &&
			(map_flag_gvg(sd->state.pmap) || map_flag_gvg(sd->bl.m) || map_getmapflag(sd->state.pmap, MF_BATTLEGROUND) || map_getmapflag(sd->bl.m, MF_BATTLEGROUND)) )
			status_calc_bl(&sd->bl, SCB_FLEE); //Refresh flee penalty

```


When the line above is commented in clif.cpp, no crash happen.

### Also, please note:
- clean rAthena without any modifications
- tested on ranger

yoonjunho72:
tested on 20170614 and 20180620, hmm there is no crash i create an instance and leave party with **/leave** command inside the instance.

teededung:
@yoonjunho72 Did you go into the map of instance?

Lemongrass3110:
Fixed your hash.

[...]

PR #3347, opened 2018-07-25
Cleaned up map data storage

aleos89:
* **Addressed Issue(s)**: #3336

  * Follow up to a942853.
  * Moved map_data storage to std::map container.
  * Created map_getmapdata to obtain pointer to map_data.
  * Got rid of MAX_MAP_PER_SERVER define.
  * Resolved deletion of instances causing other instances to lose their intended map ID.
Thanks to @teededung, @anacondaqq, and @Lemongrass3110!

### DIFF
--- a/src/char/char.hpp
+++ b/src/char/char.hpp
@@ -4,6 +4,8 @@
 #ifndef _CHAR_HPP_
 #define _CHAR_HPP_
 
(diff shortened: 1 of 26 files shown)


keitenai:
channel_setcolor is causing mapserver crash

Issue #3349, opened 2018-07-25
Important Family Share EXP Bug


gustavobrigo:
* **rAthena Hash**: latest

* **Client Date**: 
2015

* **Server Mode**: 
Both

* **Description of Issue**: 
  * Result:  There is a char server check on the family condition where the child_id is 0 it allows a party with 2 members share exp over the limit and without being a family. 
This bugs happens around the check over the leave / change master routine
  * Expected Result: 
Not able to share
  * How to Reproduce: 
You need 3 chars 2 8x(or any higher level) classes and 1 1/1 novice:
- Create a party with one of the 2 8x chars
- Invite the novice 1/1
- Invite the other member
- Change the leader to novice 1/1
- Leave the party (old leader)
- Now the novice 1/1 CAN set the exp share to equal.
  * Official Information:
  

* **Modifications that may affect results**:

PR #3350, opened 2018-07-25
Fixed EXP bug and added failsafe to family check functions

gustavobrigo:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/3349

* **Server Mode**: 

Both

This adds a failsafe function to not execute the query during the family check AND avoid an exp sharing bug when the child_id is == 0 AND the returning value from the SQL is 0 also, this makes the check on `char_child` function returns a FALSE 1(positive)

### DIFF
--- a/src/char/char.cpp
+++ b/src/char/char.cpp
@@ -1888,6 +1888,8 @@ int char_married(int pl1, int pl2)
 
 int char_child(int parent_id, int child_id)
 {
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
Could you rebase your pull and push it with the correct author please?

Thanks for your contribution!

gustavobrigo:
@Lemongrass3110 just fixed the user.

Lemongrass3110:
Thanks a lot! :+1:

PR #3351, opened 2018-07-25
Zeny Exploit on selling bullets/spheres

gustavobrigo:
* **Addressed Issue(s)**: 
Zeny Exploit on selling bullets obtained from magazines/boxes bought from Magazine Dealers

* **Server Mode**: 
Both

There is a NPC: npc\merchants\ammo_boxes.txt
Which sells ammo boxes at 500z containing 500 bullets, the selling price of the spheres and some bullets is greater than 1z making this a "zeny generator".

This pull fixes the TXT dbs, but it sould be done on the SQL files as well in a short future. I don't know for how many time this bug is around, but it is a HUGE exploit.


AzarthMZintos:
but you need the bullets for make the boxes?

cydh:
why?
player need to buy bullets first before exchange to cartridge
player need to gather materials before exchange to spheres, the material prices are higher than the spheres

gustavobrigo:
Well we had a issue with this NPC, anyway, this was a fail pull, sorry, we were just cleaning up out our commits here and trying to pass to rathena..

PR #3361, opened 2018-07-27
instance_create missing exception

sader1992:

* **Server Mode**: both

-
adding exception for if the the instance name is not correct

testing script
```
prontera,152,185,4	script	npcName	444,{
	instance_create("test");
}
```
before
image

after
image

i wanted to do this at instance.cpp
```cs
	//nullpo_retr(-1, db);
	if (!db) 
	{
		ShowError("instance_create: Invalid instance name '%s'.\n", name);
		return -5;
	}
```
but don't know what the best (also don't know if removing the `nullpo_retr` is a good thing


sader1992:
didn't knew sorry xD

Issue #3382, opened 2018-08-03
Mapflag warning after enabling PK Mode


lelouch22:
* **rAthena Hash**: 68fccf447b3c9a2aa90a3de7bdc10856400635bd

* **Client Date**: 2014-10-22bRagexe

* **Server Mode**: Pre-Renewal

* **Description of Issue**: After i enable the PK Mode this warning message shows up on the map server
  * Result: 
> [Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from aldeg_cas01.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from aldeg_cas02.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from aldeg_cas03.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from aldeg_cas04.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from aldeg_cas05.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from gefg_cas01.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from gefg_cas02.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from gefg_cas03.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from gefg_cas04.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from gefg_cas05.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from payg_cas01.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from payg_cas02.
[Warning]: npc_parse_mapflag: Unable to set PvP and GvG(null) Castle flags for the same map! Removing PvP flag from payg_cas03.
[...]

Issue #3370, opened 2018-07-31
Unable moving character and keeping lag after using @spawn command


teededung:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/3165e521237c50e8053723639e2a74c8353f36e1

* **Client Date**: 20170614

* **Server Mode**: Both

* **Description of Issue**: 
  * Result: Unable moving character and keeping lag after using ```@spawn``` command
  * Expected Result: No lag
  * How to Reproduce: ```@spawn orc_lady 100```
 
Test on clean rAthena. When I switch back to https://github.com/rathena/rathena/commit/7717c1e6bc8f067625bbf588a1b07c58f9f8ebaf it turns to normal.

aleos89:
What environment are you testing this on?

teededung:
I tested on Windows  @aleos89

Issue #3369, opened 2018-07-31
Map Server Crash Instance


mazvi:
* **rAthena Hash**: 374bfa7

* **Client Date**: 20170517RE

* **Server Mode**: RENEWAL

* **Server Information**: 
centos-release-7-5.1804.el7.centos.2.x86_64
gcc (GCC) 5.3.1 20160406 (Red Hat 5.3.1-6)

* **Description of Issue**: 
  * Result: Map Server Crash On Instance OGH
  * Expected Result: 
  * How to Reproduce: This player dual login and party to enter OGH, then 1st Comannder of Destruction and 2nd not apear, THEN HE WAITING FOR LONG! on OGH after that map server crash.
  * Official Information:

LOG UPDATE:

```
(08/01/2018 20:57:59) [ Error ] : buildin_getnpcid: No such NPC 'Wandering Minstrel#mal'.
(08/01/2018 20:57:59) [ Debug ] : Source (NPC): Wanderer#mal at malangdo (149,120)
(08/01/2018 22:49:51) [ Error ] : script:op_2: invalid data for operator C_SUB
(08/01/2018 22:49:51) [ Debug ] : Data: string value="Thanatos Sword:Thanatos Great Sword:Thanatos Spear:Thanatos Long Spear:Thanatos Staff:Thanatos Two-handed Staff:Thanatos Bow:Thanatos Dagger:Thanatos Katar:Thanatos Knuckle:Thanatos Hammer:Thanatos Axe:Thanatos Violin:Thanatos Whip:Gray Shield:Gray Armor:Gray Robe:Gray Cloak:Gray Shoes:Gray Helmet:Cancel"
(08/01/2018 22:49:51) [ Debug ] : Data: number value=1
(08/01/2018 22:49:51) [ Debug ] : Source (NPC): King#dk11 at 0011@spa (204,29)
(08/02/2018 03:05:34) [ Error ] : script:op_2: invalid data for operator C_SUB
[...]

PR #3384, opened 2018-08-03
Cleaned up map storage type

aleos89:
* **Addressed Issue(s)**: #3369, #3370, #3382

  * Converted maps storage container back to static array.
  * Converted mapflag storage containers to unordered_map.
  * Removed a lot of redundancy and recursive lookups.
  * Fixed a couple short-circuit checks to help with performance.
  * Fixed PK Mode attempting to set PvP on flags that are already active.
  * Getting MF_RESTRICTED will return the map's zone value instead of the state.
Thanks to @teededung, @lelouch22, @Lemongrass3110, and @cydh!


teededung:
@aleos89 Script command `aremonster` **in Instance** is not working (no monsters).

LiamKarlMitchell:
e_mapflag could be used as a type in the std::unordered_map<e_mapflag, int> flag; ?

When pk_mode is set check not GVG, GVG_DUNGEON, GVG_CASTLE, BATTLEGROUND flags before setting pvp to 1?

Issue #3385, opened 2018-08-03
reserved identifier violation


elfring:
I would like to point out that identifiers like “[`_BATTLE_HPP_`](https://github.com/rathena/rathena/blob/8dea04a8b3895398a5c62e06dc8cf920cebd43df/src/map/battle.hpp#L4 "Update candidate")” and “[`_CONFIG_CORE_HPP_`](https://github.com/rathena/rathena/blob/93139226564233c35dc8359a1ccb747f8106e934/src/config/core.hpp#L4 "Another update candidate")” [do not fit](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51-CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29 "Do not declare an identifier which is reserved for the compiler implementation.") to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?

elfring:
Thanks for [your adjustment of affected include guards](https://github.com/rathena/rathena/commit/967fe6f5ab8c946d863bc90d52a5d52843cdad85 "Cleaned up header guard names").

PR #3395, opened 2018-08-08
Cleaned up header guard names

aleos89:
* **Addressed Issue(s)**: #3385

  * Updated the header guard names to C++ standards.
Thanks to @elfring's suggestion!


elfring:
Thanks for your source code improvement.

Issue #3400, opened 2018-08-12
inarray char temp is changing for unknown reason(unknown for me at least xD)


sader1992:
* **rAthena Hash**: 9ca000a4a751bad8b5e576ea67cd73740ef81442

* **Client Date**: 2018-05-02d

* **Server Mode**: Both

* **Description of Issue**: 
at `inarray` command script
https://github.com/rathena/rathena/blob/4befcf747fa0c6eb22c6c1dcc31c5f55486dcf91/src/map/script.cpp#L6401

  * How to Reproduce: 
```cs
-	script	test_string	-1,{
OnInit:
	.@s$ = "test";
	if(inarray(.@s$,"test") == -1)
		debugmes "It's False";
	else
		debugmes "It's True";
end;
}
```
the script should return true
however
this the result with some debug lines
```cs
[Error]: buildin_inarray: temp1 'test'.
[Error]: buildin_inarray: temp2 '²²²²²²²²²²²²²²²²²²²²╕░═@M¤'.
[Error]: buildin_inarray: value 'test'.
[Debug]: script debug : 0 110000253 : It's False
```

this happen only on string case , the int case work fine.

  * Modifications
adding debug lines
```cs
		for (i = 0; i <= array_size; ++i)
		{
			temp = (char*)get_val2(st, reference_uid(id, i), ref);
			ShowError("buildin_inarray: temp1 '%s'.\n", temp);
			script_removetop(st, -1, 0);
			ShowError("buildin_inarray: temp2 '%s'.\n", temp);
			ShowError("buildin_inarray: value '%s'.\n", value);
			if (!strcmp(temp, value))
			{
				script_pushint(st, i);
				return SCRIPT_CMD_SUCCESS;
			}
			
		}
```

i don't know if that is intended or not
i did try to see why this happen but i have failed xD

PR #3401, opened 2018-08-12
Fix string case in inarray command script

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/3400

* **Server Mode**: Both

maybe this fix if there is no other way ? as i fail to know a better fix


anacondaqq:
Like I said here: https://github.com/rathena/rathena/pull/3071#discussion_r187475158 is not it? 😄

sader1992:
@anacondaqq `Got a HeadShot On` @sader1992 xD

i misunderstood that comment xD
i expected to work with char same as int , i still don't why this happen
also when i tested it i test on int ❌ 
the more i learn

PR #3411, opened 2018-08-19
Item scripts from kRO 2017.05.10 update

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal


Lemongrass3110:
Maybe add them to the monster's card drop slot as well? :+1:

PR #3417, opened 2018-08-21
Malangdo enchant weapon update

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal

Added more weapon to enchant with Mayomayo NPC
_This update need to be verified on kRO_


admkakaroto:
has a problem with this npc, you can enchant too a few items.

An example: 13412 | 13413, these items have 3 slots and it is possible to put 2 incantations, thereby saving 1 slot of the item and if you use npc of card removal the spell will come out together.

Atemo:
It will be fine for now.

Issue #3426, opened 2018-08-23
Gospel can be refreshed after a deco reco


FriggRM:
* **rAthena Hash**: c856b07c45aed6c5842ee19a940f7e76f780221c

* **Client Date**: 2015-11-04

* **Server Mode**: Pre-Renewal

* **Description of Issue**: When a player deco and then reco he can instantly have a new gospel buff.
If the player is slow enough and is not banned from spamming the deco/reco he can stack all the buff from gospel. (ATK*2/HP*2/SP*2/ALLSTAT+20 ...)
  * Result: 
  * Expected Result: Gospel buff attribution cooldown should not be reset on dc
  * How to Reproduce: Simply deco reco on a gospel to have a new buff, rince and repeat
  * Official Information:
  

* **Modifications that may affect results**:

syracuase:
I can confirm to this

aleos89:
```lua
EFST_GOSPEL={
	ResetDead = true;
	ResetDispel = true;
	Save = false;
	Send = false;
	IgnorePretendDead = false;
	DeBuff = true;
	ResetCLEARANCE = true;
	ActorAppearance = false;
	SendMultiCast = false;
	["backward compatibility send"] = true; -- 이전버젼 호환성을 위해 추가합니다.
};
```
According to kRO, Gospel isn't saved on logout.

mrjnumber1:
the person who gets effected by gospel also gets 10000 ms of invincibility to any jt_gospel aoe. though whether or not that is also saved, i'm unsure of.

Daraen1:
Sorry to bump this old issue but I think this is insanely broken...

Can someone confirm this is an expected behavior here?
Rathena had a lot update since the issue was created.

I read on aleos89's pull request that it was difficult to fix.
Is this still the case now?

Playtester:
To me the PR looks pretty simple and I have no opposing information that would require ATKpercent to be saved on logout.

Daraen1:
Can we have any official information that confirms or denies whether it is normal or abnormal to have a new buff instantly after logging out and back in, please?

Playtester:
> Can we have any official information that confirms or denies whether it is normal or abnormal to have a new buff instantly after logging out and back in, please?

This is official. However, you lose most of your previous buffs when you do that, officially.

Daraen1:
It is the case atm on rathena?
If you lose all buff when logging out, getting new buff when logging in is not really a problem.
[...]

PR #3433, opened 2018-08-27
Removed Gospel buffs on logout

aleos89:
* **Addressed Issue(s)**: #3426

  * Remove miscellaneous buffs given by Gospel on logout to prevent stacking.
Thanks to @FriggRM and @mrjnumber1!


Mammouth-ROO:
Hello, What is delaying the fix?

aleos89:
The reason this is on hold is because I learned there was a lot more to this issue than just ending the status. On Aegis, SC_INCATKRATE and others behave as arrays, where a skill can store it's attack increase to the EFST separately. In the case of Gospel, only Gospel is cleared while other skills wouldn't be cleared.

Mammouth-ROO:
oh ! I understand it's hard to fix , but it's a pity! :( , this abuse bug are exploit by all , and destroy all mvp with disconcerting ease. For me it's a high priority emergency.
thank you for your work !

PR #3477, opened 2018-09-07
autopot script improvment

sader1992:
* **Server Mode**: both

while the original script sometimes work, but after testing it does not work all the time
or it does not work as the players normally want to.

**the rewrite info:**

1. the script now hook the player to an event with timer
2. the timer = the delay
3. adding minimum delay as config to the script
4. adding blacklist items so if someone don't want an item to work with the script
5. adding sp option
6. changing the commands as following
7. this is a down but the effect is white for hp and blue for sp (before was switching between some effects)
```
@autopot <hp|sp> <item id> {<min hp % [1..100]> {<delay [minimum delay ..1000]>}}
@autopot <hp|sp> list
@autopot <hp|sp> <on|off>
@autopot info
@autopot help
@autopot blacklist
```
for test
```
@str 9999
@item 502 500
@autopot hp 502
@monster orc_lord
```

ofc i would like to get feedback

### DIFF
--- a/npc/custom/etc/autopot.txt
+++ b/npc/custom/etc/autopot.txt
@@ -157,10 +157,10 @@ OnCommand:
 end;
 
 OnStart:
-	if (!getstatus(SC_BERSERK) || !getstatus(SC_SATURDAYNIGHTFEVER) || !getstatus(SC_GRAVITATION) ||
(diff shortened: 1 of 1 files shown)


sader1992:
also i will add SP option too for this script [in progress]

Issue #3484, opened 2018-09-11
EDP effect when switching weapon


Atemo:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/d17ddf41fe3f1a4eb0fcb75d82888ea89dc697ef

* **Client Date**: -

* **Server Mode**: renewal

* **Description of Issue**: 
  * Result: On rathena EDP effect remains when switching weapon.
  * Expected Result: On aegis **switch** to a different weapon or **unequip** a weapon cancel EDP. Also the message `The skill cannot be used with this weapon.` is sent to the client when the character try to use **EDP with fist** (note that the skill is not casted and poison bottle is not consumed).
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**:

PR #3485, opened 2018-09-12
EDP renewal change

Atemo:
* **Addressed Issue(s)**: #3484

* **Server Mode**: renewal

Switch to a different weapon or unequip a weapon now cancel EDP
EDP can't be used with fist (right hand check only)

### DIFF
--- a/conf/battle/skill.conf
+++ b/conf/battle/skill.conf
@@ -377,9 +377,9 @@ banana_bomb_duration: 0
 skill_drop_items_full: no
 
 // EDP setting: (Note 3)
-// On official renewal server switch/unequip right hand weapon should remove EDP and EDP can't be used with bare hand.
-// 0: Disabled (rAthena legacy and pre-renewal behaviour).
-// 1: Enable the setting on pre-renewal compilation.
-// 2: Enable on renewal compilation.
+// When switching/unequipping a right hand weapon, should EDP be removed? EDP can't be used with bare hand.
+// 0: Disabled (rAthena legacy and pre-renewal behavior).
+// 1: Enabled on pre-renewal.
+// 2: Enabled on renewal.
 // 3: 1+2
 switch_remove_edp: 2
--- a/conf/battle/skill.conf
+++ b/conf/battle/skill.conf
@@ -376,6 +376,10 @@ banana_bomb_duration: 0
 // Legacy rAthena logic: yes
 skill_drop_items_full: no
 
-// Switch/unequip right hand weapon should remove EDP and EDP cannot be used with fist? (Note 1)
-// Renewal official behavior is "yes".
-switch_remove_edp: yes
+// EDP setting: (Note 3)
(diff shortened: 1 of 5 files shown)


conf/battle/skill.conf
```diff
@@ -375,3 +375,7 @@ banana_bomb_duration: 0
 // Official: no
 // Legacy rAthena logic: yes
 skill_drop_items_full: no
+
+// Switch/unequip right hand weapon should remove EDP and EDP cannot be used with fist? (Note 1)
```
cydh: `fist` -> `bare hand`?

src/map/battle.cpp
```diff
@@ -8522,6 +8522,7 @@ static const struct _battle_data {
 	{ "autoloot_adjust",                    &battle_config.autoloot_adjust,                 0,      0,      1,              },
 	{ "broadcast_hide_name",                &battle_config.broadcast_hide_name,             2,      0,      NAME_LENGTH,    },
 	{ "skill_drop_items_full",              &battle_config.skill_drop_items_full,           0,      0,      1,              },
+	{ "switch_remove_edp",                  &battle_config.switch_remove_edp,               0,      0,      1,              },
```
cydh: default to yes, so `1, 0, 1`

Atemo:
I was wondering if the setting
```
// EDP setting: (Note 3)
// On official renewal server switch/unequip right hand weapon should remove EDP and EDP can't be used with bare hand.
// 0: Disabled (rAthena legacy and pre-renewal behaviour).
// 1: Enable the setting on pre-renewal compilation.
// 2: Enable on renewal compilation.
// 3: 1+2
switch_remove_edp: 2
```
is better. Disabled by default on pre-renewal compilation, enabled by default on renewal

conf/battle/skill.conf
```diff
@@ -375,3 +375,11 @@ banana_bomb_duration: 0
 // Official: no
 // Legacy rAthena logic: yes
 skill_drop_items_full: no
+
+// EDP setting: (Note 3)
+// On official renewal server switch/unequip right hand weapon should remove EDP and EDP can't be used with bare hand.
+// 0: Disabled (rAthena legacy and pre-renewal behaviour).
+// 1: Enable the setting on pre-renewal compilation.
+// 2: Enable on renewal compilation.
+// 3: 1+2
```
aleos89:
```
// EDP setting: (Note 3)
// When switching/unequipping a right hand weapon, should EDP be removed? EDP can't be used with bare hand.
// 0: Disabled (rAthena legacy and pre-renewal behavior).
// 1: Enabled on pre-renewal.
// 2: Enabled on renewal.
// 3: 1+2
```

Daegaladh:
I think it should have a 4th option, disabled on pre-renewal + enabled in renewal, and be the default one, so it have the official behaviour for the 2 modes by default.

Issue #3499, opened 2018-09-18
TK_MISSION mob id select to match aegis


mrjnumber1:
* **rAthena Hash**:  current
* **Client Date**: x
* **Server Mode**: either
* **Description of Issue**: 
  * Result: TK_MISSION pulls a random mob from MOBG_Branch_Of_Dead_Tree
  * Expected Result: to match aegis, of course
  * How to Reproduce: use tk mission
  * Official Information:  The formula extracted from 14.2 zone is a bit oddly written, but it can be simplified as follows:
    - The mob id is in the range [MOBID_PORING, MOBID_GREEN_IGUANA]; 
    - The mob must have below 30,000 HP.
    - The mob must give Base EXP.
    - The mob must be CLASS_NORMAL.
* **Modifications that may affect results**: x

PR #3509, opened 2018-09-19
Corrected Taekwon Mission mob generator

aleos89:
* **Addressed Issue(s)**: #3499

  * Corrected the way target monsters are generator for Taekwon Mission.
  * Requires: The mob ID is in the range [MOBID_PORING, MOBID_GREEN_IGUANA], the mob must have below 30,000 HP, the mob must give Base EXP, the mob must be CLASS_NORMAL.
Thanks to @mrjnumber1!


src/map/skill.cpp
```diff
@@ -6797,17 +6797,23 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			int id;
+			struct mob_data *tk_md;
+
+			do {
+				id = rnd_value(MOBID_PORING, MOBID_GREEN_IGUANA);
+				tk_md = map_id2md(id);
+			} while (!tk_md || (id >= MOBID_DESERT_WOLF_2 && id <= MOBID_ANTONIO) || id == MOBID_SWITCH || // These mobs automatically fail
+				tk_md->status.class_ != CLASS_NORMAL || // Must be a Normal mob
+				tk_md->status.hp >= 30000 || // Must be less than 30,000 HP
+				tk_md->db->base_exp <= 0); // Must give Base EXP
```
anacondaqq: must give exp, and used `<=`  instead of `>=` ?

aleos89: It's because I used `||` instead of `&&`. I could adjust the `while()` to `&&` and inverse the checks so it better matches the comments. :)

cydh:
sorry just simply against this
just add new random mob list for tk mission

mrjnumber1:
yeah that's what i think too, that way people can add to it easily

aleos89:
That's one of the reason I didn't originally create the mob group for this PR. Partly because of laziness to query the data and with a YAML mob_db we could easily have `Group: [ MOBG_Taekwon_Mission ]`.

cydh:
lol. u can ask me such "yeah I planned to, but lazy to make a new group, do it for me!" :rofl:

Issue #3513, opened 2018-09-20
Warning Installing Rathena


RadianFord:
* **rAthena Hash**: a12f1ce02f2c2bdeea3a09c3f4ea06f89a6762ca

* **Client Date**: 20170125ragexe

* **Server Mode**: Pre Renewal

* **Description of Issue**:  This happens while im compiling a fresh trunk.
Ubuntu 18.04 x64 version

```
int_mail.cpp: In function ‘void mapif_Mail_return(int, uint32, int)’:
int_mail.cpp:475:6: warning: ‘%s’ directive output may be truncated writing up t                                      o 39 bytes into a region of size 37 [-Wformat-truncation=]
 void mapif_Mail_return(int fd, uint32 char_id, int mail_id)
      ^~~~~~~~~~~~~~~~~
In file included from /usr/include/stdio.h:862:0,
                 from ../common/../../3rdparty/libconfig/libconfig.h:48,
                 from ../common/showmsg.hpp:7,
                 from int_mail.cpp:10:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:65:44: note: ‘__builtin___snprintf_c                                      hk’ output between 4 and 43 bytes into a destination of size 40
        __bos (__s), __fmt, __va_arg_pack ());
                                            ^
```

```
mob.cpp: In function ‘int mobskill_use(mob_data*, unsigned int, int)’:
mob.cpp:3551:5: warning: ‘%s’ directive output may be truncated writing up to 25                                      5 bytes into a region of size 253 [-Wformat-truncation=]
 int mobskill_use(struct mob_data *md, unsigned int tick, int event)
     ^~~~~~~~~~~~
In file included from /usr/include/stdio.h:862:0,
                 from /usr/include/c++/7/cstdio:42,
                 from /usr/include/c++/7/ext/string_conversions.h:43,
                 from /usr/include/c++/7/bits/basic_string.h:6349,
                 from /usr/include/c++/7/string:52,
                 from /usr/include/c++/7/stdexcept:39,
                 from /usr/include/c++/7/array:39,
                 from /usr/include/c++/7/tuple:39,
                 from /usr/include/c++/7/unordered_map:41,
                 from map.hpp:9,
                 from skill.hpp:12,
                 from unit.hpp:11,
                 from mob.hpp:13,
                 from mob.cpp:4:
[...]

Issue #2787, opened 2018-01-08
Warnings compiling on gcc 7.2.0


Daegaladh:
* **rAthena Hash**: a65e7dbf4697b64f835965d1e96b9502d1b7369e

* **Client Date**: not relevant

* **Server Mode**: both

* **Description of Issue**: 
  * Result: It seems GCC 7.2.0 became more strict:
```
int_mail.cpp: In function ‘void mapif_Mail_return(int, uint32, int)’:
int_mail.cpp:475:6: warning: ‘%s’ directive output may be truncated writing up to 39 bytes into a region of size 37 [-Wformat-truncation=]
 void mapif_Mail_return(int fd, uint32 char_id, int mail_id)
      ^~~~~~~~~~~~~~~~~
In file included from /usr/include/stdio.h:862:0,
                 from ../common/../../3rdparty/libconfig/libconfig.h:48,
                 from ../common/showmsg.h:11,
                 from int_mail.cpp:10:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:65:44: note: ‘__builtin___snprintf_chk’ output between 4 and 43 bytes into a destination of size 40
        __bos (__s), __fmt, __va_arg_pack ());
                                            ^
```
```
mob.cpp: In function ‘int mobskill_use(mob_data*, unsigned int, int)’:
mob.cpp:3551:5: warning: ‘%s’ directive output may be truncated writing up to 255 bytes into a region of size 253 [-Wformat-truncation=]
 int mobskill_use(struct mob_data *md, unsigned int tick, int event)
     ^~~~~~~~~~~~
In file included from /usr/include/stdio.h:862:0,
                 from /usr/include/c++/7/cstdio:42,
                 from /usr/include/c++/7/ext/string_conversions.h:43,
                 from /usr/include/c++/7/bits/basic_string.h:6349,
                 from /usr/include/c++/7/string:52,
                 from /usr/include/c++/7/stdexcept:39,
                 from /usr/include/c++/7/array:39,
                 from /usr/include/c++/7/tuple:39,
                 from /usr/include/c++/7/bits/stl_map.h:63,
                 from /usr/include/c++/7/map:61,
                 from mob.cpp:7:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:65:44: note: ‘__builtin___snprintf_chk’ output 4 or more bytes (assuming 259) into a destination of size 256
        __bos (__s), __fmt, __va_arg_pack ());
                                            ^
```
```
[...]

PR #3522, opened 2018-09-26
Corrected some GCC 7.1+ compile warnings

aleos89:
* **Addressed Issue(s)**: #2787, #3513

  * Resolved some format-truncation warnings.
Thanks to @RadianFord!


aleos89:
You're right. We should check for `npos` in case that character isn't in the name.

PR #3541, opened 2018-10-03
Renewal Monster's ATK & MATK Calculation Updates

cydh:
* **Addressed Issue(s)**: 
* **Server Mode**: Renewal
-----
* Updated renewal min & max calculation for ATK and MATK calculation
* Massive updates on ATK2 in db/re/mob_db.txt
  * The value is currently enforce to [ATK2 = ATK2 - ATK1]
  * ATK2 from Aegis data is base MATK, while ATK1 is base ATK, not as ATK range
* Nothing was changed in pre-renewal system
* The tool/reset_mob_atk.pl is originally from convert_itemtype.pl by lighta


slyx88:
Shouldn't max damage of ifrit be 6450?
`return status->rhw.atk * 120 / 100;`
`5375 * 1.2 = 6450`
screenrathena000

cydh:
in this mob_db updates, most changes are in atk, some monsters have atk increased.
and of course the matk fix, it'll gives big different!

---
asd
the left is before the updates, and right is after. exactly how I felt when soloing TE Gloria Guild Dungeon, without wind element armor, Frozen + Jupitel Thunder from Wizard Katrinn = sent back to inn xD

officialwedevgames:
Still affecting the pre-renewal.

Questune09:
i got 1 shot both times 35k dmg with 115 int and more than 230 mdef with fireball. lord of vermilion and meteor storm instant kill as well. i had pain killer as well. so a character with low int would get dmg around 50k to 70k from a fireball. is that normal?

officialwedevgames:
Look like you need to revert this and its affecting the pre-re.

Issue #2962, opened 2018-03-08
Pets loyal cordial bonus system and more, pet db, e evolution


AzarthMZintos:
* **rAthena Hash**: 

Any

* **Client Date**: Any

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: the pet bonus only work for loyal, no option for cordial and bonus missing at those pets db
  * Expected Result: Pet Loyal and Cordial Bonuses db and working system(besides some have awkaward and neutral bonus) pet evolution system
  * How to Reproduce: Using pets and look at db
  * Official Information:https://www.divine-pride.net/database/item/9069/mastering-egg (not official info) mastering is evolution pet and have 4 intimacy diferente lvl bonus I look at the forum for info and only found this: https://rathena.org/board/topic/105838-renewal-pet-bonus-implementation/?tab=comments#comment-342885
  

* **Modifications that may affect results**:

aleos89:
Can you please provide the hash you're using.

AzarthMZintos:
I use Github desktop i have no ideia how to get that info, i just update if that tell you something, if you know how to see let me know

aleos89:
The new pet evolution system has already been started on. I'll just adjust some labels for now.

OptimusM:
Any updates on this?

aleos89:
Implemented in ac558d7.

PR #3544, opened 2018-10-04
Initial implementation of pet evolution system + Pet DB conversion to YAML

secretdataz:
* **Addressed Issue(s)**: #2962

* **Server Mode**: Renewal.

    - Converted pet database to YAML format.
    - Added new pet hunger mechanic by @aleos89.
    - Implemented pet evolution system.
    - Added new pets.

Some parts are still hacky. Do not merge.

**If you found a bug or have a suggestion, please open an issue instead of commenting here.**
And no, we will not discuss about whether the DB should be converted to YAML or not. Discussing about the structure is fine.


src/map/battle.cpp
```diff
@@ -8520,6 +8520,8 @@ static const struct _battle_data {
 	{ "allow_bound_sell",                   &battle_config.allow_bound_sell,                0,      0,      0xF,            },
 	{ "event_refine_chance",                &battle_config.event_refine_chance,             0,      0,      1,              },
 	{ "autoloot_adjust",                    &battle_config.autoloot_adjust,                 0,      0,      1,              },
+    { "feature.petevolution",               &battle_config.feature_petevolution,            1,      0,      1,              },
```
Atemo: Indentation

src/map/battle.cpp
```diff
@@ -8659,6 +8661,17 @@ void battle_adjust_conf()
...
 #endif
 
+#if PACKETVER < 20141008
+	if (battle_config.feature_petevolution) {
+		ShowWarning("conf/battle/feature.conf petevolution is enabled but it requires PACKETVER 2014-10-08 or newer, disabling...\n");
+		battle_config.feature_petevolution = 0;
+	}
+	if (battle_config.feature_petevolution) {
+		ShowWarning("conf/battle/feature.conf pet auto feed is enabled but it requires PACKETVER 2014-10-08 or newer, disabling...\n");
+		battle_config.feature_petevolution = 0;
```
Atemo:
```
	if (battle_config.feature_pet_auto_feed) {
		ShowWarning("conf/battle/feature.conf feature_pet_auto_feed is enabled but it requires PACKETVER 2014-10-08 or newer, disabling...\n");
		battle_config.feature_pet_auto_feed= 0;
```

src/map/map-server.vcxproj
```diff
@@ -351,7 +351,7 @@
     <Copy SourceFiles="$(SolutionDir)db\import-tmpl\mob_race2_db.txt" DestinationFolder="$(SolutionDir)db\import\" ContinueOnError="true" Condition="!Exists('$(SolutionDir)db\import\mob_race2_db.txt')" />
     <Copy SourceFiles="$(SolutionDir)db\import-tmpl\mob_random_db.txt" DestinationFolder="$(SolutionDir)db\import\" ContinueOnError="true" Condition="!Exists('$(SolutionDir)db\import\mob_random_db.txt')" />
     <Copy SourceFiles="$(SolutionDir)db\import-tmpl\mob_skill_db.txt" DestinationFolder="$(SolutionDir)db\import\" ContinueOnError="true" Condition="!Exists('$(SolutionDir)db\import\mob_skill_db.txt')" />
-    <Copy SourceFiles="$(SolutionDir)db\import-tmpl\pet_db.txt" DestinationFolder="$(SolutionDir)db\import\" ContinueOnError="true" Condition="!Exists('$(SolutionDir)db\import\pet_db.txt')" />
+    <Copy SourceFiles="$(SolutionDir)db\import-tmpl\pet_db.yml" DestinationFolder="$(SolutionDir)db\import\" ContinueOnError="true" Condition="!Exists('$(SolutionDir)db\import\pet_db.txt')" />
```
Atemo: Import `.yml`

anacondaqq:
I'm so sorry, please try to understand me correctly, no negative, etc.
But is it really necessary to convert each DB to yaml? What main purpose? Just job for job? Or for commits resume history? There are a lot of warnings and issues with yaml parser in the code since the first conversion which is not fixed at all, there are many other issues too.

Is it easier to read? Absolutely no. Wall of text of nonsense info, with super stupid format where each space/tab / symbol is very important. I was tried to work with achievement db when tried to implement around 100 custom achievements and to be clear, I did not feel so many pain with yaml ever.

Sorry, don't get me rude, but I do not see any reason of breaking current libconfig format in any other DB, because it does not bring any benefits at all because of the DB still not usable and hard to read and hard to maintain for example on a customizable server. 

Maybe better to write one GUI tool like SDE to work with whole DB, instead of converting from nonsense info to another nonsense format where need read hell a lot to understand what is going on.

db/import-tmpl/pet_db.yml
```diff
@@ -0,0 +1,39 @@
+###########################################################################
+# Pet Database
+###########################################################################
+#
+# Pet Settings
+#
+###########################################################################
+# - MobID                    Monster ID of the pet
+#   Name                     the monster as defined in the database.
```
Atemo: Maybe the docs should say when a field is optional

secretdataz:
@anacondaqq 
With the CSV format, it gets exponentially harder to read or customize the DB when an array type data is added, and YAML is perfect for handling that.
And you can't mess up YAML formatting unless you are editing a YAML file on Windows 98 SE with Notepad. Most modern text editors supports automatic formatting of YAML file.
Oh, and it surely helps with readability.
Unlike YAML, most modern editors does not support libconfig out of the box or with an extension. Try searching for libconfig support for Atom, Sublime Text, VSCode, etc. :)

anacondaqq:
@secretdataz about csv, I do not agree.

Check:

What is not readable here? 

Or for example another example:



About LibConfig i do not agree twice.



And yml here:

cydh:
1. CSV is totally readable __BUT__ need xyz, must install blaha blah blah blah. so incorrect. while YAML u can open it with mere notepad with easy to understand "this hell: have this value"
2. YAML tab and spaces are the prbolem... ur text editor must screw up. YAML uses TXT if ur text editor is good, it'll read our editconfig file that use spaces for YAML that is the standard. there are bunch of options that __**ARE DEFAULT YAML**__ string parsing. what u see on herc, iirc is custom grammar for string.
3. you compare it libconfig? what a low-measurement, such a turn down of event. if u prefer listing sequence and map using bracketz, u can simply use it, YAML parser support it in its grammar, in multi-line string it proves more rich shits that libconfig that always need escaping

you can read everything about YAML structure & *grammar* here http://yaml.org/spec/1.2/spec.html
if u understand libconfig... ugh, sorry, wrong approach, **if u understanding the multi-dimensional array**, you can understand json, yaml, php array itself, lua list, and even libconfig. so first u want to talk high about list, comparing with libconfig is wrong, please learn *listing* in general

anacondaqq:
1. Nope, not need x,y,z, and csv in *athena for around 14 years, and never was a problem until someone start to convert 1 to another without any reason, and in result broken several tools which does not work anymore with DB, broken a lot of other things related to 14 years old csv format what was OKAY for more than dozen of years to edit, read, check. And did not require to have x,y,z to work with it. Never.
2. Nope. You do not understand the key problem. Again key problem - spaces/tabs = a super big problem for separation logical structures which can be very easily broken and you will spend a lot of time for finding where it is broken with a YAML supported text editor (code, atom, etc). `{ , ; "" }` or even simple comma makes the whole format much more readable and easy to work and foolproof that you will not do some mistake because of missing tab or space.
3. Nope. I do not compare, I demonstrate. Compassion and demonstration are totally different things. And yaml parser [super bad and problematic too](https://github.com/jbeder/yaml-cpp/issues). What means that the even current yaml format NOT properly implemented and work in the rAthena. That why I talk and will repeat it again and again, that moving to YAML is a mistake which will provoke a lot of problems about doing the same thing again and again while staying on the same place without any progress in real community requested features and things. Running on the spot.

anacondaqq:
@cydh lol, this is so dirty man, You have edited your comment above and changed the sense of it and added links, etc after I answered to the previous sense of the message, and now after editing your comment above you write: 



Oh man, this is the super dirty thing.

cydh:
I edited for typos, grammar, markdown editing, and correction because the clipboard was changed :D

u guys are simple don't want to care that the CSV is such a pain. there too much limitation

- look how ugly we add atk:matk, how stupid we add eLvMin:eLvMax
- when I and lighta add inf3 for skill_db, we had to append all files
- then on skill_tree for job level requirement
- skill_requirement need to be altered to remove hardcoded check between level dependant or multi-requirements, this CSV limits us
- we also need to random option drop for monster db (I know you will cry more for this, just be prepared)
- quest db need to expanded, adding a value in CSV then edit others for empty comma are just same change the while file

oh whether u two understand or not, it means nothing wkwk

src/map/clif.cpp
```diff
@@ -2783,6 +2783,18 @@ void clif_inventorylist(struct map_session_data *sd) {
 			n++;
 		}
 	}
+
+	// Send fake egg data
+	if (battle_config.feature_petevolution && sd->pd) {
+		int pos = ne * se + 4;
```
Lemongrass3110: This is causing overflowed pointer if the player has 100 items in his inventory.

src/map/clif.hpp
```diff
@@ -1113,4 +1123,8 @@ void clif_equipswitch_add( struct map_session_data* sd,uint16 index, uint32 pos,
 void clif_equipswitch_remove( struct map_session_data* sd, uint16 index, uint32 pos, bool failed );
 void clif_equipswitch_reply( struct map_session_data* sd, bool failed );
 
+/// Pet evolution
+void clif_parse_pet_evolution(int fd, struct map_session_data *sd);
```
Lemongrass3110: No need to export the parse function.

src/map/clif.cpp
Daegaladh: Aegis manages pet eggs by reserving a item slot for the egg when the pet is hatched, so you never could have 100 items + a pet

secretdataz: This will most likely need a personalized maximum inventory slot system. Also to prepare for the inventory expansion system kRO has.

Lemongrass3110:
Blocked until #3694 was merged.

Haikenz:
@Lemongrass3110 Something also interfered in the pet attack, when pet starts to attack the server crash and map-server.exe closes alone

attackjom:
I have test in KRO today with capture of Novice Poring in training ground. and see this.
When hatch the egg -> egg disappear.
When return pet to egg -> The egg return to inventory.

Lemongrass3110:
> 
> 
> I have test in KRO today with capture of Novice Poring in training ground. and see this.
> When hatch the egg -> egg disappear.
> When return pet to egg -> The egg return to inventory.

I guess you are referring to the change of behavior of AEGIS I introduced now. This is exactly what AEGIS is doing behind the scenes. It only looks like it is removed from your inventory, because the client is programmed to hide it, when it is set to "broken" and it's type is petegg.

Zarbony:
Does there can be add an Option where everyone can choose by themselfe If aegis behavior is in use or Not ?

I would Preffer the aegis mode

Lemongrass3110:
There will be no option, because we want to get rid of more of those invalid behaviors that drop items if the inventory is full. This never happens on officials as confirmed by @Daegaladh and others as well.
But one step after the other.

Zarbony:
Then another Question, i want to add an Custom Pet with Evolution System.
The Pet is Ingame and the one who is Evo target to.
In Client i add at every File that i found the information for evolution.
But Ingame the Pet hasnt get the Evolution Option such as Mastering and so on.
Whish one is wrong?

I Added the Infos at:
luafiles/datainfo/petinfo.lub
 - Under PetnameTable = {
	[jobtbl.JT_PIKATCHU] = "pikachu",
	[jobtbl.JT_REALPIKATSCHU] = "realpikatschu",
- The Same under PetStringTable 
- under PetEggItemID_PetJobID = {
	[22012] = jobtbl.JT_PIKATCHU,
	[22013] = jobtbl.JT_REALPIKATSCHU,
- and under PetFoodTable = {
	[jobtbl.JT_PIKATCHU] = 22100,
	[jobtbl.JT_REALPIKATSCHU] = 22100,

and then in PetEvolutionCln_sak.lub
function main ()
	InsertEvolutionRecipeLGU(22012, 22013, 22100, 10)
	InsertPetAutoFeeding(22013)

I tested out Whish lub works and this one does, because i changed some Evo Recipes from other Pets whish works

then in my import/pet_db.yml

```
- MobID: 20912
  Name: PIKATCHU
  JName: Pikachu
  TameWithID: 0
  EggItemID: 22012
  EquipItemID: 0
  FoodItemID: 22100
  Fullness: 4
  HungryDelay: 90
  Intimacy:
    Start: 250
    Fed: 50
    Overfed: -100
    Hungry: -20
    OwnerDie: -20
  CaptureRate: 2000
  Speed: 50
  SpecialPerformance: true
  TalkConvertClass: 0
  AttackRate: 250
  RetaliateRate: 4000
  ChangeTargetRate: 8000
  Script: '{ .@i = getpetinfo(PETINFO_INTIMATE); if(.@i >= PET_INTIMATE_LOYAL){ bonus
    bAllStats,20; } }'
  SupportScript: '{ petskillattack "MG_LIGHTNINGBOLT",7,100,0; }'
  Evolution:
    - TargetID: 20913
      ItemsRequirement:
        - ItemID: 22100
          Amount: 10
- MobID: 20913
  Name: REALPIKATSCHU
  JName: Pikachu
  TameWithID: 0
  EggItemID: 22013
  EquipItemID: 0
  FoodItemID: 22100
  Fullness: 4
  HungryDelay: 90
  Intimacy:
    Start: 250
    Fed: 50
    Overfed: -100
    Hungry: -20
    OwnerDie: -20
  CaptureRate: 2000
  Speed: 50
  SpecialPerformance: true
  TalkConvertClass: 0
  AttackRate: 2500
  RetaliateRate: 4000
  ChangeTargetRate: 8000
  Script: '{ .@i = getpetinfo(PETINFO_INTIMATE); if(.@i >= PET_INTIMATE_LOYAL){ bonus
    bAllStats,20; } }'
  SupportScript: '{ petskillattack "TK_TURNKICK",7,100,0; }'

```

Lemongrass3110:
I did not yet try to add a custom pet to evolution, but please ask this kind of questions in the forum or in discord's support channel. Comments here should be development related.

From my experience though you should make sure that you are using the correct file in /System. Maybe try by adding invalid lua code to see if you get an error when starting the client so you can be sure you are editing the correct file.

Haikenz:
with these latest new changes
src\map\status.cpp(2971): error C2039: 'speed': is not a member of 's_pet_db'
`pd->status.speed = pd->get_pet_db()->speed;`

src\map\pet.hpp(26): note: see declaration of 's_pet_db'
`struct s_pet_db {`

src/map/clif.cpp
```diff
@@ -20369,6 +20379,34 @@ void clif_achievement_reward_ack(int fd, unsigned char result, int achievement_i
 	WFIFOSET(fd, packet_len(0xa26));
 }
 
+/// Process the pet evolution request
+/// 09fb <packetType>.W <packetLength>.W <evolutionPetEggITID>.W (CZ_PET_EVOLUTION)
+void clif_parse_pet_evolution( int fd, struct map_session_data *sd ){
+#if PACKETVER > 20141008
```
Atemo: it's not `>=` ?

src/map/pet.cpp
```diff
@@ -31,12 +35,434 @@ using namespace rathena;
...
+				uint32 amount;
+
+				if( !this->asUInt32( requirementNode, "Amount", amount ) ){
+					return 0;
+				}
+
+				if( amount > MAX_AMOUNT ){
+					this->invalidWarning( requirementNode["Amount"], "Amount %u exceeds the maximum of %d.\n", amount, MAX_AMOUNT );
+					return 0;
+				}
```
Atemo: Maybe skip the requirement if amount is 0

db/re/pet_db.yml
```diff
@@ -0,0 +1,1733 @@
...
+#   IntimacyStart            Amount of Intimacy the pet starts with. (Default: 250)
+#   IntimacyFed              Amount of Intimacy that is increased when fed.
+#   IntimacyOverfed          Amount of Intimacy that is increased when over-fed. (Default: -100)
+#   IntimacyHungry           Amount of Intimacy that is increased when the pet is hungry. (Default: -5)
+#   IntimacyOwnerDie         Amount of Intimacy that is increased when the pet owner dies. (Default: -20)
+#   CaptureRate              Capture success rate (10000 = 100%)
+#   SpecialPerformance       If a pet has a Special Performance. (Default: true)
+#   AttackRate               Rate of which the pet will attack (requires at least pet_support_min_friendly intimacy).
+#   RetaliateRate            Rate of which the pet will retaliate when master is being attacked (requires at least pet_support_min_friendly intimacy).
+#   ChangeTargetRate         Rate of which the pet will change its attack target.
```
Atemo: The 3 rates are missing `(10000 = 100%)`

db/re/pet_db.yml
```diff
@@ -0,0 +1,1733 @@
...
+#   AllowAutoFeed            Allows turning automatic pet feeding on. (Default: false)
+#   Script                   Bonus script to execute when the pet is alive. (Default: null)
+#   SupportScript            Legacy pet script for pet* script commands. (Default: null)
```
Atemo: The `SupportScript` is ambiguous (need to give more infos about what it does)

conf/battle/feature.conf
```diff
@@ -71,6 +71,18 @@ feature.achievement: on
...
+// Requires: 2014-10-08aRagexe or later
+feature.petevolution: on
+
+// Automatic Pet Feeding (Note 1)
+// Requires: 2014-10-08aRagexe or later
+feature.petautofeed: on
+
+// At which rate should pet autofeeding trigger? (Note 2)
+// Default: 89
+feature.pet_autofeed_rate: 89
```
Atemo: It should say `At which hunger should pet autofeeding trigger? (Note 2)`, even if the hunger is the rate

src/map/pet.cpp
```diff
@@ -31,12 +35,434 @@ using namespace rathena;
...
+
+	if( this->nodeExists( node, "HungryDelay" ) ){
+		uint32 delay;
+
+		if( !this->asUInt32( node, "HungryDelay", delay ) ){
+			return 0;
+		}
+
+		pet->hungry_delay = delay * 1000;
+	}
```
Atemo: The default `hungry_delay` value is missing

Atemo:
Another 3 points - sorry for the erratic review.
* `@reloadmobdb` doesn't update some of the change of pet_db.yml for the current pet alive. For example the current pet need to be re-hatch (or player relog) to run the bonus "SupportScript" added in pet_db after `@reloadmobdb`

* During a session, when a pet evolve the egg doesn't appear in the inventory at first. It appears as broken when trying to hatch it in the same session (but it hatches). Everything is fine after another "return to egg".
screenrAthena001
screenrAthena003

* idk if it's my client (2015-11-04) the automatic feeding client side doesn't update `sd->pd->pet.autofeed` clif side (clif_parse_configuration never called)

db/pet_db.yml
```diff
@@ -0,0 +1,63 @@
...
+# - Mob                      Monster that can be used as pet
+#   TameItem                 Pet Tame Item.
+#   EggItem                  Pet Egg Item.
+#   EquipItem                Pet Accessory Item.
+#   FoodItem                 Pet Food Item.
+#   Fullness                 The amount of hunger is decreased every [HungryDelay] seconds.
+#   HungryDelay              The amount of time it takes for hunger to decrease after feeding. (Default: 60 seconds)
+#   HungerIncrease           The amount of hunger that is increased every time the pet is fed (Default: 20)
+#   IntimacyStart            Amount of Intimacy the pet starts with. (Default: 250)
+#   IntimacyFed              Amount of Intimacy that is increased when fed.
```
Atemo: Some infos about default value are missing for IntimacyFed and below

Haikenz:
Sem título
@Lemongrass3110 Lasted merge (484897d)

aleos89:
That's because you're missing this at the top of your pre-re/pet_db.yml database:
```yaml
Header:
  Type: PET_DB
  Version: 1
```

zackdreaver:
Just wondering if "Speed" field left empty is intended? or did miss something?
Because pet's movement speed is following its monster counterpart.

I'm not really following up this PR in the past.

Btw, isn't "Tikbalang_Pet" should be "Tikbalang_Egg"?

Sorry, I'm not creating new issue, just want clarification first.

Lemongrass3110:
@zackdreaver
We dropped the field in the new database files and now load it directly from mob db.

zackdreaver:
So, the slower movement speed is intended?

officialwedevgames:
Can't add a new pet egg.

  - Mob: Q_MONKEY
    TameItem: Quest_Golden_Banana
    EggItem: Quest_Fish_Con
    Fullness: 3
    IntimacyFed: 50
    CaptureRate: 10000
    SpecialPerformance: false

100% chance rate but still cant get it.

PR #3548, opened 2018-10-04
Initial Release of Equipment Switch

Lemongrass3110:
* **Addressed Issue(s)**: cant find the correct one - if someone does please link it.

* **Server Mode**: renewal

First implementation of the new kRO feature.

Thanks to everyone who contributed to this release in any way, be it donations, information or testing.


anacondaqq:
Okay, founded problem (I think people will report about it with a time)

It's because of MAX_STORAGE > 850, while almost all uses 900-1000 slots for storage. I think these people will report about the issue. I don't think it's possible somewhat to do with it instead of decreasing max storage size a little bit.

Lemongrass3110:
@anacondaqq yeah totally forgot about this, since it has been this long already. I think the new maximum was somewhere between 820~850 items.

conf/battle/feature.conf
```diff
@@ -67,6 +67,10 @@ feature.roulette: on
 // Requires: 2015-05-13aRagexe or later
 feature.achievement: on
 
+// Equipment Switch (Note 1)
+// Requires: 2017-02-08bRagexeRE or later
+feature.equipswitch: off
```
Atemo: The feature is not available on official server by default?

aleos89: I think this is because of the warning that’s presented at start up since our packetver is 2015~ and it’s causing the build bots to “fail”.

Balferian:
When server start I have some SQL error with cart_inventory.
All work fine, but need to update **upgrade_20181004.sql**
add column to **cart_inventory**: **favorite** and **unique_id**

aleos89:
Sounds like you applied the diff wrong. Cart Inventory doesn't have a favorite tab so that column is not needed in SQL. Look at the diff again and make sure it only applies to `TABLE_INVENTORY` when parsing char-server side.

Questune09:
why i dont have this?

image


using latest rathena rev

src/map/clif_packetdb.hpp
```diff
@@ -2327,18 +2327,22 @@
...
 // 2016-10-26bRagexeRE
 #if PACKETVER >= 20161026
 	packet(0x0AA5,-1);
 #endif
 
+// 2017-02-08bRagexeRE
+#if PACKETVER >= 20170208
+	parseable_packet(0x0A97,8,clif_parse_equipswitch_add,2,4);
+	packet(0x0A98,10);
+	parseable_packet(0x0A99,4,clif_parse_equipswitch_remove,2,4);
```
MathReaper: Trying to unequip boots:
`[Warning] clif_parse: Received unsupported packet (packet 0x0040, 4 bytes received), disconnecting session #3.`

0x0A99 have size 8 on client 20170215.

BruSantato:
I added the correct Diff all in a clean rAthena, when connecting the emulator does not have any error, but it is not changing the equipments when clicking on CHANGE
I use a client: 20180530

image

anacondaqq:
@BruSantato check feature.conf feature.equip_switch must be on, by default it's off.

BruSantato:
@anacondaqq Sorry, I forgot to mention this, it's already activated, and even then it does not make the exchange of equipment

MathReaper:
2H weapons actually conflicts with 1H weapons or shields (switching 2H weapon for weapon + shield) and the 1H weapon is removed from equipment switch interface.

BruSantato:
> I added the correct Diff all in a clean rAthena, when connecting the emulator does not have any error, but it is not changing the equipments when clicking on CHANGE
> I use a client: 20180530
> 
> image

?

src/map/clif_packetdb.hpp
Lemongrass3110: Sorry I do not have that client to confirm this. :(

BruSantato:
@MathReaper Sorry, I did not know they were conflicting: /

src/map/clif_packetdb.hpp
MathReaper: I can upload here. I will open a repo of clients unpacked by myself soon.
[2017-02-15aRagexeRE](https://www.mediafire.com/file/9sfvlf8u02u0hrl/) - MediaFire

MathReaper:
@BruSantato I'm just reporting an issue, this isn't related to what you're reporting. xD

Ryuhorox:
sem titulo

help me .. I applied the diff but of these errors

MathReaper:
@ecdarreola I think that this happens because `pc_equipswitch` function is called two times by the skill because the equipment is in another position too, and the weapon is switched 2 times?

Lemongrass3110:
@MathReaper & @ecdarreola the problems should be fixed now.

src/map/clif_packetdb.hpp
Lemongrass3110: This has been fixed as well. :)

ecdarreola:
@Lemongrass3110 : The equipment switch does't work now on the latest revision. No errors on the console though.

Lemongrass3110:
@ecdarreola which packet version?

ecdarreola:
@Lemongrass3110 : The packet version is 20180621 or 20180621aRagexe client.

ecdarreola:
Tested right now 20180620RagexeRE equip switch doesn't work anymore. No errors in console.

mjonrest:
i found this error. ( it must be something happen with my src file.. but if someone can tell me what should i do for this, it will be great.. thanks )

error_equip_switch

Lemongrass3110:
@ecdarreola it still works fine for me. It has to be some issue on your end. Make sure you test this branch in a clean state to make sure you do not apply anything wrong.

ecdarreola:
@Lemongrass3110 : Still the same. I have already used a fresh copy of rA.

https://youtu.be/qSA4sTLDLdc uploaded a video of the behavior in-game.

Things done:
1. Install a fresh rathena
2. Pull equip switch using git pull
3. Compile using 20180621

zellkennedy:
Test:

I have the problem that after changing one time the last items are not more change able without drag this again. Is this the default behavior?

https://vimeo.com/297988385

zellkennedy:
> @zellkennedy I think that this problem was already reported, it's because the two-handed weapons. ;)

Hmm, @Lemongrass3110  sayed that was fixed but still, I can reproduce it and ecdarreola problem is other.

mjonrest:
error
i found this error.. this is cause by monster that spamming waterball like drake,high wizard katrinn,strouf,etc

BruSantato:
Hello, I just applied the system in a clean rathena, the items are added to the replace slots but I can not change the items, the system is active in features.conf and the sql table is imported
I'm using a client 2018-06-20 and the system does not work
teste replace equipament

BruSantato:
?

RadianFord:
I've tested this one and the only problem was.

```
1. Place main gauche in the replacement equipment.
2. Equip infiltrator in the normal equipment window.
3. Switch/Change equipment's.
4. Result: dagger switched perfectly.
5. Switch it again into the infiltrator weapon.
6. Result: Dagger will go back to normal state.
```

Mateuus:
For those who are in trouble not working with 2018-06-20
Comment this part

clif.cpp:
```suggestion
/*
	if( DIFF_TICK(tick, sd->equipswitch_tick) < 0 ) {
	// Client will not let you send a request
		return;
	}
 	sd->equipswitch_tick = tick + skill_get_cooldown( skill_id, skill_lv );
 	*/
```

RadianFord:
I noticed somethings.

1. If mapflag noskill are set, of course this wont work since its a skill.
Suggestion: add map_msg `The Switch Feature wont work in this map.`

2. If your character is sitting, this wont work either.
Suggestion: add map_msg `The Switch Feature wont work while in sitting mode.`

Lemongrass3110:
@RadianFord it is actually easier to reproduce.
Dagger on main window -> Katar on switch window -> switch

Lemongrass3110:
So guys everything should be fixed now.
Please test it again and if no further reports come in, we can merge this to master after review.

RadianFord:
@Lemongrass3110 Yeah, i just found out that tooday..

PR #3549, opened 2018-10-07
BlackSmith quest skills

Balferian:
* **Addressed Issue(s)**: 

* **Server Mode**: Pre-re/Re

* **Description of Pull Request**: After reborn players that have quest skills before cant learn it after reborn or if player use [Neuralizer](https://www.divine-pride.net/database/item/12213)/reset skills.

### DIFF
--- a/npc/quests/skills/blacksmith_skills.txt
+++ b/npc/quests/skills/blacksmith_skills.txt
@@ -15,7 +15,7 @@
 
 geffen,178,72,3	script	Akkie#qsk_bs	726,{
 	if (BaseJob == Job_Blacksmith) {
-		if (BLACK_SK == 7 && getskilllv("BS_UNFAIRLYTRICK") == 0) {
+		if (BLACK_SK >= 7 && Upper == 1 && getskilllv("BS_UNFAIRLYTRICK") == 0) {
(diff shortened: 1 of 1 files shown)


npc/quests/skills/blacksmith_skills.txt
```diff
@@ -15,7 +15,7 @@
 
 geffen,178,72,3	script	Akkie#qsk_bs	726,{
 	if (BaseJob == Job_Blacksmith) {
-		if (BLACK_SK == 7 && getskilllv("BS_UNFAIRLYTRICK") == 0) {
+		if (BLACK_SK >= 7 && Upper == 1 && getskilllv("BS_UNFAIRLYTRICK") == 0) {
```
Atemo: On aegis
- `BLACK_SK` 7 and 8, `brack_greed` 2 and 3 can only be reachable by rebirth and transcendant 3rd class
- neuralizer doesn't reset quests skills

=
- add a condition for theses variable to only be reachable by rebirth and transcendant 3rd class
- I would prefer to simply check BLACK_SK case 7 for non rebirth and case 8 for transcendant if the skill is learned and give it accordly

Issue #3540, opened 2018-10-02
IM_NONE idle timer


Atemo:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/daebd30049993e68404f41391281482847ce42c4

* **Client Date**: -

* **Server Mode**: -

* **Description of Issue**: 
  * Result: Idle timer doesn't re-start when all players leave the instance map (IM_NONE mode)
  * Expected Result: 
  * How to Reproduce: 

Test sample:
instance_db.txt
```
100,abababababab,14400,1,prontera,150,150
```
NPC
```
izlude,125,140,0	script	abababababab#A	56,{
	.@instance_id = instance_create( "abababababab", IM_NONE );
	instance_enter( "abababababab", 0, 0, getcharid(0), .@instance_id );
	end;
}
```

Additionnal check should be added for instance type none (and clan?) in `pc_setpos`
https://github.com/rathena/rathena/blob/daebd30049993e68404f41391281482847ce42c4/src/map/pc.cpp#L5497
  * Official Information:
  

* **Modifications that may affect results**:

PR #3550, opened 2018-10-08
Fix for issue #3540 "IM_NONE idle timer"

Atemo:
* **Addressed Issue(s)**: #3540

* **Server Mode**: -

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -5494,21 +5494,21 @@ enum e_setpos pc_setpos(struct map_session_data* sd, unsigned short mapindex, in
 	if( sd->state.changemap ) { // Misc map-changing settings
 		int i;
 
-		if(map_getmapdata(sd->bl.m)->instance_id && !mapdata->instance_id) {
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -5494,21 +5494,21 @@ enum e_setpos pc_setpos(struct map_session_data* sd, unsigned short mapindex, in
 	if( sd->state.changemap ) { // Misc map-changing settings
 		int i;
 
-		if(map_getmapdata(sd->bl.m)->instance_id && !mapdata->instance_id) {
-			bool instance_found = false;
+		if (map_getmapdata(sd->bl.m)->instance_id && map_getmapdata(sd->bl.m)->instance_id != mapdata->instance_id) {
```
aleos89: You could store this value like: `if ((instance_id = map_getmapdata(sd->bl.m)->instance_id) && instance_id != mapdata->instance_id)`.

Then `instance_id` could be used in the checks below as well.

PR #3556, opened 2018-10-10
Updated Firewall cast check

aleos89:
* **Addressed Issue(s)**: N/A

  * Firewall should check at cast begin if unit max count has been reached.
  * Removed redundancy with the unit max count checks by merging into a single function.
Thanks to @mrjnumber1!


ecdarreola:
This PR creates a crash log. But doesn't have a full BT of it. (See snipit below)

image

ecdarreola:
Attaching the full bt log of the crash. 
[bt full of crash log.txt](https://github.com/rathena/rathena/files/2595842/bt.full.of.crash.log.txt)

PR #3557, opened 2018-10-10
Proposal to change the default packet version

Lemongrass3110:

* **Server Mode**: Both

We support all required features or have them ready in some pull requests.
The packet version after this will require a huge overhaul since item id sizes have been changed in all packets.

### DIFF
--- a/conf/battle/feature.conf
+++ b/conf/battle/feature.conf
@@ -69,12 +69,12 @@ feature.achievement: on
 
 // Homunculues Autofeeding (Note 1)
 // Requires: 2017-09-20bRagexeRE or later
-feature.homunculus_autofeed: off
+feature.homunculus_autofeed: on
 
 // Attendance System (Note 1)
 // Requires: 2018-03-07bRagexeRE or later
-feature.attendance: off
+feature.attendance: on
 
 // Private Airship System (Note 1)
 // Requires: 2018-03-21aRagexeRE or later
-feature.privateairship: off
+feature.privateairship: on
--- a/src/config/packets.hpp
+++ b/src/config/packets.hpp
@@ -13,7 +13,7 @@
 	/// Do NOT edit this line! To set your client version, please do this instead:
 	/// In Windows: Add this line in your src\custom\defines_pre.hpp file: #define PACKETVER YYYYMMDD
 	/// In Linux: The same as above or run the following command: ./configure --enable-packetver=YYYYMMDD
-	#define PACKETVER 20151104
+	#define PACKETVER 20180620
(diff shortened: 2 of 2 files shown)


anacondaqq:
Strongly against this (2018) Client version is SUPER bad and weak. You will have hell a lot of complaints.
Please leave 20170614 exe as default one, because it is MUCH stable. I telling you as a guy which supported hell a lot of people with this game (2017) client and have a lot of troubles with 2018 game clients.
2018 clients are not production ready at all, they need some time to make them good,  there is hell a lot of problems with 2018 clients mostly because of not supported features, because of removed content (AI support is a mega critical problem for RE and PRE servers), etc. Also, there is no stable client-base for these clients, so this packetver will be useless. 20150411 pretty stable and not bad game client (yes it has some bugs and problems). But better to pick something really stable, and as for me, something really stable is 2016-2017 game clients.

Lemongrass3110:
@anacondaqq I regret having to tell you that rAthena is a server software and has nothing to do with the client side. If you do not want to use the default settings you are free to override/change them. But I have not found a lot of client diffs that are still broken.

admkakaroto:
Does it make any difference to use the version defined in the emulator after this one?

`#elif PACKETVER > 20180307`

after this date all the clients use the same packages, so I think if it is to change it should be for this date.

Lemongrass3110:
@admkakaroto private airship is 03-21 for example.

MathReaper:
20180620 isn't too far?

secretdataz:
@MathReaper Any reason why you consider it "too far"?

anacondaqq:
@secretdataz go ahead, at least you (not personally you, but dev team) will face tons of negative here at on board too because it's too early add as default packetver a client which does not work well at all for now.

Lemongrass3110:
@anacondaqq it is okay. I will take the blame. This way we will get feedback for the new clients at least. Sorry to say this, but we need to stay up to date and not intentionally stay behind for 3 years just because some features were dropped that some pre-renewal servers still want. That might be hard to accept, but as long as there is no open source client we will always have to accept the changes Gravity does to the client.

zackdreaver:
Well I don't mind really if you guys changed the default packetserver to this,
Anything for the greater good of community, I'll try to support it.
If you guys don't mind, can you guys make or update the announcement regarding this?
Either minimum version or where to download the client?
https://rathena.org/board/topic/86650-officially-supported-client-versions/

p.s: The new Archbishop skills now can't be used normally if you use older client without hack,
like 2015-11-04aRagexe.

PR #3605, opened 2018-10-21
Fixed cashshop NPC exploit

secretdataz:


anacondaqq:
LOOOOL, thx for the fix, nice catch. I think affected server without packet encryption.

anacondaqq:
LOOOOL, thx for the fix, nice catch. I think affected server without packet encryption.

anacondaqq:
LOOOOL, thx for the fix, nice catch. I think affected server without packet encryption.

anacondaqq:
LOOOOL, thx for the fix, nice catch. I think affected server without packet encryption.

Lemongrass3110:
> 
> 
> How can i try this one?

@RadianFord might be better if we do not expose how it can be done here in public. Basically this issue only affects the old cash shop not the new one(button).

Issue #3626, opened 2018-10-28
There is no protection of players' personal data!!


mrjnumber1:
* **rAthena Hash**: current
* **Client Date**: x
* **Server Mode**: x
* **Description of Issue**: 
  * Result: As it is now, players' passwords are stored as md5 hashes, or worse, plain text. This is quite an ugly thing, unfortunately.
  * Expected Result: It is irresponsible to continue with it as is. Using a slow or iterative COULD be an issue even though the login server itself doesn't do much. Something as simple as what's suggested on https://crackstation.net/hashing-security.htm would be good enough. In my opinion, pincodes and email addresses should be have an option to be stored behind a hash as well. i understand why emails may not need to be, as it's quite common for servers' dbs to leak and suddenly peoples' emails get farmed by scammers.
  * How to Reproduce: x
  * Official Information: x
* **Modifications that may affect results**: x

vykimo:
I agree with you, at least propose different protection for passwords. And propose mail/pin protections too.

Lemongrass3110:
@mrjnumber1 I totally share your concerns regarding all of this. But this would mean we have to drop some packet support for old (still used) login packets. And I guess I do not have to tell you that MD5 is a worthless approach.

Since server owners hate to use launchers that is a bit of a problem and launchers would only solve the login issue. I have not checked if it would be possible for pin code or mail yet.

vykimo:
At least stockage in SQL should be encrypted for mails and pincodes.

mrjnumber1:
we don't have to drop support for those old packets. we can still take the plain text passwords in, and then hash+salt those for storage in the db. of course, it's highly recommended to put a big annoying warning message about using plaintext passwords, since anyone who's on wireless will have their details easily sniffable (although it's highly unlikely that someone will do this for a ro server lol).
in fact, we may as well do the same even for the md5 sending login packets, too, unfortunately.

Everade:
Would be great to put our focus a little bit more towards players data privacy and security.
I doubt that players expect even private servers to run with plaintext or badly hashed passwords these days.

@gepard-me brought this up **almost 6 years ago** on hercules and went into some details on handling options. It might help.
https://github.com/HerculesWS/Hercules/issues/74

PR #3639, opened 2018-10-30
MD5 emails protection

vykimo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/3626
* **Server Mode**: BOTH
* **Description of Pull Request**: Protect emails stored in DB by allowing administrator to store emails in MD5 or not.
It's a partial answer to the issue concerning user personal data protection (next step is pincode).
_TODO :_ In case of old clients which ask for mail to delete character, add md5 support.

### DIFF
--- a/conf/login_athena.conf
+++ b/conf/login_athena.conf
@@ -106,9 +106,10 @@ vip_char_increase: -1
 // 0 or more: new accounts automatically expire after the given value, in seconds
 start_limited_time: -1
 
-// Store passwords as MD5 hashes instead of plain text?
-// NOTE: Will not work with clients that use <passwordencrypt>
+// Store passwords/emails as MD5 hashes instead of plain text?
+// NOTE: use_MD5_passwords will not work with clients that use <passwordencrypt>
 use_MD5_passwords: no
+use_MD5_emails: no
 
 // Ipban features
 ipban_enable: yes
--- a/src/char/char_clif.cpp
+++ b/src/char/char_clif.cpp
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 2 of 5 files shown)


Lemongrass3110:
Personally I am a fan of data protection, but this opens another problem.
How will you be able to reset your password now? Sending out a link becomes impossible like this.
Also the administrator loses any means of contacting the player outside of the game.

Also do not waste your time on pin code please. Hashing a 4 character string will make hash crackers laugh... Even if you salt it and store the in another column... :-)

vykimo:
@Lemongrass3110 I agree with you, I think for pincode it's useless... But for email it is not... If someone loose password, player will click "forgot password", and his nickname + mail will be asked for recovery.

For admin point of contact.... That's the point, following the law (eu : GDPR), admin shouldn't be able to use mail to send promo/newsletter without player conscent. AND I've almost never seen mention of this in player registration form. Currently mail is only used as a security for delete character function.
Conversely, forum registration forms are mentionning this contact point. Most of the time as an extra box to check for newsletter, and using forum mail DB is more legal.

I don't know if you understand my purpose...

secretdataz:
rAthena does have the functionality to send an email. So, technically, this concern is not valid in rAthena's context. The appropriate place would be FluxCP issue tracker. Also, hashing the email doesn't really make sense. It is not our duty to prevent any server admin from abusing the email they collected in rAthena's database.

And another thing, MD5 in 2019.

vykimo:
My opinion md5 better than nothing.
Sure its not enough, but even for passwords

Lemongrass3110:
Sorry but we have to reject this for the time being, until we address #3626 seriously.

PR #3640, opened 2018-10-30
Implemented Blocking Play feature

aleos89:
* **Addressed Issue(s)**: N/A

  * Adds the SC_BLOCKING_PLAY status that keeps players from moving, using/picking/dropping items, using skills, chatting, and swapping equipment while changing maps.
  * Disables the ability for players to pre-cast skills when changing maps to avoid cast times.
Thanks to @mrjnumber1!

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -18590,11 +18590,10 @@ void clif_parse_reqworldinfo(int fd,struct map_session_data *sd) {
 		clif_ackworldinfo(sd);
 }
 
-/// unknown usage (CZ_BLOCKING_PLAY_CANCEL)
-/// 0447
+/// Enable playing abilities of a player if map is done loading.
+/// 0447 (CZ_BLOCKING_PLAY_CANCEL)
 void clif_parse_blocking_playcancel(int fd,struct map_session_data *sd) {
-	//if(sd)
-	;
+	status_change_end(&sd->bl, SC_BLOCKING_PLAY, INVALID_TIMER);
 }
 
 /// req world info (CZ_CLIENT_VERSION)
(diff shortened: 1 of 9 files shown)


luan122:
nice SC is it to be used with instances to avoid the blank mes?

aleos89:
It's meant to keep players from pre-casting skills before warping to another map. This PR isn't fully complete yet as there are some extra checks that will need to be added.

mrjnumber1:
it's worth noting that the portal casting that was available was a bug on aegis that athena never had. however, it's possible it has other implications, like a quick simple anti-bot checking, or making it so the player doesn't move or do anything between map changes when everything didn't load yet to prevent bugging out. 

at the very least, it could be completely useless for us, but it's to match aegis better, and COULD be important, after all!

src/map/clif.cpp
```diff
@@ -18591,11 +18591,10 @@ void clif_parse_reqworldinfo(int fd,struct map_session_data *sd) {
...
 }
 
-/// unknown usage (CZ_BLOCKING_PLAY_CANCEL)
-/// 0447
+/// Enable playing abilities of a player if map is done loading.
+/// 0447 (CZ_BLOCKING_PLAY_CANCEL)
 void clif_parse_blocking_playcancel(int fd,struct map_session_data *sd) {
-	//if(sd)
-	;
+	status_change_end(&sd->bl, SC_BLOCKING_PLAY, INVALID_TIMER);
```
Lemongrass3110: Should we check sd warping state here?
So that we ignore faked packets from bots?

Badarosk0:
@aleos89 Could you update this PR with the current version of rAthena?
Thank you

Issue #3612, opened 2018-10-24
The mess that is countitem.


kisuka:
So ummm... not to be a dick but whose idea was this? Combining literally every item counting related script command into one  function. This looks and feels incredibly messy.

I understand not wanting to have a lot of duplicate code lying around, but this seems particularly not the way to solve that issue. You're literally tying 10 different script commands into one function and determining their functionality based off the start of the command's string and then doing a bunch of if-else statements. I feel like this completely goes against an idea of clean coding standards.

Also goes against the principals of separation of concerns. Instead of one function doing one thing you now have one function doing 10 things. I'd highly recommend breaking these out into their own functions.

If you disagree with this I'd love to hear your opinion as to why.

https://github.com/rathena/rathena/blob/fe197bfa120aef5fd31c6896122b35fdc06774b4/src/map/script.cpp#L6726

PR #3657, opened 2018-11-05
Cleaned up countitem-type script commands

aleos89:
* **Addressed Issue(s)**: #3612

  * Separated the countitem-type script commands into their own functions.
  * Created a global count function to be used between them.
Thanks to @kisuka's suggestion!


RadianFord:
Is this good to go?

aleos89:
Should be, yes.

lighta:
Will there be a depreciation notice about the old countitem for the script using them ?

aleos89:
The commands aren't deprecated, just move into their own functions.

src/map/script.cpp
```diff
@@ -6870,39 +6795,229 @@ BUILDIN_FUNC(countitem)
...
+		script_pushint(st, -1);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (script_isstring(st, 2)) // item name
+		id = itemdb_searchname(script_getstr(st, 2));
+	else // item id
+		id = itemdb_exists(script_getnum(st, 2));
+
+	if (id == NULL) {
```
Normynator: redundant nullptr check

src/map/script.cpp
```diff
@@ -6870,39 +6795,229 @@ BUILDIN_FUNC(countitem)
...
+		if (!script_rid2sd(sd))
+			return SCRIPT_CMD_FAILURE;
+	}
+
+	if (script_isstring(st, 2)) // item name
+		id = itemdb_searchname(script_getstr(st, 2));
+	else // item id
+		id = itemdb_exists(script_getnum(st, 2));
+
+	if (id == NULL) {
```
Normynator: redundant nullptr check

src/map/script.cpp
```diff
@@ -6870,39 +6795,229 @@ BUILDIN_FUNC(countitem)
...
+		if (!script_rid2sd(sd))
+			return SCRIPT_CMD_FAILURE;
 	}
 
+	if (script_isstring(st, 2)) // item name
+		id = itemdb_searchname(script_getstr(st, 2));
+	else // item id
+		id = itemdb_exists(script_getnum(st, 2));
+
+	if (id == NULL) {
```
Normynator: redundant nullptr check

Issue #3649, opened 2018-11-03
Electric Shocker + MVP's


admkakaroto:
* **rAthena Hash**: Any
* **Client Date**: Any
* **Server Mode**: RE
* **Description of Issue**: The Electric Shocker trap is not arresting, according to boss-type monsters.

PR #3660, opened 2018-11-05
Corrected Electric Shocker versus MVP

aleos89:
* **Addressed Issue(s)**: #3649

  * Electric Shocker is able to stop MVP as well.
  * Status should not stack with itself.
  * Status should be cleared by normal means.
Thanks to @admkakaroto!


slyx88:
MVP and Boss protocol monsters still able to move after this commit

admkakaroto:
this not work ^^

slyx88:
Thanks.It works as intended.

PR #3661, opened 2018-11-05
Corrected Enchant Arms status icon

aleos89:
* **Addressed Issue(s)**: N/A

  * Enchant Arms should display a status icon depending on the element (skill level) being used.
  * EFST_WEAPONPROPERTY makes it change the EFST to start from EFST_ATTACK_PROPERTY_NOTHING + val1.
Thanks to @mrjnumber1!


slyx88:
All the 4 converters (fire,water,wind & earth) and aspersio converter and work just fine. They override the previous endow but it fails to override those elements if you use cursed water

aleos89:
@slyx88 Could you explain a little better? I don't think I was able to reproduce this. I spawned an Increase Soil, noted my damage, used a Fire Scroll, noted the increased damage, use a Cursed Water and my damage went down to what it was prior to using the Fire Scroll.

Issue #3102, opened 2018-05-06
disablenpc with chat room


sader1992:
* **rAthena Hash**: 8d638b9f726e3863be5109c6075c9a8458f326f2

* **Client Date**: 2015-11-04

* **Server Mode**: both

* **Description of Issue**: 

  * Result: the npc not disabled for the players not in it's map 
  * Expected Result: the npc should be disabled for for everyone
  * How to Reproduce: 
```
prontera,152,187,4	script	disablenpc_test	444,{
	disablenpc strnpcinfo(3);
end;
OnInit:
	waitingroom "chat room", 1;
}
```
go to a map other map >> hide the npc using `disablenpc` >> back to where the npc is
the npc is disabled as expected for the players in it's map
the players that were not in it's map when using the command still can see the npc

  * Official Information: no information

* **Modifications that may affect results**: 
non

however we can get around this issue by adding `delwaitingroom strnpcinfo(3);` before using `disablenpc strnpcinfo(3);`

Issue #3645, opened 2018-10-31
Crash client (disablenpc()/enablenpc() + delay timer)


Badarosk0:
* **rAthena Hash**: 

https://github.com/rathena/rathena/commit/aa275c57c939ef94d83dc3d2b1cbf34d59115d96

* **Client Date**: 

20180621

* **Server Mode**: 
Renewal

* **Description of Issue**: 
  * Result: 
If you add any delay between disablenpc and enablenpc there is a chance of random crash client generation.

**Example 01 reproduce:**
```
prontera,106,242,3	script	Teste bau BUG#1	4_TREASURE_BOX,{
specialeffect EF_COIN;
disablenpc( strnpcinfo(0) );
sleep 100;
enablenpc( strnpcinfo(0) );
end;
}

```
**Example 02 reproduce:**
```
prontera,106,242,3	script	Teste bau BUG#1	4_TREASURE_BOX,2,2,{
OnTouch:
specialeffect EF_COIN;
disablenpc( strnpcinfo(0) );
end;
}
-	script	#timeteste	-1,{
end;

	OnTimer100:
		stopnpctimer;
		setnpctimer 0;
		startnpctimer;

			enablenpc "Teste bau BUG#1";

	end;

OnInit:
initnpctimer;
end;
}
```

**In this example below the crash client does not occur:**
```
prontera,106,242,3	script	Teste bau BUG#1	4_TREASURE_BOX,{
specialeffect EF_COIN;
disablenpc( strnpcinfo(0) );
//sleep 100;
enablenpc( strnpcinfo(0) );
end;
}
```

  * Expected Result: 
Do not generate crash client when clicking NPC in disablenpc + delay script.

vykimo:
Not confirmed with latest rAthena and 20180620.

zellkennedy:
Can't reprocue in 20180620 too.

Badarosk0:
It is not a simple problem to reproduce but it does exist. It seems to me that it is linked to the disable script effect along with the NPC hide. This has existed for some years now. For those of you who use rAthena as an emulator you should have noticed a report of players reporting that the game closed in the middle of some instance for example. It is not constant but there is.
Try to use the script I provided above and try to play as if you were a player. Click the chests repeatedly until the problem occurs.

vykimo:
If its such a random problem. It should exist on official server too and maybe its clientside more than server side...

cydh:
^depends, if the server sends wrong packet -> client crash

[...]

PR #3692, opened 2018-11-14
Bugfix disable npc

vykimo:
Bugfix disable npc

* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/3102

* **Server Mode**: BOTH

* **Description of Pull Request**: Fix the bug addressed in issue 3102. Normal behavior is that maya purple can't see NPCs disabled but only hidden npc. 
OPTION_INVISIBLE is associated to disabled npc (OPTION_HIDE for hidden npc), so invisible NPCs must be skipped when changing map/refreshing even if they have a chat room. 
With this modification, hidden npc with a chatroom will still display it (for Roleplay purpose and players with maya purple card will see black npc form ) but disabled npc will not (and packet won't be send).

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -4619,7 +4619,7 @@ void clif_getareachar_unit(struct map_session_data* sd,struct block_list *bl)
 	/**
 	* Hide NPC from maya purple card.
 	**/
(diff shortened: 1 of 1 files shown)


Badarosk0:
This solves the crash client crash problem by clicking on NPCs as an instance chest.

https://github.com/rathena/rathena/issues/3645

aleos89:
Taken over in #5269. Thank you for the contribution!

PR #3698, opened 2018-11-23
Changed skip into error

Normynator:
* **Addressed Issue(s)**: 
if common fails target map/char/login will be disabled.

* **Server Mode**: 
Both

if common fails stop cmake execution and throw an error, this will avoid building only mapcache and common_base.


Normynator:
@Lemongrass3110

PR #3714, opened 2018-11-28
Corrected possible item loss during char login

aleos89:
* **Addressed Issue(s)**: N/A

  * Resolved a rare case if a player logs in and the server hasn't fully initialized all of the player's items that the auto-save timer fires off and saves the player with no items.
Thanks to @Tokeiburu!


anacondaqq:
Hmm, I can confirm this, but I thought this is "default behavior" of how the thing is work. Anyway, thank you for the fix.

aleos89:
It can only happen on high load which is why I would have never noticed it during local testing.

anacondaqq:
To be clear I noticed it even on single char in the game while doing some stuff on windows/linux too, but after fast crash/restart

Issue #3490, opened 2018-09-14
Rental Item Restriction


cydh:
* **rAthena Hash**: ca83d35
* **Client Date**: -
* **Server Mode**: Both
* **Description of Issue**: 
  * Result: Rental items (item with expire_time) are simply permitted from any *move* actions (everything that defined in item_trade)
  * Expected Result: Just like other items, proper rental items are separated from normal item and listed in item_trade db.
  * How to Reproduce: 
  * Official Information: 
      * That's why Aegis have many *same* items with different ID & 'name'
      * ^and most of those items are 0 weight
      * ^also have 0 sell price (in common, or at least the items are give small amount of zeny than the 'real' item)
      * ^as example fro **Deviruchi Cap**, this is [the rental item](https://www.divine-pride.net/database/item/5281/), and this is [the non rental](https://www.divine-pride.net/database/item/5038/).

* **Modifications that may affect results**: -

PR #3749, opened 2018-12-12
Changed item with expire_time sell value

Atemo:
* **Addressed Issue(s)**: #3490

* **Server Mode**: 

Rental item sell value is supposed to be 0 in item_db. This is an extra check.


Lemongrass3110:
This partially addresses #3490.
Maybe we can check if we can fix what cydh mentioned with this pull request or with #3738?

cydh:
you shouldn't set this like this, if u want, we can make it as config.
rentals items have its own value defined in db, there's no 'force to value' (all rental items have 0 price)

cydh:
1. default to `no`: use price from item_db sell price
2. as lemongrass mentioned, if set to `yes` also need the check on https://github.com/rathena/rathena/blob/ce1508a01c68dee80544242b470987742f3b3ece/src/map/npc.cpp#L2105-L2107
```cpp
		if (battle_config.rental_item_novalue)
			continue;
		value = pc_modifysellvalue(sd, sd->inventory_data[idx]->value_sell);

		z+= (double)value*amount;
```

Issue #1661, opened 2016-10-27
Bug Instances


admkakaroto:
Is there a way of abusing some instances that need to be rectified.

A player can leave a group and enter the instance with another character while the instance is still generated, so if a player up the endless tower, after descending the group leader to stay up there, the player who left can get another character, the leader invites him to the group, and this guest player climbs the tower with the other character ... I believe this should not happen.

Jeybla:
Related: 109b30dd5184e06806253e692fa0d893b24282e5

Why do you call this abuse if its simply a character joining another et group? This should be okay since you don't get any more MvPs.

From http://irowiki.org/wiki/Endless_Tower:

> Warning: If any party member leaves their party while inside the dungeon, they will be kicked out and sent back to their save point! However, they can be reinvited to the party and reenter the tower.
> [...]
> There is no limit on how many instances a party can create within the 4 hour limit. A new instance can be created under a different party leader if the previous instance was destroyed. Once the time limit is up, there is a 6 day 20 hour cooldown period for all party members in which they cannot enter a new Endless Tower instance. **However, additional instances will not allow access to additional MVPs.** 

But currently it's possible to kill many MvPs by just staying in the party and reopen the instance with a dummy party leader.

Quickfix: Disabling re-entering the instance completely.

Official: Extend the fix from @Atemo : Killed MvPs should be safed to a "party-attached" variable so they won't spawn by reopening the instance.

cydh:
Well, it's official way to abuse the official server :D
but for PS owner, maybe will find this thing is annoying, you can add permanent char variable for quest delay when player enter the instance, and check it again before player re-enter the instance.

Atemo:
The previous update was based on iro. However a player should get 2 ashes at the start and use 1 ashe to re-enter.

admkakaroto:
regardless of the official, I think half wrong can abuse the ability to re-enter

Lemongrass3110:
Officials solved this problem in 2017. Now you cannot leave, invite or expel from parties while an instance is active. I will be creating a pull request for this soon.

PR #3753, opened 2018-12-15
Added configs for instance limitations

Lemongrass3110:
* **Addressed Issue(s)**: #1661

* **Server Mode**: Both

Adds configurations to prevent party, guild and clan modifications while an instance is running.
On officials party modifications are blocked now.

Thanks to @admkakaroto

Still misses some messages for custom scenarios(guilds and clans). Maybe @aleos89 can help me out on this.


Badarosk0:
Could you upgrade to the new review of rAthena for testing? @Lemongrass3110

Badarosk0:
It appears that it is not yet compatible with the current version of rAthena.
```
bool guild_has_permission( struct map_session_data* sd, enum e_guild_permission permission ){
	nullpo_retr( false, sd );

	**_int position = guild_getposition( sd );_**

	if( position < 0 ){
		return false;
	}

	return ( sd->guild->position[position].mode & permission ) != 0;
}
```

Badarosk0:
Currently I can still use the /invite command even with the generated instance.
@Lemongrass3110

Badarosk0:
Could you update this PR with the current version of rAthena?
@Lemongrass3110

sader1992:
it would be nice if the block can be set via command script.
for example
`party_block(<party_id>,<true|false>)`
same for guild/clan
we can for the instance call the party block if the config is set
we also can use this system for things other then instances, where players loss the direct control on party/guild/clan, the party/guild/clan can only be edited by command scripts while the block is applied
also it might be a good idea to export it to sql , so incase of server reload (optional maybe?)
this will give the server owner the ultimate control over those things if he Choose to use it for other systems.

I wanted to make this kind of system however this PR is already exist , it would be much more efficient to apply it to this directly.
if not I would create it and open a new PR for review after this is merged.

src/map/clif.cpp
```diff
@@ -13069,93 +13065,54 @@ void clif_parse_CreateParty2(int fd, struct map_session_data *sd){
...
-	name[NAME_LENGTH-1] = '\0';
-
-	if(map_getmapflag(sd->bl.m, MF_PARTYLOCK)) {// Party locked.
-		clif_displaymessage(fd, msg_txt(sd,227));
-		return;
-	}
-
-	t_sd = map_nick2sd(name,false);
+void clif_parse_PartyInvite2( int fd, struct map_session_data *sd ){
+	char name[NAME_LENGTH];
```
lighta, line 13077: = {0}
It bug me so much all those uninatialized variable.

src/map/guild.cpp
```diff
@@ -695,103 +695,150 @@ int guild_recv_info(struct guild *sg) {
...
-	if( sd->guild_invite != guild_id )
-		return 0; // mismatch
+	if( sd->guild_invite != guild_id ){
+		return false; // mismatch
+	}
 
 	// look up the person who sent the invite
 	//NOTE: this can be NULL because the person might have logged off in the meantime
-	tsd = map_id2sd(sd->guild_invite_account);
+	struct map_session_data *tsd = map_id2sd(sd->guild_invite_account);
```
lighta, line 786: can this lookup be moved after "already in another guild" check ?
Since the another guild is not dependant.

src/map/guild.cpp
```diff
@@ -880,62 +927,95 @@ int guild_member_added(int guild_id,uint32 account_id,uint32 char_id,int flag) {
...
+	if( !guild_has_permission( sd, GUILD_PERM_EXPEL ) ){
+		return false;
+	}
+
+	if( g->instance_id > 0 && battle_config.instance_block_expulsion ){
+		return false;
+	}
+
+	// TODO: for leave this is different messages
+	if( map_getmapflag( sd->bl.m, MF_GUILDLOCK ) || sd->bg_id ){
```
lighta, line 991: `if(sd->bg_id || map_getmapflag( sd->bl.m, MF_GUILDLOCK )`
would be faster, there is no full elevation in c++

src/map/guild.cpp
```diff
@@ -1999,8 +2082,19 @@ int guild_break(struct map_session_data *sd,char *name) {
...
+	if( g->instance_id ){
+		if( battle_config.instance_block_leave ){
+			return 0;
+		}else{
+			instance_destroy(g->instance_id);
+		}
+	}
+
+	// Guild locked.
+	if( map_getmapflag( sd->bl.m, MF_GUILDLOCK ) ){
```
lighta, line 2094: why isn't this before the `instance_destroy(g->instance_id);`

src/map/party.cpp
```diff
@@ -569,31 +594,40 @@ int party_member_added(int party_id,uint32 account_id,uint32 char_id, int flag)
...
-	p = party_search(sd->status.party_id);
+	struct party_data *p = party_search(sd->status.party_id);
 
-	if( p == NULL )
-		return 0;
+	if( p == nullptr )
+		return false;
+
+	// Party locked.
+	if( map_getmapflag( sd->bl.m, MF_PARTYLOCK ) ){
```
lighta, line 606: this could be move before `party_search`

src/map/pc.hpp
```diff
@@ -313,6 +313,7 @@ struct map_session_data {
 		bool cashshop_open;
 		bool sale_open;
 		unsigned int block_action : 10;
+		bool refuse_party;
```
lighta, line 316: I didn't saw the real usage of the var.
Can't this create false positive and refuse all subsequent party request ?

lighta (changes requested):
1 regression spotted in `instance_destroy`
Various optimisation could be done in functions, the fastest check should be done on top.

Badarosk0:
Could you please update this PR? This one is out of date with the src. Thanks

Badarosk0:
When possible, could you update this PR?
If you use "Adventure's Agency Party" you can recruit another player from within the instance and there is no limitation for this system.
https://github.com/rathena/rathena/commit/39d5a02a3176875257ae8b59383fa7ff846e09a1
I believe you need to limit the entry of new players to the party if the instance is generated.

admkakaroto:
> When possible, could you update this PR please? If you use "Adventure's Agency Party" you can recruit another player from within the instance and there is no limitation for this system. [39d5a02](https://github.com/rathena/rathena/commit/39d5a02a3176875257ae8b59383fa7ff846e09a1) I believe you need to limit the entry of new players to the party if the instance is generated.

If add mapflag partylock not avoid this problem?

Badarosk0:
> > When possible, could you update this PR please? If you use "Adventure's Agency Party" you can recruit another player from within the instance and there is no limitation for this system. [39d5a02](https://github.com/rathena/rathena/commit/39d5a02a3176875257ae8b59383fa7ff846e09a1) I believe you need to limit the entry of new players to the party if the instance is generated.
> 
> If add mapflag partylock not avoid this problem?

Yes, adding it to src will partially resolve it. but the mapflag is inside the instance, so in the pre-map when the instance is generated, it will still be possible to recruit players with the generated instance

Lemongrass3110:
```
on standby/inside instance:
can't invite (메모리얼 던전 이용 중엔 초대 및 탈퇴가 되지 않습니다 => MSI_CANNOT_JOIN_AND_LEAVE_PARTY_IN_MAKING_MEMORIALDUNGEON)
can't leave as both member and leader (메모리얼 던전 이용 중엔 초대 및 탈퇴가 되지 않습니다 => MSI_CANNOT_JOIN_AND_LEAVE_PARTY_IN_MAKING_MEMORIALDUNGEON)
can't expel (no message)
can't change leader (no message)
```

aleos89 (approved):

PR #3754, opened 2018-12-16
Added support for viewpointvalue command

Lemongrass3110:

* **Server Mode**: Both

Additionally added a script command to update the player's camera.
Added a new atcommand camerainfo for this, which is limited to admins.

Thanks to @LunarSHINING for bringing this to my attention.


Lemongrass3110:
@zellkennedy maybe just in case someone wants to zoom in 
grafik
Looks great, doesnt it?

This is a GM only command though so maybe it is only for taking screenshots or something similar.

Issue #3774, opened 2018-12-20
Some BS/WS skill doesn't give effect while in Madogear


laziem:
* **rAthena Hash**: f4442fd
* **Client Date**: 20180621
* **Server Mode**: RE
* **Description of Issue**: 
  * Result: Adrenaline Rush, Power Thrust, Maximum Power Thrust, Cart Boost, Full Adrenaline Rush give no effect while using Madogear
  * Expected Result: working
  * How to Reproduce: The skill can be cast but didn't give any effect
* **Modifications that may affect results**: -

maqubeks:
Is this really fixed? The latest revision still doesn't give any effect.

PR #3775, opened 2018-12-20
Corrected some statuses and Mado Gear interaction

aleos89:
* **Addressed Issue(s)**: #3774

  * Resolved a few statuses getting removed when using Mado Gear that should be enabled.
  * Added the ability to remove these statuses if the skills are disabled.
Thanks to @laziem and @Lemongrass3110!


slyx88:
How about additional check on status.cpp?

Issue #3804, opened 2018-12-31
bVariableCastrate and bFixedCastrate did not working


laziem:
* **rAthena Hash**: f4442fd
* **Client Date**: 20180621
* **Server Mode**: RE
* **Description of Issue**: 
  * Result: `bonus2 bFixedCastrate,sk,n; ` and `bonus2 bVariableCastrate,sk,n;` did not work on some skill  (magical)
  * Expected Result: working
  * How to Reproduce: Edit item 1631 script to `bonus2 bVariableCastrate,"AL_HOLYLIGHT",-100;` or any value, then cast the skill
* **Modifications that may affect results**: -

PR #3808, opened 2019-01-02
Corrected item bonus FixedCastRate

aleos89:
* **Addressed Issue(s)**: #3804

* **Server Mode**: Renewal

  * Corrected FixedCastRate attempting to save into VariableCastRate bonus array.
Thanks to @laziem!

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -3713,7 +3713,7 @@ void pc_bonus2(struct map_session_data *sd,int type,int type2,int val)
 			break;
 		}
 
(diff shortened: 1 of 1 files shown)


laziem:
I try the fix, it work on magic skill like Magnus Exorcismus, but now, bonus2 bVariableCastrate didn't work for skill like Arrow Storm or Cart Cannon (Genetic).

aleos89:
I just tested on 7eed88c with the fix and was not able to see any problem using `bonus2 bVariableCastrate` with Cart Cannon or Arrow Storm. I noticed there is an issue with `@reloaditemdb` and it not re-parsing items currently equipped to players, which might be the issue to your case.

PR #3832, opened 2019-01-04
Fixed Hunting Missions

sigtus:
- Fixed getmapxy getting the first player map and coordinates in every instance
- Optimization of shop by using inarray script command
- Changed some values to constants


Atemo:
Thanks !

PR #3843, opened 2019-01-07
Add optional argument {<instance mode>} to instance_id()

vykimo:
* **Server Mode**: Both

Modification of instance_id({<instance mode>}) :
- Optional argument {<instance mode>} added
You can specify the instance mode because else, instance_id will check instances in this order : single, party, guild, then clan. So this pull request allows npc to select the right instance id attached to the player. (in case of different instances opened at the same time but not attached at the same mode).
+ Documentation in script_command.txt

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -378,7 +378,7 @@ static struct linkdb_node *sleep_db; // int oid -> struct script_state *
  *------------------------------------------*/
 const char* parse_subexpr(const char* p,int limit);
 int run_func(struct script_state *st);
-unsigned short script_instancegetid(struct script_state *st);
+unsigned short script_instancegetid(struct script_state *st, bool use_mode = false, enum instance_mode mode = IM_PARTY);
 
 const char* script_op2name(int op)
 {
@@ -19705,7 +19705,7 @@ BUILDIN_FUNC(bg_get_data)
  *------------------------------------------*/
 //Returns an Instance ID
 //Checks NPC first, then if player is attached we check
-unsigned short script_instancegetid(struct script_state* st)
+unsigned short script_instancegetid(struct script_state* st, bool use_mode, enum instance_mode mode)
 {
 	unsigned short instance_id = 0;
 	struct npc_data *nd;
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -19705,7 +19705,7 @@ BUILDIN_FUNC(bg_get_data)
  *------------------------------------------*/
 //Returns an Instance ID
 //Checks NPC first, then if player is attached we check
-unsigned short script_instancegetid(struct script_state* st)
+unsigned short script_instancegetid(struct script_state* st, bool use_mode, enum instance_mode mode)
```
Lemongrass3110: Instead of two parameters make it default IM_NONE and check if it is set to something else.

vykimo: Good idea !

src/map/script.cpp
```diff
@@ -19901,7 +19922,16 @@ BUILDIN_FUNC(instance_mapname)
...
 {
-	script_pushint(st, script_instancegetid(st));
+	if (script_hasdata(st, 2)) {
+		enum instance_mode mode = static_cast<instance_mode>(script_getnum(st, 2));
+		
+		if (mode < IM_NONE || mode >= IM_MAX) {
+			ShowError("buildin_instance_create: Unknown instance mode %d for '%s'\n", mode, script_getstr(st, 2));
+			return SCRIPT_CMD_FAILURE;
+		}
+		script_pushint(st, script_instancegetid(st, true, mode));
```
aleos89: Your function here is pushing 3 arguments still.

src/map/script.cpp
Atemo: The 'auto-detect' of the instance ID of the player for `instance_enter`, `instance_destroy` etc.. commands who are using `script_instancegetid` will still be broken if the player owns 2 instances (for example IM_PARTY and IM_GUILD instances types).

For me the 'auto-detect' lead to bug, I would prefer to remove it and make, instead, IM_PARTY by default in `script_instancegetid`. The instance ID will always be required for the instances commands for instances types other than IM_PARTY (=add explanations in the docs).

Sample of the idea:
```
izlude,100,100,0	script	test instance	56,{
	// If player owns IM_PARTY and IM_GUILD, by default the command will destroy IM_PARTY instance.
	instance_destroy;
	end;
}

izlude,101,100,0	script	test instance 2	56,{
	// ID always required to destroy IM_GUILD instance.
	instance_destroy instance_id(IM_GUILD);
	end;
}

izlude,101,100,0	script	test instance 3	56,{
	// If player owns IM_PARTY and IM_GUILD, by default the command warp the player in IM_PARTY instance.
	instance_enter("Endless Tower");
	end;
}

izlude,101,100,0	script	test instance 4	56,{
	// ID always required to warp in IM_GUILD instance.
	instance_enter( "Endless Tower", -1, -1, getcharid(0), instance_id(IM_GUILD) );
	end;
}
```

vykimo: Yes good idea

src/map/script.cpp
```diff
@@ -19901,7 +19922,16 @@ BUILDIN_FUNC(instance_mapname)
  *------------------------------------------*/
 BUILDIN_FUNC(instance_id)
 {
-	script_pushint(st, script_instancegetid(st));
+	if (script_hasdata(st, 2)) {
+		enum instance_mode mode = static_cast<instance_mode>(script_getnum(st, 2));
```
Lemongrass3110: Please treat mode as int here and only do a static_cast after you validated it(check below).
This can cause undefined behavior and is a no-go.

src/map/script.cpp
```diff
@@ -19901,7 +19922,16 @@ BUILDIN_FUNC(instance_mapname)
...
+		
+		if (mode < IM_NONE || mode >= IM_MAX) {
+			ShowError("buildin_instance_create: Unknown instance mode %d for '%s'\n", mode, script_getstr(st, 2));
```
Lemongrass3110: buildin_instance_create should be buildin_instance_id

Additionally your output is weird. %d mode=getnum(2) and %s for getstr(2). So you treat the same parameter in two ways.

aleos89:
Took over in #4090! Thank you for your contribution!

Issue #3589, opened 2018-10-17
BUG on bonus2 bSkillAtk with Soul Destroyer


laziem:
* **rAthena Hash**: 67139f590

* **Client Date**: 2017

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Soul Destroyer damage get boosted way more that it should be
  * Expected Result: Boosted correctly 
  * How to Reproduce: Try to put `bonus2 bSkillAtk,"ASC_BREAKER",100;` or with any value on any item. With value of 100, the Soul Destroyer damage get boosted like 400%, when it should only 100%.
  * Official Information: -

* **Modifications that may affect results**: -

PR #3868, opened 2019-01-15
Corrected item bonuses SkillAtk and SubSkill

aleos89:
* **Addressed Issue(s)**: #3589

  * Corrected item bonuses SkillAtk and SubSkill granting too much damage.
Thanks to @laziem!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -5169,6 +5169,12 @@ static void battle_calc_weapon_final_atk_modifiers(struct Damage* wd, struct blo
 #endif
 	}
 
+	// add any miscellaneous player ATK bonuses
+	if (sd && (skill_damage = pc_skillatk_bonus(sd, skill_id)))
+		ATK_ADDRATE(wd->damage, wd->damage2, skill_damage);
+	if (tsd && (skill_damage = pc_sub_skillatk_bonus(tsd, skill_id)))
+		ATK_ADDRATE(wd->damage, wd->damage2, -skill_damage);
+
 	// Skill damage adjustment
 	if ((skill_damage = battle_skill_damage(src, target, skill_id)) != 0)
 		ATK_ADDRATE(wd->damage, wd->damage2, skill_damage);
@@ -5405,15 +5411,6 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 				ATK_ADD(wd.weaponAtk, wd.weaponAtk2, sstatus->matk_min);
 		}
 #endif
-		// add any miscellaneous player ATK bonuses
-		if( sd && skill_id && (i = pc_skillatk_bonus(sd, skill_id))) {
-			ATK_ADDRATE(wd.damage, wd.damage2, i);
-			RE_ALLATK_ADDRATE(&wd, i);
-		}
-		if (tsd && (i = pc_sub_skillatk_bonus(tsd, skill_id))) {
-			ATK_ADDRATE(wd.damage, wd.damage2, -i);
-			RE_ALLATK_ADDRATE(&wd, -i);
-		}
 
 #ifdef RENEWAL
 		// In Renewal we only cardfix to the weapon and equip ATK


laziem:
soul_destroyer

I try to test on Jur (1250) and add script:

```
bonus2 bSkillAtk,"ASC_BREAKER",100;
bonus2 bSkillAtk,"AS_SONICBLOW",100;
bonus2 bSkillAtk,"ASC_METEORASSAULT",100;
```

All worked except for Soul Destroyer, damage increase more than 100%

aleos89:
I only tested with base stats. But after the src adjustment the 4.4x~ damage reduced down to the 2x it should have been.

What stats/equips are you specifically using so I can do more tests. Thanks!

laziem:
I use weapon +0 Jur (id: 1250), allstat 90, and skill Maximize Power so I get constant damage (without Power Maximize same result).

**Before:** **_with 100% bSkillAtk_**
Sonic Blow : 4240 > 8576 [~102% or ~2.02x]
Soul Destroyer : 6430 > 26364 [~310% or ~4.1x]
Meteor Assault : 2144 > 4383 [~104% or ~2.04]

**After:** **_with 100% bSkillAtk_**
Sonic Blow : 4240 > 8480 [100% or 2x]
Soul Destroyer : 6430 > 16604 [~158% or ~2.58x]
Meteor Assault : 2144 > 4288 [100% or 2x]

From the test, only Soul Destroyer that didn't work.

nubspixel:
I was concerned that the bSkillAtk might have affected other skills as well, hence. I did some testing:

Rune Knight lv174 wearing an Old Rune Criclet (modified to only have bSkillAtk):
**Before patch:**
_Hundred Spears:_ `bonus2 bSkillAtk,"RK_HUNDREDSPEAR",150;`
25245 > 63150 (150% increase | difference of 0%)
_Ignition Break:_ `bonus2 bSkillAtk,"RK_IGNITIONBREAK",200;`
14131 > 42443 (200% increase | difference of 0%)

**After patch:**
_Hundred Spears:_ `bonus2 bSkillAtk,"RK_HUNDREDSPEAR",150;`
24710 > 61775 (150% increase | difference of 0%)
_Ignition Break:_ `bonus2 bSkillAtk,"RK_IGNITIONBREAK",200;`
16628 > 49884 (200% increase | difference of 0%)

Which made me curious what the bug actually is:
Guillotine Cross lv174 using jur[3] and jur[2] (modified to have bSkillAtk):
**Before patch:**
_Meteor Assault:_ `bonus2 bSkillAtk,"ASC_METEORASSAULT",100;`
2290 > 4606 (101% increase | difference of 1%)
_Soul Breaker:_ `bonus2 bSkillAtk,"ASC_BREAKER",100;`
6351> 25678 (304% increase | difference of 204%)
_Sonic Blow:_ `bonus2 bSkillAtk,"AS_SONICBLOW",100;`
4464 > 8952 (101% increase | difference of 1%)

**After patch:**
_Meteor Assault:_ `bonus2 bSkillAtk,"ASC_METEORASSAULT",100;`
2386 > 4772 (100% increase | difference of 0%)
_Soul Breaker:_ `bonus2 bSkillAtk,"ASC_BREAKER",100;`
6679 > 17246 (158% increase | difference of 58%)
_Sonic Blow:_ `bonus2 bSkillAtk,"AS_SONICBLOW",100;`
4632 > 9264 (100% increase | difference of 0%)

aleos89:
The extra damage probably has to do with the second ATK from Soul Destroyer. Will look into it further.

cydh:
tbh need confirmation, the skill atk bonus will increase by parts (if skill has magic & weapon dmg parts) or only on final.
assumed like our current behavior by excluding `misc`-type skills in specific lines for weapon or magic formula, add switch-case for these lines
https://github.com/rathena/rathena/blob/ce1508a01c68dee80544242b470987742f3b3ece/src/map/battle.cpp#L6239-L6241
become
```cpp
			//Damage bonuses
			switch (skill_id) {
#ifdef RENEWAL
				case NJ_ISSEN:
				case ASC_BREAKER:
				case CR_ACIDDEMONSTRATION:
				case GN_FIRE_EXPANSION_ACID:
					break; // These skills will do a damage adjustment later
#endif
				default:
					if ((i = pc_skillatk_bonus(sd, skill_id)))
						ad.damage += (int64)ad.damage*i/100;
			}
```

Issue #3100, opened 2018-05-05
getmapxy script update


pajodex:
Hi, 

This is not really an issue, again, but rather a suggestion..

I've been playing with `getmapxy` and I found out that when used in mobs, it returns to `"null",0,0` value.. Here is how I coded to further explain my self ...
So I spawned a BG monster using this code
```bg_monster 0, "bat_c01", 99, 92, "MONSTER", 1001, strnpcinfo(3)+"::OnMonsDown";```
Then added something for getmapxy..
```getmapxy(.@mapname$, .@mapx, .@mapy, UNITTYPE_NPC, "MONSTER");```
```announce "Test" + .@mapname$ +" , "+ .@mapx +" , "+ .@mapy +" end test",bc_all;```

But it returned to `"null",0,0` .. So, here is what I suggest.. 
instead of using `*getmapxy("<variable for map name>",<variable for x>,<variable for y>,<type>{,"<search string>"}) `.. why not search for `parameters` instead of a `string` just like this (from Herc) `*getmapxy("<variable for map name>",<variable for x>,<variable for y>,<type>{,"<search parameter>"}) `
it is convenient to scripters who have something in mind but can't do because of some limitations to commands..

Well, this is just merely a suggestion... but it would be great to have this in rAthena as well.. Have a great day!

Source : https://github.com/HerculesWS/Hercules/pull/871/files

PR #3890, opened 2019-01-25
Updated script command getmapxy parameters

aleos89:
* **Addressed Issue(s)**: #3100

  * Script command getmapxy will now allow integer values for the search value.
  * Deprecated UNITTYPE constants as they were replaced by BL.
  * Updated script command getunittype return value to return BL constants.
  * Updated scripts to new BL constants.
Thanks to @pajodex's suggestion!


Atemo:
Changing the target by default from BL_NPC to BL_PC doesn't improve anything, I'm against this change.

PR #3910, opened 2019-02-02
Cleaned up script command argument parsing

aleos89:
* **Addressed Issue(s)**: #3918

  * Cleaned up the way script commands parse arguments.
  * script_isstring and script_isnum will now properly check variable references.
  * This allows script command arguments to contain variables without having to make source side adjustments.
  * Reverted several script commands that reference the data directly.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -2678,14 +2678,14 @@ static bool script_rid2sd_( struct script_state *st, struct map_session_data** s
  * @param data Variable/constant
  * @param sd If NULL, will try to use sd from st->rid (for player's variables)
  */
-void get_val_(struct script_state* st, struct script_data* data, struct map_session_data *sd)
+struct script_data *get_val_(struct script_state* st, struct script_data* data, struct map_session_data *sd)
 {
 	const char* name;
 	char prefix;
 	char postfix;
 
 	if( !data_isreference(data) )
-		return;// not a variable/constant
+		return data;// not a variable/constant
 
 	name = reference_getname(data);
 	prefix = name[0];
@@ -2703,7 +2703,7 @@ void get_val_(struct script_state* st, struct script_data* data, struct map_sess
 				data->type = C_INT;
 				data->u.num = 0;
 			}
-			return;
+			return data;
 		}
 	}
 
@@ -2811,12 +2811,12 @@ void get_val_(struct script_state* st, struct script_data* data, struct map_sess
 	}
 	data->ref = NULL;
 
-	return;
+	return data;
 }
 
-void get_val(struct script_state* st, struct script_data* data)
+struct script_data *get_val(struct script_state* st, struct script_data* data)
 {
-	get_val_(st,data,NULL);
+	return get_val_(st,data,NULL);
 }
 
 struct script_data* push_val2(struct script_stack* stack, enum c_op type, int64 val, struct reg_db* ref);
@@ -3250,7 +3250,7 @@ const char* conv_str_(struct script_state* st, struct script_data* data, struct 
 	}
 	else if( data_isreference(data) )
 	{// reference -> string
-		//##TODO when does this happen (check get_val) [FlavioJS]
+		//##TODO when does this happen (check get_val) [FlavioJS] -- at getd!!
 		data->type = C_CONSTSTR;
 		data->u.str = reference_getname(data);
 	}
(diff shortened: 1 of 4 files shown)


src/map/script.cpp
```diff
@@ -10053,14 +9987,11 @@ BUILDIN_FUNC(itemskill)
...
 	TBL_PC* sd;
-	struct script_data *data;
 
 	if( !script_rid2sd(sd) || sd->ud.skilltimer != INVALID_TIMER )
 		return SCRIPT_CMD_SUCCESS;
 
-	data = script_getdata(st, 2);
-	get_val(st, data); // Convert into value in case of a variable
-	id = ( data_isstring(data) ? skill_name2id(script_getstr(st,2)) : script_getnum(st,2) );
+	id = ( script_isstring(st, 2) ? skill_name2id(script_getstr(st,2)) : script_getnum(st,2) );
```
Lemongrass3110: Skill ID is not validated. skill_name2id returns 0 in case it does not exist.

src/map/script.cpp
```diff
@@ -9339,9 +9287,7 @@ BUILDIN_FUNC(autobonus3)
 
 	rate = script_getnum(st,3);
 	dur = script_getnum(st,4);
-	data = script_getdata(st, 5);
-	get_val(st, data); // Convert into value in case of a variable
-	atk_type = ( data_isstring(data) ? skill_name2id(script_getstr(st,5)) : script_getnum(st,5) );
+	atk_type = ( script_isstring(st, 5) ? skill_name2id(script_getstr(st,5)) : script_getnum(st,5) );
```
Lemongrass3110: Skill ID is not validated. skill_name2id returns 0 in case it does not exist.
Bad variable name as well.

src/map/script.cpp
```diff
@@ -9181,9 +9136,7 @@ BUILDIN_FUNC(bonus)
 		case SP_SKILL_USE_SP:
 		case SP_SUB_SKILL:
 			// these bonuses support skill names
-			data = script_getdata(st, 3);
-			get_val(st, data); // Convert into value in case of a variable
-			val1 = ( data_isstring(data) ? skill_name2id(script_getstr(st,3)) : script_getnum(st,3) );
+			val1 = ( script_isstring(st, 3) ? skill_name2id(script_getstr(st,3)) : script_getnum(st,3) );
```
Lemongrass3110: Skill ID is not validated. skill_name2id returns 0 in case it does not exist.

src/map/script.cpp
```diff
@@ -18368,7 +18425,21 @@ BUILDIN_FUNC(unitskillusepos)
...
+		const char *name = script_getstr(st, 3);
+
+		if (!(skill_id = skill_name2id(name))) {
+			ShowError("buildin_unitskillusepos: Invalid skill name %s.\n", name);
+			return SCRIPT_CMD_FAILURE;
+		}
+	} else {
+		int skill_id_lookup = script_getnum(st, 3);
+
+		if (!(skill_id = skill_get_index(skill_id_lookup))) {
```
Lemongrass3110: Same

src/map/script.cpp
```diff
@@ -20206,7 +20277,21 @@ BUILDIN_FUNC(areamobuseskill)
...
+		const char *name = script_getstr(st, 7);
+
+		if (!(skill_id = skill_name2id(name))) {
+			ShowError("buildin_areamobuseskill: Invalid skill name %s.\n", name);
+			return SCRIPT_CMD_FAILURE;
+		}
+	} else {
+		int skill_id_lookup = script_getnum(st, 7);
+
+		if (!(skill_id = skill_get_index(skill_id_lookup))) {
```
Lemongrass3110: Same

src/map/script.cpp
```diff
@@ -20934,7 +21019,21 @@ BUILDIN_FUNC(npcskill)
...
+		const char *name = script_getstr(st, 2);
+
+		if (!(skill_id = skill_name2id(name))) {
+			ShowError("buildin_npcskill: Invalid skill name %s.\n", name);
+			return SCRIPT_CMD_FAILURE;
+		}
+	} else {
+		int skill_id_lookup = script_getnum(st, 2);
+
+		if (!(skill_id = skill_get_index(skill_id_lookup))) {
```
Lemongrass3110: Same

Issue #3908, opened 2019-02-01
Checking card positions (Accessories)


Badarosk0:
* **rAthena Hash**: 
https://github.com/rathena/rathena/commit/298cfefd2d43fa1966b5c50fdaa5a654aebd7782

* **Client Date**: 
2018-06-20

* **Server Mode**: 
RE

* **Description of Issue**: 
  * Result:   Use one of these cards that I sent you up on 2 accessories and try to equip. You will notice that the card that could equip only on one side is equipping on both sides. It appears that a check in the src is missing in addition to checking the accessory, also checking the position of the card.

  * Expected Result: As with accessories, some cards must fit on only one side.
  * How to Reproduce:  Equip two of these cards on any accessory side.
```
27170,Shotgun_Buffalo_Bandit_Card,Shotgun Buffalo Bandit Card,6,20,,10,,,,,,,,128,,,,,{ bonus bAgi,7; bonus bHit,7; if (isequipped(27171)) { bonus bStr,3; bonus bAtkRate,3; } },{},{}
27171,Revolver_Buffalo_Bandit_Card,Revolver Buffalo Bandit Card,6,20,,10,,,,,,,,8,,,,,{ bonus bStr,4; bonus bAtkRate,4; },{},{}
```
  * Official Information:
  

* **Modifications that may affect results**:

PR #3913, opened 2019-02-02
Fixed accessory equipping with cards

aleos89:
* **Addressed Issue(s)**: #3908

  * Accessories that have cards which require specific left/right locations must take priority.
Thanks to @Badarosk0!

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -9839,10 +9839,26 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos,bool equipswit
 		clif_notify_bindOnEquip(sd,n);
 	}
 
-	if(pos == EQP_ACC) { //Accesories should only go in one of the two,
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -9839,10 +9839,26 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos,bool equipswit
...
+
+				if (card_pos == EQP_ACC_L || card_pos == EQP_ACC_R) {
+					pos = card_pos; // Use the card's equip position
+					break;
+				}
+			}
+		}
 	}
 
 	if(pos == EQP_ARMS && id->equip == EQP_HAND_R) { //Dual wield capable weapon.
```
Lemongrass3110: Just seeing this now. Shouldnt all those that compare with == be else if cases to speed things up?

PR #3932, opened 2019-02-08
Corrected script commands containing arrays

aleos89:
* **Addressed Issue(s)**: N/A

  * Corrects a potential map-server crash with script commands min, max, getpartymember, getguildmember, getunits, getmapunits, getareaunits, getunitdata, and channel_setgroup.
Thanks to @Tokeiburu and @Atemo!

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -22430,6 +22430,11 @@ BUILDIN_FUNC(minmax){
 			// Get the session data, if a player is attached
 			sd = st->rid ? map_id2sd(st->rid) : NULL;
 
+			if (not_server_variable(*name) && sd == NULL) {
+				ShowError("buildin_%s: No player attached for player variable '%s'\n", functionname, name);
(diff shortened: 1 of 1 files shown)


src/map/script.cpp
```diff
@@ -8269,6 +8269,10 @@ BUILDIN_FUNC(getpartymember)
 				ShowError("buildin_getpartymember: The array %s is not string type.\n", varname);
 				return SCRIPT_CMD_FAILURE;
 			}
+			if (not_server_variable(*varname)) {
+				ShowError("buildin_getpartymember: Requires a server variable, player variable '%s' is invalid.\n", varname);
```
Atemo: For me blocking player variable is overkill. How about `script_rid2sd` check like https://github.com/rathena/rathena/blob/c5a5c545181bcb1dae9b33131acf137f5033a59f/src/map/script.cpp#L6035

Issue #3788, opened 2018-12-22
Camouflage + Aloevera


Everade:
* **rAthena Hash**: 
01f61cf

* **Client Date**: 
not relevant

* **Server Mode**: 
Renewal

* **Description of Issue**: 
  * Result:
Casting ![camouflage](https://user-images.githubusercontent.com/12261601/50379258-ba52a800-0645-11e9-861e-2366eca1bedf.png) Camouflage skill while under the effects of ![606](https://user-images.githubusercontent.com/12261601/50379259-c2aae300-0645-11e9-90b2-039d1ca391a1.png) Aloevera cancels the provoke effect.
If you get inflicted with a status (i.e. Blinded) while under the effects of Aloevera, Aloevera gets canceled.

  * Expected Result: Shouldn't de-active effect.

  * Official Information:
https://irowiki.org/wiki/Camouflage

* **Modifications that may affect results**: 
none

PR #3933, opened 2019-02-08
Fixed Camouflage

aleos89:
* **Addressed Issue(s)**: #3788

  * Fixes the Provoke icon getting removed when clif_changeoption is called and it's not an infinite duration.
  * Fixed Camouflage not ending after 10 seconds.
Thanks to @Everade!


Everade:
@aleos89 
Why does it cancel the skill after 10 seconds?
iRO Wiki claims that duration is: SP depletion or player is revealed

Was this a kRO update i've missed?

aleos89:
You're right, sorry. I confused it when reading with the 10 seconds it lasts for increasing the stats.

PR #3209, opened 2018-06-16
pcblockall command.

crazystorm2017:
Block players action same like opening a mes box.

* **Addressed Issue(s)**: 
None.

* **Server Mode**: 
Both.

* **Description of Pull Request**: 
A script command for simulating the mes box.
I believe this is useful for the upcoming episode updates.
Credits mainly to joecalis of rathena, just added some parts to it.

PR #3951, opened 2019-02-18
Added pcblock and checkpcblock script commands.

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

## Overview
This PR take over https://github.com/rathena/rathena/pull/3209 (branch deleted)
Merge from https://github.com/HerculesWS/Hercules/pull/842

## Simple test script
```
izlude,123,140,6	script	pcblock	763,{
	setarray .@type_name$[0],
		"PCBLOCK_MOVE",
		"PCBLOCK_ATTACK",
		"PCBLOCK_SKILL",
		"PCBLOCK_USEITEM",
		"PCBLOCK_CHAT",
		"PCBLOCK_IMMUNE",
		"PCBLOCK_SITSTAND",
		"PCBLOCK_COMMANDS",
		"PCBLOCK_NPCCLICK";
		// "PCBLOCK_ALL";
	.@size = getarraysize(.@type_name$);
	.@menu_type$ = implode( .@type_name$, ":" );

	while (true) {
		mes "Current flag : ";
		for ( .@i = 0; .@i < .@size; .@i++ ) {
			if (checkpcblock() & (1 << .@i))
				mes .@type_name$[.@i];
		}
		.@s = select(.@menu_type$) - 1;
		pcblock (1 << .@s), !(checkpcblock() & (1 << .@s));
		next;
	}
	end;
}
```


doc/script_commands.txt
```diff
@@ -6133,6 +6133,58 @@ Examples:
...
+	pcblock PCBLOCK_ATTACK|PCBLOCK_SKILL, true;
+
+// Re-enables attack, skills and item use
+	pcblock PCBLOCK_ATTACK|PCBLOCK_SKILL|PCBLOCK_ITEM, false;
+
+// checkpcblock related checks
+	if (checkpcblock() & PCBLOCK_IMMUNE)
+		mes "You are invulnerable!";
+
+	if (checkpcblock() & (PCBLOCK_MOVE|PCBLOCK_SITSTAND))
```
secretdataz: If only one of MOVE and SITSTAND flag is set, the expression will still be true.

doc/script_commands.txt
```diff
@@ -6133,6 +6133,58 @@ Examples:
...
+		mes "You can't walk and sit.";
+
+	if ((checkpcblock() & (PCBLOCK_ATTACK|PCBLOCK_SKILL)) == PCBLOCK_NONE)
```
secretdataz: Comparing to 0 is more expressive imo

src/map/pc.hpp
```diff
@@ -730,6 +730,15 @@ struct map_session_data {
...
 	uint8 hatEffectCount;
 #endif
+
+	struct {
+		unsigned attack   : 1;
+		unsigned useitem  : 1;
+		unsigned chat     : 1;
+		unsigned sitstand : 1;
+		unsigned commands : 1;
+		unsigned npcclick : 1;
```
aleos89: Not a huge deal, but maybe make a note that `move` and `skill` restrictions are in `unit.hpp::unit_data`.

```suggestion
		// move and skill are defined in unit.hpp::unit_data as it can be applicable to all objects.
```

src/map/script.cpp
```diff
@@ -17234,6 +17234,88 @@ BUILDIN_FUNC(pcblockskill)
...
 
+BUILDIN_FUNC(pcblock)
+{
+	TBL_PC *sd;
+
+	if (!script_mapid2sd(4, sd))
+		return SCRIPT_CMD_SUCCESS;
+
+	enum pcblock_action_flag type = (pcblock_action_flag)script_getnum(st, 2);
+	int state = (script_getnum(st, 3) > 0) ? 1 : 0;
```
aleos89: state could be defined as bool.

```suggestion
	bool state = (script_getnum(st, 3) > 0) ? true : false;
```

src/map/pc.hpp
```diff
@@ -730,6 +730,15 @@ struct map_session_data {
 	uint32* hatEffectIDs;
 	uint8 hatEffectCount;
 #endif
+
+	struct {
+		unsigned attack   : 1;
```
aleos89: If we continue support for `unitblockmove` and `unitblockskill`, would attack be an option to throw into `unit.hpp::unit_data` too?

src/map/script.hpp
```diff
@@ -1898,6 +1898,23 @@ enum e_hat_effects {
 	HAT_EF_MAX
 };
 
+/**
+ * Player blocking actions related flags.
+ */
+enum pcblock_action_flag {
```
aleos89: Prefix the enum to match naming standard.

```suggestion
enum e_pcblock_action_flag {
```

src/map/pc.hpp
Atemo: If we implement a command to block the attack of a unit, eventually. For now no need to move it to `unit_data` I guess.

Atemo: `unitblockmove` and `unitblockskill` could be dropped. The options are supported largely by `setunitdata` if I'm not wrong.

aleos89: I know we can currently manipulate movement with `*_CANMOVE_TICK` and we have `*_IMMUNE_ATTACK` which isn't necessarily equal to blocked skill usage. But the `unit_data::state` could most definitely be added to `setunitdata`/`getunitdata`.

doc/script_commands.txt
```diff
@@ -6133,6 +6133,58 @@ Examples:
...
+'checkpcblock' command return the bit-mask value of the currently
+enabled block flags.
+
+The available <type> are:
+	PCBLOCK_MOVE
+	PCBLOCK_ATTACK
+	PCBLOCK_SKILL
+	PCBLOCK_USEITEM
+	PCBLOCK_CHAT
+	PCBLOCK_IMMUNE
```
AnnieRuru: imo, rathena `*setunitdata` use `UMOB_DMGIMMUNE`
so I guess `PCBLOCK_DMGIMMUNE` might sound consistent with setunitdata counterpart

PS: I am posting once here because Atemo helped me fix PCBLOCK_EMOTION

Atemo: It wasn't out of curiosity? Thanks anyway

Atemo:
The bool part was to force to the new status, your sample force to use the check with the set but yeah that could do it.
How about the command name, you're using another name you don't like the current?

doc/script_commands.txt
```diff
@@ -6133,6 +6133,58 @@ Examples:
...
+Examples:
+
+// Make the attached player invulnerable to monster (same as @monsterignore)
+	pcblock PCBLOCK_IMMUNE, true;
+
+// Prevents the attached player from attacking and using skills
+	pcblock PCBLOCK_ATTACK|PCBLOCK_SKILL, true;
+
+// Re-enables attack, skills and item use
+	pcblock PCBLOCK_ATTACK|PCBLOCK_SKILL|PCBLOCK_ITEM, false;
```
sigtus: pcblock PCBLOCK_ATTACK|PCBLOCK_SKILL|PCBLOCK_USEITEM, false;

PR #3961, opened 2019-02-21
Definition path of DBPATH location in level_penalty.txt

marky291:
This uses a condition to check if renewal, and although it's true you only need it for renewal, it allows easier to edit the folder structure of DBPATH.


secretdataz:
Can we keep all relevant changes in one PR please?

marky291:
@secretdataz I was only suppose to make a single change... I'm using the web browser... I really can't be bothered otherwise. :)

PR #3962, opened 2019-02-21
Use DBPATH Definitions.

marky291:
Same reason as comment in #3961.

Use DBPath to allow easier to restructure files with the src config.


secretdataz:
#3961

PR #3964, opened 2019-02-21
Definitions for DB Pathing

marky291:


marky291:
Was suppose to be one commit using the web browser interface, but found few more and the web browser allows one commit per edit?, sorry for the 8 commits, found a way to merge in the end

Daegaladh:
> 
> 
> Was suppose to be one commit using the web browser interface, but found few more and the web browser allows one commit per edit?, sorry for the 8 commits, found a way to merge in the end

It's ok as long as you keep them all in the same PR, we can squash and merge them all together.

rfperuch:
Actually there's a problem on this
In the file ``pc.cpp`` while reading the file ``job_db1.txt``, if the macro #RENEWAL_ASPD is defined, the serve will read the ``job_db1.txt`` from the renewal db instead of pre-renewal, that will cause some problems on the players attributes.

rfperuch:
@aleos89 @Daegaladh @marky291

Issue #3015, opened 2018-03-29
Monster Power Up Bug (With ScreenShot)


slyx88:
* **Client Date**: 
2017-06-14

* **Server Mode**: 
Renewal

* **Description of Issue**: 
Monster Power Up skill does not triple the attack power of the monster. Nevertheless the hit bonus works as intended

* Expected Result: 
Melee attack and range attack of the monsters including their skills should do 3 times higher in damage.
* How to Reproduce: 
Summon ifrit and check the damage before and after it uses power up skill.
 * Official Information:
  

* **Modifications that may affect results**: 
none
 
  
![screenrathena006](https://user-images.githubusercontent.com/37906520/38225292-d63a8144-3726-11e8-9db3-4819a25aaae0.jpg)
![screenrathena004](https://user-images.githubusercontent.com/37906520/38225296-df2d831e-3726-11e8-8526-c8c73e771987.jpg)

slyx88:
from status.c it seems the attack of monster is neither watk nor batk. The ATK and ATK2 of the monster are defined as rhw.atk and rhw.atk2

Lemongrass3110:
Could you check if your problem is still there after fe197bf?
This should have fixed a lot of MATK problems with renewal mobs.

slyx88:
The damage increased slightly. I spawned harword bio3 monster. Let him use Maximize Power 1st for damage consistency and check the damage before and after power up. tested on warlock level 175 with 1000 VIT.
Before powerup : 3005
After powerup : 3565

Powerup only affects ATK status and apparently the skill of the mobs are not affected. I managed to get the damage up to official's one by adding this in battle.cpp under //SC effects from caster side.

`if (sc->data[SC_INCATKRATE] ) {
			if (src->type == BL_MOB && !skill_id)	
				damage += damage * sc->data[SC_INCATKRATE]->val1 / 100;	
		}`

PR #3968, opened 2019-02-21
setunitdata script command monster stat recaulculation fixes

SeravySensei:
* **Addressed Issue(s)**: 

 https://github.com/rathena/rathena/issues/3015
https://github.com/rathena/rathena/issues/2866
maybe more

* **Server Mode**: 

Renewal

Today I tried to make a script raise monster stats and ran into several problems. This commit fixes the following problems I had to deal with :
-Monster stat recalculation is cut short and doesn't happen at all if the unit recalculated has no regen (monsters don't) and no status active on it (scripts changing stats with setunitdata do not cause status changes). I commented out this code until there is a more appropriate solution, as it outright makes setunitdata not function at all - according to the docs folder, it is supposed to trigger a recalculation and wouldn't be meaningful otherwise anyway. While it sounds correct to simply copy the base_status to status, it is not : this results in a monster who has correct rhw.matk but incorrect min and max matk, as that is being calculated within the base_status itself.
-When MATK is calculated, rhk.matk from status is used instead of base_status. This makes changes in base_status have no effect at all on calculated matk values, and scripts change base_status, not sure what else. 
-Monsters have their attack min and max stored in rhw.atk and rhw.atk2 as far as I see when running in debug and watching data in memory. The script command setunitdata also sets these values.  Commit https://github.com/rathena/rathena/commit/fe197bfa120aef5fd31c6896122b35fdc06774b4 removed the recalculation for them, which means monster attack no longer moves from base_status to status and no longer has effects applied to it. I restored these 2 lines which made setunitdata for them function correctly. I suspect this one is responsible for the linked NPC_POWERUP issue.
-There is no MOB_MATK for scripts. Now that monster matk is no longer their ATK2 stat, it need to be added.
-There is no UMOB_MATK. A min and max version does exist, but those are calculated values, setting them is meaningless and should get lost as soon as a recalculation of stats happen (note the first problem blocked recalculations on monsters, suppressing this problem as well). 

Note, I'm a noob who has no idea about how things are done here. Feel free to deny this and make your own pull request if it's not correctly formatted, or you are more experienced with how monster stat recalculations work. If you are wondering what's this whole weird mess of stuff, I just wanted to have a script command raise my monster's MATK by a percentage. It's not my fault I ran into 5 separate problems trying to get something that simple done. I also have no idea if the described problems affect unit types other than monsters (if they use rhw.atk, atk2, matk then most likely yes), in which case some of those might also get fixed. This aims to be an emergency fix to what I consider a critcial set of problems, it's by no means a well tested and perfect solutions. Also, I haven't yet tested setunitdata on other monster stats. There might be more similar bugs lurking around.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -17034,7 +17034,8 @@ BUILDIN_FUNC(getmonsterinfo)
 		case MOB_JOBEXP:	script_pushint(st,mob->job_exp); break;
 		case MOB_ATK1:		script_pushint(st,mob->status.rhw.atk); break;
 		case MOB_ATK2:		script_pushint(st,mob->status.rhw.atk2); break;
-		case MOB_DEF:		script_pushint(st,mob->status.def); break;
+		case MOB_MATK:		script_pushint(st,mob->status.rhw.matk); break;
+		case MOB_DEF:		script_pushint(st, mob->status.def); break;
 		case MOB_MDEF:		script_pushint(st,mob->status.mdef); break;
 		case MOB_STR:		script_pushint(st,mob->status.str); break;
 		case MOB_AGI:		script_pushint(st,mob->status.agi); break;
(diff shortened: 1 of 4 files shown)


teededung:
@SeravySensei Missing this line to `getunitdata` function
```
getunitdata_sub(UMOB_MATK, md->status.rhw.matk);
```
@aleos89 Please merge this fix, I just solved my issue when using `setunitdata UMOB_ATKMIN, UMOB_ATKMAX and a new UMOB_MATK from SeravySensei`

Please note that setunitdata with UMOB_MATKMIN and UMOB_MATKMAX doesn't work

aleos89:
Part of this requires the new mob data for ATK1 and ATK2. We were waiting for mob_db.yml before we try to apply these fixes. They haven’t been forgotten.

zh82558223:
> Part of this requires the new mob data for ATK1 and ATK2. We were waiting for mob_db.yml before we try to apply these fixes. They haven’t been forgotten.

@aleos89 
hello, sir,

atk

In the case of using mob_db.yaml, using setunitdata to modify the int of the monster can improve matk, but it is invalid for str, and atk and matk cannot be modified directly. Did you forget it?
Please fix this problem.

PR #3971, opened 2019-02-22
Better itemdb_read_flag error message

marky291:
Should this be a warning message?, since technically you can ignore this and no issues will occur.


aleos89:
I'll leave it as an error type message for now. We are beginning work on rewriting a lot of things and that will get covered later on. Thanks for your PRs!

PR #3977, opened 2019-02-24
Update mob_db.txt

plsfixrito:
* **Addressed Issue(s)**: Outdated /re/mob_db.txt

* **Server Mode**: Renewal

* **Description of Pull Request**: Complete update for /re/mob_db.txt based on divine-pride.net

Here is a list of all notable changes https://puu.sh/CRgCU/baebf05ff1.txt


SeravySensei:
Haven't looked at other stats nor can I judge them but the atk and atk2 doesn't seem to consider the recent update of atk being the monster's base attack and atk2 being their base matk.

plsfixrito:
> Haven't looked at other stats nor can I judge them but the atk and atk2 doesn't seem to consider the recent update of atk being the monster's base attack and atk2 being their base matk in renewal databases.

Well, the entire changes are based on divine-pride.
ATK1 stands for MAX ATK while ATK2 is for MIN ATK

SeravySensei:
Unfortunately that's not what the monster database uses these for right now in renewal. Min/Max ATK and MATK are both stats calculated from other stats - in case of matk, from int, level and the base matk which is stored in atk2. So the min/max stats are not in the monster database at all.

plsfixrito:
> Unfortunately that's not what the monster database uses these for right now in renewal. Min/Max ATK and MATK are both stats calculated from other stats - in case of matk, from int, level and the base matk which is stored in atk2. So the min/max stats are not in the monster database at all.

I'll update the database again later with this information that you provided thanks.
It seems that the information provided in the docs and was provided to me by the devs were wrong.
I was told that ATK1 is max ATK and ATK2 is min

SeravySensei:
That was certainly true half a year ago. This is a new change. This commit here : https://github.com/rathena/rathena/commit/fe197bfa120aef5fd31c6896122b35fdc06774b4

plsfixrito:
@SeravySensei thanks !
Now I have updated the database again and here is the complete log of what changed https://puu.sh/CRgCU/baebf05ff1.txt

Updated the log on the main post also.

Lemongrass3110:
Sorry to turn you down, but this will declined because the next update to our mob database will be #2780.

After we did a 1:1 conversion to yaml we will gladly compare it to divine pride(we already have tools to do this automatically) and recheck it with the yaml version as you will see exactly which database columns will be changed which makes version tracking much easier.

Thanks for understanding and keep up the good work!

Issue #3953, opened 2019-02-19
setunitdata doesn't change monster mode


teededung:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/915c84f10fa6840441807726cd082550e1a5a2f8
* **Client Date**: 20180621
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: setunitdata doesn't change monster mode
  * How to Reproduce: 
1. Test with this script
```
prontera,156,186,4	script	stest	857,{
	monster "prontera",156,187,"Dummy",1002,1,"stest::OnMyMobDead";
	.@GID = $@mobid[0];
	getunitdata .@GID,.@mobData;

	debugmes "Mode: " + .@mobData[UMOB_MODE]; // 131
	setunitdata .@GID,UMOB_MODE﻿,(.@mobData[UMOB_MODE] | MD_STATUS_IMMUNE | MD_DETECTOR | MD_KNOCKBACK_IMMUNE);

	getunitdata .@GID,.@mobData;
	debugmes "New Mode: " + .@mobData[UMOB_MODE];
	end;
OnMyMobDead:
	end;
}
```
2. Use a weapon with a bonus `bonus2 bAddClass,Class_Boss,25;`
3. The damage does not increase on boss monster

aleos89:
This is a two part issue.

1. The item bonus is specifically looking for CLASS_BOSS which gets applied to `addclass[CLASS_MAX]`. The only time we assign CLASS_BOSS is during mob_db parsing. The quick fix would be to add a check for the 3 modes which specifies a "boss" class and change the class during SCB_MODE calculation.
2. Because the monsters don't have the BL_REGEN flag (described in #3968 by @SeravySensei) it's not going to hit SCB_MODE.

Quick and easy fixes. I'm more leaning toward just removing the BL_REGEN check since processing should be quick enough.

Zarbony:
Then this would explain the Issue with the Abyssmal Knight Card what i describe in #3955 .
How does this can fixed?

I checked the battle folder and found MvP und Superboss.
Does this both uses the same Class_Boss oder is there any Class_MVP flag?

aleos89:
You can see the list of all mob modes in doc/mob_db_mode_list.txt. CLASS_BOSS is a series of monster modes combined into one. There is no CLASS_MVP, but there is a MVP mode which makes the monster immune to coma and gives the MVP effect.

[...]

PR #3978, opened 2019-02-24
Corrected mob class calculation with setunitdata

aleos89:
* **Addressed Issue(s)**: #3953

  * Resolved an issue with monster's class not being properly recalculated when using script command setunitdata.
  * For now, just allow switching between CLASS_NORMAL and CLASS_BOSS.
Thanks to @teededung and @SeravySensei!


teededung:
Switching between CLASS_NORMAL and CLASS_BOSS by setunitdata is fine now.

teededung:
setunitdata UMOB_SIZE and UMOB_RACE doesn't working, when I checked the monster size and race after changing size and race it was the same

Issue #3974, opened 2019-02-23
MAX_WALKPATH bug


SeravySensei:
* **rAthena Hash**: 

0c8999466

* **Client Date**: 

2018-06-20

* **Server Mode**: 

Renewal

* **Description of Issue**: 
  * Result: 
If I change #DEFINE MAX_WALKPATH to a higher or lower number, character fails to move. Depending on the number, this might be something like specific cells in a specific distance and direction becoming impossible to walk to, or if set to other numbers, for example 60, the character won't be able to walk any further than 2 tiles. Meanwhile the official number of 32, and another number I used so far, 128, works without problems.
  * Expected Result:  Any number should allow walkpaths up to that distance to work, without having any effect on the client if the number is higher than what clients normally allow. If there is a special requirement for the number (divisible by 8? 16? exponent of 2?), it should be in a comment above the definition, but ideally such a restriction should not exist.
  * How to Reproduce:  Change the constant to 60 then try to walk to further than 2 tiles of distance
  * Official Information: Not applicable
  

* **Modifications that may affect results**: 
  
   I have OFFICIAL_WALKPATH turned off. I believe I have no custom modifications in my path.cpp.  I asked on the forum to make sure it's not a bug that only affects me, but had no replies : https://rathena.org/board/topic/118293-max_walkpath/

aleos89:
I would assume there's some client limitation behind this. Like you said, it could be something limited to 2^x.

We can add a comment to state this if it's really the case.

anacondaqq:
Client limitation is 40 cells, I did test it several times a long time ago (a year ago).

SeravySensei:
Here is a specific example. Max_Walkpath of 20. From Aldebaran 140,116, I cannot move to Aldebaran 140,111. I can move to 112 and higher, but not 110 or below. So I can move no further than 5 tiles to the south. I also can't walk to 143,116, only 142,116 so in that direction, 2 tiles.
I added a breakpoint to path_seach and it triggered on clicking the unreachable tile. So the client properly sends the server the request to walk. However, path_search returns false. So it fails to find the actual path.

vstumpf:
You can use backticks (\`) as code tags, surround the code with it.
Alternatively, you can prepend all the asterisks (\*) with backslashes (\\), like so: `(MAX \* MAX)` will render as (MAX \* MAX)

anacondaqq:
[...]

PR #3985, opened 2019-02-27
Max Walkpath Fix

SeravySensei:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/3974

* **Server Mode**: 

Probably both, tested in renewal

Attempted to make the declared array large enough to have no index conflicts, as instructed by the comment. Tested for MAX_WALKPATH values of 3 and 17, both seems to work as intended, while they did not work before. 
If this solution does work, I suggest removing the fixme comment and replacing the "if e return false" part with outputting an error message, as it should never actually happen anymore. (In fact if that was known to be an exit from the function due to a bug, it should have done that in the first place instead of merely returning a "no can't walk there" result.)

### DIFF
--- a/src/map/path.cpp
+++ b/src/map/path.cpp
@@ -48,7 +48,7 @@ static BHEAP_STRUCT_VAR(node_heap, g_open_set);	// use static heap for all path 
 /// Comparator for binary heap of path nodes (minimum cost at top)
 #define NODE_MINTOPCMP(i,j) ((i)->f_cost - (j)->f_cost)
 
-#define calc_index(x,y) (((x)+(y)*MAX_WALKPATH) & (MAX_WALKPATH*MAX_WALKPATH-1))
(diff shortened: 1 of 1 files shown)


src/map/path.cpp
```diff
@@ -329,7 +329,7 @@ bool path_search(struct walkpath_data *wpd, int16 m, int16 x0, int16 y0, int16 x
 		// FIXME: This array is too small to ensure all paths shorter than MAX_WALKPATH
 		// can be found without node collision: calc_index(node1) = calc_index(node2).
 		// Figure out more proper size or another way to keep track of known nodes.
-		struct path_node tp[MAX_WALKPATH * MAX_WALKPATH];
+		struct path_node tp[(2 * MAX_WALKPATH + 1) * (2 * MAX_WALKPATH + 1)];
```
AoShinRO, line 332:
```suggestion
		struct path_node tp[(2 * MAX_WALKPATH + 1) * (2 * MAX_WALKPATH * distance_xy(x0, y0, x1, y1) + 1)];
```

Issue #2613, opened 2017-11-19
Drop rate bonuses bDropAddRace and bDropAddClass are ignored


Jeybla:
* **rAthena Hash**: a49a1a52a3e0036c9cad3794c4db5628f088b443

* **Client Date**: -

* **Server Mode**: Both

* **Description of Issue**: 
Current drop rate calculation:
```C
int drop_rate_bonus = 0;

// pk_mode increase drops if 20 level difference [Valaris]
if( battle_config.pk_mode && (int)(md->level - sd->status.base_level) >= 20 )
	drop_rate = (int)(drop_rate*1.25);

// Add class and race specific bonuses
drop_rate_bonus += sd->dropaddclass[md->status.class_] + sd->dropaddclass[CLASS_ALL];
drop_rate_bonus += sd->dropaddrace[md->status.race] + sd->dropaddrace[RC_ALL];

// Increase drop rate if user has SC_ITEMBOOST
if (&sd->sc && sd->sc.data[SC_ITEMBOOST])
	drop_rate_bonus += sd->sc.data[SC_ITEMBOOST]->val1;

drop_rate_bonus = (int)(0.5 + drop_rate * drop_rate_bonus / 100.);
// Now rig the drop rate to never be over 90% unless it is originally >90%.
drop_rate = i32max(drop_rate, cap_value(drop_rate_bonus, 0, 9000));

if (pc_isvip(sd)) { // Increase item drop rate for VIP.
	drop_rate += (int)(0.5 + drop_rate * battle_config.vip_drop_increase / 100.);
	drop_rate = min(drop_rate,10000); //cap it to 100%
}
```
Example drop rate bonus from the item_db: `bonus2 bDropAddRace,RC_All,5;`
Therefore the calculation would look like:
`drop_rate*5/100` which is always lower than `drop_rate`, so because of the i32max operation drop_rate simply does not change due to bDropAdd bonuses.

Steps to do (discussed with @cydh ):
1. initialize `int drop_rate_bonus = 0;` with 100 instead of 0.
2. Recheck and sanitize all drop rate bonuses in scripts, items, etc. (Bubble Gum needs to be lowered from 200 to 100)
3. Make the drop rate cap/drop rate vip cap configureable (battle config).
4. Refactor the code so it is reusable in `@monsterinfo`. Drop bonuses should also be displayed in `@monsterinfo` to make the bonus process more transparent.

luan122:
@cydh @aleos89
[...]

Issue #2481, opened 2017-10-06
EXP Calculation is wrong


cydh:
* **rAthena Hash**: Since its birth

* **Client Date**: Nothing to do with this

* **Server Mode**: Both

### Result: 
Current EXP calculation
1. All bonuses in percent is multiplied by monster's EXP directly
2. Guild EXP by tax is not affected by EXP modifiers

### Expected Result: 
(of course not as the result...)
* **Base_EXP** always be 100% although the server rate is 3x just keep in mind that the 100% based on the base.
* **VIP_EXP** multiplies the `Base_EXP`. If the VIP is 2x (200%) means `2 * Base_EXP`.
* **Battle_Manual** bonus multiplies the `Base_EXP` only!! If it's 50%, that means `50% * Base_EXP`
* **Equip_Bonus** multiplies the `Base_EXP` and `VIP_EXP`, so if you are wearing a pair of Bloody Rings (each adds 15% EXP) you will gain bonus exp as `30% * (Base_EXP + VIP_EXP)`
* **Event_EXP** just like VIP, this is just conditional event EXP for server. I don't think rA users will use this, I bet people will just simple adjust the base_exp.
(I said base exp is not as base level exp)

And then, if player is taxed for a guild, the EXP will be divided after all those calculation.

### How to Reproduce: 
Kill monster

### Official Information:
Play official servers, try kill a monster. Kill again while you have VIP enabled, then try consume Battle Manual, then wear EXP Rings.
Not just recent, just *lazy* to care about this. You can simply check by playing iRO's RE:Start or other 'classic renewal' official servers which has *renewal exp* disabled (no level penalty)
  

Just here's the example for RO Gravindo 'Klasik' server
* `Base_EXP` is 100%
* `VIP_EXP` is 100% of `Base_EXP`
* A Bloody Ring adds 15% of `Base_EXP + VIP_EXP`, is `15% * (100 + 100) = 30%`
* A Battle Manual with 50% of `Base_EXP`

If you want put all that percents to single calculation, will be
`100% + 100% + 30% +50% = 280%` of **Mon_BEXP**
Orc Warrior has 408 **Mon_BEXP**, it'll be = `1,142.4` and without the ring will be `1,020`
As detailed numbers `408 + 408 + 122.4 + 204`

The left is without ring, and the right side is wearing a Bloody Ring
![exp](https://cdn.discordapp.com/attachments/120596466661195778/365879402707550210/image.png)

### **Modifications that may affect results**: 
  
  
None

PR #3986, opened 2019-02-27
drop rate fix

SeravySensei:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/2613

* **Server Mode**: 

Both
Does what was described in the issue as the desired solution. Also noticed drop rate is halved for Medium size monsters while description says it should halve Small monsters, fixed that too.
mobinfo will display drop rates as current(base) is the two are not equal for any reason.
I'm unsure if the VIP bonus is additive or multiplicative, or other. Please check.


rAthenaAPI:
This pull request **introduces 1 alert** when merging 69ee584f898c97bd40ad95b4e7af2a95abad1a85 into e41ce87213043ca4c36196c9e41223ecf2e18b21 - [view on LGTM.com](https://lgtm.com/projects/g/rathena/rathena/rev/pr-db0c5cbbc94e11f2d953b5882e0802ef213f7b27)

**new alerts:**

* 1 for Comparison result is always the same

---

yashimwong:
this pull causes a nullpo error in mob.cpp:2447: in func `mob_getdroprate' when killing monsters using @killmonster command

ecdarreola:
@SeravySensei : Causes nullpo error in mob.cpp after testing this pr.

SeravySensei:
> 
> 
> @SeravySensei : Causes nullpo error in mob.cpp after testing this pr.

Aleos89 changed most of this PR for reasons only he knows, so I'll leave the debugging to him. @killmonster works fine on my system so the bug was introduced by his changes.

aleos89:
Should be good to go now. `status_kill` sends a NULL source object so `@killmonster` which uses that function was sending it as such and nullpo was just catching it.

ecdarreola:
@aleos89 question.
Let's say at base 1x the item is dropped at 50% and if changing the base rates to 2x it should be 100% right?
The cap 9000 will apply to the base? or just modifier (Using bubble gum)? I'm quite confused. Thanks!

aleos89:
Correct, adjusting the item config drop rate will ultimately increase the drop rate based on this formula: `rate = rate * rate_adjust/100;` (rate = mob_db drop rate; rate_adjust is the battle_config value). The 9000 (90%) cap is applied to the total drop rate (base + bonuses) as long as the mob_db drop is below 90%.

The system needs a bit of cleanup, as seen in #2481.

ecdarreola:
> Correct, adjusting the item config drop rate will ultimately increase the drop rate based on this formula: `rate = rate * rate_adjust/100;` (rate = mob_db drop rate; rate_adjust is the battle_config value). The 9000 (90%) cap is applied to the total drop rate (base + bonuses) as long as the mob_db drop is below 90%.
> 
> The system needs a bit of cleanup, as seen in #2481.

A little follow-up will clearly fix my confusion.

By mob_db is that original drop rate right? e.g. mob poring (1002)
`1002,PORING,Poring,Poring,1,60,1,27,20,1,8,1,2,5,6,1,1,0,6,5,10,12,1,3,21,0x83,400,1872,672,480,0,0,0,0,0,0,0,909,7000,1202,100,938,400,512,1000,713,1500,512,150,619,20,0,0,0,0,4001,1
`

The base drop (rate) is 70% so if I just modify it on the battle > drop.conf (rate_adjust) to 3x, it should be 100%. However since this PR fixes something, it should only be capped at 90%?

SeravySensei:
According to my understanding the drop.conf multiplier is part of the base drop rate. So in that case the drop rate remains 100%. The cap only applies to modifiers from in-game effects like bubble gum. I might remember that wrong though, I worked on this PR half a year ago.

yashimwong:
@aleos89 
crash

src/map/mob.cpp
```diff
@@ -2416,6 +2416,81 @@ void mob_damage(struct mob_data *md, struct block_list *src, int damage)
...
+	}
+
+	if (src) {
+		if (battle_config.drops_by_luk) // Drops affected by luk as a fixed increase [Valaris]
+			drop_rate += status_get_luk(src) * battle_config.drops_by_luk / 100;
+		if (battle_config.drops_by_luk2) // Drops affected by luk as a % increase [Skotlex]
+			drop_rate += (int)(0.5 + drop_rate * status_get_luk(src) * battle_config.drops_by_luk2 / 10000.);
+	}
+
+	if (src->type == BL_PC) { // Player specific drop rate adjustments
```
yashimwong: This check block shld be added under if (src) to check whether the src has a value before doing any other checks

ecdarreola: What's the fix for this? just add nullptr?

SeravySensei: Done, moved the end of the if (src) block to the right position

db/pre-re/item_db.txt
```diff
@@ -4829,7 +4829,7 @@
 12261,Secret_Medicine,Leap of Fantasy,2,0,,0,,,,,0xFFFFFFFF,7,2,,,,,,{ getexp2(2000000,1000000); },{},{}
 12262,Inspector_Certificate_,Authoritative Badge,2,0,,10,,,,,0xFFFFFFFF,7,2,,,,,,{ sc_start SC_SPEEDUP0,540000,25; },{},{}
 12263,Comp_Battle_Manual,Field Manual,2,2,,10,,,,,0xFFFFFFFF,7,2,,,,,,{ sc_start SC_EXPBOOST,1800000,50; },{},{}
-12264,Comp_Bubble_Gum,Bubble Gum,2,2,,10,,,,,0xFFFFFFFF,7,2,,,,,,{ sc_start SC_ITEMBOOST,1800000,200; },{},{}
+12264,Comp_Bubble_Gum,Bubble Gum,2,2,,10,,,,,0xFFFFFFFF,7,2,,,,,,{ sc_start SC_ITEMBOOST,1800000,50; },{},{}
```
aleos89, line 4832: Shouldn't this be 100? Just curious where 50 came from. The description states:
```
30분간 몬스터가 아이템을 드롭할 확률이 100% 증가한다.
```

db/re/item_db.txt
```diff
@@ -6491,7 +6491,7 @@
 12261,Secret_Medicine,Leap of Fantasy,2,0,,0,,,,,0xFFFFFFFF,63,2,,,,,,{ getexp2(2000000,1000000); },{},{}
 12262,Inspector_Certificate_,Authoritative Badge,2,0,,10,,,,,0xFFFFFFFF,63,2,,,,,,{ sc_start SC_SPEEDUP0,540000,25; },{},{}
 12263,Comp_Battle_Manual,Field Manual,2,2,,10,,,,,0xFFFFFFFF,63,2,,,,,,{ sc_start SC_EXPBOOST,1800000,200; },{},{}
-12264,Comp_Bubble_Gum,Bubble Gum,2,2,,10,,,,,0xFFFFFFFF,63,2,,,,,,{ sc_start SC_ITEMBOOST,1800000,200; },{},{}
+12264,Comp_Bubble_Gum,Bubble Gum,2,2,,10,,,,,0xFFFFFFFF,63,2,,,,,,{ sc_start SC_ITEMBOOST,1800000,50; },{},{}
```
aleos89, line 6494: Same as the pre-re database change.

db/pre-re/item_db.txt
SeravySensei: Unfortunately, this PR is more than a year old. I don't remember but I have to guess the item effect changed meanwhile or was mistranslated at the time I made this. Change it to what the correct number currently is please.

Masao87:
How save is it to pull this request already in its current state? Should we rather wait until you guys put it into master?

aleos89:
It's ready for review by any other dev.

Masao87:
I guess this would need some adjustment now with the convertion to the yaml database?

Masao87:
I guess this commit is obsolete by now and can be deleted? @aleos89

ecdarreola:
This still works if needed on your end.

aleos89 (approved):

db/re/item_db_usable.yml
```diff
@@ -6704,7 +6704,7 @@ Body:
       NoMail: true
       NoAuction: true
     Script: |
-      sc_start SC_ITEMBOOST,1800000,200;
+      sc_start SC_ITEMBOOST,1800000,100;
```
Atemo, line 6707: +200% for 30mins
```suggestion
      sc_start SC_ITEMBOOST,1800000,200;
```

Atemo (changes requested):

Issue #1310, opened 2016-05-26
Battleground Queue System


secretdataz:

admkakaroto:
similar to that eAmod?

anacondaqq:
I guess he means battleground button in alt+v interface, and all system related to it.

Everade:
Would be better to get BG fixed in the first place, before implementing the queue system.
https://github.com/rathena/rathena/issues/1240

iubantot:
any news on this feature? <3

reigneil:
info extracted in msgstring:
`Entering the battleground registering system#
Not possible to register entering and move to the battleground#
1. It is not possible to register several battlegrounds at the same time.#
2. You cannot register the battleground for Solo / Party / Guild at the same time.#
3. Party battleground is only available for registration from the party leader,#
the party members who are offline will miss the queue.#
4. Registering the battleground is possible from everywhere except the battleground map,#
if the entering location is not a town, field or normal dungeon, character will be transferred to saved coordinates when returning.#
5. the rewards of battleground or other services can be found from battleground waiting room through the recruiting officer.#
Entering the battleground registering Help word#`

other info related to battleground queue system:
https://github.com/zackdreaver/ROenglishRE/blob/master/data/luafiles514/lua%20files/battlefield/battlefieldinfo.lub

iubantot:
up

Daegaladh:
Any news on this? I'd like to help, to speed things up

aleos89:
This was on hold because we wanted to utilize the C++ queue class rather than implementing another common library that we have to maintain. It requires a bit more work to convert `script.c` to `script.cpp`. Once that's done it should be relatively easy to implement the rest of the feature.

luan122:
@aleos89 will this come back? i think you guys already converted script.c to script.cpp right?

playpsro:
Any update for this bg?

PR #3995, opened 2019-03-02
Implemented the Battleground Queue System

aleos89:
* **Addressed Issue(s)**: #1310

  * Added a battle config which allows toggling of the Battleground Queue client feature and also toggles between the modes of scripts.
  * Added script commands bg_reserve, bg_unbook, and bg_desert.
  * Removed old custom Battleground scripts.
Thanks to @Lemongrass3110, @Locien, and @secretdataz!


rAthenaAPI:
This pull request **introduces 5 alerts** when merging 975def7be01d3ecbdf3fd1c3a347e8ff49ee247f into 36d47c70e4c9f62c37bdd2295eed566e3a776465 - [view on LGTM.com](https://lgtm.com/projects/g/rathena/rathena/rev/pr-f841328608562f48e7427968b86bb8822a059916)

**new alerts:**

* 5 for Catching by value

---

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+{
+	struct party_data *p = party_search(sd->status.party_id);
+
+	if (!p)
+		return BG_APPLY_INVALID_APP; // Someone has bypassed the client check for being in a party
+
+	uint32 pl_cid = 0;
+
+	for (const auto &it : p->party.member) {
+		if (it.leader) {
```
InusualZ: Checking for the party guild leader should be moved up here like so
```suggestion
		if (it.leader) {
		    if (sd->status.char_id != it.char_id) {
		        return BG_APPLY_PARTYGUILD_LEADER;
		}
```

This would eliminate the need of `pl_cid` which is only used once

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+		}
+	}
+
+	if (sd->status.char_id != pl_cid)
+		return BG_APPLY_PARTYGUILD_LEADER; // Not the party leader
+	else {
+		int p_online = 0;
+
+		if (auto bg = bg_search_name(name).lock()) {
+			for (const auto &it : p->party.member) {
```
InusualZ: This for loop can be easily replaced with an `std::count_if`
```suggestion
			int player_online = std::count_if(p->party.member.begin(), p->party.member.end(), [](const auto& member) { return member.online; })
```

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+				if (it.online)
+					p_online++;
+			}
+
+			if (p_online > bg->required_players)
+				return BG_APPLY_PLAYER_COUNT; // Too many party members
+			else {
```
InusualZ: This else clause is unnecessary. The if statement would return the function...

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+				std::vector<struct map_session_data *> list;
+
+				for (const auto &it : p->party.member) {
+					if (it.online) {
+						struct map_session_data *pl_sd = map_charid2sd(it.char_id);
+
+						if (pl_sd && list.size() < bg->required_players)
+							list.push_back(pl_sd);
```
InusualZ: When the list if full you should consider breaking the loop.

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+	if (strcmp(sd->status.name, sd->guild->master) != 0)
+		return BG_APPLY_PARTYGUILD_LEADER; // Not the guild leader
+	else {
+		struct guild *g = guild_search(sd->status.guild_id);
+		std::vector<struct map_session_data *> list;
+
+		if (auto bg = bg_search_name(name).lock()) {
+			int g_online = 0;
+
+			for (const auto &it : g->member) {
```
InusualZ: Again, this for loop can be replaced with `std::count_if`

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+			}
+
+			if (g_online > bg->required_players)
+				return BG_APPLY_PLAYER_COUNT; // Too many guild members online
+
+			for (const auto &it : g->member) {
+				if (it.online) {
+					struct map_session_data *pl_sd = map_charid2sd(it.char_id);
+
+					if (pl_sd && list.size() < bg->required_players)
```
InusualZ: Again, consider breaking the loop when the list is full.

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+	if (bg->max_lvl && sd->status.base_level > bg->max_lvl)
+		return BG_APPLY_PLAYER_LEVEL; // Level too high
+
+	if (bg_queues.empty()) {
+		std::shared_ptr<s_battleground_queue> queue = bg_queue_create(bg);
+		bool r = rnd() % 2 != 0;
+
+		if (!r) {
+			while (!list.empty() && queue->teama_members.size() < bg->required_players) {
+				struct map_session_data *sd2 = list.back();
```
InusualZ: What's the reasoning behind not using `list.pop_back()` here instead of accessing through `list.back()`?

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+				continue;
+			}
+
+			if (queue->in_ready_state)
+				continue;
+
+			if (queue->teama_members.size() + list.size() <= bg->required_players || queue->teamb_members.size() + list.size() <= bg->required_players) { // Make sure there's enough space on one side to join as a party/guild in this queue
+				if (!r && queue->teama_members.size() + list.size() <= bg->required_players) {
+					while (!list.empty() && queue->teama_members.size() < bg->required_players) {
+						struct map_session_data *sd2 = list.back();
```
InusualZ: Again, what's the reasoning behind not using list.pop_back() here instead of accessing through list.back()? Ditto elsewhere

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+				sd->bg_queue->in_ready_state = false;
+				sd->bg_queue_accept_state = false;
+			}
+
+			util::erase_at(list, i);
+
+			if (list.empty() && listb.empty()) { // If there are no players left in the queue, discard it
+				std::vector<std::shared_ptr<s_battleground_queue>>::iterator it = bg_queues.begin();
+
+				while (it != bg_queues.end()) {
```
InusualZ: You should replace all of this while loop with an `std::remove_if()`, Like
```suggestion
				std::remove_if(bg_queues.begin(), bg_queues.end(), [](const auto& q) {
		                    if (sd->bg_queue == q) {
				        if (q->tid_requeue && get_timer(q->tid_requeue)) {
					    delete_timer(q->tid_requeue, bg_on_ready_loopback);
					    q->tid_requeue = 0;
			                }
			                it = bg_queues.erase(it);
			            }
				    return true;
				});
```

src/map/battleground.cpp
```diff
@@ -222,59 +621,863 @@ int bg_team_get_id(struct block_list *bl)
...
+ */
+bool bg_queue_leave(struct map_session_data *sd)
+{
+	if (!sd || !sd->bg_queue)
+		return false;
+
+	if (!bg_queue_leave_sub(sd, sd->bg_queue->teama_members, sd->bg_queue->teamb_members) || !bg_queue_leave_sub(sd, sd->bg_queue->teamb_members, sd->bg_queue->teama_members)) {
+		ShowError("bg_queue_leave: Fatal error. Couldn't find player %s in battlegrounds queue.\n", sd->status.name);
+		return false;
+	} else
```
InusualZ: Else clause is unnecessary

aleos89:
@InusualZ Thanks for the reviews! For random, we are using mt19937 provided by C++. You can check it out in src/common/random.cpp. :)

As for shared_ptr, we sort of stuck with them when refactoring other parts of the code for safe construction and deconstruction. We can definitely watch for performance changes.

src/map/battleground.cpp
aleos89: This won't work since party member data is not a vector or something that contains iterator form (begin, end). This could definitely be something refactored in the future when party is converted to C++.

src/map/battleground.cpp
aleos89: This if statement is actually at fault. This should be:
```cpp
if (p_online > bg->max_players)
```

src/map/battleground.cpp
aleos89: My understanding is that `pop_back()` returns void which is why we have to use `back()` to access before hand.

src/map/battleground.cpp
aleos89: This and the `count_if` I'm 50/50 on. While I understand their uses, might be better to get more opinions on this from other developers.

src/map/battleground.cpp
InusualZ: Yep, you are right. For some reason I thought the contrary.

src/map/battleground.cpp
```diff
@@ -222,59 +624,853 @@ int bg_team_get_id(struct block_list *bl)
...
+ * Leave a Battleground queue
+ * @param sd: Player data
+ * @return True on success or false otherwise
+ */
+bool bg_queue_leave(struct map_session_data *sd)
+{
+	if (!sd || !sd->bg_queue)
+		return false;
+
+	if (!bg_queue_leave_sub(sd, sd->bg_queue->teama_members, sd->bg_queue->teamb_members) || !bg_queue_leave_sub(sd, sd->bg_queue->teamb_members, sd->bg_queue->teama_members)) {
```
Atemo: mapserv is crashing when the user press 'cancel' to leave the queue

iubantot:
im having this error after changing the min players to 1 is this normal? my map server crashes
image

admkakaroto:
If you enter the queue some strange characters will appear in the dialog, if you click 'cancel' the game crashes and the map_server crashes. You may need some updating.

aleos89:
Can you post your crash dump?

noobzter003:
screenrAthena000
Help me!!

noobzter003:
Untitled

admkakaroto:
If you register on a battlefield, then refresh or if you click the bg icon again, it looks like this:

before:
a2


after:
a1

RagnaWay:
> If you register on a battlefield, then refresh or if you click the bg icon again, it looks like this:
> 
> before:
> a2
> 
> after:
> a1

Same problem with me

violent01:
@admkakaroto  @RagnaWay its not a problem. What you are seeing on the first try is the first pop up window which tells you what bg you are queueing while the second pop up window is the info for the remaining players need to join the queue.

violent01:
@aleos89  i think there are errors on the script part of flavius, KVM, tiera. Below is a screenshot of it.
error bg

aleos89:
@violent01 I didn't have that problem. Make sure you merged properly. Looks like you didn't define `bg_leave` / `bg_desert`.

src/common/utilities.hpp
```diff
@@ -133,6 +146,7 @@ namespace rathena {
 				vector.clear();
 			else
 				vector.erase(vector.begin() + index);
+      vector.shrink_to_fit();
```
InusualZ: I don't think this is a good idea. You are reallocating the memory, every time you erase something. That could be heavy...

PR #4021, opened 2019-03-18
Fix wrong `next` function call

keitenai:
* **Addressed Issue(s)**: 

* **Server Mode**: renewal

Fix script error "npc_scriptcont: Guard Leader#buwaya (sd->npc_id=110011242) is not 'Unknown NPC' (id=0)

### DIFF
--- a/npc/re/quests/quests_malaya.txt
+++ b/npc/re/quests/quests_malaya.txt
@@ -4589,9 +4589,9 @@ malaya,290,340,3	script	Guard Leader#buwaya	571,{
 		next;
 		mes "[Guard Leader]";
 		mes "We are doing our best in securing the safety of our village so please enjoy your stay in Port Malaya.";
+		next;
 	}
 	if (malaya_buwaya == 10) {
-		next;
 		if (select("Tell him about Buwaya.:Done with conversation.") == 2) {
 			close;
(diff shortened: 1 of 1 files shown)


sader1992:
if this causes that warning , wouldn't be also better to ignore the next command when there is no mes active ?

aleos89:
I don't think we want to suppress this type of warning as it would promote bad scripting behavior.

sader1992:
i agree , but this warning happens a lot , so by not making a custom message for it or just remove it at this case , it will get mixed with other and more important errors with the same message in my opinion

secretdataz:
More annoying messages = More server owners asking their scripter to fix the problematic script.

sader1992:
yes but people will get an error anyway
i would much rather someone asking me to fix this error
```
buildin_next: Guard Leader#buwaya 'mes' is not active.
```
or
```
buildin_clear: Guard Leader#buwaya 'mes' is not active.
```
instead of this 
```
npc_scriptcont: Guard Leader#buwaya (sd->npc_id=110011242) is not 'Unknown NPC' (id=0)
```

secretdataz:
I agree with your point. A pull request is welcome and I'd be happy to review it.

PR #4043, opened 2019-03-26
Implementing atcommand_disable_npc config

sader1992:
ignore atcmmand if player with group id is attached to npc

* **Server Mode**: both

* most of the warning where `sd->npc_id != npc_id`
this can happen when the player is attached to npc (uses sleep or timer) that use @refresh or other command like script commands from bindatcmd and it deattach the player from the npc

* ps: the player can use atcommand by making it as shortcut on the (Alt+M) menu ingame and this patch will ignore the atcommand using this trick
some npcs after it do x use sleep to do y > if player deattach before y than it's bug
ofc the player can disconnect from the server , while i don't know how to prevent that i can prevent the atcommands

I defaulted it to 0 cuz players are used to using @refresh and @load etc while stuck on npc
i think the official should be 99 as normal players does not have @commands enabled

* this PR is not done yet
as it should allow `*atcommand "<command>";` command script to progress
i will add support for that if the developers agree about it

* Update:
while checking , i found this
https://github.com/rathena/rathena/blob/9f4a569ecc0af179e6926c2b31062b2fff3bd771/doc/script_commands.txt#L5357
however it's not implemented in the src
maybe it would be a better idea ?

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -4235,7 +4235,7 @@ void script_attach_state(struct script_state* st){
 		sd->st = st;
 		sd->npc_id = st->oid;
 		sd->npc_item_flag = st->npc_item_flag; // load default.
-		sd->state.disable_atcommand_on_npc = (!pc_has_permission(sd, PC_PERM_ENABLE_COMMAND));
+		sd->state.disable_atcommand_on_npc = battle_config.atcommand_disable_npc;
 #ifdef SECURE_NPCTIMEOUT
 		if( sd->npc_idle_timer == INVALID_TIMER )
 			sd->npc_idle_timer = add_timer(gettick() + (SECURE_NPCTIMEOUT_INTERVAL*1000),npc_secure_timeout_timer,sd->bl.id,0);
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -4235,7 +4235,7 @@ void script_attach_state(struct script_state* st){
 		sd->st = st;
 		sd->npc_id = st->oid;
 		sd->npc_item_flag = st->npc_item_flag; // load default.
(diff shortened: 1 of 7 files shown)


secretdataz:
Real solution: stop abusing sleep2 and use timers instead

sader1992:
i discovered this while seen the error from the official npcs
most of the new npc uses that
https://github.com/rathena/rathena/blob/master/npc/re/quests/quests_16_1.txt

players get sleep2 sometimes think they are stuck and do refresh or other commands
while in the official does not allow that as far as i know

sader1992:
after sreaching , i found atcommand_enable_npc config in rathena but it's not fully implemented
https://github.com/rathena/rathena/blob/9f4a569ecc0af179e6926c2b31062b2fff3bd771/doc/script_commands.txt#L5357
i will change this PR to implement `atcommand_enable_npc`

sader1992:
Done.

src/map/script.cpp
```diff
@@ -4235,7 +4235,7 @@ void script_attach_state(struct script_state* st){
 		sd->st = st;
 		sd->npc_id = st->oid;
 		sd->npc_item_flag = st->npc_item_flag; // load default.
-		sd->state.disable_atcommand_on_npc = (!pc_has_permission(sd, PC_PERM_ENABLE_COMMAND));
+		sd->state.disable_atcommand_on_npc = battle_config.atcommand_disable_npc;
```
Lemongrass3110: Probably better to put this into an "and" relation?
For example a server owner could want to disable it for all players by default, but wants to use it himself or allow it for GMs.
Like this:
```
sd->state.disable_atcommand_on_npc = battle_config.atcommand_disable_npc && (!pc_has_permission(sd, PC_PERM_ENABLE_COMMAND));
```
This would require to change the default option to "false" in groups.conf as well though.

Issue #4041, opened 2019-03-26
token of Siegfried not working. Roulette version.


daisyanne1380:
* **rAthena Hash**: 

* **Client Date**: 
2018-06-20e

* **Server Mode**: 
Renewal

* **Description of Issue**: 
  * Result: 
i get the resurrect menu but when i click on it, it does nothing. I'm using client 2018-0620e.  I didn't diff skip resurrect button either. And im using a svn from about 2 weeks ago so fairly new.

  * Expected Result:
to revive.

  * How to Reproduce:  
Ok i figured out whats wrong, svn is set to revive if you have token of Siegfried id 7621 But roulette give you a non trade-able version id 6316. I think both id should be added to the src.

  * Official Information:

* **Modifications that may affect results**: 
  Havent made any modifications

Normynator:
Solved: https://rathena.org/board/topic/118604-is-token-of-siegfried-not-working/

daisyanne1380:
hello, i apply this fix and now it does work BUT i notice that it doesn't take the token away from you when you revive. thanks

daisyanne1380:
hello did anyone else verify that the item isnt removed from inventory upon death and revival

aleos89:
I was unable to reproduce your issue. Make sure you pulled the latest version from the PR.

daisyanne1380:
yea thats odd it does work but doesn't delete the token for me upon use.

PR #4047, opened 2019-03-26
Added an item group for Token of Siegfried

aleos89:
* **Addressed Issue(s)**: #4041

  * Created an item group used to remove hard coding of Token of Siegfried and add the others to work with this feature.
Thanks to @daisyanne1380!


rAthenaAPI:
This pull request **introduces 2 alerts** when merging f2f0a5e2e80dc5c86879bbbebe0de6550e5cb2db into 9f4a569ecc0af179e6926c2b31062b2fff3bd771 - [view on LGTM.com](https://lgtm.com/projects/g/rathena/rathena/rev/pr-e10422e3a37db1704f85aeea08e3e281fbb52e79)

**new alerts:**

* 2 for Comparison result is always the same

---

mrjnumber1:
i do like this work. i realize that this is sort of off topic for this PR,  but there are other things which would benefit from being placed in item groups as well due to how they are used in the source. the ones i can think of are (and i realize some of these may already be grouped!):

- Magnifiers
- Fly Wings
- Butterfly Wings/Teleport Scrolls
- Bubble Gum
- EXP/JEXP Manuals
- Mercenary Potions
- Slug Shot Slugs
- Trap/Alloy Traps
- Weapon Repair/Refine items in skill level order (instead of them being hard coded)
- Item produces probably would be better to work with in their own yaml dbs... who knows the best way for now, it's out of the scope of this.
- Cooking Mix failures

in general, the fewer item ids hardcoded to the map server, the better, of course. 

carry on!

PR #4055, opened 2019-03-27
update aldebaran.txt

userid0:
* **Addressed Issue(s)**: 

* **Server Mode**:  All

* **Description of Pull Request**: Add missing script


Atemo:
That follow the others warps definition but actually the facing is ignored during the parsing.
Anything else?

PR #4062, opened 2019-03-28
Added several warning messages

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

# To quote the commits
- [X] `areamonster` now checks if the monster ID entry exists.
- [X] `enablenpc` / `disablenpc` / `hideoffnpc` / `hideonnpc` now display the npc source when the npc name (param) doesn't exist.
- [X] Added clif_name_area to `UMOB_LEVEL` on `setunitdata` and a capvalue to `UPET_HUNGER`.
- [X] Added the NPC source to status_set_viewdata when no view data has been found.
- [X] Added `export_deprecated_constant3` in `script_constants.hpp` to display the constant name replacing the deprecated.
- [X] Removed the debug message from `@warp` when the map name is wrong.


src/map/script.cpp
```diff
@@ -481,12 +482,11 @@ static void script_dump_stack(struct script_state* st)
...
 {
-	struct block_list* bl;
-
 	if( st->oid == 0 )
 		return; //Can't report source.
 
-	bl = map_id2bl(st->oid);
+	struct block_list* bl = map_id2bl(st->oid);
+
 	if( bl == NULL )
```
Normynator: Redundant nullptr check

src/map/atcommand.cpp
```diff
@@ -504,11 +504,11 @@ ACMD_FUNC(mapmove)
 			return -1;
 	}
 
-	mapindex = mapindex_name2id(map_name);
+	mapindex = mapindex_name2idx(map_name, NULL);
```
Normynator: Better use nullptr instead of NULL

src/map/script.cpp
```diff
@@ -210,6 +210,7 @@ static struct str_data_struct {
 	int (*func)(struct script_state *st);
 	int val;
 	int next;
+	const char *deprecated_name;
 	bool deprecated;
 } *str_data = NULL;
```
Normynator: I know its not your edit but using a nullptr here would be great :) <3

PR #4070, opened 2019-04-02
Fixes #4066

### LINKED ISSUE #4066: Devotion Inherits Status
* **rAthena Hash**: https://github.com/rathena/rathena/commit/ac558d7c1e58c817744968f70f487c3d3d16aac2
* **Client Date**: 20180621
* **Server Mode**: RE
* **Description of Issue**: The effect of SC_DEFENDER is being added twice when inherited from Devotion.
  * Result: Double damage reduction.
  * How to Reproduce: Activate SC_DEFENDER, use Devotion on a party member, now disable SC_DEFENDER and re-enable it, that's it, you'll be double the dev status against long-range attacks.
  * Prints:
Enabled SC_DEFENDER first time:

a2

DMG: 2300

Second time with SC_DEFENDER enabled:
a3

DMG: 460

The problem can only be reproduced if you enable SC_DEFENDER before using Devotion, then deactivate and activate SC_DEFENDER again.

cydh:
* **Addressed Issue(s)**: #4066

* **Server Mode**: Both

  * Fixed Defending Aura effect taking twice
  * Removed incorrect Devotion behavior (leftover from old fixes)

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -1301,22 +1301,6 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 				status_change_end(bl,SC_VOICEOFSIREN,INVALID_TIMER);
 		}
 
-		if( sc->data[SC_DEVOTION] ) {
-			struct status_change_entry *sce_d = sc->data[SC_DEVOTION];
-			struct block_list *d_bl = map_id2bl(sce_d->val1);
-
-			if( d_bl &&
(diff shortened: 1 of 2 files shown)


admkakaroto:
Sorry but it still has not resolved, the problem persists, test using an RK for example, as in the images I posted in the issue.

to reproduce you need to enable defending first, then use devotion on the party member, disable and activate defender, and you will get twice the def ranged.

cydh:
https://youtu.be/VMoU3oC81f4

admkakaroto:
Sorry I'm unleashing, I forgot to do something here, now it worked perfectly, sorry.

PR #4079, opened 2019-04-04
Fixed battle.cpp blocks

keitenai:
* **Addressed Issue(s)**: expanding / collapsing block is broken.

* **Server Mode**: ANY

fixes issue with folding (collapses)/ expanding Blocks


Lemongrass3110:
Sorry, but this will be rejected until we decided for a styling for all source files.

PR #4080, opened 2019-04-04
Fixed skill.cpp blocks

keitenai:
* **Addressed Issue(s)**:  expanding / collapsing block is broken.

* **Server Mode**: ANY

fixes issue with folding (collapses)/ expanding Blocks


Lemongrass3110:
Sorry, but this will be rejected until we decided for a styling for all source files.

anacondaqq:
@Lemongrass3110 clang-format vs default will be great.
With small rule for battle_config flags (otherwise they will look too long).
Anyway, this is nice to see that you thinks about one style for all code.

Normynator:
Its kinda a hard decision to take. Code style depends only on personal perferences, there is (more or less) no right or wrong.
I also hope for a single codestyle for all files.

Issue #4069, opened 2019-04-01
cap value missing in reseffect


Litro:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/8580287286405a2a21ba21d604d71e2a2ae32485

* **Client Date**: any

* **Server Mode**: any

* **Description of Issue**: 
  * Result: value over `10000`
  * Expected Result: shouldn't it capped 
  * How to Reproduce: change max limit your stats to 250, and try use 3 orc hero card, with vit 100 and luk 200, mini stun will occur 
  * Official Information:
  

* **Modifications that may affect results**:  change max stats
  
  

it was linking to this commit https://github.com/rathena/rathena/commit/c71ef92fb0aa50119484d5576853eb0c79baa397#diff-49937e616f59c440ac7bc7d0d5ec2237R3267
```
-		i = sd->reseff[type2]+val;
-		sd->reseff[type2]= cap_value(i, -10000, 10000);
+		if (sd->reseff.size() == MAX_PC_BONUS) {
+			ShowWarning("pc_bonus2: SP_RESEFF: Reached max (%d) number of resistance bonuses per character!\n", MAX_PC_BONUS);
+			break;
+		}
+
+		val = cap_value(val, -10000, 10000);
+		pc_bonus_itembonus(sd->reseff, type2, val);
```

PR #4088, opened 2019-04-06
cap value of resistance effect from item

Litro:
* **Addressed Issue(s)**: #4069

* **Server Mode**: both

* **Description of Pull Request**: cap value of resistance effect from item


aleos89:
This is the same as #4081. Thanks for your contribution!

PR #4090, opened 2019-04-06
Cleaned up instance script commands

aleos89:
* **Addressed Issue(s)**: N/A

  * Script command instance_id now supports an instance mode argument. If no mode is given it will return the attached script's instance.
  * Script commands instance_destroy, instance_npcname, instance_mapname, instance_announce now attach themselves to the script's instance by default.
  * Script commands instance_enter and instance_warpall attach themselves to the player's IM_PARTY instance by default.
  * Update documentation to match changes.
Thanks to @vykimo and @Atemo!

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -378,7 +378,7 @@ static struct linkdb_node *sleep_db; // int oid -> struct script_state *
  *------------------------------------------*/
 const char* parse_subexpr(const char* p,int limit);
 int run_func(struct script_state *st);
-unsigned short script_instancegetid(struct script_state *st, enum instance_mode mode);
+unsigned short script_instancegetid(struct script_state *st, enum instance_mode mode = IM_NONE);
 
 const char* script_op2name(int op)
 {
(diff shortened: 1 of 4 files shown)


src/map/script.cpp
```diff
@@ -19921,7 +19945,19 @@ BUILDIN_FUNC(instance_mapname)
  *------------------------------------------*/
 BUILDIN_FUNC(instance_id)
 {
-	script_pushint(st, script_instancegetid(st));
+	enum instance_mode mode = IM_NONE; // Default to the attached NPC
+
+	if (script_hasdata(st, 2)) {
+		mode = static_cast<instance_mode>(script_getnum(st, 2));
```
Lemongrass3110: Unspecified C++ behavior if the value is not in the range of the existing enum values.

src/map/script.cpp
```diff
@@ -378,7 +378,7 @@ static struct linkdb_node *sleep_db; // int oid -> struct script_state *
  *------------------------------------------*/
 const char* parse_subexpr(const char* p,int limit);
 int run_func(struct script_state *st);
-unsigned short script_instancegetid(struct script_state *st);
+unsigned short script_instancegetid(struct script_state *st, enum instance_mode mode);
```
Lemongrass3110: Please provide the new parameter with a default value.
This will allow revert of a few calls to this function as well.

PR #4092, opened 2019-04-08
Ensure the mysql server host support full length domain

CairoLee:
* **Server Mode**:  Both

rAthena is support use the domain to connect mysql server, but some cloud services (like RDS in ALIYUN from China), they will provide a supper long domain for you (far more than 32 bytes), our mysql server host ip buffer is too small, use the RDS domain will be truncated.

So I think we can extend it for support full length domain.

### DIFF
--- a/src/char/inter.cpp
+++ b/src/char/inter.cpp
@@ -45,7 +45,7 @@ InterServerDatabase interServerDb;
 Sql* sql_handle = NULL;	///Link to mysql db, connection FD
 
 int char_server_port = 3306;
-char char_server_ip[32] = "127.0.0.1";
+char char_server_ip[64] = "127.0.0.1";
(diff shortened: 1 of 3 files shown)


aleos89:
Good for a temporary solution. Best results would be to finally move these to string format.

CairoLee:
@aleos89 Thank you for approved. 

Do you mean std::string? I have considered it, but i think that requires a lot of changes. I don't have the confidence to modify so much content at once. 

Because... 

My English is not very well. Although I hope to communicate with experts like you, but the more discussion the more difficult i feel. I am worried that I can't accurately convey my meaning and waste everyone's precious time.

Power by Google Translation. XD

aleos89:
This PR is fine the way it is.

The change I'm suggesting is for later down the line for rAthena devs to work on. :)

PR #4094, opened 2019-04-08
Corrected guild level increases

aleos89:
* **Addressed Issue(s)**: N/A

  * Resolves guild levels being capped to various hard coded limits rather than using the constant.
Thanks to @datleeroy!


src/char/int_guild.cpp
```diff
@@ -827,9 +827,12 @@ bool guild_check_empty(struct guild *g)
 
 unsigned int guild_nextexp(int level)
 {
-	if (level == 0)
+	if (!level)
```
InusualZ: I really think this change should be reverted

PR #4097, opened 2019-04-10
init guidelines.md

Normynator:
* **Addressed Issue(s)**: 
Missing code guidelines for rAthena

* **Server Mode**: Both

Initial base for guidelines.

### DIFF
--- a/guidelines.md
+++ b/guidelines.md
@@ -1,0 +1,156 @@
+# Welcome
+Welcome to the rAthena Code Guidelines. 
+This is here to help all developers in creating safe, maintainable, easily readable and understandable code.
+
+# References
+These guidelines are inspired by the CppCoreGuidelines. Some of the examples below maybe taken directly from there.
+If you need more information that may not be listed here, please be sure to check out [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md).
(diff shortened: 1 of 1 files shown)


guidelines.md
```diff
@@ -0,0 +1,156 @@
...
+
+##### Reason
+
+Commits should be on point, documented and as small as possible.
+
+##### Enforcement
+
+Check for removed or added blank lines on each commit.
+
+### 2. Use std::remove_if, std::remove & ... within std::vector::earse
```
InusualZ: The are typo on the word `erase`, in this line and in the comment below

guidelines.md
```diff
@@ -0,0 +1,156 @@
...
+
+##### Example
+
+    vec.erase(std::remove_if(vec.begin(), vec.end(), [](int a) {return a % 2 == 0;}), vec.end()); // good
+
+##### Enforcement
+
+Check for `std::remove` on datatype `std::vector` without being surrounded by an `erase`.
+
+### 3. Don't add redundant `==` or `!=` to conditions
```
InusualZ: I really think that the opposit should be enforced. If you guys want to move to C++, there are going to be a lot of variable with type `auto`. So being explicit in the comparison that you are trying to accomplish would give you a hint in the type of variable that you are dealing with. Even more so, the type could override the `operator bool` and that would complicate things a little bit more...

Normynator: ``auto`` is only used if the type is obvious. So the type would be clear in following ``if-statements``.
If the type overrides the ``operator bool`` then its not redundant.

PR #4099, opened 2019-04-11
Reduced MAX_MAP_SERVERS and MAX_SERVERS size

aleos89:
* **Addressed Issue(s)**: N/A

  * Reduced MAX_MAP_SERVERS down to 2.
  * Reduced MAX_SERVERS down to 5.

### DIFF
--- a/src/char/char.hpp
+++ b/src/char/char.hpp
@@ -193,7 +193,7 @@ struct CharServ_Config {
 };
 extern struct CharServ_Config charserv_config;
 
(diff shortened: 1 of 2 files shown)


luan122:
any reason to this?

aleos89:
Many private servers out there aren't using more than 1 map server, let alone anything close to 30. It's more common for having a single login server connect to multiple character servers though which is why I reduced that to 5, and once again nothing close to 30 is being used.

Issue #4087, opened 2019-04-06
npc unitwarp and reloadscript


Litro:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/fded1aa114e5cb0e321d4ace4a374440bcb07122

* **Client Date**: any

* **Server Mode**: any

* **Description of Issue**: 
  * Result: `[Warning]: npc_reload: 8 npcs weren't removed at map prontera!`
  * Expected Result: npc should be removed
  * How to Reproduce: try `@warpthatnpc` 3 times in game and see map console
```
prontera,150,150,0	script	unitwarp test	112,{
	end;
	
OnInit:
	if (strnpcinfo(2) == "") {
		bindatcmd "warpthatnpc",strnpcinfo(3) + "::OnAtcommand",99;
		setarray .maps_list$, "izlude", "geffen", "morocc";
	}
	else {
		.gid[.count] = getnpcid(0);
		.count++;
	}
	end;
	
OnAtcommand:
	.@dest_map$ = .maps_list$[rand(getarraysize(.maps_list$))];
	for (.@i = 0; .@i < .count; .@i++) {
		unitwarp .gid[.@i], .@dest_map$, -1, -1;
		getunitdata .gid[.@i], .@data;
		announce "NPC Coordinat "+.@i+" ->  map : "+.@dest_map$+" x : "+.@data[UNPC_X]+", y : "+.@data[UNPC_Y], 0;
	}
	end;
}

prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#01	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#02	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#03	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#04	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#05	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#06	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#07	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#08	112
prontera,151,151,0	duplicate(unitwarp test)	unitwarp test#09	112
```
  * Official Information: none

* **Modifications that may affect results**:  none

PR #4104, opened 2019-04-12
unitwarp npc - issue #4087

Atemo:
* **Addressed Issue(s)**: #4087

* **Server Mode**: N/A

Move the npc warped by `unitwarp` to the new map.

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -3118,7 +3118,7 @@ int unit_remove_map_(struct block_list *bl, clr_type clrtype, const char* file, 
 			break;
 		case BL_MOB:
 			// /BL_MOB is handled by mob_dead unless the monster is not dead.
-			if (status_isdead(bl) {
+			if (status_isdead(bl)) {
 				map_delblock(bl);
 				break;
 			}
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -3112,11 +3112,21 @@ int unit_remove_map_(struct block_list *bl, clr_type clrtype, const char* file, 
 		skill_cleartimerskill(bl);
 	}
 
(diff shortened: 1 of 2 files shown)


Litro:
i have tested this, seem fine for me

src/map/unit.cpp
```diff
@@ -3093,6 +3096,11 @@ int unit_remove_map_(struct block_list *bl, clr_type clrtype, const char* file,
 			}
 			break;
 		}
+		case BL_NPC: {
```
aleos89: You don't have to wrap this in brackets since you aren't defining any variables.

src/map/unit.cpp
```diff
@@ -3101,11 +3111,14 @@ int unit_remove_map_(struct block_list *bl, clr_type clrtype, const char* file,
...
-	if( bl->type != BL_MOB || !status_isdead(bl) )
-		clif_clearunit_area(bl,clrtype);
 
-	map_delblock(bl);
+	if (bl->type != BL_NPC) {// already handled by npc_remove_map
+		// /BL_MOB is handled by mob_dead unless the monster is not dead.
+		if( bl->type != BL_MOB || !status_isdead(bl) )
+			clif_clearunit_area(bl,clrtype);
+		map_delblock(bl);
+	}
```
aleos89: You could adjust this to a switch case for readability:
```cpp
switch (bl->type) {
	case BL_NPC:
		// already handled by npc_remove_map
		break;
	case BL_MOB:
		// /BL_MOB is handled by mob_dead unless the monster is not dead.
		if (status_isdead(bl) {
			map_delblock(bl);
			break;
		}
		// Fall through
	default:
		clif_clearunit_area(bl, clrtype);
		map_delblock(bl);
		break;
}
```

Issue #260, opened 2015-01-30
Unofficial walkpath


MrAntares:
Git Version: **12628 - 2014-12-04 05:04:30 (luciar)**
Client Date: **2012-04-10**
Src Mods: **NONE**
General settings: 
- **Full Renewal**
- **1/1/1/1/1 rate**
- **PK server (not in towns, above lvl 30)**
- **PVP ATK/Skill modifers: 100%**
- **WOE ATK/Skill modifers: rA default**
- **Official walk path and stacking**
- **Exp and drop penalty table values below 100% set to 100%**
- **Party and attacker exp bonus: +40%**
- More at (Hungarian, use translate): http://wiki.ragna.ro/

How to reproduce & Description:

I believe it was Playtester who fixed the official walkpath's monster lagging through walls issue (official path search). Unfortunately this fix was not implemented for the unofficial walkpath setting. So if we want to use this settings, mosters will lag through walls, when they find a way around, but they shouldn't.

I think the only difference what should be between official and non-official walkpath, is that on non-official setting there shouldn't be a walkpath length check, so if theres a way on the screen, you should walk around obstacles and corners. Maybe this is not the correct way to describe it, but I hope you get it.

Playtester:
The client simply won't display any complex walk path longer than 14 cells properly. It doesn't provide enough memory to do a larger path search than that, so if you want the walk path to be longer, you will have to live with not seeing actual movement of this length.

I'd say it's impossible to fix without client modification.

MrAntares:
Okay, to make it simpler:
Unofficial walkapth needs to be exactly like official EXCEPT when there is a corner or a simple rock in the way, the character should go around. 

Now it stops and freezes for one or two secs on official settings, so I want to turn it off, but on unofficial setting monsters go crazy. So anything else that comes from the official path is perfectly fine, including the mob's pathsearch.

Btw all started with this:
https://rathena.org/board/tracker/issue-5830-weird-movement-bug/?gopid=9618#entry9618

So on unofficial settings we should do exactly what the blue line showed before. The red line (official walk) was intended to counter the issue when you click on the stair or wall, and the character starts to go around by itself.... but this is still not the case sometimes. Just try to go downstairs in pyramids or geffen tower.

Playtester:
In theory you could make the official walkpath only work for monsters and not for players by adding the corresponding checks, but I know from some servers that it can also lead to position lag for players.

anacondaqq:
[...]

PR #4120, opened 2019-04-22
Fix monster warp through walls bug Issue 260

SeravySensei:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/260

* **Server Mode**: 
Both

As comments suggested, monsters always take official walkpath while non-monsters take whichever OFFICIAL_WALKPATH defines.
Ideal solution would be, of course, to hex the client to properly display complex paths of more than 14 tiles of movement, but I'm going to accept the comment saying it's not possible. If anyone who does client editing reads this, please check if it is indeed impossible to fix on client side.

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -89,12 +89,15 @@ int unit_walktoxy_sub(struct block_list *bl)
 	if( !path_search(&wpd,bl->m,bl->x,bl->y,ud->to_x,ud->to_y,ud->state.walk_easy,CELL_CHKNOPASS) )
 		return 0;
 
-#ifdef OFFICIAL_WALKPATH
-	if( !path_search_long(NULL, bl->m, bl->x, bl->y, ud->to_x, ud->to_y, CELL_CHKNOPASS) // Check if there is an obstacle between
+	if (!path_search_long(NULL, bl->m, bl->x, bl->y, ud->to_x, ud->to_y, CELL_CHKNOPASS) // Check if there is an obstacle between
 		&& wpd.path_len > 14	// Official number of walkable cells is 14 if and only if there is an obstacle between. [malufett]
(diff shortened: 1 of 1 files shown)


SeravySensei:
I realized there is one problem with this : If monsters can't walk more than 14 tiles then attacking them from 15 or more distance is safe - they won't be able to find the attacker and keep walking randomly. Archers with Eagle Eye have a range of 15 for example. Is this really how it works on official servers? Is 14 tiles really the limitation in the client?

SeravySensei:
I still observe monsters warping through walls even with this unfortunately - although I'm not sure but it feels less frequent now.

cydh:
client default is 14, so does that normal player for first requesting hit-lock. but if bysome accident if monster or player went out of 14 cells as long as the cells are shootable, the attack is continue. except monster have teleport condition on idle/rude/cant attack back.
but monster never chases player back if it's out of range.
that's the official behavior

SeravySensei:
I'm not talking about attacking.
The client doesn't display monsters moving if they walked from one side of a wall to another, only after they arrived and started moving again. It was explained this happens because clients can't display moves longer than 14 if it's not a straight line. So this PR aims to ensure monsters never walk a path longer than 14 even if official walkpath is off. Yet, the monsters moving around walls isn't displayed by the client anyway. So either the client can't display monsters walking around a wall regardless of distance, or this PR isn't working as intended for some reason.
Whichever is the case, monsters walking not being displayed by the client is a serious issue. I need to find out how to fix it. Any ideas?

SeravySensei:
Update, the problems I observed after this PR were caused by a different local bug. It seems to work correctly now. Also added an improvement : now if the monsters would walk a path the client can't show, they will walk the first 14 tiles of it instead of assuming the target is unreachable.

PR #4136, opened 2019-04-30
Persistent favorites

SeravySensei:
* **Addressed Issue(s)**: 
None, this is a new optional feature. 

* **Server Mode**: 

Both.

When items are put into the favorites tab in the inventory, players expect them to actually stay there. This is not the case for items that get consumed : once you used up your last gemstone, even if you buy more, it'll not be in favorites anymore, so if you rely on favorites to sell or storage everything that's not in favorites, you'll end up going into battle without the items.
This pull request adds a new option to items.conf that allows items to be memorized when put into the favorites tab, so even if the item leaves and reenters the inventory for any reason (dropped, consumed and rebought, etc), it'll stay in the tab where it originally was.
To achieve this, the itemID/charID pairs have to be stored in the database - the root of the current behavior lies in the fact favorites are stored in the actual instance of the inventory item, so it's lost as soon as that item is moved.

Note : I have no idea how RAthena creates the databases when it's first installed - the commands to add the favorites table are in favs.sql, if they need to be moved to a specific file, please move it there.
I've been using this feature to play in the past 2 months without issues, since then the only change I added was to make it optional.

### DIFF
--- a/sql-files/favs.sql
+++ b/sql-files/favs.sql
@@ -1,0 +1,7 @@
+DROP TABLE IF EXISTS `favorites`;
+CREATE TABLE favorites(
+    char_id INT,
+    itemid INT,
+    PRIMARY KEY (char_id,itemid)
+    );
+	
--- a/src/char/char.cpp
+++ b/src/char/char.cpp
@@ -512,6 +512,27 @@ int char_mmo_char_tosql(uint32 char_id, struct mmo_charstatus* p){
 			strcat(save_status, " hotkeys");
 	}
 #endif
+	// item favorites
+	if (SQL_ERROR == Sql_Query(sql_handle, "DELETE FROM `%s` WHERE char_id='%d'",
+		schema_config.favs_db, p->char_id))
+	{
+		Sql_ShowDebug(sql_handle);
+		errors++;
+	}
+	for (i = 0; i < MAX_FAVORITES; i++) {
+		if (p->favs[i]>0) {
+			if (SQL_ERROR == Sql_Query(sql_handle, "INSERT INTO `%s`(`char_id`,`itemid`) VALUES ('%d','%d');",
+				schema_config.favs_db, p->char_id, p->favs[i]))
+			{
+				Sql_ShowDebug(sql_handle);
+				errors++;
+			}
+		}
+	};
+	
+	
+	strcat(save_status, " favorites1");
+
 	StringBuf_Destroy(&buf);
 	if (save_status[0]!='\0' && charserv_config.save_log)
 		ShowInfo("Saved char %d - %s:%s.\n", char_id, p->name, save_status);
@@ -1020,6 +1041,7 @@ int char_mmo_char_fromsql(uint32 char_id, struct mmo_charstatus* p, bool load_ev
 	struct hotkey tmp_hotkey;
 	int hotkey_num;
 #endif
+	int fav, noffav;
 	StringBuf msg_buf;
 	char sex[2];
 
@@ -1220,6 +1242,19 @@ int char_mmo_char_fromsql(uint32 char_id, struct mmo_charstatus* p, bool load_ev
 	}
 	StringBuf_AppendStr(&msg_buf, " hotkeys");
 #endif
+
+	if (SQL_ERROR == SqlStmt_Prepare(stmt, "SELECT `itemid` FROM `%s` WHERE `char_id`=?", schema_config.favs_db)
+		|| SQL_ERROR == SqlStmt_BindParam(stmt, 0, SQLDT_INT, &char_id, 0)
+		|| SQL_ERROR == SqlStmt_Execute(stmt)
+		|| SQL_ERROR == SqlStmt_BindColumn(stmt, 0, SQLDT_INT, &fav, 0, NULL, NULL)
+		) SqlStmt_ShowDebug(stmt);
+	noffav = 0;
+	while (SQL_SUCCESS == SqlStmt_NextRow(stmt))
+	{
+		if (noffav < MAX_FAVORITES) { p->favs[noffav] = fav; }
+		else ShowWarning("mmo_char_fromsql: Attempt to load more favorites than MAX_FAVORITES!");
+	}
+	StringBuf_AppendStr(&msg_buf, " favorites");
 
 	/* Mercenary Owner DataBase */
 	mercenary_owner_fromsql(char_id, p);
@@ -2616,7 +2651,9 @@ void char_sql_config_read(const char* cfgName) {
 			safestrncpy(schema_config.friend_db, w2, sizeof(schema_config.friend_db));
 		else if(!strcmpi(w1,"hotkey_db"))
 			safestrncpy(schema_config.hotkey_db, w2, sizeof(schema_config.hotkey_db));
-		else if(!strcmpi(w1,"quest_db"))
+		else if (!strcmpi(w1, "favs_db"))
+			safestrncpy(schema_config.favs_db, w2, sizeof(schema_config.favs_db));
+		else if (!strcmpi(w1, "quest_db"))
 			safestrncpy(schema_config.quest_db,w2,sizeof(schema_config.quest_db));
 		else if(!strcmpi(w1,"homunculus_db"))
 			safestrncpy(schema_config.homunculus_db,w2,sizeof(schema_config.homunculus_db));
@@ -2681,7 +2718,8 @@ void char_set_default_sql(){
 	safestrncpy(schema_config.auction_db,"auction",sizeof(schema_config.auction_db)); // Auctions System
 	safestrncpy(schema_config.friend_db,"friends",sizeof(schema_config.friend_db));
 	safestrncpy(schema_config.hotkey_db,"hotkey",sizeof(schema_config.hotkey_db));
-	safestrncpy(schema_config.quest_db,"quest",sizeof(schema_config.quest_db));
+	safestrncpy(schema_config.favs_db, "favorites", sizeof(schema_config.favs_db));
+	safestrncpy(schema_config.quest_db, "quest", sizeof(schema_config.quest_db));
 	safestrncpy(schema_config.homunculus_db,"homunculus",sizeof(schema_config.homunculus_db));
 	safestrncpy(schema_config.skill_homunculus_db,"skill_homunculus",sizeof(schema_config.skill_homunculus_db));
 	safestrncpy(schema_config.mercenary_db,"mercenary",sizeof(schema_config.mercenary_db));
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -18560,9 +18560,12 @@ void clif_parse_MoveItem(int fd, struct map_session_data *sd) {
 	}
 	else if (type == 0) {
 		sd->inventory.u.items_inventory[index].favorite = 1;
-		for (j = 0; (j < MAX_FAVORITES) && (sd->status.favs[j] > 0); j++) { ; }
+		if (battle_config.persistent_favorites)
+			if (!(itemdb_isequip2(sd->inventory_data[index])) || battle_config.persistent_favorites_equipment) {
+			for (j = 0; (j < MAX_FAVORITES) && (sd->status.favs[j] > 0); j++) { ; }
 
-		if (j < MAX_FAVORITES) (sd->status.favs[j] = (sd->inventory.u.items_inventory[index].nameid));
+			if (j < MAX_FAVORITES) (sd->status.favs[j] = (sd->inventory.u.items_inventory[index].nameid));
+		}
 	}
 	else
 		return;/* nothing to do. */
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -18544,6 +18544,7 @@ void clif_parse_MoveItem(int fd, struct map_session_data *sd) {
 	struct s_packet_db* info = &packet_db[RFIFOW(fd,0)];
 	int index = RFIFOW(fd,info->pos[0]) - 2;
 	int type = RFIFOB(fd, info->pos[1]);
+	int j;
 
 	/* can't move while dead. */
 	if(pc_isdead(sd)) {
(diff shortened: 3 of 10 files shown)


rAthenaAPI:
This pull request **introduces 1 alert** when merging b5c0b51c51a9e5cd728ec24d60c3f7a324687933 into 7f772c32d3be201861946bb64720c231828465ac - [view on LGTM.com](https://lgtm.com/projects/g/rathena/rathena/rev/pr-f0c14a58c28efc89d3250a4585bfaf3e2490b74f)

**new alerts:**

* 1 for Comparison result is always the same

---

src/map/clif.cpp
```diff
@@ -18577,6 +18577,7 @@ void clif_parse_MoveItem(int fd, struct map_session_data *sd) {
 	struct s_packet_db* info = &packet_db[RFIFOW(fd,0)];
 	int index = RFIFOW(fd,info->pos[0]) - 2;
 	int type = RFIFOB(fd, info->pos[1]);
+	int j;
```
aleos89: This could be moved into their respective cases below.

src/map/pc.cpp
```diff
@@ -4803,6 +4803,7 @@ short pc_search_inventory(struct map_session_data *sd, unsigned short nameid) {
 char pc_additem(struct map_session_data *sd,struct item *item,int amount,e_log_pick_type log_type) {
 	struct item_data *id;
 	int16 i;
+	int j;
```
aleos89: Can be moved into it's respective case below.

sql-files/favs.sql
```diff
@@ -0,0 +1,7 @@
+DROP TABLE IF EXISTS `favorites`;
+CREATE TABLE favorites(
+    char_id INT,
+    itemid INT,
+    PRIMARY KEY (char_id,itemid)
+    );
+
```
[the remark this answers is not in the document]
SeravySensei: moved it, hope I got the syntax right

src/char/char.cpp
```diff
@@ -1221,6 +1243,19 @@ int char_mmo_char_fromsql(uint32 char_id, struct mmo_charstatus* p, bool load_ev
...
 
+	if (SQL_ERROR == SqlStmt_Prepare(stmt, "SELECT `itemid` FROM `%s` WHERE `char_id`=?", schema_config.favs_db)
+		|| SQL_ERROR == SqlStmt_BindParam(stmt, 0, SQLDT_INT, &char_id, 0)
+		|| SQL_ERROR == SqlStmt_Execute(stmt)
+		|| SQL_ERROR == SqlStmt_BindColumn(stmt, 0, SQLDT_INT, &fav, 0, NULL, NULL)
+		) SqlStmt_ShowDebug(stmt);
+	noffav = 0;
+	while (SQL_SUCCESS == SqlStmt_NextRow(stmt))
+	{
+		if (noffav < MAX_FAVORITES) { p->favs[noffav] = fav; }
```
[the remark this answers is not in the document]
SeravySensei: This looks like a bug, I have no idea why that didn't cause problems. Noffav++ is missing.
Also, not sure how c++ handles that, is it safe to assume the favs array has all zeroes here or should I clear it first?

src/map/pc.cpp
```diff
@@ -4860,6 +4861,10 @@ char pc_additem(struct map_session_data *sd,struct item *item,int amount,e_log_p
 		if( item->equipSwitch )
 			sd->inventory.u.items_inventory[i].equipSwitch = 0;
 
+		for (j = 0; j < MAX_FAVORITES; j++) {
+			if (sd->status.favs[j] == item->nameid) sd->inventory.u.items_inventory[i].favorite = 1;
```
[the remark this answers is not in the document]
SeravySensei: absolutely no idea what you want here, no else branch. To be honest formatting is of zero importance to me so it might be more effective if you submitted the change you want like in the first two cases so I can just click the commit button and have it done.

sql-files/favs.sql
```diff
@@ -0,0 +1,7 @@
+DROP TABLE IF EXISTS `favorites`;
+CREATE TABLE favorites(
+    char_id INT,
+    itemid INT,
```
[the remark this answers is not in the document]
SeravySensei: Prefer to keep the name, "item" tells more about what it is. "nameid "Can be confusing, characters and items both have names. Finding and replacing all uses is not worth the time either. I added the other suggested changes.

src/map/pc.cpp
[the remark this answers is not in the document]
aleos89: Yes, I meant formatting, sorry for not being clear. It makes it hard to read when everything is bunched together on a single line. Remember, white space is free!

src/char/char.cpp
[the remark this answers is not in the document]
aleos89: `memset(p, 0, sizeof(struct mmo_charstatus));`
The struct is memset to 0 which is why all your values are 0.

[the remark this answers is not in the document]
SeravySensei: perfect

src/map/pc.cpp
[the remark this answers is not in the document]
SeravySensei: added a line break. Personally I prefer if there is a single instruction in an if statement to be in the same line - makes sure you don't accidentally read the second line without noticing the condition part.

src/map/clif.cpp
```diff
@@ -18586,10 +18586,22 @@ void clif_parse_MoveItem(int fd, struct map_session_data *sd) {
 	if (index < 0 || index >= MAX_INVENTORY)
 		return;
 
-	if ( sd->inventory.u.items_inventory[index].favorite && type == 1 )
+	if (sd->inventory.u.items_inventory[index].favorite && type == 1) {
+		int j;
 		sd->inventory.u.items_inventory[index].favorite = 0;
-	else if( type == 0 )
+		for (j = 0; j < MAX_FAVORITES; j++) if (sd->status.favs[j] == (sd->inventory.u.items_inventory[index].nameid)) sd->status.favs[j] = 0;
```
vstumpf: Put this on multiple lines.

src/map/clif.cpp
```diff
@@ -18586,10 +18586,22 @@ void clif_parse_MoveItem(int fd, struct map_session_data *sd) {
...
+	}
+	else if (type == 0) {
+		int j;
 		sd->inventory.u.items_inventory[index].favorite = 1;
+		if (battle_config.persistent_favorites)
+			if (!(itemdb_isequip2(sd->inventory_data[index]) && !(sd->inventory_data[index]->type==IT_AMMO))
+				|| battle_config.persistent_favorites_equipment) {
+			for (j = 0; (j < MAX_FAVORITES) && (sd->status.favs[j] > 0); j++) { ; }
+
+			if (j < MAX_FAVORITES) { sd->status.favs[j] = (sd->inventory.u.items_inventory[index].nameid); };
```
vstumpf: Multiple lines

src/char/char.cpp
```diff
@@ -1221,6 +1241,21 @@ int char_mmo_char_fromsql(uint32 char_id, struct mmo_charstatus* p, bool load_ev
...
+		|| SQL_ERROR == SqlStmt_BindColumn(stmt, 0, SQLDT_INT, &fav, 0, NULL, NULL)
+		) SqlStmt_ShowDebug(stmt);
+	noffav = 0;
+	while (SQL_SUCCESS == SqlStmt_NextRow(stmt))
+	{
+		if (noffav < MAX_FAVORITES)
+		{ p->favs[noffav] = fav; noffav++;  }
+		else
+		{ ShowWarning("mmo_char_fromsql: Attempt to load more favorites than MAX_FAVORITES!");	}
+	}
```
vstumpf: Should look something like this
```suggestion
		if (noffav < MAX_FAVORITES) { 
			p->favs[noffav] = fav; noffav++;  
		} else {
			ShowWarning("mmo_char_fromsql: Attempt to load more favorites than MAX_FAVORITES!");	
		}
	}
```

src/char/char.cpp
```diff
@@ -512,6 +512,25 @@ int char_mmo_char_tosql(uint32 char_id, struct mmo_charstatus* p){
...
+		if (p->favs[i]>0) {
+			if (SQL_ERROR == Sql_Query(sql_handle, "INSERT INTO `%s`(`char_id`,`itemid`) VALUES ('%d','%d');",
+				schema_config.favs_db, p->char_id, p->favs[i]))
+			{
+				Sql_ShowDebug(sql_handle);
+				errors++;
+			}
+		}
+	};
+	strcat(save_status, " favorites1");
```
vstumpf: Why favorites1? Why not just favorites

src/char/char.cpp
```diff
@@ -512,6 +512,25 @@ int char_mmo_char_tosql(uint32 char_id, struct mmo_charstatus* p){
 			strcat(save_status, " hotkeys");
 	}
 #endif
+	// item favorites
```
vstumpf: With this block of code you're always clearing and writing item favorites. Maybe only do this if there's a change? See friends or hotkeys above.

sql-files/main.sql
```diff
@@ -1101,3 +1101,13 @@ CREATE TABLE IF NOT EXISTS `vendings` (
...
+
+--
+-- Table structure for table `favorites`
+--
+
+CREATE TABLE IF NOT EXISTS `favorites` (
+    `char_id` int(11) unsigned NOT NULL default '0',
+     `itemid` smallint(5) unsigned NOT NULL default '0',
+    PRIMARY KEY (char_id,itemid)
+    );
```
vstumpf: Set the engine=MyISAM

sql-files/main.sql
```diff
@@ -1101,3 +1101,13 @@ CREATE TABLE IF NOT EXISTS `vendings` (
...
   PRIMARY KEY (`id`)
 ) ENGINE=MyISAM;
+
+--
+-- Table structure for table `favorites`
+--
+
+CREATE TABLE IF NOT EXISTS `favorites` (
+    `char_id` int(11) unsigned NOT NULL default '0',
+     `itemid` smallint(5) unsigned NOT NULL default '0',
```
vstumpf: We normally use nameid as the variable name.

sql-files/main.sql
```diff
@@ -1101,3 +1101,13 @@ CREATE TABLE IF NOT EXISTS `vendings` (
   `autotrade` tinyint(4) NOT NULL,
   PRIMARY KEY (`id`)
 ) ENGINE=MyISAM;
+
+--
+-- Table structure for table `favorites`
+--
+
+CREATE TABLE IF NOT EXISTS `favorites` (
+    `char_id` int(11) unsigned NOT NULL default '0',
```
vstumpf: The spacing for this file is two spaces, please abide.

ecdarreola:
@SeravySensei i'm getting this debug after updating it to the latest. 
image

SeravySensei:
> 
> 
> @SeravySensei i'm getting this debug after updating it to the latest.
> image

One of the revieweres requested the field to be renamed. If you created the table with the old field name before then, it's not going to work unless you update the field name in your database.
...and this is exactly why i believe randomly renaming stuff is a bad idea. Well, it's better early than later but "never" is the best in my opinion.

ecdarreola:
@SeravySensei 
Is this normal? 
image

SeravySensei:
> 
> 
> @SeravySensei
> Is this normal?
> image

Somehow you managed to drag one item into the favorites tab multiple times. I guess I should update this to prevent that from being possible.

ecdarreola:
@SeravySensei Yeah seems like it. It appears when doing that hehe.

ecdarreola:
@SeravySensei : still the same after updating it.

image

SeravySensei:
I wasn't able to reproduce it.
Maybe you need to take the duplicated item out of your favorites tab (possibly even more than once) to actually get rid of your duplicate entries, or delete them from your database.  (also make sure you recompiled)

ecdarreola:
> I wasn't able to reproduce it.
> Maybe you need to take the duplicated item out of your favorites tab (possibly even more than once) to actually get rid of your duplicate entries, or delete them from your database. (also make sure you recompiled)

Probably deleting favs entry will be the best way.

EvilCloudHermit:
> @SeravySensei : still the same after updating it.
> 
> image

It's an old think, but for all who had the same problem, i managed to get it fixed with:

```
diff = 0; // Initialize variable to track differences

for (i = 0; i < MAX_FAVORITES; i++) {
    if (p->favs[i] != cp->favs[i]) { // Check for differences in favorite items
        diff = 1;
        break;
    }
}

if (SQL_ERROR == Sql_Query(sql_handle, "DELETE FROM `%s` WHERE char_id='%d'",
    schema_config.favs_db, p->char_id))
{
    Sql_ShowDebug(sql_handle);
    errors++;
}

if (diff == 1) { // Save only if there are differences
    // Then insert
    StringBuf_Clear(&buf); // Clear the buffer
    StringBuf_Printf(&buf, "INSERT INTO `%s` (`char_id`, `nameid`) VALUES ", schema_config.favs_db); // Prepare INSERT statement
    for (i = 0, count = 0; i < MAX_FAVORITES; ++i) {
        if (p->favs[i] > 0) {
            // Check if the entry already exists
            if (SQL_SUCCESS != Sql_Query(sql_handle, "SELECT * FROM `%s` WHERE char_id='%d' AND nameid='%d'",
                schema_config.favs_db, p->char_id, p->favs[i]))
            {
                // Query error
                Sql_ShowDebug(sql_handle);
                errors++;
            } else {
                int num_rows = Sql_NumRows(sql_handle);
                if (num_rows == 0) {
                    // Entry does not exist, so insert
                    if (count)
                        StringBuf_AppendStr(&buf, ",");
                    StringBuf_Printf(&buf, "('%d','%d')", p->char_id, p->favs[i]);
                    count++;
                }
            }
        }
    }
    if (count) {
        if (SQL_ERROR == Sql_QueryStr(sql_handle, StringBuf_Value(&buf))) {
            Sql_ShowDebug(sql_handle);
            errors++;
        }
        strcat(save_status, " favorites");
    }
}
```
I believe there is still a missing function, when the player drags an item from the favorites tab, the corresponding entry of that item should be deleted directly from the SQL database. This would address the issue of reaching the maximum limit of 128 entries after a certain period, while also removing unnecessary entries that are no longer needed by the player.

PR #4139, opened 2019-05-02
Fixed potential race-condition possibility in cart item transaction

secretdataz:
* **Addressed Issue(s)**: -
* **Server Mode**: Both
Fixed potential race-condition possibility in cart item transaction which I will not list the steps to reproduce here.

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -5447,7 +5447,7 @@ void pc_putitemtocart(struct map_session_data *sd,int idx,int amount)
 
 	item_data = &sd->inventory.u.items_inventory[idx];
 
-	if( item_data->nameid == 0 || amount < 1 || item_data->amount < amount || sd->state.vending )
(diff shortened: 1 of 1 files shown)


attackjom:
Is this the dupe protection ?

src/map/pc.cpp
```diff
@@ -5487,21 +5487,25 @@ int pc_cartitem_amount(struct map_session_data* sd, int idx, int amount)
...
+
+	if (pc_checkadditem(sd, item_data->nameid, amount) == CHKADDITEM_OVERAMOUNT) {
+		return;
+	}
+
+	item item_copy = *item_data;
+
+	pc_cart_delitem(sd, idx, amount, 0, LOG_TYPE_NONE);
+	char flag = pc_additem(sd, &item_copy, amount, LOG_TYPE_NONE);
+	if(flag != 0) {
```
aleos89:
```suggestion
	if(flag != ADDITEM_SUCCESS) {
```

PR #4143, opened 2019-05-03
viewpoint accept char id

sader1992:

* **Server Mode**: both

* **Description of Pull Request**: Allow viewpoint to accept an account id

```cs
prontera,155,190,0	script	viewpoint_charid	444,{
	.@v = 40;
	while(true){
		viewpoint 2,30,.@v -1,1,0xFF0000;
		viewpoint 1,30,.@v,1,0xFF0000;
		sleep2 100;
		.@v += 10;
		if(.@v > 1000)
			end;
	}

end;
OnInit:
	.v = 40;
	while(true){
		getmapunits(BL_PC,"prontera",.@charName$);
		for(.@i=0;.@i<getarraysize(.@charName$);.@i++){
			viewpoint 2,30,.v -1,1,0xFF0000,getcharid(0,.@charName$[.@i]);
			viewpoint 1,30,.v,1,0xFF0000,getcharid(0,.@charName$[.@i]);
		}	
		sleep 100;
		.v += 10;
	}
}
```

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -6687,12 +6687,12 @@ BUILDIN_FUNC(viewpoint)
 	int type,x,y,id,color;
 	TBL_PC* sd;
 
-	if (script_hasdata(st, 7)) {
-		script_charid2sd(7, sd)
+	if (script_hasdata(st, 7))
+		script_charid2sd(7, sd);
 	else
-		script_rid2sd(sd))
+		script_rid2sd(sd);
 
-	if(!sd){
+	if (!sd) {
 		st->state = END;
 		return SCRIPT_CMD_FAILURE;
 	}
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -6687,16 +6687,16 @@ BUILDIN_FUNC(viewpoint)
 	int type,x,y,id,color;
 	TBL_PC* sd;
 	if (script_hasdata(st, 7)) {
-        if (!(sd = map_id2sd(script_getnum(st, 7)))) {
-            ShowError("buildin_viewpoint: player not found (AID=%d).\n",script_getnum(st, 7));
-            st->state = END;
-            return SCRIPT_CMD_FAILURE;
-        }
-    }
-    else {
-        if (!script_rid2sd(sd))
-            return SCRIPT_CMD_FAILURE;
-    }
+		if (!(sd = map_id2sd(script_getnum(st, 7)))) {
+			ShowError("buildin_viewpoint: player not found (AID=%d).\n",script_getnum(st, 7));
+			st->state = END;
+			return SCRIPT_CMD_FAILURE;
+		}
+	}
+	else {
+		if (!script_rid2sd(sd))
+			return SCRIPT_CMD_FAILURE;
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -6686,16 +6686,24 @@ BUILDIN_FUNC(viewpoint)
...
+        if (!(sd = map_id2sd(script_getnum(st, 7)))) {
+            ShowError("buildin_viewpoint: player not found (AID=%d).\n",script_getnum(st, 7));
+            st->state = END;
+            return SCRIPT_CMD_FAILURE;
+        }
+    }
+    else {
+        if (!script_rid2sd(sd))
+            return SCRIPT_CMD_FAILURE;
+    }
```
aleos89: Could you fix the indentation here.

sader1992: sorry , getting used to a new editor
you mean like the last commit right ?

aleos89:
I just realized you use account ID rather than char ID. Is there a reason for this choice? Pretty much all commands extend with char ID.

sader1992:
nope, there is no reason , i just think it's better for the commands to use gid (just my preferences xD), i would change it if it's better to use char id
also i am planing to `cutin` command with the same idea `cutin "<filename>",<position>{,account id};` but didn't have time to prepare the testing files
so if it's better to use char id i would do that too there.

sader1992:
now the command uses the char id

src/map/script.cpp
```diff
@@ -6686,16 +6686,24 @@ BUILDIN_FUNC(viewpoint)
 {
 	int type,x,y,id,color;
 	TBL_PC* sd;
+	if (script_hasdata(st, 7)) {
+		if (!script_charid2sd(7, sd)) {
+			ShowError("buildin_viewpoint: player not found (CID=%d).\n",script_getnum(st, 7));
```
Atemo: `script_charid2sd` already show the message
```
ShowError("%s: Player with char id '%d' is not found.\n", func, id_);
```
```suggestion
```

sader1992:
i wander if it's a good idea to end the script if the command fail , given that this not a critical command

Atemo:
Actually `script_charid2sd_` is using `script_rid2sd_` when there is no data in the loc. `script_rid2sd_` end the script if there is no sd however `script_charid2sd_` doesn't end it if there is data in loc and no sd. Let's say it's fine to end it

src/map/script.cpp
```diff
@@ -6687,15 +6687,22 @@ BUILDIN_FUNC(viewpoint)
...
 
+	if (script_hasdata(st, 7)) {
+		script_charid2sd(7, sd)
+	else
+		script_rid2sd(sd))
+
+	if(!sd){
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
```
Atemo: After re-reading the check just need to be
```

	if (!script_charid2sd(7, sd)) {
		st->state = END;
		return SCRIPT_CMD_FAILURE;
	}
```

sader1992: MVP

Issue #4146, opened 2019-05-04
Cart vs Overweight BUG item disapear


mazvi:
* **rAthena Hash**: 7f772c3

* **Client Date**: 20170517RE

* **Server Mode**: RENEWAL

* **Description of Issue**: 
  * Result:  item disapear if move from cart to inventory when overweight for example weight 9092/9100 the item weight is 20. item will disapear.
  * Expected Result: 
  * How to Reproduce:  Save many item on cart, make your weight 99% example 9092/9100. then move your item from cart to your inventory. Bingo you have skill hand of thanos, your item will be disapear...
  * Official Information:
  

* **Modifications that may affect results**:

PR #4148, opened 2019-05-04
Additional fixes for cart exploit

secretdataz:
Issue: #4146

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -4584,7 +4584,7 @@ char pc_can_add_item_to_inventory(map_session_data* sd, storage_type type, unsig
 
 	switch (type) {
 		case TABLE_INVENTORY:
-			if (sd->weight + itemdb_weight(nameid) > sd->max_weight)
+			if (sd->weight + (amount * itemdb_weight(nameid)) > sd->max_weight)
 				return ADDITEM_OVERWEIGHT;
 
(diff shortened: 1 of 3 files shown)


admkakaroto:
The problem has not yet been resolved, if you are overweight and pick up an item from the cart it disappears.

aleos89:
@admkakaroto It should be good now.

admkakaroto:
@aleos89 
a part is fixed, picking the items from the cart to inventory is 'ok'.

however if you send overweight items to the cart they are deleted, see:

a1

now, send yggdrasil to cart:
a2

thanos power in action:
a3

aleos89:
I realized I forgot the amount factor into the weight.

cahya1992:
this update solved the item dissapear but when cart get overweight item got error can not be use or drag
you should use  @refresh or @go to use that item again

laziem:
Item with cart restriction gone missing when try to put on cart

admkakaroto:
Continues with only a small problem, when you are overweight in the cart and send an inventory item over there, the item exits from the inventory, you need to use the @refresh command for the item to reappear, including it changes the inventory item count, this can cause problems.

Normal cart:
a1

Now, send Yggdrasil to Cart:
a2

Yggdrasil Have gone, item count decreased too:
a3



Sending from cart to inventory seems to be running 100% well.

PR #4155, opened 2019-05-09
fix getareaunits coordinates

sader1992:
Co-Authored-By: AnnieRuru <annieruru@users.noreply.github.com>

* **Server Mode**: both

thanks to @AnnieRuru she warned me about it time ago and i didn't think about the issue much
test script:
```cs
-	script	getareaunits	-1,{
OnInit:
	for(.@i = 152;.@i<156;.@i++){
		for(.@m = 162;.@m<166;.@m++){
			monster "prontera",.@i,.@m,"--en--",1084,1;
		}
	}
	//start :	152,162
	//end	:	156,166
	debugmes "==============================================";
	debugmes "Wrong way of typing it:";
	debugmes "" + getareaunits(BL_MOB,"prontera",152,166,156,162);
	debugmes "==============================================";
	debugmes "Right way of typing it:";
	debugmes "" + getareaunits(BL_MOB,"prontera",152,162,156,166);
	debugmes "==============================================";
end;
}
```
Before:
image

After:
image

Edit:
now the command should accept both ways , so it would be backward compatible and work the right way thanks to @Atemo
image

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -11385,7 +11385,7 @@ BUILDIN_FUNC(getunits)
 
 	for (bl = (struct block_list*)mapit_first(iter); mapit_exists(iter); bl = (struct block_list*)mapit_next(iter))
 	{
-		if (!m || (m == bl->m && !x0 && !y0 && !x1 && !y1) || (bl->m == m && (bl->x >= x0 && bl->y <= y0) && (bl->x <= x1 && bl->y >= y1)))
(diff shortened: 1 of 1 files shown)


sader1992:
maybe we could add a warning message too would be a good idea ?

Atemo:
Why not check and swap the coordinates?

sader1992:
i think like this , there is no need for the warning message anymore

Issue #4142, opened 2019-05-03
Item ID:5384 have wrong script


AzarthMZintos:
* **rAthena Hash**: dont matter

* **Client Date**: dont matter

* **Server Mode**: dont matter

* **Description of Issue**: 
  * Result: the script of the item is wrong
  * Expected Result: the script should be corrected
  * How to Reproduce: look at divine-pride / descripition of the item and compare to rathene db
  * Official Information:

No official information: https://www.divine-pride.net/database/item/5384 (the KRO version not others)

  

* **Modifications that may affect results**:

PR #4157, opened 2019-05-10
Corrected Twin Pompom By JB item bonus

aleos89:
* **Addressed Issue(s)**: #4142

* **Server Mode**: Renewal

  * Corrected Twin Pompom By JB item bonus to match kRO description.
Thanks to @AzarthMZintos!


aleos89:
Now that I look back at the description, anything after the LUK bonus is a Christmas Event bonus. I'll move these bonuses to the import/item_db along with the other event item bonuses.

Issue #3037, opened 2018-04-06
Stylist Shop (Stylist UI)


BrOgBr:
* **rAthena Hash**: 

* **Client Date**: 2015+

* **Server Mode**: Renewal

* **Description of Issue**: 
 This is a very nice Ui for the Stylist _(Stylist Shop)_.

<img width="298" alt="zzz" src="https://user-images.githubusercontent.com/34192842/38410369-c21dce06-395a-11e8-8101-6a28eefcd5f5.png">

More Info:
https://forums.warpportal.com/index.php?/topic/191935-kro-styling-shop-renovation-jan-27th-2016/

Preview:
https://youtu.be/kbCa5qefWMg

It was released about six days ago in herc by the Asheraf at this PR:
HerculesWS/Hercules#2004

Angelic234:
[https://github.com/idathena/trunk/commit/4bdd552e65a710d0347247aa7c6affe76fc1159a](https://github.com/idathena/trunk/commit/4bdd552e65a710d0347247aa7c6affe76fc1159a)

Haikenz:
Thanks @Angelic234

RagnaWay:
Any news ???

dimasshotta:
this is how i fix in rathena. Please check if it has something wrong:

static void pc_readdb_stylist_libconfig_sub(struct config_setting_t *it, int16 count, const char *source) {
	int i32, Id = 0, zeny = 0, itemID = 0, boxItemID = 0;
	bool allow_doram = false;
	const char *name;
	int64 type;
	int16 idx;	

	nullpo_retv(it);
	nullpo_retv(source);

	if (!config_setting_lookup_string(it, "Type", &name)) {
		ShowWarning("pc_readdb_stylist_libconfig_sub: Missing Type in \"%s\", entry #%d.\n", source, count);
		return;
	}
	if (!script_get_constant(name, &type)) {
		ShowWarning("pc_readdb_stylist_libconfig_sub: Invalid Type '%s' in \"%s\", entry #%d.\n", name, source, count);
		return;
	}
	if (type < 0 || type >= MAX_STYLIST_TYPE) {
		ShowWarning("pc_readdb_stylist_libconfig_sub: Out of range Type '%s' in \"%s\", entry #%d.\n", name, source, count);
		return;
	}
	if (!config_setting_lookup_int(it, "Id", &i32) || i32 < 0) {
[...]

PR #4164, opened 2019-05-12
Implemented Stylist UI

Haikenz:
* **Addressed Issue(s)**: #3037

* **Description of Pull Request**: Preview: https://www.youtube.com/watch?v=mptcwFzWtGE

Special thanks to @JohnnyPlayy  and @vykimo


conf/battle/feature.conf
```diff
@@ -103,3 +103,7 @@ feature.attendance: on
 // Private Airship System (Note 1)
 // Requires: 2018-03-21aRagexeRE or later
 feature.privateairship: on
+
+// Stylist UI (Note 1)
+// Requires: 2015-11-04aRagexeRE or later
+feature.stylistui: on
```
sctnightcore: It look like https://github.com/idathena/trunk/commit/4bdd552e65a710d0347247aa7c6affe76fc1159a
I think you forgot. Thank you idathena.

aleos89:
Another thing to note, the database will need to be converted to YAML since we will be dropping support for libconfig.

conf/battle/feature.conf
4144: It's should be also thanks hercules?

exneval: shall address to hercules instead for the original code, but like @aleos89 said, it's better to use YAML format

sctnightcore: xD

violent01:
Map crash whenever i run the server. I'm using the latest rathena version.

yashimwong:
@violent01 a crash/error message or gdb would help.

violent01:
@yashimwong here's what's happening with my mapserver.
githubpic

yashimwong:
Can you run the map server with the Visual Studio Debugger?  Follow the steps here: https://github.com/rathena/rathena/wiki/Visual-Studio-Debugger

violent01:
@yashimwong this is what it reproduced. Btw how should i read the call stacks? is it descending or ascending because there are tons of things in the call stack and the guide in rathena debugging says read the last call in the msvs call stack.
githubpic2

AsurielRO:
Maybe its time to reopen this?

hannicaldummy:
When change anything map-server crashes, anyone knows what can it be?

Issue #4178, opened 2019-05-16
Bug Pet Egg (@storeall)


Badarosk0:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/9aa5f7d8345d9d89ec9e8b44060e92703f3d43b8

* **Client Date**: 20180621

* **Server Mode**: RE

* **Description of Issue**: 
  * Result: If you are with a Pet and use the @storeall command the Pet will not go back to Egg.
  * Expected Result: 
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**:

sader1992:
well not go back to Egg ?
what do you mean ?
the pet when you hatch it = the egg no longer in your inventory , so do you mean that after you do the command , than you are trying to get the pet back to an egg , you can't do it ?

it would be much better if you fill the **How to Reproduce** point

Badarosk0:
1- Create an egg with the command @item 9003
2- Use an incubator to hatch an egg. @Item command 643
3. Try to return the pet to the egg after colliding. (It will work perfectly.)
4- Hatch the egg again.
5- Use the @storeall command
6- Try to turn the pet to the egg. (It will not work and the egg will not go back to the egg and will be stuck in your char)
@sader1992

pigati23:
have you tried to create the egg with @makeegg?

sader1992:
i can confirm the issue with PRE-RE 4ba54838fe78d86fd5e03ebe5ffcd650e092cac9

PR #4179, opened 2019-05-16
prevent moving/dropping the pet egg if it's hatched

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/4178

* **Server Mode**: both

the server is allowed moving the egg pet after hatched
to test this
```
@item 9003
@hatch 
@storall / @dropall
```

returning the pet to the egg afterword will fail

 
 [ x ] player to trade `trade_tradeadditem`
 [ x ] player to cart `pc_cart_additem`
 [ x ] player to sell `pc_can_sell_item`
 [ x ] player to storage `storage_canAddItem`
 [ x ] and droping the item `pc_candrop`
 [ ] ?

### DIFF
--- a/src/map/itemdb.cpp
+++ b/src/map/itemdb.cpp
@@ -574,7 +574,7 @@ bool itemdb_isrestricted(struct item* item, int gmlv, int gmlv2, bool (*func)(st
 }
 
 bool itemdb_ishatched_egg(struct item* item) {
-	if (item->card[0] == CARD0_PET && item->attribute == 1)
+	if (item && item->card[0] == CARD0_PET && item->attribute == 1)
 		return true;
 	return false;
 }
--- a/src/map/itemdb.cpp
+++ b/src/map/itemdb.cpp
@@ -573,6 +573,12 @@ bool itemdb_isrestricted(struct item* item, int gmlv, int gmlv2, bool (*func)(st
 	return true;
 }
 
+bool itemdb_ishatched_egg(struct item* item) {
+	if (item->card[0] == CARD0_PET && item->attribute == 1)
+		return true;
+	return false;
+}
+
 /** Specifies if item-type should drop unidentified.
 * @param nameid ID of item
 */
--- a/src/map/pet.cpp
+++ b/src/map/pet.cpp
@@ -925,7 +925,7 @@ bool pet_return_egg( struct map_session_data *sd, struct pet_data *pd ){
 	it.card[0] = CARD0_PET;
 	it.card[1] = GetWord(pd->pet.pet_id, 0);
 	it.card[2] = GetWord(pd->pet.pet_id, 1);
-	it.card[3] = (unsigned int)MakeDWord(it.card[1], it.card[2]);//FOR TESTING
+	it.card[3] = (unsigned int)MakeDWord(it.card[1], it.card[2]);
 	//need to check bound state for the item ? //the best option is to make it with the pet info TODO: sql
 
 	//Delete Egg from the Player
--- a/src/map/pet.cpp
+++ b/src/map/pet.cpp
@@ -925,7 +925,8 @@ bool pet_return_egg( struct map_session_data *sd, struct pet_data *pd ){
 	it.card[0] = CARD0_PET;
 	it.card[1] = GetWord(pd->pet.pet_id, 0);
 	it.card[2] = GetWord(pd->pet.pet_id, 1);
-	//need to check bound state for the item ?
+	it.card[3] = (unsigned int)MakeDWord(it.card[1], it.card[2]);
+	//need to check bound state for the item ? //the best option is to make it with the pet info TODO: sql
 
 	//Delete Egg from the Player
 	if (pc_additem(sd, &it, 1, LOG_TYPE_OTHER))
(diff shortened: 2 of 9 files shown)


src/map/storage.cpp
```diff
@@ -214,6 +214,9 @@ static enum e_storage_add storage_canAddItem(struct s_storage *stor, int idx, st
 
 	if (!stor->state.put)
 		return STORAGE_ADD_NOACCESS;
+	
+	if (items[idx].card[0] == CARD0_PET && items[idx].attribute == 1)
+		return STORAGE_ADD_INVALID;
```
aleos89: If you could move this up above the `STORAGE_ADD_NOACCESS` check I would appreciate it!

sader1992:
this not ready yet , it's the same problem with @dropall also i was thinking about WPE and trade
so i am thinking about making the egg bound to char , this would fix anything we didn't think about too

sader1992:
also deleting the item using commands like delitem should not work on it

i was thinking the best idea would be is making a new bound element that bound the item to char and prevent anything from editing it

or we find a new way to make the egg usable like making a new sd item to save the egg value , i think this would be the **most reasonable solution**

src/map/pet.cpp
```diff
@@ -913,18 +913,32 @@ static int pet_performance(struct map_session_data *sd, struct pet_data *pd)
...
+	//sd->inventory.u.items_inventory[i].attribute = 0;
+	
+	struct item it;
+	memset(&it, 0, sizeof(it));
+	it.nameid = pd->pet.egg_id;
+	it.identify = 1;
+	it.card[0] = CARD0_PET;
+	it.card[1] = GetWord(pd->pet.pet_id, 0);
+	it.card[2] = GetWord(pd->pet.pet_id, 1);
+	//need to check bound state for the item ?
```
sader1992: here i was thinking what the best way to save the bound stat of the item

sader1992:
this just how the idea of the code
ofc if it's a good idea , i need to retrieve the first commits/clean the code
and search for any other function that use the hidden egg

sader1992:
i wonder if the way it's right now is coming from the official server

Daegaladh:
> 
> 
> i wonder if the way it's right now is coming from the official server

On official servers, when you hatch an egg the item apparently is still there in your inventory, just hidden.
I've tested this by having 99 items + an egg, and I still wasn't able to get any more items after hatching it. Also, when you get the pet back to the egg, there's no "you got X egg" message.
Btw, on Aegis if you have overweight or your inventory full, you never get the items dropped on the floor in any case, I was planning on getting rid of all those `map_addflooritem`...

Lemongrass3110:
Somehow the subject of this pull request changed into a strange direction. The behavior is 100% correct, only some places that might have not been taken into account need to be fixed. If you keep it like this(reverting to the old wrong Athena behavior), then we will have to reject and close this pull request and fix it ourselves. If you want to fix the issues, then please feel free to go on with it.

sader1992:
sorry didn't know that the old code was like that xD
i just saw https://github.com/eathena/eathena/blob/0cb07145ea1431528d0591b6a8706856915b3782/src/map/pet.c#L297

so the correct thing to do is just retrieving to the start , fixing the problem of @dropall and @storeall
unfortunately i don't have the list of all the problem , i only seen the issue by Badarosk0 and the drop thing just came in mind
didn't think about it more as i thought if we just change the behavior any unknown issue would be fixed xD

so i guess this PR is ready like this

sader1992:
@Lemongrass3110 i am not sure why (about pc_can_sell_item) but if it's something about WPE we might need to add to pc_cart_additem and trade_tradeadditem

sader1992:
i just tested it ,and yes
so 
player2cart (no need to vending skill)
player2trade

Lemongrass3110:
> i just tested it ,and yes
> so
> player2cart (no need to vending skill)
> player2trade
> player2mail
> 
> i was thinking of making a function that do the check , so any update later would be in one place , and it's better than copy paste the code all over the place
> maybe
> 
> ```c
> bool is_hatched_egg(struct item *item);
> 
> /*
>  return true if the item is hatched egg
>  return false if not
> */
> bool is_hatched_egg(struct item *item){
> 	if(item->card[0] == CARD0_PET && item->attribute == 1)
> 		return false;
> 
> 	return true;
> }
> 
> if(is_hatched_egg(item))
> 	return INVALID;
> ```
> 
> I don't know what the best place for it , the logical place , i think it's pet.cpp , however that mean adding the pet header to the other files

Probably either pet.hpp or itemdb.hpp
But I think that there might already be similar functions in the itemdb one. So maybe that one would be better. I was also thinking about this, when I changed it, but ended up being lazy and not wanting to change that many places.

src/map/itemdb.cpp
```diff
@@ -573,6 +573,12 @@ bool itemdb_isrestricted(struct item* item, int gmlv, int gmlv2, bool (*func)(st
 	return true;
 }
 
+bool itemdb_ishatched_egg(struct item* item) {
+	if (item->card[0] == CARD0_PET && item->attribute == 1)
```
aleos89: Either use nullpo to catch an invalid item or do like:
```suggestion
	if (item && item->card[0] == CARD0_PET && item->attribute == 1)
```

admkakaroto:
Is it possible to block use the item Pet incubator if you already have the window hatching the eggs open?

PR #4196, opened 2019-06-02
Corrected Enchant Arms for older clients

aleos89:
* **Addressed Issue(s)**: N/A

  * Fixes Enchant Arms displaying an incorrect status icon for clients older than 2015-11-04.
Thanks to @Daegaladh!


src/map/status.cpp
```diff
@@ -10325,6 +10325,10 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				val1 = val1%ELE_ALL;
 			else if (val1 < 0)
 				val1 = rnd()%ELE_ALL;
+
+#ifdef PACKETVER < 20151104
```
Daegaladh: if, not ifdef ;)

PR #4200, opened 2019-06-03
Empty message in announce script command

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

This PR added a check to fail silently when the message input is empty.


secretdataz:
Would comparison to null character be more appropriate and more expressive here?
@Normynator

Normynator:
Not sure if script_getstr can return a nullptr but I would suggest:
Keep the nullptr check and add a strlen > 0 check.

Issue #3901, opened 2019-01-31
[suggestion] announce to array group of players gid


sader1992:
the idea is to have another announcement command
`*announce "<text>",<flag>{,<fontColor>{,<fontType>{,<fontSize>{,<fontAlign>{,<fontY>}}}}};`
where the falg would be .@GID_Array

this would be helpful in a lot of things
for example like the eamod bgannounce send the announce to players who have sd->state.bg_listen = true

but this way is very limited
instead we can do something like this
example for how it would look:

```cs
bindatcmd("listenbg",strnpcinfo(3)+"::Onlistenbg",0,99);
bindatcmd("listenpvp",strnpcinfo(3)+"::Onlistenpvp",0,99);
bindatcmd("listengvg",strnpcinfo(3)+"::Onlistengvg",0,99);

Onlistenbg:
	.@ndx = inarray($BG_Group,getcharid(3));
	if(.@ndx == -1){
		$BG_Group[getarraysize($BG_Group)] = getcharid(3);
		message "<character name>","BattleGround announcement ON";
	}else{
		message "<character name>","BattleGround announcement OFF";
		$BG_Group[.@ndx] = 0;
	}
end;

OnEvent_Somthing_Happen_In_Bg:
	announce "Somthing Happen In Bg!!!",$BG_Group;
end;
```
same code for the other commands above
this would expand the announce
ofc it would be another name

_____________________________________________________
while typing the suggestion
I thought why not making the list of players in the server and in each map as an array and we can then edit the command to work on everything ,this would need full edit on the clif_broadcast<< I won't add any info as it just got me , so just saying
something like GID_GROUP
```cs
*announce "<text>",<flag>{,<fontColor>{,<fontType>{,<fontSize>{,<fontAlign>{,<fontY>{,.@array}}}}}};
announce "Somthing Happen In Bg!!!",GID_GROUP,0,0,0,0,0,$BG_Group;
```

I will try to work on the idea above if no one did and if rathena devs are OK with it

luan122:
Hey sader not rushing anything, just curious and excited about it, how is the progress of this feature? :D

PR #4203, opened 2019-06-04
Announce Accept Char id

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/3901

* **Server Mode**: both

the original idea is to make a new GID like `GID_GROUP` and hook it to vector of players
but instead of making things complicated
this will allow `<char_id>` in the `announce` command as it already accept attached char on `bc_self`

* with this PR we can do this:
```cs
OnInit:
	bindatcmd("listenpvp",strnpcinfo(3)+"::Onlistenpvp",0,99);
end;

Onlistenpvp:
	.@ndx = inarray($pvplisteners,getcharid(0));
	if(.@ndx == -1){
		$pvplisteners[getarraysize($pvplisteners)] = getcharid(0);
		dispbottom "PVP Announcement : ON";
	}else{
		deletearray $pvplisteners[.@ndx],1;
		dispbottom "PVP Announcement : OFF";
	}
end;
OnSomethingHappen:
	for(.@i=0;.@i<getarraysize($pvplisteners);.@i++){
		announce "Something happen in PVP!",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,$pvplisteners[.@i];
	}
end;
```

* **test script**:
```cs
//with both online char id 150000 , 150003
prontera,157,170,4	script	announce_to_id	444,{
	announce "this for the char id 150000",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,150000;
	announce "this for the char id 150003",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,150003;
}
```

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6799,7 +6799,7 @@ as soon as the progress bar activates.
 //
 ---------------------------------------
 
-*announce "<text>",<flag>{,<fontColor>{,<fontType>{,<fontSize>{,<fontAlign>{,<fontY>}}}}};
+*announce "<text>",<flag>{,<fontColor>{,<fontType>{,<fontSize>{,<fontAlign>{,<fontY>{,<char_id>}}}}}};
 
 This command will broadcast a message to all or most players, similar to
 @kami/@kamib GM commands.
@@ -6819,7 +6819,7 @@ Target flags:
 You cannot use more than one target flag.
 
 Source flags:
-- bc_pc: Broadcast source is the attached player (default).
+- bc_pc: Broadcast source is the attached player or the char id if it's provided (default).
 - bc_npc: Broadcast source is the NPC, not the player attached to the script
   (useful when a player is not attached or the message should be sent to those
   nearby the NPC).
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -6858,7 +6858,7 @@ but it can be used instead in NPCs to "preview" an announce.
 	announce "This is my message just for you people here",bc_npc|bc_area;
 
 	// This will be a private message to the player with character id 150000
-	announce "This is my message just for char id 150000",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,150000;
+	announce "This is my message just for char id 150000",bc_self,0xFFF618,FW_NORMAL,12,0,0,150000;
 
 ---------------------------------------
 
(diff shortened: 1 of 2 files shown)


Litro:
how much loop with out sleep it will need freeloop ? 
```
OnSomethingHappen:
	for(.@i=0;.@i<getarraysize($pvplisteners);.@i++){
		announce "Something happen in PVP!",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,$pvplisteners[.@i];
	}
end;
```
can this go like just send array name and server will handle the loop
```
OnSomethingHappen:
	announce "Something happen in PVP!",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,$pvplisteners;
end;
```

sader1992:
@Litro 
nope , the command accept int char id , it wont pass the reference of the array

for the free loop , i don't think there is a static number , in general as long as your array is above 50 use a free loop
something like this
```cs
OnSomethingHappen:
	.@size = getarraysize($pvplisteners);
	if(.@size > 50)
		freeloop(1);
	for(.@i=0;.@i<.@size;.@i++){
		announce "Something happen in PVP!",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,$pvplisteners[.@i];
	}
	if(.@size > 50)
		freeloop(0);
end;
```

doc/script_commands.txt
```diff
@@ -6856,6 +6857,9 @@ but it can be used instead in NPCs to "preview" an announce.
 	// This will be shown on everyones screen that is in sight of the NPC.
 	announce "This is my message just for you people here",bc_npc|bc_area;
 
+	// This will be a private message to the player with character id 150000
+	announce "This is my message just for char id 150000",bc_self,0xFFF618,FW_NORMAL,12,NULL,NULL,150000;
```
Atemo: `NULL` hasn't been defined as constant (yet!)
```suggestion
	announce "This is my message just for char id 150000",bc_self,0xFFF618,FW_NORMAL,12,0,0,150000;
```

src/map/script.cpp
```diff
@@ -11103,10 +11103,10 @@ BUILDIN_FUNC(announce)
 		}else{
 			struct map_session_data* sd;
 
-			if( script_rid2sd(sd) )
-				bl = &sd->bl;
-			else
+			if (!script_charid2sd(9, sd))
```
Lemongrass3110: Can you restore the original order here? Inverting the "if" adds a useless "not"-operation and actually all you do is exchange the called function from script_rid2sd to script_charid2sd.

PR #4229, opened 2019-06-28
Small isequippedcnt script command update

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

Updated `isequippedcnt` docs
Updated the command to only count the unique IDs in the `isequippedcnt` parameter

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -15410,38 +15410,43 @@ BUILDIN_FUNC(isday)
 BUILDIN_FUNC(isequippedcnt)
 {
 	TBL_PC *sd;
-	int i, id = 1;
-	int ret = 0;
 
-	if (!script_rid2sd(sd)) { //If the player is not attached it is a script error anyway... but better prevent the map server from crashing...
+	if (!script_rid2sd(sd)) {
 		script_pushint(st,0);
 		return SCRIPT_CMD_SUCCESS;
 	}
 
-	for (i=0; id!=0; i++) {
-		short j;
-		FETCH (i+2, id) else id = 0;
-		if (id <= 0)
-			continue;
+	int ret = 0;
+	int total = script_lastdata(st);
+	std::vector<int32> list;
 
-		for (j=0; j<EQI_MAX; j++) {
+	for( int i = 2; i <= total; ++i ) {
+		int tmp = script_getnum(st,i);
+		if (tmp <= 0)
+			continue;
+		list.push_back(tmp);
+	}
+    std::sort(list.begin(), list.end());
+    list.erase(std::unique(list.begin(), list.end()), list.end());	// remove duplicates ID
+
+	for( auto &id : list ) {
+		for (short j = 0; j < EQI_MAX; j++) {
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -15410,38 +15410,43 @@ BUILDIN_FUNC(isday)
...
 		return SCRIPT_CMD_SUCCESS;
 	}
 
-	for (i=0; id!=0; i++) {
-		short j;
-		FETCH (i+2, id) else id = 0;
-		if (id <= 0)
+	int ret = 0;
+	int total = script_lastdata(st);
+	std::vector<int32> list;
```
secretdataz:
```suggestion
	std::vector<int32> list(total);
```

src/map/script.cpp
```diff
@@ -15410,38 +15410,43 @@ BUILDIN_FUNC(isday)
...
+
+	for ( int i = 2; i <= total; ++i ) {
+		int tmp = script_getnum(st,i);
+		if (tmp <= 0)
 			continue;
+		list.push_back(tmp);
+	}
+	std::sort(list.begin(), list.end());
+	list.erase(std::unique(list.begin(), list.end()), list.end());	// remove duplicates ID
```
secretdataz: If you want an ordered list with unique elements, consider using std::set.

src/map/script.cpp
```diff
@@ -15410,38 +15410,43 @@ BUILDIN_FUNC(isday)
...
 			if (itemdb_type(id) != IT_CARD) { //No card. Count amount in inventory.
 				if (sd->inventory_data[index]->nameid == id)
-					ret+= sd->inventory.u.items_inventory[index].amount;
+					ret += sd->inventory.u.items_inventory[index].amount;
 			} else { //Count cards.
-				short k;
 				if (itemdb_isspecial(sd->inventory.u.items_inventory[index].card[0]))
 					continue; //No cards
-				for(k=0; k<sd->inventory_data[index]->slot; k++) {
+				for ( short k = 0; k < sd->inventory_data[index]->slot; k++ ) {
```
secretdataz:
```suggestion
				for (short k = 0; k < sd->inventory_data[index]->slot; k++) {
```
If you are not going to put a leading space and a trailing space in every parenthesis in this function, don't put one in the for loop.
Also check the loop above.

src/map/script.cpp
```diff
@@ -15410,38 +15410,43 @@ BUILDIN_FUNC(isday)
...
 
-	for (i=0; id!=0; i++) {
-		short j;
-		FETCH (i+2, id) else id = 0;
-		if (id <= 0)
+	int ret = 0;
+	int total = script_lastdata(st);
+	std::vector<int32> list;
+
+	for ( int i = 2; i <= total; ++i ) {
```
secretdataz:
```suggestion
	for (int i = 2; i <= total; ++i) {
```
If you are not going to put a leading space and a trailing space in every parenthesis in this function, don't put one in the for loop.

Issue #3324, opened 2018-07-18
Boss Damage Update


admkakaroto:
* **rAthena Hash**: Any
* **Client Date**: Any
* **Server Mode**: RE
* **Description of Issue**: Now in the kRO the mvp's only receive 10% of the player's damage.
  * Official Information: http://ro.gnjoy.com/news/devnote/View.asp?category=2&seq=3891525&curpage=1

https://media.giphy.com/media/7zSzFATT4pVbJFpv6M/giphy.gif

aleos89:
The Korean translation says _some MVP_. It says the ones who are reduced have a green aura, so I would assume this doesn't apply to all MVP then. We'll need info on which ones are affected.

admkakaroto:
hmmm is now what i saw really seems to be, the mvp's get an icon with a different color to indicate this, i will try to get information and update here.

aleos89:
Interesting. Thanks for the info!

Asheraf:
This is still a sakray only patch I don't think it will pass at least in the current format (I tried to kill an mvp.. it's ridiculous) but who knows maybe they want to kill solo mvp farming or make it very end game.

Lemongrass3110:
Maybe they just try it on more MVPs now and then only put it on special instances or hard modes etc.

bruno199:
https://www.divine-pride.net/forum/index.php?/topic/3318-kro-mvp-bosses-adjustment/

Lemongrass3110:
Putting this on hold until it gets merged from sakray to main.

BrOgBr:
@Asheraf I was thinking that it can have a good side and a very bad one.
Transclasses will a very hard difficult to kill this MVPs, but it could "balance" the high 3rd's damage.
Unless I did misunderstood how it'll work. xD

@Lemongrass3110 On instances would be very nice.

MrAntares:
If this gets implemented we will need an off scwitch for it, otherwise low player count and rate servers will rebel, starting with me :)

AzarthMZintos:
Comment about what MrAntares, if possible will be nice to have a on/off switch and a way to include/turn off all/more mvp/boss, but that maybe be custom feature...

admkakaroto:
@Atemo Let the boys dream!

MrAntares:
We are only discussing the possibilities. It's nice to have a plan, in case this gets implemented ;)

Lemongrass3110:
[...]

PR #4231, opened 2019-07-02
Fix for issue #3324

SonaSonaSona:
* **Addressed Issue(s)**: #3324

* **Server Mode**: Renewall

This pull request adds the "green aura" system, or at least the server side part. 
Some MVPs in newer clients will have green auras that should reduce all damage they take by 90%, so this pull request makes that damage reduction work. (I checked the client files to know which MVPs had the aura)
PS: This doesn't enable the visuals for the auras, that's on the client side.


src/map/battle.cpp
```diff
@@ -1525,7 +1525,11 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 
 		if (flag&BF_MAGIC && bl->type == BL_PC && sc->data[SC_GVG_GIANT] && sc->data[SC_GVG_GIANT]->val4)
 			damage += damage * sc->data[SC_GVG_GIANT]->val4 / 100;
-
+
```
Daegaladh: This is a Renewal only change, so please add `#ifdef RENEWAL`

SonaSonaSona: I dont think that's necessary cause the race2 mobs were added in the re/mob_race2_db.txt

Daegaladh: Oh, that's right, nevermind

SonaSonaSona:
> Battle config(s) please.
> Minimum is on/off.
> A good addition would be to be able to specify the reduction in the config as well

Done, tested and It works with @reloadbattleconf atcommand too

conf/battle/monster.conf
```diff
@@ -262,3 +262,9 @@ monster_hp_bars_info: yes
 // This can be legit gameplay (e.g. players keeping an MVP stuck inside icewall), but if you want to prevent any
 // exploits and be notified about them, you can set this to yes.
 monster_stuck_warning: no
+
+// Green Aura Reduction: (Note 2)
+// How much damage do MVPs with green aura (RC2_GREENAURA) should ignore.
+// 0: Disabled/Take full damage
+// 1~100: % of the damage they ignore (official is 90)
+greenaura_reduction_rate: 0
```
Daegaladh: The default value should be the official

SonaSonaSona: I left it that way because that change is implemmented on 17.1 and people could get confused with MVPs taking significant less damage, since rathena is missing some 17.1 content.
Let me know if it still needs to change.

laziem:
Should the reduction works for all kind of damage (skill)? I run a few test, when I use Cart Cannon - genetic, the damage get reduced, but when I use Spear Boomerang / Axe Boomerang / Arrow Shower / Bash  / Dragon Breath etc, the damage did not get reduced.

SonaSonaSona:
For some reason it is only working if the user have some type of status change or a cart (I believe it counts as a status change). I'm trying to figure out why this happens, and then fix.

> Should the reduction works for all kind of damage (skill)? I run a few test, when I use Cart Cannon - genetic, the damage get reduced, but when I use Spear Boomerang / Axe Boomerang / Arrow Shower / Bash / Dragon Breath etc, the damage did not get reduced.

Thats's why Cart Cannon was working, because you need a cart to use the skill.

Questune09:
Is this also related to this?

https://www.divine-pride.net/forum/index.php?/topic/3318-kro-mvp-bosses-adjustment/

conf/battle/monster.conf
```diff
@@ -262,3 +262,9 @@ monster_hp_bars_info: yes
 // This can be legit gameplay (e.g. players keeping an MVP stuck inside icewall), but if you want to prevent any
 // exploits and be notified about them, you can set this to yes.
 monster_stuck_warning: no
+
+// Green Aura Reduction: (Note 2)
+// How much damage do MVPs with green aura (RC2_GREENAURA) should ignore.
```
aleos89:
```suggestion
// How much damage MVPs with a green aura (RC2_GREENAURA) ignore.
```

db/re/mob_race2_db.txt
```diff
@@ -51,3 +51,6 @@ RC2_ROCKRIDGE,3736,3737,3738,3739,3740,3741,3742,3743,3744,3745,3746,3747,3748,3
 //RC2_HEARTHUNTER,3626,3627,3628,3629
 // Werner Laboratory
 //RC2_WERNER_LAB,3621,3622,3631,3632,3633 // miss ID 3826 - YSF Seyren monster
+// Green Aura 90% Reduced damage
```
aleos89:
```suggestion
// Green Aura MVP
```

db/re/mob_race2_db.txt
```diff
@@ -51,3 +51,6 @@ RC2_ROCKRIDGE,3736,3737,3738,3739,3740,3741,3742,3743,3744,3745,3746,3747,3748,3
...
+// Just uncomment the line below to enable reduced damage. PS: The green aura effect is clientside.
```
aleos89: This comment is useless now. Also, don't forget to uncomment the group below since the battle_config manages this now.

src/map/battle.cpp
```diff
@@ -1512,6 +1512,13 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 			status_change_spread(bl, src, 1); // Deadly infect attacked side
 
 	} //End of target SC_ check
+	
+	if (status_get_race2(bl) == RC2_GREENAURA) {
+		int finaldmg;
+		finaldmg = 100 - battle_config.greenaura_reduction_rate;
+		if (status_get_race2(src) != RC2_GREENAURA) // [Sona]: This is for MvP vs MvP event purposes. (damage doesn't get reduced when attacking eachother)
+			damage = damage > 100 ? damage * finaldmg / 100 : 1; // Reduces damage received by the ammount specified on conf/battle/monster.conf
```
aleos89:
```suggestion
			damage = damage > 100 ? damage * (100 - battle_config.greenaura_reduction_rate) / 100 : 1; // Reduces damage received by the amount specified in conf/battle/monster.conf
```
You could remove the declaration of the variable and just put the config inside the formula.

src/map/battle.cpp
```diff
@@ -1525,7 +1532,7 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 
 		if (flag&BF_MAGIC && bl->type == BL_PC && sc->data[SC_GVG_GIANT] && sc->data[SC_GVG_GIANT]->val4)
 			damage += damage * sc->data[SC_GVG_GIANT]->val4 / 100;
-
+
```
aleos89:
```suggestion

```
Left over blank spacing.

src/map/battle.cpp
```diff
@@ -1512,7 +1512,11 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 			status_change_spread(bl, src, 1); // Deadly infect attacked side
 
 	} //End of target SC_ check
-
+	
+	if (status_get_race2(bl) == RC2_GREENAURA) {
+		if (status_get_race2(src) != RC2_GREENAURA) // [Sona]: This is for MvP vs MvP event purposes. (damage doesn't get reduced when attacking eachother)
```
aleos89: Just curious, could this case even happen where a monster attacks another? I can't think of anything on kRO that does this.

SonaSonaSona: I've seen some cases where bosses battle eachother if hit by a sanctuary of another one

saya9200:
Man this feature fuilly implemented would be so good. GX just shredding everything on every server with minimal efforts, doing numbers that are absolutely braindead. this could help. I wonder how difficult it is to add the new skills.

aleos89:
Taken over in #5749. Thank you for your contribution!

PR #4248, opened 2019-07-15
Corrected Magnum Break delay

### LINKED ISSUE #4191: Magnum Break Delay
* **rAthena Hash**:  a3fc44e

* **Client Date**: No client

* **Description of Issue**: 
  * Result: Magnum Break has a hardcoded skill-specific cooldown of 2 seconds (Duration1) which cannot be reduced by Bragi. 
  * Expected Result: Magnum Break should have an aftercast delay of 2 seconds which can be reduced by Bragi (like all other skills) and no cooldown. 
  * How to Reproduce: Cast Magnum Break, then try to use another skill during the delay (works even if though it shouldn't). Spam Magnum Break when having Bragi Buff (cannot spam, still have to wait 2 seconds). 
  * Official Information: Magnum Break on emulators always had this weird skill-specific hardcoded delay of 2 seconds. This implementation was done right at the start (probably came straight from jAthena). Players from the old euRO (pre-re) and iRO classic both confirmed that the delay was a normal 2 second global aftercast delay that was reduced by bragi. In this thread, it is also confirmed that aftercast of Magnum Break is reduced by Bragi (Magic Strings is the iRO name): https://oldforums.irowiki.org/index.php?topic=274.0 (note: this thread is from 2007, before the renewal update!)

Either the implementation has already been wrong in jAthena und nobody ever questioned it, or the skill was changed sometime during pre-renewal times (maybe with the addition of 2-2 classes).

The hardcoded cooldown of Magnum Break doesn't make any sense anymore in the first place as we have a normal cooldown implementation in rAthena.

So what needs to be done is removing the following code in skill.cpp:

```
		if( sd )
			skill_blockpc_start(sd, skill_id, skill_get_time(skill_id, skill_lv));
		else if( bl->type == BL_MER )
			skill_blockmerc_start((TBL_MER*)bl, skill_id, skill_get_time(skill_id, skill_lv));
```

And update pre-re/skill_cast_db.txt to:

```
//-- SM_MAGNUM
7,0,2000,0,0,10000,0
```

Note: In renewal the code above is also not needed. The 2018 1st/2nd/trans skill rebalance update from 2018 finally added a 2000ms cooldown to Magnum Break, but we don't need to have this weird hardcoded delay and instead just use the cooldown column for it.

So re/skill_cast_db.txt should look like this in the rebalance update:
```
//-- SM_MAGNUM
7,0,500,0,0,10000,2000,0
```

Note: I can't really access the repository anymore, otherwise I'd fix it myself.

aleos89:
* **Addressed Issue(s)**: #4191

  * Fixes #4191.
  * Magnum Break should have an aftercast delay of 2 seconds which can be reduced by Bragi (like all other skills) and no cooldown.
Thanks to @Playtester!


db/re/skill_cast_db.txt
```diff
@@ -21,7 +21,7 @@
 //-- SM_PROVOKE
 6,0,0,0,30000,0,0,0
 //-- SM_MAGNUM
-7,0,2000,0,2000,10000,0,0
+7,0,500,0,2000,10000,2000,0
```
Playtester: I think Duration1 can be set to 0 in renewal as well.

PR #4312, opened 2019-08-22
Fixes #4311

### LINKED ISSUE #4311: Inventory check while buying item with GUID flag
* **rAthena Hash**: 92ba773

* **Client Date**: 2018-06-20 for testing

* **Server Mode**: Both

* **Description of Issue**: 
  * Result: If not enough room in inventory, player's cash/zeny will be reduced as requested number but the item's number is truncated
  * Expected Result: Fail to buy if player doesn't have enough inventory space at first
  * How to Reproduce:
    1. Set item with GUID flag into db/import/item_cash_db.txt
    2. Set inventory to 99/100, 
    3. Buy the item with amount 2 or more
    4. on purchase, it display failure message but player receive only 1 item but the cash reduced in full payment
  * Official Information: (well, offcial server always do each check-then-buy/sell item unlike our collective check, so, not related)

* **Modifications that may affect results**: none

cydh:
* **Addressed Issue(s)**: #4311

* **Server Mode**: Both

  * Added item with GUID flag as `CHKADDITEM_NEW` in pc_checkadditem
  * Adjust inventory blank check for guid item to `+n`

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -15744,7 +15744,7 @@ void clif_parse_Mail_getattach( int fd, struct map_session_data *sd ){
 
 				switch( pc_checkadditem(sd, item->nameid, item->amount) ){
 					case CHKADDITEM_NEW:
-						new_ += itemdb_inventory_slot_needed(data, item->amount);
+						new_++;
 						break;
 					case CHKADDITEM_OVERAMOUNT:
 						clif_mail_getattachment(sd, msg, 2, MAIL_ATT_ITEM);
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -15744,7 +15744,7 @@ void clif_parse_Mail_getattach( int fd, struct map_session_data *sd ){
 
 				switch( pc_checkadditem(sd, item->nameid, item->amount) ){
 					case CHKADDITEM_NEW:
-						new_++;
+						new_ += itemdb_inventory_slot_needed(data, item->amount);
 						break;
 					case CHKADDITEM_OVERAMOUNT:
 						clif_mail_getattachment(sd, msg, 2, MAIL_ATT_ITEM);
--- a/src/map/itemdb.cpp
+++ b/src/map/itemdb.cpp
@@ -1986,7 +1986,7 @@ bool item_data::isStackable()
 
 int item_data::inventorySlotNeeded(int quantity)
 {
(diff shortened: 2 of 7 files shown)


src/map/itemdb.cpp
```diff
@@ -514,6 +514,10 @@ bool itemdb_isstackable2(struct item_data *id)
 	}
 }
 
+int itemdb_inventory_slot_needed(item_data *id, int amount) {
```
Lemongrass3110: Since this is a general function:
Please add a stackable call as well in case someone calls the function without checking before.

Lemongrass3110: And since we use C++ now. Maybe change it to call by reference. That way nullptr for id is not possible.

src/map/clif.cpp
```diff
@@ -15744,7 +15744,7 @@ void clif_parse_Mail_getattach( int fd, struct map_session_data *sd ){
 
 				switch( pc_checkadditem(sd, item->nameid, item->amount) ){
 					case CHKADDITEM_NEW:
-						new_++;
+						new_ += itemdb_inventory_slot_needed(data, item->amount);
```
cydh: shouldn't need here, if the attached item is GUID'd item with a stack of 10, player should get it as 10 in 1 inventory slot
```suggestion
						new_++;
```

src/map/vending.cpp
```diff
@@ -204,7 +204,7 @@ void vending_purchasereq(struct map_session_data* sd, int aid, int uid, const ui
 		case CHKADDITEM_EXIST:
 			break;	//We'd add this item to the existing one (in buyers inventory)
 		case CHKADDITEM_NEW:
-			new_++;
+			new_ += itemdb_inventory_slot_needed(itemdb_search(vsd->cart.u.items_cart[idx].nameid), amount);
```
cydh: shouldn't need here, if the vended item is GUID'd item with a stack of 10, player should get it as 10 in 1 inventory slot
```suggestion
			new_++;
```

ecdarreola:
@cydh after the last change (**6f60e44**) an issue is introduce where in message is displaying "out of the maximum capacity" and failing to buy in NPC.

https://user-images.githubusercontent.com/48939874/65260551-1eed5580-db39-11e9-9955-f6bb631bf119.png

Before that, this message is not showing or no issue at all.

Issue #4317, opened 2019-08-25
Bug about RODEX item and weight


voyfmyuh:
- rAthena Hash: Latest

- Client Date: 20180621

- Server Mode: Renewal

When you remove an item, the weight will be 0.
![9i4lj-n30pe](https://user-images.githubusercontent.com/54499336/63653962-b237a300-c7a6-11e9-9797-174ad2011dc3.gif)

PR #4332, opened 2019-09-01
Fixed item removal in RODEX

Lemongrass3110:
* **Addressed Issue(s)**: #4317

* **Server Mode**: Both

Fixed RODEX not displaying any weight after removing any item. Thanks to @voyfmyuh for this.
Additionally fixed RODEX internally always removing the item completely, when it was only partially removed.


pigati23:
Works perfectly now.

Issue #4343, opened 2019-09-12
Bug Reflect Damage (Item bonus+Skill Reflect)


Badarosk0:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/9aa5f7d8345d9d89ec9e8b44060e92703f3d43b8

* **Client Date**: 20180621

* **Server Mode**: RE

* **Description of Issue**: When using any item that adds melee reflect damage the reflect to the target occurs only up to the HP size of the player who received the damage. 
The problem also occurs with skill Reflect.

  * Result: 
It is only reflecting up to the size of the HP from who received the damage, who had a reflect bonus. Example who has the card Orc Lord.
  * Expected Result:
 It should reflect the full percentage of the item used and not the size of the HP.
  * How to Reproduce: 
Equip an Orc Lord card. Take a Sura and Azura strike the target. The reflected damage will be the target's HP value.
  * Official Information:
  

* **Modifications that may affect results**:

aleos89:
Most likely because of this define: `int max_damage = status_get_max_hp(bl);` add here: ad11dd6.

Although, does it matter? The maximum possible damage you could do is the target's HP. It would be nice to see the initial bug report from the old tracker. Guess we need to see if Aegis is capping it as well.

aleos89:
Fixed in [33a99b5](https://github.com/rathena/rathena/commit/33a99b59931346097c684c305c58a4eae97088e7).

PR #4352, opened 2019-09-17
Cleaned up reflected damage cap

aleos89:
* **Addressed Issue(s)**: #4343

  * Reflected damage in renewal should only be capped to the target's HP.
Thanks to @exneval and @Badarosk0!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -6881,8 +6881,14 @@ struct Damage battle_calc_attack(int attack_type,struct block_list *bl,struct bl
 int64 battle_calc_return_damage(struct block_list* bl, struct block_list *src, int64 *dmg, int flag, uint16 skill_id, bool status_reflect){
 	struct map_session_data* sd;
 	int64 rdamage = 0, damage = *dmg;
-	int max_damage = status_get_max_hp(bl);
 	struct status_change *sc, *ssc;
+#ifdef RENEWAL
+	int max_damage = status_get_max_hp(bl);
+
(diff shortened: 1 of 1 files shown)


ecdarreola:
Does this cause all attack to have reflect damage of 1?

cahya1992:
> After applying this PR, all attacks may it be normal or not using different classes receives reflect damage of 1.

Confirm too this issue

Badarosk0:
True, it is not working.

teededung:
Before applying this PR:
Ranger's skill Cluster Bomb is reflected up to 3 hits by Reflect Shield.

After applying the PR: 
Cluster Bomb is reflected 2 hits by Reflect Shield.

Is it true about reflected hit of the skill?

GIF

Badarosk0:
Still in trouble. Not yet reflecting as the correction.

Badarosk0:
Reflect damage taken is still limited to the target's HP. So if I have a higher HP than the target I will never die.

Badarosk0:
Can you update this PR with the latest version rA? @aleos89 
Thank you

ecdarreola:
Crash signal received. 
[bt full reflect.txt](https://github.com/rathena/rathena/files/4811494/bt.full.reflect.txt)

Badarosk0:
When applying this PR, any damage done to the target is returning 1 damage to the player. @aleos89

aleos89:
Please refer to #6717 for the damage cap fix.

Issue #4364, opened 2019-09-21
Item 1164, 2652 wrong script


SeravySensei:
These items curse the owner at a chance **when attacking**.
So their script should use "autobonus" instead of "bonus2".
I tested and attacking while wearing the item(s) causes curse significantly more often than "rate" specified using the existing script. (bonus2 is ran on each stat calculation I believe.)
Furthermore, according to Divine Pride, these effects only trigger on melee attacks.
https://www.divine-pride.net/database/item/2652/goddess-of-fortunes-cursed-brooch
https://www.divine-pride.net/database/item/1164/muramasa

cydh:
```
bonus2 bAddEff2,eff,n;      		Adds a n/100% chance to cause status eff on self when attacking
```

SeravySensei:
Then why was it triggering a lot more often than specified? I mean, I literally tried reducing the number from 50 to 5 but it still cursed me all the time when it should have been only 1 in 2000 attacks. I'm not an expect on scripting but something definitely is wrong here. Can someone else test this item to confirm it has a problem?

SeravySensei:
Did some more tests.
Using "bonus bCritical,6; bonus2 bAddEff2,Eff_Curse,50;" I got cursed roughly 5-6 times in using 100 arrows.
Using "bonus bCritical,6; bonus2 bAddEff2,Eff_Curse,5;" I still got cursed roughly 5-6 times in using 100 arrows.
Using "bonus2 bAddEff2,Eff_Curse,5;" gave me 4-5 curses.
So regardless of the contents of the script, the curse chance was roughly 5%. Unequipping the bow and attacking unarmed also had the same effect. It seems the rate parameter doesn't work.
Then I tried "bonus2 bAddEff2,Eff_Freeze,500;" (Ice Falchion for example has that effect although at a lower rate), gave me 1 freeze per roughly 20 arrows.
Next I tried "bonus2 bAddEff2,Eff_Freeze,9000;" , this should have been a 90% chance and indeed I got frozen on pretty much every attack.
Conclusion : bAddEff2 seems to work as intended on Freeze but does not for Curse. No idea why. (Note, the effect was tested on an accessory. Haven't tested other types of items.)

SeravySensei:
The problem causing this issue was not the script itself but the status resistance calculation.

PR #4365, opened 2019-09-21
Fix self-curse items

SeravySensei:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/4364

* **Server Mode**: 
Renewal

Note : I haven't checked if the bug is present in the pre-renewal database. If yes, someone should fix those as well.
I found no official information on the duration of the curse effects. This PR sets it to 20 seconds. Change that if you know the correct amount.


SeravySensei:
The problem causing this issue was not the script itself but the status resistance calculation.

PR #4373, opened 2019-09-25
Additional check in map_mapid2mapname

Atemo:
* **Addressed Issue(s)**: /

* **Server Mode**: /

The mapid2name script command didn't check if the map existed (potential crash)
```
/// Returns the map name of given map ID.
///
/// mapid2name <map ID>;
BUILDIN_FUNC(mapid2name)
{
	uint16 m = script_getnum(st, 2);

	if (m < 0) {
		script_pushconststr(st, "");
		return SCRIPT_CMD_FAILURE;
	}

	script_pushstrcopy(st, map_mapid2mapname(m));

	return SCRIPT_CMD_SUCCESS;
}
```

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3114,7 +3114,7 @@ Notice that NPC objects disabled with 'disablenpc' will still be located.
 
 *mapid2name(<map ID>)
 
-Returns the map name of the given map ID. Returns "Unknown" if the given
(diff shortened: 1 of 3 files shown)


aleos89:
Might need to adjust the script command documentation which could return "Unknown" if it's an invalid map as well.

Issue #4378, opened 2019-10-01
achievementexists does not check completion


Lemongrass3110:
* **rAthena Hash**: ac03cea

* **Client Date**: -

* **Server Mode**: Both

* **Description of Issue**: 
  * Result: If an achievement has counting variables it is added to the achievements of the player as non completed. A missing check in the script command `achievementexists` makes it look like the player has completed the achievement already though.
  * Expected Result: Only return true if the player has completed the achievement.
  * How to Reproduce: -
  * Official Information: -

* **Modifications that may affect results**: 
None

PR #4326, opened 2019-08-28
Updated Achievement EXP to official values

aleos89:
* **Addressed Issue(s)**: #4266

* **Server Mode**: Pre-renewal and Renewal

* **Description of Pull Request**: 
  * Added the official values for the AG_GOAL_ACHIEVE.

TODO:
- [x] Either move the EXP to it's own database or create a sub-label within the current Achievement Database structure to support it.

PR #4379, opened 2019-10-01
Fixed achievementexists script command

Lemongrass3110:
* **Addressed Issue(s)**:  #4378

* **Server Mode**: Both

Fixes the issue by adding the check for completing the achievement.


ecdarreola:
Works like a charm together with the following PR

#4381 and #4326 

Thanks!

PR #4390, opened 2019-10-03
Adding portal in Yuno City

JohnnyPlayy:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

Information taken directly from KRO
image
I noticed that there is a missing NPC

[Missing script](https://www.pastiebin.com/5d9688bec5ddb)


aleos89:
The NPC looks like it's part of #3037.

npc/quests/quests_lighthalzen.txt
```diff
@@ -4206,7 +4207,7 @@ lhz_que01,99,74,5	script	#li_Man	48,{
 	end;
 }
 
-lhz_que01,90,71,5	script	#li_bird	1019,3,3,{
+lhz_que01,90,71,5	script	#li_bird	PECOPECO,3,3,{
```
aleos89: You can leave this as `1019` since we don't export mob names (they are technically defined in the mob_db.

JohnnyPlayy: Change made https://github.com/rathena/rathena/pull/4392

Issue #4408, opened 2019-10-12
Blacksmith ranking point should be added on successful +10 refine, not MAX_REFINE


secretdataz:
* **rAthena Hash**: 783009d

* **Client Date**: Irrelevant

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Blacksmith ranking point is increased on successful +20 refine attempt via an NPC. Refining via WS_WEAPONREFINE behaves as expected.
  * Expected Result: Points should be given on successful +10 refine attempt.
  * How to Reproduce:
      `@produce Knife` then equip the knife you just created.
      `@refine 2 9`
      Try refining the knife one more time via NPC. Repeat until it's successful.
  * Official Information: N/A. rAthena should make the behavior consistent between WS_WEAPONREFINE and `successrefitem` script command.

* **Modifications that may affect results**: None

cydh:
How to reproduce: should increase refine success rate to +10 before
refining it xD

Well, still in idRO the fame bonus when reach +10.

PR #4409, opened 2019-10-12
Threshold for blacksmith refine fame points

secretdataz:
* **Addressed Issue(s)**: #4408

* **Server Mode**: Renewal

* **Description of Pull Request**: Adjusted default threshold to +10 for both pre-renewal and renewal. Also made it configurable.


Lemongrass3110:
@secretdataz could you hit the squash and merge button please? :-)

PR #4413, opened 2019-10-16
Small correction of ritual of blessing instance

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal

This PR essentially corrects some event timers.
The hp of  `Suspicious Intruder` monsters are changed to 10000 after being spawn (their max hp doesn't change).


Badarosk0:
I don't know if it has anything to do with this change. But I realized that if you kill Bijou too fast, you can kill her 2 times.

Atemo:
Bijou spawn in the room of consciousness (the daily instance), this PR corrected the instance used during the main quest. But I will look into it ;)

Issue #4406, opened 2019-10-12
Pet bugs


teededung:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/783009d642ecb29b0b11447167cc36cf736045ee

* **Client Date**: 20180621
* **Server Mode**: Pre-renewal, Renewal
* **Description of Issue**: 
  * Result: Pet egg bugs
  
  * How to Reproduce:
Bug 1:  No pet egg in inventory but you still have the pet
1. Hatch an egg to show Pet list menu (just showing the menu, do not choose the egg yet)
2. Request trading with another character
3. Complete the trading with the egg that you are going to incubate
4. Select the egg still displays in the pet menu and ...!!

Bug 2: You still have egg in your inventory when the pet leaves you (Intimate = 0). This will make the next incubation you will see pet name with red color in the menu list.

Badarosk0:
This problem has also happened to me.

sader1992:
i could not reproduce the issue of the trade
and in the server side , it check for the egg after selection
so even if you trade , if the `item['index'] != egg` you will get an error in the server side `wrong egg item inventory 'index'`
the client glitches if you get back that egg again , but it works fine .

maybe you have another way to reproduce the issue?

teededung:
@sader1992 Please check this video https://www.youtube.com/watch?v=sFXZgfLsSqU

sader1992:
Ok thanks , i see what you mean now.
I can confirm both issues.

PR #4422, opened 2019-10-27
fix several issues with pets

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/4406

* **Server Mode**: both

Thansk to @teededung 

1. fix where you can hatch egg while you are trading.
2. fix where egg not deleted after the pet is lost.

for test:
1. `@item 9005` , `@hatch` , trade with another player and put the egg in the trade window , before finalize the trade after the ok in both sides , complete hatching the pet , that complete the trade.

2. `@item 9005` , `@hatch` and hatch the pet , `@petfriendly 1` `@pethungry 1` and wait
also `@item 9005` `@petfriendly 0`

3. put in the item db `60110,Test_Egg,Test Egg,7,20,,0,,,,,,,,,,,,,{},{},{}` , get the item , and try hatching it without adding the pet in the pet db (thanks to @admkakaroto).

Side Note: there is an issue where the player can't hatch new pet after the old pet is lost unless he relogin , it need to set `sd->status.pet_id = 0` , however this break the random walk behavior

### DIFF
--- a/src/map/pet.cpp
+++ b/src/map/pet.cpp
@@ -606,6 +606,9 @@ void pet_set_intimate(struct pet_data *pd, int value)
 
 	struct map_session_data *sd = pd->master;
 
+	if(pd->pet.intimate <= PET_INTIMATE_NONE)
+		pc_delitem(sd, pet_egg_search(sd, pd->pet.pet_id), 1, 0, 0, LOG_TYPE_OTHER);
+
 	if (sd)
 		status_calc_pc(sd,SCO_NONE);
 }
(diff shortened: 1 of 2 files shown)


admkakaroto:
There is another problem too, if you hatch a pet egg that doesn't exist in pet_db, the egg gets broken in inventory and you can't get rid of it at all, it just disappears from inventory but you can see it in hatch window as broken.

src/map/pet.cpp
```diff
@@ -1148,6 +1151,11 @@ int pet_select_egg(struct map_session_data *sd,short egg_index)
 	if(egg_index < 0 || egg_index >= MAX_INVENTORY)
 		return 0; //Forged packet!
 
+	if(sd->trade_partner){	//The player have trade in progress.
+		clif_displaymessage(sd->fd, "You need to complete the trade first.");
```
Daegaladh: Please, do not hardcode messages, move it to map_msg.conf

sader1992: Thinking about it , it's better to keep this information from the player , i think it's better to remove the message.

Issue #4449, opened 2019-11-14
Right-click view player's equipment is crazy


cydh:
* **rAthena Hash**: ef81cefd5663445c4da745ce9bcb4e06ea3677e4
* **Client Date**: -
* **Server Mode**: Both
* **Description of Issue**: 
  * Result: 
     1. (technically) when player ask to view other's equipment, the server will sends all equip-type items, not only the equipped items only. (even in player's window will show the correct equipment, but the data was sent are everything)
     2. there's no range limit, as long as the target and who request  in same map-server, it's possible to view target's equipments
  * Expected Result: 
     1. Only send necessary item to the player who requested
     2. Limited to player in view area
  * How to Reproduce: 
    1. By using debug-mode and make breakpoints or use packet sniffer, then right-click to view other player's equipment, read the number of items were sent or the read the packet by sniffer.
    2. Use packet forger, only need to send the target's AID
  * Official Information: I believe official servers do it right, if no, we should prevent this. 
* **Modifications that may affect results**: -

mazvi:
Nice

Lemongrass3110:
Fixed in 5bcbed5

PR #4450, opened 2019-11-14
* Fixed equip view to only send equipped item

cydh:
* **Addressed Issue(s)**: #4449
* **Server Mode**: Both
  * Fixed equip view to only send equipped item
  * Limiting only can view player's equip only at same map (tested on official server, and no fail message)


src/map/clif.cpp
```diff
@@ -16698,7 +16700,11 @@ void clif_parse_ViewPlayerEquip(int fd, struct map_session_data* sd)
 	if (!tsd)
 		return;
 
-	if( tsd->status.show_equip || pc_has_permission(sd, PC_PERM_VIEW_EQUIPMENT) )
+	if (pc_has_permission(sd, PC_PERM_VIEW_EQUIPMENT))
```
Lemongrass3110: Should still be on the same map or not?

cydh: because it's admin permission, that's why ignore the map restriction.
or what? this from packet request always strict the map, then make atcommand that ignore map and map-server restrtiction /gg

Lemongrass3110: I dont think adding „features“ like this intentionally is a good idea.
If we forgot to add a check somewhere and you know of it feel free to add the missing ones though or fix the order.

cydh: :thinking: what is this, just like I'm sweeping dirty floor and someone comes ask for "are you clearing this properly" while checking the dirts and make a mess, then said "oh, seems nothing there what I'd like to find"
:rofl:

Lemongrass3110: You are not clearing it properly. You clear one dirt pile and create a dirt pile closeby. 
So telling you this is like:
You did a good start, but you might be able to improve it even further. It is no offense.

Issue #4453, opened 2019-11-16
skill crit bonus not applied (like Sharp Shooting)


sader1992:
* **rAthena Hash**: 34e5f537b73440589ba50532357b3bb7c9977cc5

* **Client Date**: 2019

* **Server Mode**: renewal

* **Description of Issue**: 
  * Result: skill crit bonus does not apply
  * Expected Result: some skills have crit and should apply crit bonus
  * Official Information: https://irowiki.org/wiki/Focused_Arrow_Strike
![image](https://user-images.githubusercontent.com/5981261/68987230-90433f00-0830-11ea-9c24-abed4f21db61.png)

  * How to Reproduce: 
this all you need to test:
`@warp prontera,110,84`
```
prontera,110,84,0	script	SN_SHARPSHOOTING_Test	444,{
	mes "change class ?";
	if(select("no:yes") == 2){
		jobchange JOB_SNIPER;
		resetstatus;
		BaseLevel = MAX_LEVEL;
		JobLevel = 70;
		atcommand "@allskill";
		getitem 1701,1;
		equip 1701;
		getitem 1753,1000;
		equip 1753;
		percentheal 100,100;
	}
	clear;
	mes "get crit item ?";
	if(select("no:yes") == 2){
		.@i0 = RDMOPT_VAR_CRITICALSUCCESSVALUE;
		.@i1 = 100;
		.@i2 = 0;
		getitem3 2220,1,1,0,0,0,0,0,0,.@i0,.@i1,.@i2;
		equip 2220;
	}
	clear;
	mes "done";
end;

OnInit:
OnKill:
	monster "prontera",102,78,"Barricade",1905,1,strnpcinfo(3) + "::OnKill";
	.monsterid = $@mobid[0];
	setunitdata .monsterid,UMOB_MAXHP,1000000000; 
	setunitdata .monsterid,UMOB_HP,1000000000; 
end;
}
```
given that you created a new char and talk to the npc > yes > yes
your damage `with Sharp Shooting level 10` on the `Barricade` will be always `662` even with the hat that apply 100 crit

* **Modifications that may affect results**: none

Issue #4128, opened 2019-04-24
Focus arrow damage should increase when crit.


attackjom:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/53cedb72af3c1ffd863c02cd716566b50cfe999d

* **Client Date**:  ragexe2018-09-19a

* **Server Mode**: Renewal

* **Description of Issue**: 
Focus arrow strike skill of Sniper class not do more damage when critical and can't take critical damage bonus from item.

  * Result: 
  * Expected Result: 
   - It should deal critical damage and gain CritAtkrate itembonus
  * How to Reproduce: 

  * Official Information:
  

Today I have test Focus arrow strike skill in KRO server, and found these information

1. Focus arrow will improve damage when do critical damage around 1.4x as you can see in the picture

// Not critical //
![screen»çÅ©¶óÀÌ010](https://user-images.githubusercontent.com/35770095/56668976-a2b22b80-66da-11e9-88dd-10c23d156a7c.jpg)

// Critical //
![screen»çÅ©¶óÀÌ009](https://user-images.githubusercontent.com/35770095/56668995-a80f7600-66da-11e9-81c3-fbce543433ba.jpg)

2. Focus arrow  have a bonus from criticaldamage itembonus ex. The Paper card
![screen»çÅ©¶óÀÌ012](https://user-images.githubusercontent.com/35770095/56669089-cf664300-66da-11e9-8f41-ae155f06cf45.jpg)

* **Modifications that may affect results**:

aleos89:
The Paper Card bonus is currently implemented to only give a bonus on physical attacks. Guess this proves it applies to skills, at least in renewal.

aleos89:
This is part of #4072.

PR #4454, opened 2019-11-16
Fixed some skill crit bonuses

sader1992:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/4453 https://github.com/rathena/rathena/issues/4460

* **Server Mode**: renewal

* **Description of Pull Request**: fix where skill crit bonus does not apply

to test
use this npc

create new char
`@warp prontera,110,84`
talk to the npc > yes > yes
given that you didn't modify your char after that
(you would be equipped with hat +100 crit )
before the fix , your damage with sharp shooting would be `662` with or without the hat
after the fix , your damage without the hat would be `662~926` and with the hat always `926`
```
prontera,110,84,0	script	SN_SHARPSHOOTING_Test	444,{
	mes "change class ?";
	if(select("no:yes") == 2){
		jobchange JOB_SNIPER;
		resetstatus;
		BaseLevel = MAX_LEVEL;
		JobLevel = 70;
		atcommand "@allskill";
		getitem 1701,1;
		equip 1701;
		getitem 1753,1000;
		equip 1753;
		percentheal 100,100;
	}
	clear;
	mes "get crit item ?";
	if(select("no:yes") == 2){
		.@i0 = RDMOPT_VAR_CRITICALSUCCESSVALUE;
		.@i1 = 100;
		.@i2 = 0;
		getitem3 2220,1,1,0,0,0,0,0,0,.@i0,.@i1,.@i2;
		equip 2220;
	}
	clear;
	mes "done";
end;

OnInit:
OnKill:
	monster "prontera",102,78,"Barricade",1905,1,strnpcinfo(3) + "::OnKill";
	.monsterid = $@mobid[0];
	setunitdata .monsterid,UMOB_MAXHP,1000000000; 
	setunitdata .monsterid,UMOB_HP,1000000000; 
end;
}
```


sader1992:
this might be related https://github.com/rathena/rathena/issues/4128

cydh:
hmm need more clean up, the NK_CRITICAL check there and call is_attack_critical is redundant. while in is_attack_critical there are too-ealry-may-not-be-used declarations

src/map/battle.cpp
```diff
@@ -2373,21 +2373,27 @@ static bool is_attack_left_handed(struct block_list *src, int skill_id)
...
 	if (!first_call)
 		return (wd->type == DMG_CRITICAL || wd->type == DMG_MULTI_HIT_CRITICAL);
 
-	if (skill_id == NPC_CRITICALSLASH || skill_id == LG_PINPOINTATTACK) //Always critical skills
-		return true;
+	if( skill_id && !skill_get_nk(skill_id)&NK_CRITICAL )
+		return false;
 
-	if( sstatus->cri && ( !skill_id || skill_get_nk(skill_id)&NK_CRITICAL ) )
+	switch(skill_id) { //Always critical skills
```
cydh: unnecessary, moved to skill_db already

sader1992: but if we removed them from here , they will enter the `sstatus->cri` check for a chance to crit
while they should always crit without checking for the chance of crit , or that what i understood from the code

in https://irowiki.org/wiki/Pinpoint_Attack
`Always deals Critical Damage, regardless of user's Critical Hit Rate.`

and https://irowiki.org/wiki/NPC_CRITICALSLASH
`Deals a Critical Attack at a target.`

cydh: revert the skill db then

sader1992: the skill db idea , is very useful , if someone want to disable the skill crit , he can just edit the skill_db
the only problem in the skill db i think is we can't control the crit chance from there

i think it's better to leave the skill_can_crit option in the skill_db , while here we control the crit chance.
ofc this only my opinion xD

Lemongrass3110: I share @sader1992's opinion on this. Having it in skill db is fine, but those 100% crit rate checks are required.

src/map/battle.cpp
```diff
@@ -3259,7 +3265,7 @@ static void battle_calc_skill_base_damage(struct Damage* wd, struct block_list *
 					RE_ALLATK_ADDRATE(wd, sd->bonus.atk_rate);
 				}
 #ifndef RENEWAL
-				if(sd->bonus.crit_atk_rate && !skill_id && is_attack_critical(wd, src, target, skill_id, skill_lv, false)) { // add +crit damage bonuses here in pre-renewal mode [helvetica]
+				if(sd->bonus.crit_atk_rate && is_attack_critical(wd, src, target, skill_id, skill_lv, false)) { // add +crit damage bonuses here in pre-renewal mode [helvetica]
```
Lemongrass3110: This is missing now. You need to add a code like:
```
#ifndef RENEWAL
	// In Pre-Renewal skills cannot crit
	if( !skill_id ){
		return false;
	}
#endif
```

Lemongrass3110: Re-reading the code it might be covered by the NK check though, because the flag is not set in pre-re.

mrjnumber1: i think it would make more sense for the flag to be set in pre-re skill_db.txt (after all, it IS critting in pre-re), and use #ifdef RENEWAL to make the crit atk rate bonus apply.

sader1992: checking the skills in ratemyserver / irowiki classic , those skills should crit
however i am not sure how accurate it is
but looking on secret's edit
https://github.com/rathena/rathena/commit/08d160aaac8cdd286bbafb32532abe42b8a42376
it look like the skills should have crit

sader1992: well , the check is covered by NK as @Lemongrass3110 mentioned , and for now seems no one reported that something wrong with the pre-re skills as they are , and this edit should not effect the pre-re anyway , so if someone reported about it , we can just change the pre-re/skill_db , but for now this would just fix the renewal issue

src/map/battle.cpp
```diff
@@ -2375,21 +2375,34 @@ static bool is_attack_left_handed(struct block_list *src, int skill_id)
...
-	struct status_change *sc = status_get_sc(src);
-	struct status_change *tsc = status_get_sc(target);
-	struct map_session_data *sd = BL_CAST(BL_PC, src);
-	struct map_session_data *tsd = BL_CAST(BL_PC, target);
-
 	if (!first_call)
 		return (wd->type == DMG_CRITICAL || wd->type == DMG_MULTI_HIT_CRITICAL);
 
+#ifdef RENEWAL
 	if (skill_id == NPC_CRITICALSLASH || skill_id == LG_PINPOINTATTACK) //Always critical skills
```
aleos89: Should we just de-hardcode these and put them into the renewal skill_db?

sader1992: i do think moving it under the `` if( skill_id && !skill_get_nk(skill_id,NK_CRITICAL) ) `` to let the skill_db decide if it can be  crit or not is better

aleos89: Right. Simply removing this check and then just adding the flags to the db should be fine. The check right below this one will catch it.

sader1992: i already added the skill_db part , however i don't think we can remove the check as those skills have 100% chance to crit , if we remove the check they wont be 100% , and there is no way to put  x% in the skill_db <details> <summary>maybe x% in the db is a good idea</summary>(it's a good idea to do in my opinion but maybe in another pr ? instead of Crecital: true we can make it Cricital: chanse , but i don't thing there is a lot of skills that would benefit from this) </details>
so i think the best is to just move the check down

aleos89: Ah, you're right. I forgot about the rate. Just leave the hardcoded parts. We can come up with something later on, it won't be related to this PR.

bRAthena-PT:
batle.cpp

//			if (skill_id == SN_SHARPSHOOTING || skill_id == MA_SHARPSHOOTING)
//				bflag &= ~(BDMG_CRIT); // Sharpshooting just ignores DEF/FLEE but damage is like a normal attack

for  rA 2024+

PR #4475, opened 2019-12-06
Fixed place names and other typos

Daegaladh:
* **Addressed Issue(s)**: 

* **Server Mode**: both

* **Description of Pull Request**: Fixes common misspelled iRO names like Schwarzwald or Ayothaya.


Daegaladh:
Should I fix other common mistranslations like **Juno**/Yuno, **Kunlun**/Gonryun, **Rune-Midgarts**/Rune-Midgard?

Daegaladh:
This one was a pain in the ass: some instances were pretty straightforward since they specify kingdom or continent, other can be guessed from the context, and for the rest I did my best guessing which one they were referring to.
I think I covered all of them, but if you find any more, let me know.

Daegaladh:
Well I think that's all.
If you can think of any more spelling errors in place names let me know and I'll add fix them too.

PR #4491, opened 2019-12-21
Corrected getfreecell command for flag 2

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

Content
-------------------------------------------------
- Fixes a part where the command required an unit ID for flag 2.
```
*getfreecell "<map name>",<rX>,<rY>{,<x>,<y>,<rangeX>,<rangeY>,<flag>};
...
 - 2 = The target should be able to walk to the target tile.
...
```
- The command itself doesn't return a value anymore, this part wasn't fully implemented.
- The command doesn't define `x` when `y` fail anymore.


src/map/script.cpp
```diff
@@ -19057,37 +19053,39 @@ BUILDIN_FUNC(getfreecell)
...
+	if (sd)
+		map_search_freecell(&sd->bl, m, &x, &y, rx, ry, flag);
+	else
+		map_search_freecell(NULL, m, &x, &y, rx, ry, flag);
 
 	// Set MapX
-	num = st->stack->stack_data[st->start + 3].u.num;
-	name = get_str(num&0x00ffffff);
-	prefix = *name;
+	num_x = st->stack->stack_data[st->start + 3].u.num;
```
Lemongrass3110: Better would be to read the reference here already with `script_getdata(st,3);`
And then parse get the rest of the info with:
```
num = reference_getuid(data);
name = reference_getname(data);
```
And when setting the value at the end simply use data->ref.

This looks like someone who understands the core of the script engine wrote this, but noone else would...

Issue #4492, opened 2019-12-24
Gatling Fever


aleos89:
* **rAthena Hash**: 464fd40

* **Client Date**: N/A

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Gatling Fever's damage is incorrect.
  * Expected Result: Target size does not affect damage.
  * How to Reproduce: Use Gatling Fever versus several different sized monsters.
  * Official Information: kRO description
```
[Level 1] : 데미지 +30, 공격속도 +1%, 회피율 -5, 지속시간 30초
[Level 2] : 데미지 +40, 공격속도 +2%, 회피율 -10, 지속시간 45초
[Level 3] : 데미지 +50, 공격속도 +3%, 회피율 -15, 지속시간 60초
[Level 4] : 데미지 +60, 공격속도 +4%, 회피율 -20, 지속시간 75초
[Level 5] : 데미지 +70, 공격속도 +5%, 회피율 -25, 지속시간 90초
[Level 6] : 데미지 +80, 공격속도 +6%, 회피율 -30, 지속시간 105초
[Level 7] : 데미지 +90, 공격속도 +7%, 회피율 -35, 지속시간 120초
[Level 8] : 데미지 +100, 공격속도 +8%, 회피율 -40, 지속시간 135초
[Level 9] : 데미지 +110, 공격속도 +9%, 회피율 -45, 지속시간 150초
[Level 10] : 데미지 +120, 공격속도 +10%, 회피율 -50, 지속시간 165초
```

* **Modifications that may affect results**: None

PR #4493, opened 2019-12-24
Corrected Gatling Fever damage

aleos89:
* **Addressed Issue(s)**: #4492

* **Server Mode**: Renewal

  * Damage no longer depends on the target's size.
Thanks to @kyurejl!


kyurejl:
is not adding +120 atk in status (Alt + A), just +20% atk speed and -50 flee.

aleos89:
Try testing the damage itself against a target. The damage doesn’t show in the status window anymore.

Issue #4327, opened 2019-08-28
Pet egg as mail attachment


RadianFord:
* **rAthena Hash**: a03d80e9b61ceb84851faa0853ac2893e4dcaae3

* **Client Date**: No applicable

* **Server Mode**: Pre Renewal

* **Description of Issue**: Making egg as achievement reward shows error in map-server.
  * Result:  `[Error]: wrong egg item inventory 25`
```
  - ID: 250007
    Group: "AG_BATTLE"
    Name: "Atroce The Warewolf"
    Target:
      - Id: 0
        MobID: 1785
        Count: 10
    Reward:
      ItemID: 40001 // Egg ID
      TitleID: 1040
    Score: 10
```
  * Expected Result: 
  * How to Reproduce: Add egg item as reward.
  * Official Information:
  

* **Modifications that may affect results**:

aleos89:
This is because when pets are created, it sends a request through the inter-server to create the SQL data for it. Since we the achievement system uses the RODEX to send the rewards, this would require a bit of modification to send any pertinent pet info.

RadianFord:
That make sense, thanks aleos!

Lemongrass3110:
To add some more information on this issue:
The message @RadianFord was talking about occurs when you use a pet incubator. Sending the egg itself works.

PR #4506, opened 2020-01-03
Fixed pet eggs as mail attachments

Lemongrass3110:
* **Addressed Issue(s)**: #4327

* **Server Mode**: Both

Fixes the issue by adding a check when getting the attachments of a mail. This additional check ensures that pet eggs that do not have a pet id stored in their card slot get their pet data/id generated by the inter server.
Thanks to @RadianFord

### DIFF
--- a/src/map/mail.cpp
+++ b/src/map/mail.cpp
@@ -16,6 +16,7 @@
 #include "itemdb.hpp"
 #include "log.hpp"
 #include "pc.hpp"
+#include "pet.hpp"
 
 void mail_clear(struct map_session_data *sd)
 {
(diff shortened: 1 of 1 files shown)


src/map/mail.cpp
```diff
@@ -289,7 +290,24 @@ void mail_getattachment(struct map_session_data* sd, struct mail_message* msg, i
...
+				// Check if it is a pet egg
+				std::shared_ptr<s_pet_db> pet = pet_db_search( item[i].nameid, PET_EGG );
+
+				// If it is a pet egg and the card data does not contain a pet id (see if clause above)
+				if( pet != nullptr ){
+					// Create a new pet
+					pet_create_egg( sd, item[i].nameid );
+				}else{
+					// Add the item normally
+					pc_additem( sd, &item[i], item[i].amount, LOG_TYPE_MAIL );
```
Daegaladh: Could you merge the 2 elses into one?

Lemongrass3110: This was a performance decission so that the pet lookup is only done in this special case and not everytime.

Lemongrass3110: I made use of the for loop to merge them as you wanted.

Daegaladh: Yeah that's what I thought. I was wondering if both things could be achieved, but nothing came to my mind, just asking in case you could find a better solution XD

PR #4514, opened 2020-01-04
Fixed official Cat Paw addition exploit

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Fixes an exploit where you can save in Manuk and Splendide fields and open Warp Portal there for other players.

### DIFF
--- a/npc/quests/quests_13_1.txt
+++ b/npc/quests/quests_13_1.txt
@@ -16,7 +16,7 @@
 //= Attitude to the New
 //= Finding a Fairy & Finding a Giant Tree
 //= The Tripartite Union's Feud
-//= Cat Hand Agent
+//= Cat Paw Agent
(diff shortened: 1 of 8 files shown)


Daegaladh:
> 
> 
> Please split the exploit fixes and the name fixes!

The new aegis script already had the Cat Paw name change, I've just applied it to all scripts to maintain consistency.

Lemongrass3110:
> 
> 
> > Please split the exploit fixes and the name fixes!
> 
> The new aegis script already had the Cat Paw name change, I've just applied it to all scripts to maintain consistency.

Thats okay for me, but it is hard to spot the exploit in between all these changes. So it would be good to have it in a separate pull request(master commit) for blame reasons.

Issue #4636, opened 2020-02-13
Custom WoE controller rewards without RODEX


Lemongrass3110:
* **rAthena Hash**: b681339

* **Client Date**: -

* **Server Mode**: Both

* **Description of Issue**: 
  * Result: Custom WoE controller does not give out the rewards if RODEX is not used.
  * Expected Result: Custom WoE controller should give out the rewards correctly.
  * How to Reproduce: Enable /npc/custom/woe_controller.txt and play WoE, you will not receive rewards and there will be an error message, that no player is attached.
  * Official Information: None, this is a custom npc.

* **Modifications that may affect results**: 
None.

Issue #4619, opened 2020-02-09
WOE CONTROLLER dont give the Reward


mazvi:
* **rAthena Hash**: b0ac171

* **Client Date**: 20200620RE

* **Server Mode**: RENEWAL

* **Description of Issue**: WOE CONTROLLER dont give the Reward
  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**:

Lemongrass3110:
This is a result of 0b3430c.
I am on it. :-)

PR #4637, opened 2020-02-13
Fixed a few bugs in custom WoE controller

Lemongrass3110:
* **Addressed Issue(s)**: #4619 and #4636

* **Server Mode**: Both

Removed hardcoded castle count
Adjusted the header format to the new header
Added the missing changelog entries as far as I was able to reconstruct them from GIT/SVN
Adjusted the script to not directly access the guild member table from SQL
Fixed a broken checkweight call that was used without an attached player
Fixed handout of rewards to an online player for each character he has in the guild
-> Should this be done/checked for RODEX or in general as well?
-> Should it be an option?


npc/custom/woe_controller.txt
```diff
@@ -168,14 +174,34 @@ OnReward:
...
+
+						cleararray( .@cid, 0, .@size );
+						cleararray( .@aid, 0, .@size );
+
+						.@size = 1;
+						.@cid[0] = .@master_cid;
+						.@aid[0] = .@master_aid;
+						break;
+					}
+				}
```
theOptimistic:
```
				.@index = inarray(.@cid, .@master_cid);
				.@master_aid = .@aid[.@index];

				cleararray( .@cid, 0, .@size );
				cleararray( .@aid, 0, .@size );

				.@size = 1;

				.@cid[0] = .@master_cid;
				.@aid[0] = .@master_aid;
```
While you're at it, I think it's better to use built-in commands.

mazvi:
Before guild sql upgrade the old woe custom controller

If dual login only first person with condition first login receive reward

But now dual login not receive reward

npc/custom/woe_controller.txt
```diff
@@ -182,29 +182,30 @@ OnReward:
...
-			.@size = getguildmember( .@gid, 2, .@aid );
+			.@size_guild = getguildmember( .@gid, 2, .@aid );
 
 			if( .Options&4 ){
 				.@master_cid = getguildmasterid( .@gid );
-				
-				for( .@j = 0; .@j < .@size; .@j++ ){
-					if( .@cid[.@j] == .@master_cid ){
-						.@master_aid = .@aid[.@j];
+				.@index = inarray(.@cid, .@master_cid);
```
sader1992:
```suggestion
				if((.@index = inarray(.@cid, .@master_cid)) != -1){
```

npc/custom/woe_controller.txt
```diff
@@ -182,29 +182,30 @@ OnReward:
...
 
-						.@size = 1;
-						.@cid[0] = .@master_cid;
-						.@aid[0] = .@master_aid;
-						break;
-					}
-				}
+				.@size_guild = 1;
+				.@cid[0] = .@master_cid;
+				.@aid[0] = .@master_aid;
```
sader1992:
```suggestion
				.@aid[0] = .@master_aid;
			}
```

npc/custom/woe_controller.txt
sader1992: also it need some <tab>s after that if you applied the changes automatically

Atemo: Like you say it shouldn't happen, I left the script without the index check on purpose to stop the script and debug just in case

Issue #4644, opened 2020-02-14
Battleground #3995 - crash map server


Balferian:
* **rAthena Hash**: 1e091cd
* **Client Date**:  2018-06-20
* **Server Mode**: Re

* **Description of Issue**: when all char register at any BG map server crashed
  * Result: when all chars register at any BG after they confirm start bg - map server crashed
  * Expected Result: start BG
  * How to Reproduce: register 10-12 chars to BG

> *** Error in `./server-map': free(): invalid pointer: 0x00000000073feb58 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81609)[0x7f3d5447f609]
./server-map[0x40f8e9]
./server-map[0x456254]
./server-map[0x656b62]
./server-map[0x407c65]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x7f3d54420495]
./server-map[0x408faf]
======= Memory map: ========
00400000-00775000 r-xp 00000000 08:01 802126                             /home/Server/renewal/server-map
00975000-00976000 r--p 00375000 08:01 802126                             /home/Server/renewal/server-map
00976000-0097c000 rw-p 00376000 08:01 802126                             /home/Server/renewal/server-map
0097c000-024fe000 rw-p 00000000 00:00 0
0403e000-09c28000 rw-p 00000000 00:00 0                                  [heap]
7f3d34000000-7f3d34021000 rw-p 00000000 00:00 0
7f3d34021000-7f3d38000000 ---p 00000000 00:00 0
7f3d3be93000-7f3d53b92000 rw-p 00000000 00:00 0
7f3d53dd9000-7f3d53de5000 r-xp 00000000 08:01 24027                      /usr/lib64/libnss_files-2.17.so
7f3d53de5000-7f3d53fe4000 ---p 0000c000 08:01 24027                      /usr/lib64/libnss_files-2.17.so
7f3d53fe4000-7f3d53fe5000 r--p 0000b000 08:01 24027                      /usr/lib64/libnss_files-2.17.so
7f3d53fe5000-7f3d53fe6000 rw-p 0000c000 08:01 24027                      /usr/lib64/libnss_files-2.17.so
7f3d53fe6000-7f3d543fe000 rw-p 00000000 00:00 0
7f3d543fe000-7f3d545c0000 r-xp 00000000 08:01 5010                       /usr/lib64/libc-2.17.so
[...]

PR #4648, opened 2020-02-16
Battleground Queue fixes

aleos89:
* **Addressed Issue(s)**: #4644

  * Initialize all battleground queues at startup to reduce overhead of creation/deletion during live server.
  * Removes lots of duplicate code between team checks.
  * Adds a check for script commands bg_leave and bg_desert to only warp players out using the Battleground Queue System.
  * Adds documentation for script commands bg_leave and bg_desert to state players will be warped out after when using the Battleground Queue System.
  * Optimizes join logic to add new players to the end of the queue rather than the front.
  * Cleans up the client messages so certain responses are cleared when they need to be.
Thanks to @Balferian!


Balferian:
With last fix all working good, didn`t found any problems.
Thanks **@aleos89**!

Issue #1863, opened 2017-01-10
The 'and' condition statement is not working as expected


Tokeiburu:
Take for example the following script:

```
prontera,150,150,0	script	test	77,{
	.@is_instance = false;
	
	if (.@is_instance && 'instance_var) {
		// ...
	}
	
	end;
}
```

The script engine will try to read the instance variable even though the condition should have stopped after the left condition failed.
```
[Warning]: script:get_val: cannot access instance variable ''instance_var', defaulting to 0
```

This is an issue if you want to checks such as:
```
prontera,150,150,0	script	test	77,{
	initnpctimer;
	end;
OnTimer1000:
	if (playerattached() && strcharinfo(0) == "Tokei") {
		// ... do something
	}
	
	initnpctimer;
	end;
}
```

which makes the script stop and gives the following error:
```
[Error]: script_rid2sd: fatal error ! player not attached!
[Debug]: Function: strcharinfo (1 parameter):
[Debug]: Data: number value=0
[Debug]: Source (NPC): test at prontera (150,150)
[Debug]: Source (NPC): test at prontera (150,150)
```

git f14b5b34ecdcd7ebd7604f0b60a911f80714b23b

vstumpf:
I was sure short-circuiting was a feature of Athena, I guess something changed?

dastgir:
This has been as it is from jA time.
Short circuiting was never a part of Athena.
But it's good if it gets supported by emulators

PR #4655, opened 2020-02-17
Introducing short-circuiting logic

Lemongrass3110:
* **Addressed Issue(s)**: #1863

* **Server Mode**: Both

Adds support for short-circuiting logic by delaying the access to the right hand argument of the logical and operator.

Thanks to @Tokeiburu for bringing this up.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -3922,24 +3922,50 @@ void op_2(struct script_state *st, int op)
 	}
 
 	get_val(st, left);
-	get_val(st, right);
+
+	bool right_is_int;
+	bool right_is_string;
+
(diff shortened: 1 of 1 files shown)


neico:
Shouldn't `C_LOR` and `C_OP3` (ternary, technically it should be an alias to `Cond && A || B`, simplified of course) short circuit as well?
A mention in the script docs also is in order, so there is no hidden features. (or absence thereof, as it currently is)

In either case, this would generally be useful for performance, and for cases where you need to check `playerattached()` before using a char/acc variable.
It's almost unbelievable that this wasn't implemented all this time.

Issue #4660, opened 2020-02-20
Skill GS_DESPERADO loses AoE after using @reloadskilldb command


teededung:
* **rAthena Hash**: 1c653616a066d2a2ea626bc894807bc68fb2e16b 
* **Client Date**: 2018
* **Server Mode**: Renewal & Pre-Renewal
* **Description of Issue**: 
  * Result: Skill GS_DESPERADO loses AoE after using `@reloadskilldb` command
  * Expected Result: Works normally
  * How to Reproduce: Use `@reloadskilldb` then cast the skill

Maybe the skill is missing
```
Flag:
        PathCheck: true
```

aleos89:
`516,0x86,    ,  3, 0, 100,enemy, 0x000	//GS_DESPERADO`
The skill never had any unit flags originally.

Balferian:
I confirm it too, after @reloadskilldb Desperado don't damage.
1. Desperado don't damage mobs.
2. Blaze shield don't cast on all area. 
(I had this problem once and I don't know how reproduce it again)
![image](https://user-images.githubusercontent.com/16213511/75033010-f1775000-54ba-11ea-8d06-4aa730185376.png)x

aleos89:
In skill.cpp, could you try finding:
```cpp
skill->unit_flag = UF_NONE;
```
and replace it with:
```cpp
skill->unit_flag.reset();
```

Balferian:
> In skill.cpp, could you try finding:
> 
> ```c++
> skill->unit_flag = UF_NONE;
> ```
> 
> and replace it with:
> 
> ```c++
> skill->unit_flag.reset();
> ```
Nothing changed.

PR #4669, opened 2020-02-27
Fixes skill database reloading

aleos89:
* **Addressed Issue(s)**: #4660

  * Resolves an issue with skill database reloading not clearing the previous data first.
  * While clearing the YAML cache it will also resize the cache to the previous capacity.
Thanks to @teededung!


teededung:
Thanks! It is okay now

PR #4673, opened 2020-02-28
Small update of npc_enable_sub

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: 

With this update a player standing on a NPCTYPE_WARP disabled will be warped when the npc is enabled (before the player need to move to be warped).

This change only affects the warps point.
```
<from mapname>,<fromX>,<fromY>,<facing>	warp	<warp name>	<spanx>,<spany>,<to mapname>,<toX>,<toY>
<from mapname>,<fromX>,<fromY>,<facing>	warp2	<warp name>	<spanx>,<spany>,<to mapname>,<toX>,<toY>
```

### DIFF
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -223,14 +223,30 @@ int npc_enable_sub(struct block_list *bl, va_list ap)
 		if (nd->sc.option&OPTION_INVISIBLE)
 			return 1;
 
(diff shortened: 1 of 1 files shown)


SOSKaito:
This introduced a bug where players in instances would be warped as soon as the warp was activated, no matter if they stood on it or not.

PR #4674, opened 2020-02-28
Updated Morse's Cave instance to match official script

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Renewal

This PR updates Morse's Cave instance to match official script.

To resume the change IG are some additionnals effects and defined warp points.


teededung:
@Atemo `if (isloggedin(.@account_id[.@i],.@char_id[.@i]) == true) {` need change to `false`

Issue #1404, opened 2016-06-29
[BUG] Skill LANDPROTECTOR failing border


Badarosk0:
Currently LANDPROTECTOR ability is a problem. The test was done with 2 Chars in PvP. 

A char uses the ability LANDPROTECTOR the other uses a skill in an area, for example, Psychic Wave with center outside the skill in LANDPROTECTOR, noting that part of the area affecting the target within the LANDPROTECTOR. 
The target takes damage. 

If the center of Psychic Wave is born in the LANDPROTECTOR does not affect the target that is in LANDPROTECTOR.

If you are not clear, I can post picture showing.

Link old bug: https://rathena.org/board/tracker/issue-5237-land-protector-border/
Link old bug 2: https://github.com/rathena/rathena/commit/e7e8b54

secretdataz:
Uh, well, I can't understand what you are trying to say.
Care to explain more?

secretdataz:
Tested on official.
If PW's center is placed outside of LP, it will still does damage to anyone in its AoE even if that player is standing in LP.
If PW's center is placed inside of LP, it will simply does nothing but showing its visual effect.

Badarosk0:
Use the protective land skill. Enter the skill area . With another character use Psychic Wave or any other skill area. The center of this another skill must be out of the protective land . If someone is on the edges of protection will be affected.


Images demonstrating the problem: http://postimg.org/gallery/wjm47igy/

secretdataz:
That's intended behavior of LP. It does not protect players standing on it from splashing effects caused by units outside it. LP just prevents placing of any unit on top of it.

anacondaqq:
Hi Dear @secretdataz 
You are totally right, and right now Land Protector works fine like it should at officials, but closing the issue is not a solution, let me explain why.

The issue with land protector were confirmed and fixed around 9-12 monthes ago, as i remember @Playtester have fixed the issue. Have fixed the issue after **ATTENTION! more then 8 years**.

A lot of people who don't playing at official servers do not understand this change, they start to blame GM's, rAthena emulator developers, and they start to insult all of people who trying to make official server for them. The reason why they do that is very simple: *They are accustomed to the old, and do not want to change anything, and they will not change anything, because they like how it was. They start to report about the issue like about critical bug, and then a lot of peole just quit a server.*

The problem is very popular, and be sure a lot of people will report about the issue again, and again, and again, here is founded by me reports about the issue from hell a lot of admins.

[...]

PR #4675, opened 2020-02-29
[Suggestion] Custom Config for Land Protector Skill Behavior

### LINKED ISSUE #4412: Land Protector area effect
* **rAthena Hash**: https://github.com/rathena/rathena/commit/9aa5f7d8345d9d89ec9e8b44060e92703f3d43b8

* **Client Date**: 20180621

* **Server Mode**: RE

* **Description of Issue**: 

If the player uses land protector and another player uses an area skill outside the land protector area the effect still occurs. Protection will not happen.

* **Result**: 

Players using land protection will not be protected in this case.

* **Expected Result**: 

Land Protector should protect against any area skills, even if the skill is used outside and the area range affects anyone in the land protector.

* **How to Reproduce**: 
Below the images showing the problem how to reproduce:

https://ibb.co/kSkC9dC
https://ibb.co/gWVbYf4

* **Official Information**: 
  

Tydus1:
* **Addressed Issue(s)**: 
None

* **Server Mode**: Pre-Renewal & Renewal

* This pull request adds a battle conf option to allow changing the land protector skill behavior since it has changed after some point on renewal. 
* The defaut config to `land_protector_behavior is 0`. Changing to 1 will change it to the old behavior.
* Fixes #4412 & #1404
* Thanks to @Cydh for the initial release.


conf/battle/skill.conf
```diff
@@ -362,6 +362,12 @@ skill_eightpath_algorithm: yes
 // damage skills (previous behavior).
 can_damage_skill: 1
 
+// Land Protector behavior
+// 0 : Officially, standing players at the border (outer cell) of the Land Protector is still can be affected/hit
+//     by AoE skills (if the skills has splash effect, such Storm Gust).
+// 1 : Old Athena style, totally ignore AoE skill from outside Land Protector area.
+land_protector_behavior: 0
```
aleos89:
```suggestion
// Land Protector behavior (Note 1)
// On official servers, players standing on the border (outer cell) of the Land Protector can still be affected/hit
// by AoE skills (if the skill has a splash effect, such as Storm Gust). The Athena behavior ignores AoE affects/hits
// while players are standing on the border.
// Official: 0
// Legacy Athena: 1
land_protector_behavior: 0
```

footloose232:
Hi, i tried to apply this but i got this error.

oer

Daemos2:
> Hi, i tried to apply this but i got this error.
> using rathena [972a14e](https://github.com/rathena/rathena/commit/972a14e4ed3c2ae15f5acb60652cf0c3db1d0deb)
> 
> oer

Use [this](https://github.com/cydh/rathena/commit/6de800aea19c99aba5b88a14bfb625a1cf870c85) instead. Is the same but for older versions.

PR #4692, opened 2020-03-07
[Suggestion] setquest, erasequest, completequest, checkquest atcommand

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

Content
--------------------------------
Suggestion to add some quests debug commands for admin.
- `@setquest` : _self explaining_
- `@erasequest` : _self explaining_
- `@completequest` : give and complete the quest
- `@checkquest` : display the quest status HAVEQUEST, HUNTING, PLAYTIME

### DIFF
--- a/conf/msg_conf/map_msg.conf
+++ b/conf/msg_conf/map_msg.conf
@@ -1706,5 +1706,17 @@
 1503: You've entered a PK Zone.
 1504: You've entered a PK Zone (safe until level %d).
 
+// @setquest, @erasequest, @completequest
+1505: Usage: %s <quest ID>
+1506: Quest %d not found in DB.
+1507: Character already has quest %d.
+1508: Character doesn't have quest %d.
+
+// @checkquest
+1509: Checkquest value for quest %d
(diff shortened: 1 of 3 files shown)


sader1992:
it would be nice if 
`@setquest <quest id><stat>` so 
and if there is no stat passed it will just give the quest as default

Atemo:
I don't get the idea, can you give an example?

Atemo:
These commands are design to debug, for admin. Your suggestion is about changing the quest data for a single player, for event script for example. A script command would be better.
Maybe in another PR ?

sader1992:
it's to switch between the quest status in PLAYTIME|HUNTING
looking into the src , the only way to do it is to change `sd->quest_log[id].time` to make the time pass by the command , it's not really a stat of the quest , it could be another command instead of using @setquest

the example that come to mind is when we test an instance , and we need to check the npc after the cooldown quest for been in the instance , or the cooldown quest for re-entering the instance

sader1992:
sorry , took sometime to check the src , yes maybe it's better for another PR.

PR #4695, opened 2020-03-08
Fixes #4693: Removed Bowling Bash Chain Damage

### LINKED ISSUE #4693: Class Re-balancing Update for Bowling Bash Knight
* **rAthena Hash**:  Latest Release.

* **Client Date**: Doesn't Matter.

* **Server Mode**: Renewal

* **Description of Issue**: Bowling Bash damage doesn't match what kRO did with their update. 
  * Result:  The damage dealt with one-hand sword or spears should be 2 hits only regarding of how many enemies within 5x5 cells of the target monster but instead the damage result is like 4 hits because of the stacked damage of the skill, likewise if you equip two-handed swords the damage can soar up to 8 hits (because of stacked damage) if there are 4+ monsters in the 5x5 cell of the target monster which i believe should be 4 hits only at max.
  * Expected Result: One-hand sword, Dagger, Spear or Two-handed Spear = maximum of 2 hits within 5x5 cells of the target monster. Two-handed sword = maximum of 4 hits if there are 4+ monsters in the 5x5 cell of the target monster. Check the kRO GIF so you can understand much better what I'm trying to say here. See the white numbers in the GIF it has no stacked damage unlike what you can see if you test it on your own server.
T0SU3XXWLNB8DCLQ9PLK

Here is an example of what i mean.
for github

  * How to Reproduce:  Just run rAthena and then test it.
  * Official Information:
  

* **Modifications that may affect results**: None modified.

yashimwong:
- Removed Bowling Bash chained damage for class rework

* **Addressed Issue(s)**:  #4693 Class Re-balancing Update for Bowling Bash Knight

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -5333,8 +5333,10 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 					break;
 				}
 			}
+#ifndef  RENEWAL
 			// Original hit or chain hit depending on flag
 			skill_attack(BF_WEAPON,src,src,bl,skill_id,skill_lv,tick,(flag&0xFFF)>0?SD_ANIMATION:0);
+#endif
 		}
 		break;
 


violent01:
this doesn't solve anything about #4693 All this Pr did was take away Bowling Bash damage.

yashimwong:
This remove the stacked damage you see as a result from chaining.  Have you tested the PR?

Here are my personal tests after PR:
This is after PR with 4 or more mobs:
image

Single mobs:
image

violent01:
looks like it was my fault. sorry yashim.

yashimwong:
No problem @violent01 , thanks for the input :)

violent01:
@yashimwong  am also testing out brandish spear and i think there are some missing updates here as well. Maybe I'll open an issue again once i confirmed it.

reigneil:
@yashimwong renewal? kRO info of this https://www.youtube.com/watch?v=JOE9qMZvHu4

PR #4703, opened 2020-03-10
Fixes #4702: Rogue Backstab Double Hit

### LINKED ISSUE #4702: Class Re-Balancing Update Back stab skill [Rogue]
* **rAthena Hash**: Latest

* **Client Date**: 20180621

* **Server Mode**: Renewal

* **Description of Issue**: Back stab behavior do not match the current update.
  * Result: Back stab behavior is currently old.
  * Expected Result:  Back stab skill should do double hit if equipped with a dagger class weapon.  Back stab should not be usable if the player is in the hiding state.
  * How to Reproduce:  
  * Official Information:  https://www.divine-pride.net/forum/index.php?/topic/3453-kro-mass-skills-balance-1st-2nd-and-transcendent-classes-skills/

9.2 Back Stab
- Changes skill mechanism, moves to target's back immediately and deals damage. Can't be used while under Hiding state.
- If user equips dagger, increases number of hit to twice.

github

  

* **Modifications that may affect results**: None

yashimwong:
- Back Stab now deals 2 hits visually when using dagger as per issue report.

* **Addressed Issue(s)**:  https://github.com/rathena/rathena/issues/4702

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -6304,10 +6304,6 @@ Body:
     MaxLevel: 10
     Type: Weapon
     TargetType: Attack
-    DamageFlags:
-      IgnoreFlee: false
-    Flags:
-      AllowWhenHidden: false
     Range: -1
     Hit: Single
     HitCount: 1
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -6307,7 +6307,7 @@ Body:
     DamageFlags:
       IgnoreFlee: true
     Flags:
-      AllowWhenHidden: true
+      AllowWhenHidden: false
     Range: -1
     Hit: Single
     HitCount: 1
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -6305,7 +6305,7 @@ Body:
     Type: Weapon
     TargetType: Attack
     DamageFlags:
-      IgnoreFlee: true
+      IgnoreFlee: false
     Flags:
       AllowWhenHidden: false
     Range: -1
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3654,10 +3654,6 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 				skillratio += (200 + 40 * skill_lv) / 2;
 			else
 				skillratio += 200 + 40 * skill_lv;
-#ifdef RENEWAL
-			if (sd && sd->status.weapon == W_DAGGER)
-				skillratio *= 2;
-#endif
 			break;
 		case RG_RAID:
 #ifdef RENEWAL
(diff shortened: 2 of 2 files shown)


src/map/battle.cpp
```diff
@@ -5358,6 +5358,10 @@ static struct Damage initialize_weapon_data(struct block_list *src, struct block
 		wd.flag |= battle_range_type(src, target, skill_id, skill_lv);
 		switch(skill_id)
 		{
+			case RG_BACKSTAP:
+				if (skill_id == RG_BACKSTAP && sd && sd->status.weapon == W_DAGGER)
+				wd.div_ = -2;
+			break;
```
aleos89: This case needs to be wrapped in `#ifdef RENEWAL` since this is a renewal only feature of the skill.

src/map/battle.cpp
```diff
@@ -5358,6 +5358,12 @@ static struct Damage initialize_weapon_data(struct block_list *src, struct block
 		wd.flag |= battle_range_type(src, target, skill_id, skill_lv);
 		switch(skill_id)
 		{
+#ifdef RENEWAL
+			case RG_BACKSTAP:
+				if (skill_id == RG_BACKSTAP && sd && sd->status.weapon == W_DAGGER)
```
aleos89:
```suggestion
				if (sd && sd->status.weapon == W_DAGGER)
```

yashimwong: I'm still not sure whether it is 2 separate hits or just 200% damage showing 2 hits. In this pull request I assumed it is 200% in one hit and client shows 2 hit.

aleos89: From the GIF, and the kRO description `단검을 장착하여 사용할 경우 데미지가 2회 발생한다.` it would seem it sends 2 separate hits now, not double damage. So the solution is to remove the double damage.

db/re/skill_db.yml
```diff
@@ -6305,9 +6305,9 @@ Body:
     Type: Weapon
     TargetType: Attack
     DamageFlags:
-      IgnoreFlee: true
+      IgnoreFlee: false
```
aleos89: I would be inclined to remove this flag all together instead of flipping it to false. False is default so the only reason to explicitly define it as false is if you're using import/skill_db. Don't forget to remove the head label as well (`DamageFlags`).

db/re/skill_db.yml
```diff
@@ -6305,9 +6305,9 @@ Body:
...
     Flags:
-      AllowWhenHidden: true
+      AllowWhenHidden: false
```
aleos89: Same as above. Don't forget to remove the head label as well (`Flags`).

violent01:
It’s working now as intended. /no1 @yashimwong

db/re/skill_db.yml
```diff
@@ -6305,7 +6305,7 @@ Body:
     Type: Weapon
     TargetType: Attack
     DamageFlags:
-      IgnoreFlee: true
+      IgnoreFlee: false
```
aleos89: Do you have source for this change? I was looking back at the changelog but I don't see it stating anything about a change with flee.

violent01: @aleos89  I think I read that somewhere I just can’t remember where. I’ll try to dig that info but it’s correct that they dropped the perfect hit for Backstab cause I’m sure I read that.

yashimwong: @aleos89  There's a increase hit bonus according to level of skill based on the translate kRO patch notes. Not sure why it would give hit bonus if it ignores flee

PR #4709, opened 2020-03-11
Job Improvement Project - Rune Knight (Part 2)

aleos89:
* **Addressed Issue(s)**: #4708

* **Server Mode**: Renewal

  * kRO Changelog: http://ro.gnjoy.com/news/update/View.asp?seq=243&curpage=1


ecdarreola:
Hundred Spear doesn't seem to do splash damage upon testing. Wind Cutter/Ignition Break working fine.

samers1:
Hello, where's part 1? could you put the pull link?

aleos89:
It was applied a long time ago in #2181.

src/map/battle.cpp
```diff
@@ -2876,6 +2880,21 @@ static int battle_get_weapon_element(struct Damage* wd, struct block_list *src,
...
+					element = ELE_HOLY;
+			}
+			break;
+		case RK_DRAGONBREATH_WATER:
+			if (sc) {
+				if (sc->data[SC_LUXANIMA]) // Lux Anima has priority over Fighting Spirit
+					element = ELE_NEUTRAL;
+				else if (sc->data[SC_FIGHTINGSPIRIT])
+					element = ELE_GHOST;
+			}
```
Litro: Missing break ? or fall through?

src/map/battle.cpp
```diff
@@ -3440,6 +3459,9 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
 	}
 
 	switch (skill_id) {
+		case RK_WINDCUTTER:
+			if (sd && sd->weapontype1 == W_2HSWORD)
+				wd->div_ = 2;
```
Litro: here too missing break?

Angelic234:
1. use item Runstone_Lux(ID 22540)
2. attack a monster
3. map-server crash

Ahmed-Chawqi:
Lux Anima Runestone gives a chance to autospell storm blast lvl 1
actually shouldn't works with skills.
just when use auto attack with weapon.

When you use lux anima stone then use a skill like Dragon breath, map-serv crash

LordWhiplash:
In hundred spears change, HitCount is 5, this doesnt means that it will hit 5 times with the full damage on each hit? isn't it supposed to be -5 and divide full damage in five hits?

ecdarreola:
Is it normal that RK Storm Blast specifically the one that procs from Lux Anima Runestone multiplies itself by the amount of units it hits?

ecdarreola:
Storm blast processed by SC_LUXANIMA is abnormal. It multiplies by the number of hits or number of mobs on the area.

Not sure if this is the official behavior however when I test the Pertz Runestone, storm blast doesn't multiply its hit but only once.

Video: https://youtu.be/Q10zwyY2qPk

LordWhiplash:
lux anima effect are missing this buffs:
+30% physical damage against all enemies sizes 
+30% critical damage
+30% melee physical damage
+30% long ranged physical damage

LordWhiplash:
Physical damage against all enemies sizes is working great, thanks!
but the +30% critical damage, +30% melee physical damage, and +30% long ranged physical damage are still missing

LordWhiplash:
Now is working fine, thanks!

Masao87:
Do you have any ETA on this maybe? :)

Issue #4707, opened 2020-03-11
Class Re-Balancing Update Brandish Spear [Knight]


violent01:
* **rAthena Hash**: Latest

* **Client Date**: 20180621

* **Server Mode**: Renewal

* **Description of Issue**: Brandish  Spear skill doesn't match what kRO did to their update.
  * Result: Brandish Spear is still a Melee Physical Attack, Single Hit, Low Damage.
  * Expected Result:  Brandish Spear should now be a Long Range Physical Attack, Multi-Hit (I believe the damage should be divided into 3 hits like how Spiral Pierce hit works) and Strength Stat should affect the damage increase.
  * Official Information: https://www.divine-pride.net/forum/index.php?/topic/3453-kro-mass-skills-balance-1st-2nd-and-transcendent-classes-skills/

1.2 Brandish Spear
- Changes damage type to become long ranged physical damage.
- Changes damage formula, damage is increased depends on user's Str. New damage formula : (400 + (100 x skill level))%Atk

![github2](https://user-images.githubusercontent.com/30449376/76391283-3a1a7e80-6380-11ea-8f34-41603eafcc18.gif)

  

* **Modifications that may affect results**: None

PR #4711, opened 2020-03-12
Fixes Brandish Spear from renewal balance

aleos89:
* **Addressed Issue(s)**: #4707

* **Server Mode**: Renewal

  * Skill is now long range.
  * Skill now does 3 hits.
  * Target's placement in the AoE does not affect damage; it's all the same for each target.
Thanks to @violent01!

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -2112,7 +2112,7 @@ Body:
       NoDamage: true
     Flags:
       IncreaseGloomyDayDamage: true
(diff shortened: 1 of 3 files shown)


mrjnumber1:
https://docs.google.com/spreadsheets/d/1xTiAQmct9hmxeSNf_TK46Fnnf9pdMCP_RzsR1kDfGzU/edit#gid=0
Just taking a guess, but based on this testing on dummies it needs to be sstatus->str*3 ?

Issue #4727, opened 2020-03-25
Custom config for NPC EMOTION skill


victorgawk:
* **rAthena Hash**: 14d2686cf

* **Client Date**: 2015-11-04aRagexe

* **Server Mode**: Pre-Renewal

* **Description of Issue**: Add a custom config for NPC EMOTION skill
  * Result: On official servers, certain mobs cast NPC EMOTION skill which displays an emoticon and change their mode from Aggressive to Passive for a certain time.
  * Expected Result: The old Athena behavior do not change their mode to Passive.
  * How to Reproduce: Try mob a few High Orcs in `gef_fild14` map. You notice that they will completely drop the aggro, and even though you go back to them they will ignore you for a certain time.
  * Official Information: The current result is working like official. This issue is intended to create a custom battle config option which allows to modify NPC EMOTION skill to work like Athena behavior.

  

* **Modifications that may affect results**: 
Create a config which allows to change NPC EMOTION to work like Athena behavior.

SeravySensei:
Agreed, such an option would be extremely useful.
I had to remove that NPC_EMOTION from all my mobs one at a time.

Lemongrass3110:
Fixed in 044aceb

PR #4728, opened 2020-03-25
[Suggestion] Custom Config for NPC EMOTION Behavior

victorgawk:
* Adds config for NPC EMOTION behavior.
* This battle config option adds the ability to modify NPC EMOTION skill to the Athena behavior.

* **Addressed Issue(s)**: #4727

* **Server Mode**: Both

On official servers, certain mobs cast NPC EMOTION skill which displays an emoticon and change their mode from Aggressive to Passive for a certain time. The Athena behavior does not change their mode to Passive.

The default config to `npc_emotion_behavior` is `0`. Changing to `1` will modify it to the Athena behavior.

### DIFF
--- a/conf/battle/skill.conf
+++ b/conf/battle/skill.conf
@@ -370,6 +370,13 @@ can_damage_skill: 1
 // Legacy Athena: 1
 land_protector_behavior: 0
 
+// NPC EMOTION behavior (Note 1)
+// On official servers, certain mobs cast NPC EMOTION skill which displays an emoticon and change their mode from
+// Aggressive to Passive for a certain time. The Athena behavior does not change their mode to Passive.
+// Official: 0
+// Legacy Athena: 1
+npc_emotion_behavior: 1
+
 // Should Tarot Card of Fate have the same chance for each card to occur? (Note 1)
 // Official chances: 15%: LOVERS | 10%: FOOL, MAGICIAN, HIGH PRIESTESS, STRENGTH, SUN | 8%: TEMPERANCE
 // 7%: CHARIOT | 6%: THE HANGED MAN | 5%: DEATH, STAR | 2%: TOWER | 1%: WHEEL OF FORTUNE, DEVIL
--- a/conf/battle/skill.conf
+++ b/conf/battle/skill.conf
@@ -375,7 +375,7 @@ land_protector_behavior: 0
 // Aggressive to Passive for a certain time. The Athena behavior does not change their mode to Passive.
 // Official: 0
 // Legacy Athena: 1
-npc_emotion_behavior: 1
+npc_emotion_behavior: 0
 
 // Should Tarot Card of Fate have the same chance for each card to occur? (Note 1)
 // Official chances: 15%: LOVERS | 10%: FOOL, MAGICIAN, HIGH PRIESTESS, STRENGTH, SUN | 8%: TEMPERANCE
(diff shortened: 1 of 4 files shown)


conf/battle/skill.conf
```diff
@@ -370,6 +370,13 @@ can_damage_skill: 1
 // Legacy Athena: 1
 land_protector_behavior: 0
 
+// NPC EMOTION behavior (Note 1)
+// On official servers, certain mobs cast NPC EMOTION skill which displays an emoticon and change their mode from
+// Aggressive to Passive for a certain time. The Athena behavior does not change their mode to Passive.
+// Official: 0
+// Legacy Athena: 1
+npc_emotion_behavior: 1
```
Daegaladh: Please, default this to 0.

victorgawk: My bad, @Daegaladh, Commit 577dbf9 fix this.

src/map/skill.cpp
```diff
@@ -8838,6 +8838,9 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			if(md->db->skill[md->skill_idx].val[4] && tsce)
 				status_change_end(bl, type, INVALID_TIMER);
 
+			if (battle_config.npc_emotion_behavior && skill_id == NPC_EMOTION && md->db->skill[md->skill_idx].val[1] && md->db->skill[md->skill_idx].val[1] == 0x81)
```
Daegaladh: I don't get why you do this

src/map/skill.cpp
```diff
@@ -8838,6 +8838,9 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+				break;
+
 			//If mode gets set by NPC_EMOTION then the target should be reset [Playtester]
 			if(skill_id == NPC_EMOTION && md->db->skill[md->skill_idx].val[1])
```
Daegaladh: If you want to achieve the old behavior, just change this line to
```
if(!battle_config.npc_emotion_behavior && skill_id == NPC_EMOTION && md->db->skill[md->skill_idx].val[1])
```

and line 8852 to
```
//Reset aggressive state depending on resulting mode
			if (!battle_config.npc_emotion_behavior)
				md->state.aggressive = status_has_mode(&md->status,MD_ANGRY)?1:0;
```

src/map/skill.cpp
victorgawk: I've created this condition because I wanted to check exactly when NPC EMOTION change the mode into 0x81 (which means MD_CANMOVE plus MD_CANATTACK mode). This was the mode I've observed that turns the mob into a passive state on mob_skill_db.

src/map/skill.cpp
victorgawk: Just tested here and got the same result. I will commit the changes because your code look better.

src/map/skill.cpp
victorgawk: I've updated the code following the other changes you requested on commit a441fb2.

mrjnumber1: dont put magic numbers then.
and rename the battle config to "disable aggressiveness toggling" or something

victorgawk: The magic number is already taken off (check commit a441fb2). About the name, I've choosen "npc_emotion_behavior" initially because it's directly related to the NPC EMOTION skill. And it disables passiveness, not the opposite. What about the name "npc_emotion_disable_passiveness"?

PR #4738, opened 2020-03-26
Cleanup for the new_account feature

Lemongrass3110:

* **Server Mode**: Both

Disabled by default - as on official servers.
Gives a better warning if it is turned on and client side password encryption is enabled and the account cannot be created because of this.

### DIFF
--- a/src/login/login.cpp
+++ b/src/login/login.cpp
@@ -316,9 +316,14 @@ int login_mmo_auth(struct login_session_data* sd, bool isServer) {
 	// Account creation with _M/_F
 	if( login_config.new_account_flag ) {
 		if( len > 2 && strnlen(sd->passwd, NAME_LENGTH) > 0 && // valid user and password lengths
-			sd->passwdenc == 0 && // unencoded password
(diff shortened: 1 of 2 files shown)


src/login/login.cpp
```diff
@@ -316,9 +316,14 @@ int login_mmo_auth(struct login_session_data* sd, bool isServer) {
 	// Account creation with _M/_F
 	if( login_config.new_account_flag ) {
 		if( len > 2 && strnlen(sd->passwd, NAME_LENGTH) > 0 && // valid user and password lengths
-			sd->passwdenc == 0 && // unencoded password
 			sd->userid[len-2] == '_' && memchr("FfMm", sd->userid[len-1], 4) ) // _M/_F suffix
 		{
+			// Encoded password
+			if( sd->passwdenc != 0 ){
+				ShowError( "Account '%s' could not be created, because client side password encryption is enabled.\n", sd->userid );
```
aleos89:
```suggestion
				ShowError( "Account '%s' could not be created because client side password encryption is enabled.\n", sd->userid );
```

Issue #4743, opened 2020-03-27
Reopen my pervious issues with respect #4741 and #4742


Surefirer:
* **rAthena Hash**: latest

* **Client Date**: 20180831

* **Server Mode**: Renewal

* **Description of Issue**: I previously opened 2 issues regarding skill AL_DECAGI and WZ_JUPITEL no effect. I received notification that these 2 issues #4741 and #4742 was closed for the reason of can not be reproduced. I have downloaded the latest clean version of ra, tested again, recorded a video to prove these two issues exist. Please take a moment to watch my video. I reopen the issue with respect. Thanks. 
  * Result: 
  * Expected Result: 
  * How to Reproduce: video link: https://youtu.be/LLyg2cNdnQI, please let me know if you are not able to watch the video.

WZ_JUPITEL part: 1:45 ~ 2:30
AL_DECAGI part: 3:53 and 9:06

  * Official Information:
  

* **Modifications that may affect results**:

Atemo:
You're right, I tested with skills of low level but it seems there is an issue with skill of higher level, sorry !

Surefirer:
These skills do not work beyond level 10.

PR #4751, opened 2020-03-30
Fixes issue with power skills

aleos89:
* **Addressed Issue(s)**: #4743

  * Resolves an issue with power skills and default data defined in the skill's array.
Thanks to @Surefirer!


Hope8888:
@aleos89 
1) `@monster 1639`
2) The monster casting lv.28 Jupitel Thunder

Should be 30 hit, but its only one hit
and map-server  `[warning]: path_blownpos: count too many 63 !`

Issue #4744, opened 2020-03-28
Star Emperor Skills


iubantot:
* **rAthena Hash**: 

* **Client Date**: bdd6302

* **Server Mode**: 20180621RagexeRE

* **Description of Issue**: Star Emperor Skills not working as intended.
  * Result: Star Emperor skills are usable even without using the stance skill first.
  * Expected Result:  sun moon star skills should be usable when not using the appropriate stance
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**:

PR #4753, opened 2020-03-31
Corrects Star Emperor stances

aleos89:
* **Addressed Issue(s)**: #4744

* **Server Mode**: Renewal

  * Adds missing stance skill requirements.
Thanks to @iubantot!

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -29412,6 +29412,7 @@ Body:
         Time: 60000
     Requires:
       SpCost: 40
+      State: Moonstance
   - Id: 2575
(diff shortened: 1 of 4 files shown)


attackjom:
It need to update csv2yaml.cpp too.

under the line of ST_PECO

aleos89:
The reason I didn't is because technically SE/SR was released to master post the skill YAML change.

attackjom:
> The reason I didn't is because technically SE/SR was released to master post the skill YAML change.

Now, both has merge to master , so I think it should fix in this pull request.

Issue #4756, opened 2020-04-01
Autoloot commands not working correctly.


ecdarreola:
* **rAthena Hash**:  6edcf4a

* **Client Date**: 20180621a or 20180919

* **Server Mode**: Renewal

* **Description of Issue**: When using `@autoloot`, `@alootid` or `@aloottype` sometimes the items are not going to the inventory directly. It is shown on the floor.
  * Result: e.g. at autoloot 100, there's an instance that items are not being looted and just appear on the ground.
  * Expected Result: be looted immediately. 
  * How to Reproduce: Use the command.
  * Official Information: N/A
  

* **Modifications that may affect results**: PR requests (Class Improvements - RK, GX, Wanderer, Kagerou/Oboro, Rebellion, Achievements) - I don't see this as conflict since it doesn't touch any of the autoloot code.

Daegaladh:
Do you have homunculus_autoloot and hom_idle_no_share enabled?

ecdarreola:
@Daegaladh I have them both disabled.

aleos89:
Are you the one making the kill or are you using some support (pet, homunculus, mercenary) to make the kill?

ecdarreola:
@aleos89 character making the kill no support from mobs, mercenary or homunculus.

Caveat: haven't tried autoloot using homunculus, or mercenary.

Daegaladh:
@ecdarreola have you tried my PR #4761 ?

ecdarreola:
Haven't tried yet. Just saw the PR. Will try later. Thank you.

ecdarreola:
@Daegaladh : PR didn't work or didn't solve the issue.

Additional findings.. both tested with or without renewal drop.
- Loots are not going directly to the inventory if the difference between the mob and player level is 30.

I assume that the autoloot is being affected by penalty table even it is commented or not.

Testing scenario:
Job Rune Knight at level 1 allstats 25000 and uses @autoloot 100
Summon Raydric which is level 115
Kill monster until level 85 - it will not automatically loot. At level 86. it works pefectly.

ecdarreola:
Additional findings:
Which invalidates above, even at same level with raydric there are instance that the item are not being autolooted to the inventory.

Not sure if drop rate is the culprit, I've tested drop rate of 10x - 100x and 1000x

ecdarreola:
I uploaded a video of what is happening lol. See link below.

https://youtu.be/6NJ4lT7OMq0

[...]

PR #4761, opened 2020-04-02
Added idletime settings for homunculus idle no share

Daegaladh:
* Follow up to 921b3e2c3e9eae36cbf2ca35bcede6e0d73ba074

* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Adds idletime settings for homunculus idle no share separated from the party share idletime settings.


Daegaladh:
Probably fixes #4756

src/map/script.cpp
```diff
@@ -17385,6 +17385,17 @@ BUILDIN_FUNC(checkidle)
 	return SCRIPT_CMD_SUCCESS;
 }
 
+BUILDIN_FUNC(checkidlehom)
+{
+	TBL_PC *sd = NULL;
+
+	if( script_nick2sd(2,sd) )
+		script_pushint(st, DIFF_TICK(last_tick, sd->idletime_hom));
```
Lemongrass3110, line 17393: Shouldnt this simply return pc_isidle_hom?

src/map/battle.cpp
```diff
@@ -8875,6 +8875,7 @@ static const struct _battle_data {
 	{ "boss_nopc_idleskill_rate",           &battle_config.boss_nopc_idleskill_rate,        100,    0,    100,              },
 	{ "boss_nopc_move_rate",                &battle_config.boss_nopc_move_rate,             100,    0,    100,              },
 	{ "hom_idle_no_share",                  &battle_config.hom_idle_no_share,               0,      0,      INT_MAX,        },
+	{ "idletime_hom_option",                &battle_config.idletime_hom_option,             0x25,   1,      INT_MAX,        },
```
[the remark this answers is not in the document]
Daegaladh: I just copypasted from the normal idletime_option and didn't think much of it, I'll fix both

src/map/script.cpp
Daegaladh: No, that would only return if the player is idle or not, not the idle time

Issue #4763, opened 2020-04-02
*bg_get_data command crashes map server


voyfmyuh:
* rAthena Hash: https://github.com/rathena/rathena/commit/4f11020a4e59704f1909ec615857b6af098b3683

* Client Date: 20180621

* Server Mode: Renewal

* Description of Issue: When reporting this code, the map server crashes!

L_Reward:
	bg_get_data getarg(0), 1;
	for ( .@i = 0; .@i < $@arenamemberscount; ++.@i )
		getitem .rewarditem[0], .rewarditem[1], $@arenamembers[.@i];
	return;

![1585839997349](https://user-images.githubusercontent.com/54499336/78265923-aaf42700-7537-11ea-9d93-c7b1649a3fbe.jpg)

PR #4764, opened 2020-04-02
Fixes to Battleground script commands

aleos89:
* **Addressed Issue(s)**: #4763

  * Cleans up script commands bg_getareausers and bg_get_data.
Thanks to @voyfmyuh!

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -19933,8 +19933,7 @@ BUILDIN_FUNC(bg_getareausers)
 {
 	const char *str = script_getstr(st, 3);
 	int16 m, x0, y0, x1, y1;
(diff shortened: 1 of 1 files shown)


src/map/script.cpp
```diff
@@ -19947,12 +19947,8 @@ BUILDIN_FUNC(bg_getareausers)
 	x1 = script_getnum(st,6);
 	y1 = script_getnum(st,7);
 
-	for( i = 0; i < bg->members.size(); i++ )
```
Lemongrass3110, line 19950: declaration of "i" can be removed now.

voyfmyuh:
This code will also cause a crash!

Issue #4808, opened 2020-04-16
GC_WEAPONCRUSH is not working


wdivet:
* **rAthena Hash**: 44b80448df07bec285b1a5056ceae519265bd19b & 2b49d12d7384a12ec0c8b1f9ce1c7db713741fc6

* **Client Date**: 20180620

* **Server Mode**: renewal

* **Description of Issue**: 
  * Result: This skill do nothing.
  * Expected Result: It suposed to be earn damage (?) and strip weapon while in Weapon Blocking state.
  * How to Reproduce: Use Weapon Block -> let mobs hit you until Weapon Block state -> press Weapon Crush and nothing happen.
  * Official Information: It suposed to be earn damage (?) and strip weapon while in Weapon Blocking state.
  

* **Modifications that may affect results**: -

PR #4809, opened 2020-04-16
Fixes Weapon Crush and Weapon Blocking combo

aleos89:
* **Addressed Issue(s)**: #4808

  * Weapon Crush now uses SC_WEAPONBLOCK_ON to track the target instead of SC_COMBO.
  * Also fixes Quick Draw Shot target selection.
Thanks to @wdivet!

### DIFF
--- a/db/pre-re/skill_db.yml
+++ b/db/pre-re/skill_db.yml
@@ -18583,6 +18583,8 @@ Body:
     Duration1: 60000
     Requires:
       SpCost: 20
+      Status:
+        Weaponblock_On: true
   - Id: 2031
     Name: GC_VENOMPRESSURE
     Description: Venom Pressure
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -19325,6 +19325,8 @@ Body:
     FixedCastTime: -1
     Requires:
       SpCost: 20
+      Status:
+        Weaponblock_On: true
   - Id: 2031
     Name: GC_VENOMPRESSURE
     Description: Venom Pressure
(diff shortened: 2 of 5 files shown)


ecdarreola:
@aleos89 error on SC_QD_SHOT_READY causing crash.

image

aleos89:
How did you trigger the crash? After Chain Action triggered I used Quick Draw Shot and it casted just fine. I use it against a living target and attempted a couple times on a dying target.

ecdarreola:
> How did you trigger the crash? After Chain Action triggered I used Quick Draw Shot and it casted just fine. I use it against a living target and attempted a couple times on a dying target.

Haven't noticed. I was not the one who actually cause the crash. I just monitored the console and found this out.

wdivet:
> How did you trigger the crash? After Chain Action triggered I used Quick Draw Shot and it casted just fine. I use it against a living target and attempted a couple times on a dying target.

I don't know this was related or not, but i experienced same problem with my current version (https://github.com/rathena/rathena/commit/2b49d12d7384a12ec0c8b1f9ce1c7db713741fc6) manual update to this commit. And yes, If i use skill Weapon Crush or QDshot, map crash.

ecdarreola:
QSD caster receiving own damage from the latest commit.
Adding NoTargetSelf DamageFlag will make the skill not working.

ecdarreola:
Latest commit makes the skill RL_QD_SHOT not usable. When casting skill nothing happens even the condition is met with SC_ and after chain action.

LordWhiplash:
Applied and is working fine, but i've notice that i can use RL_QD_SHOT 3 times in one unique double hit from GS_CHAINACTION  if i start spamming RL_QD_SHOT even before the double hit proc...
I'll try to record a video of it... kinda hard to explain...
I don't know if it is suposed to happen (never player Rebellion on official servers)

ecdarreola:
Works fine now.

Issue #3176, opened 2018-05-30
Earthquake don't make any damage


poporing:
* **rAthena Hash**:  c68da31f67b205854ead7d10e5efb4d6c8bac19d

* **Client Date**: 20170614

* **Server Mode**: pre-renewal

* **Description of Issue**: lot of the time mvp that cast Earthquake make no damage
  * Result: when mvps ex: ifrit, rsx ect... cast earthquake player didn't get any damage
  * Expected Result: should make high damage
  * How to Reproduce: Just let a mvp cast earthquake on you
  * Official Information: I know that earthquake should do high damage.
  

* **Modifications that may affect results**: 
 I test it in a 100% clean rathena git

poporing:
I'm the only one with this issue?

poporing:
[Error]: 0 enemies targeted by 653:NPC_EARTHQUAKE, divide per 0 avoided!

KrokusPokus:
Your error message basically seems to originate from the lines
```
if (nk&NK_SPLASHSPLIT) { // Divide MATK in case of multiple targets skill
	if (mflag>0)
		ad.damage /= mflag;
	else
		ShowError("0 enemies targeted by %d:%s, divide per 0 avoided!\n", skill_id, skill_get_name(skill_id));
}
```
in battle_calc_magic_attack(). I don't know what situation would cause a mflag == 0, exactly.

There is a general problem with too low damage in rAthena's Earthquake calculation, though. From what I can tell, rAthena has no special handling for the damage of Earthquake, which is a magical attack, being based on a mob's physical atk instead of its matk (sources: https://irowiki.org/wiki/NPC_EARTHQUAKE and https://irowiki.org/wiki/NPC_EARTHQUAKE state "Neutral element magical damage based on the monsters physical attack power"). This could properly be fixed like this:

```
--- E:/Temp2/battle.cpp	Mon Jun 11 11:41:56 2018
+++ E:/Temp2/battle_new.cpp	Mon Jun 11 11:44:00 2018
@@ -5826,10 +5826,19 @@
 				ad.damage = status_get_lv(src) * 10 + sstatus->int_;
 				break;
 			default: {
-				if (sstatus->matk_max > sstatus->matk_min) {
-					MATK_ADD(sstatus->matk_min+rnd()%(sstatus->matk_max-sstatus->matk_min));
-				} else {
-					MATK_ADD(sstatus->matk_min);
+				if (skill_id == NPC_EARTHQUAKE) {
[...]

PR #4818, opened 2020-04-19
Bug fix for npc_earthquake skill: MATK calculation should be based on …

arkauss83:
…user's physical attack.

* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/3176

* **Server Mode**: Both

This bug fix should fix the NPC_EARTHQUAKE matk dmg calculation, which should be based on user's physical atk instead of matk. This patch could be improved by adding a specific block for any skill based on physical attack instead of magic, but this is the only one I know which should be like this... Hope it helps!


aleos89:
This was taken over in #5017. There were several other parts of the skill that were wrong and needed to be fixed. Thank you for your contribution!

PR #4828, opened 2020-04-21
Fixed Max HP & SP rate bonus

### LINKED ISSUE #4264: HP bonuses formula incorrect for pre-Renewal
* **rAthena Hash**: Latest

* **Client Date**: 2017-06-14

* **Description of Issue**: 
  * Result:  In a similar fashion to the Heal formula, the HP bonuses formula is also wrong for pre-RE. In 96443cd71073ad752c65d4d6705a61ec1747735f some changes were made which resulted in a combination such as Tao Gunka with Berserk/Frenzy not working correctly anymore. On rAthena, it applies one of them on the Base HP only.

As reported by one of our players, Base HP = 10969. With only Berserk = 32907. With only Tao Gunka = 21938. With both = 43878. This should be 65814.

We confirmed this on our 11.3 Aegis server. Base HP was 10670. With Berserk = 32010. With only Tao Gunka = 21340. With both = 64020.
  * Expected Result: Berserk applies to HP including bonuses, not on base.
  * How to Reproduce:  Test Tao Gunka + Berserk/Frenzy on Pre-Renewal Aegis.
  * Official Information: Aegis 11.3.
  

cydh:
* **Addressed Issue(s)**: #4264

* **Server Mode**: Both

  - Just trying to fix #4264

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -1147,6 +1147,8 @@ void initChangeTables(void)
 	StatusIconChangeTable[SC_SPL_MATK] = EFST_SPL_MATK;
 	StatusIconChangeTable[SC_ATKPOTION] = EFST_PLUSATTACKPOWER;
 	StatusIconChangeTable[SC_MATKPOTION] = EFST_PLUSMAGICPOWER;
+	StatusIconChangeTable[SC_INCREASE_MAXHP] |= EFST_ATKER_ASPD;
+	StatusIconChangeTable[SC_INCREASE_MAXSP] |= EFST_ATKER_MOVESPEED;
 
 	/* Cash Items */
 	StatusIconChangeTable[SC_FOOD_STR_CASH] = EFST_FOOD_STR_CASH;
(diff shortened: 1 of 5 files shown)


src/map/status.cpp
```diff
@@ -1147,6 +1147,8 @@ void initChangeTables(void)
 	StatusIconChangeTable[SC_SPL_MATK] = EFST_SPL_MATK;
 	StatusIconChangeTable[SC_ATKPOTION] = EFST_PLUSATTACKPOWER;
 	StatusIconChangeTable[SC_MATKPOTION] = EFST_PLUSMAGICPOWER;
+	StatusIconChangeTable[SC_INCREASE_MAXHP] |= EFST_ATKER_ASPD;
```
Daegaladh, line 1150: `|=` copypaste fail

src/map/status.cpp
```diff
@@ -1147,6 +1147,8 @@ void initChangeTables(void)
...
+	StatusIconChangeTable[SC_INCREASE_MAXSP] |= EFST_ATKER_MOVESPEED;
```
Daegaladh, line 1151: `|=` copypaste fail

Daegaladh: Also Gold Pills SCs are wrong on pre-re item_db

Issue #4833, opened 2020-04-23
WOE Controller Reward Issue


mazvi:
* **rAthena Hash**: 4a3dac6

* **Client Date**: 20180620RE

* **Server Mode**: RENEWAL

* **Description of Issue**: 

sending gifts woe using email, wrong recipient when 1 account has 2 char in the same guild.

The email is sent not to an online account but to an offline account.

I use option 1 | 2 | 8

I think the problem also impact if not use mail the player not receive if have 2 char on same account but join on same guild

  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**:

PR #4843, opened 2020-04-25
Fixes Woe_Controller (#4833)

akiyamamio16:
Fix some bugs when 1 account has 2 character with same guild while Duplicate IP Check is Enabled.
- Player will now receive the reward to the online character of same guild if "[4] Only reward Guild Masters" option is Disabled.


mazvi:
Is this fixes my issue? #4833

Atemo:
Repo deleted, takeover in https://github.com/rathena/rathena/pull/5071

Issue #4852, opened 2020-04-27
lord of vermilion calculation


mazvi:
* **rAthena Hash**: 05ebde0

* **Client Date**: 20180620RE

* **Server Mode**: RENEWAL

* **Description of Issue**: skill does not match the description of the kRO
  * Result:  LOV LV1
![screenRENEWAL1145](https://user-images.githubusercontent.com/8316103/80337353-d42c7c80-8883-11ea-8d5e-a7b2714114d9.jpg)

  * Result:  LOV LV10
![screenRENEWAL1144](https://user-images.githubusercontent.com/8316103/80337360-d8f13080-8883-11ea-9c2d-4ebcf8c44510.jpg)
  * Expected Result: 
[SKID.WZ_VERMILION] = {"로드 오브 버밀리온(Lord of Vermilion)", "MAX Lv : 10", "^777777습득조건 : 선더 스톰 1, 유피텔 선더 5^000000", "계열 : ^993300액티브^000000", "타입 : ^777777마법^000000", "대상 : ^777777지면 1셀^000000", "내용 : ^777777지정 셀 주변 13X13범위에 낙뢰를 모아 떨어뜨려 풍속성 마법 데미지를 주고, 확률적으로 [상태이상 : 암흑]을 건다.", "암흑에 걸릴 확률은 대상의 상태이상 내성에 의해 감소된다.^000000", "^ffffff_^000000", "[Lv 1] : ^777777MATK 500%, 암흑확률:15%^000000", "[Lv 2] : ^777777MATK 600%, 암흑확률:20%^000000", "[Lv 3] : ^777777MATK 700%, 암흑확률:25%^000000", "[Lv 4] : ^777777MATK 800%, 암흑확률:30%^000000", "[Lv 5] : ^777777MATK 900%, 암흑확률:35%^000000", "[Lv 6] : ^777777MATK 1000%, 암흑확률:40%^000000", "[Lv 7] : ^777777MATK 1100%, 암흑확률:45%^000000", "[Lv 8] : ^777777MATK 1200%, 암흑확률:50%^000000", "[Lv 9] : ^777777MATK 1300%, 암흑확률:55%^000000", "[Lv10] : ^777777MATK 1400%, 암흑확률:60%^000000"} 
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**:

Latiosu:
Happy to look into this issue.

mazvi:
[...]

PR #4855, opened 2020-04-27
Fix Lord of Vermilion damage

Latiosu:
* **Addressed Issue(s)**: #4852 
* **Server Mode**: Renewal
* **Description of Pull Request**: Corrects player WZ_VERMILION skill damage ratio to match kRO (source: divine-pride).

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -6247,7 +6247,7 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 						break;
 					case WZ_VERMILION:
 						if(sd)
-							skillratio += 400 + skill_lv * 100;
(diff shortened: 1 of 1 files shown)


aleos89:
This change means that all 20 hits will be doing 600%+ damage each.

qwerty7vp:
shouldn't it be
skillratio += 300 + skill_lv * 100;

aleos89:
Right, good catch. I just looked at the description again. `skillratio` is defaulted to 100 so any `+=` need to account for that default value.

Latiosu:
Ah yes, I totally missed: https://github.com/rathena/rathena/blob/816728eee1c8bfa777ec0f10b5df48903e8f061a/src/map/battle.cpp#L6017
I'll update PR.

Issue #4863, opened 2020-04-29
MO_FINGEROFFENSIVE Lv2+ doesnt work as it should be


qwerty7vp:
* **rAthena Hash**: 4829f3775

* **Client Date**: 20180919Ragexe, 20180620RagexeRE

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: All level damage almost the same and only 1 hit
  * Expected Result: Hit and Damage should raise according to skill level
  * How to Reproduce: Create Sura/Champion character and test the skill
  * Official Information:
  

Video for my testing : https://youtu.be/-p40Qx2_Dow

* **Modifications that may affect results**: Nothing I tried set finger_offensive_type: both 0 and 1 and `@reloadbattleconf` , `@reloadskilldb`
`

Daegaladh:
I already fixed this on 38cf16edbaa36836072780b4481023e57e014892
I've just tested it on a clean rAthena and works fine, setting `finger_offensive_type` to 0 hits depending on the number of spheres you have, and setting it to 1 does all the damage in 1 hit.

qwerty7vp:
but I'm using latest rA and it has your fix but it still hit only once with both setting to 0 and 1
![image](https://user-images.githubusercontent.com/64515036/80631811-7ca73000-8a80-11ea-869d-4b98180e86c6.png)

Daegaladh:
You must be doing something wrong
https://i.imgur.com/rkb3nNB.mp4

qwerty7vp:
what's diff client and kRO version you're using ?
I tried to clone clean latest rathena 2 minutes ago
it's still not working like yours

Daegaladh:
I've tested both on 2018-06-21aRagexe and 2015-05-13aRagexe
But I don't think it's something related to the client

qwerty7vp:
this is weird I tried many things
2019 kRO , 2020 kRO
20180620RagexeRE, 20180621Ragexe/RagexeRE, 20180919Ragexe
with clean latest rA ( only conf/import folder that edited)
still cant make it multihit
is there a condition of the skill to make it multihit ? like weapon or buff or something ?

Daegaladh:
The only condition is the spheres, the hits depend on the number of spheres you have

PR #4866, opened 2020-04-29
Fixed MO_FINGEROFFENSIVE

Daegaladh:
* **Addressed Issue(s)**: #4863

* **Server Mode**: Both

* **Description of Pull Request**: The number of hits depends on the spheres the player has, but the max number of hits should be the same as the skill level.

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -7519,7 +7519,17 @@ Body:
           Amount: 24
         - Level: 5
           Amount: 28
-      SpiritSphereCost: 1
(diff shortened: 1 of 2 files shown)


LordWhiplash:
About the damage, i've checked that at master that it is `skillratio += 500 + skill_lv * 2`, when it should be `skillratio += 500 + skill_lv * 200` according to Divine Pride and http://kafra.kr/#!/en/KRO/skillinfo/267

I've just tested the changes, and is working fine, thanks!

PR #4871, opened 2020-04-30
Add dockerised local development environment

Latiosu:
* **Server Mode**: Both
Adds an easy way for developers new to rAthena to get a local development environment up and running with only a single pre-requisite (Docker Desktop).

**Instructions for use:**
1. `docker-compose up -d` to spin up database (automatically imports `/sql-files/*.sql` on first load)
2. `docker exec -it rathena bash` to connect to lightweight linux container with all rathena dependencies (plus gdb)
3. Devs can now run usual commands (`./configure`, `make`, etc ...) or start server (`./athena-start`, `gdb map-server`, etc ...)
4. `docker-compose down` to free resources when done

**Demonstration:** https://terminalizer.com/view/c05443023904

**Additional Information:**
- Database is persisted to local disk so no need to recreate database every time.
- First time the database volume is created, all `*.sql` files are imported into the database. This does not happen on future start ups unless the volume has been deleted (devs can delete volume with `docker-compose down --volumes`)
- Devs can inspect state of containers with `docker-compose ps` and modify the setup as they see fit (Official Docs: https://docs.docker.com/compose/compose-file/)
- It would be ideal to have a page of documentation to accompany this feature (if it is approved), which I'm happy write but not sure where it would best live; perhaps in `CONTRIBUTING.md`?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,36 @@
+version: "3.7"
+
+services:
+    db:
+        image: "mariadb:bionic"
+        container_name: "rathena_db"
+        ports:
+            - "3306:3306" # allow DB connections from host
+        volumes:
+            - "rathenadb:/var/lib/mysql" # save database to local disk
+            - "./sql-files/:/docker-entrypoint-initdb.d" # initialize db with ./sql-files
+        environment:
+            MYSQL_ROOT_PASSWORD: ragnarok
+            MYSQL_DATABASE: ragnarok
+            MYSQL_USER: ragnarok
+            MYSQL_PASSWORD: ragnarok
+    server:
+        image: "rathena:local"
+        container_name: "rathena"
+        ports:
+            - "5121:5121" # map server
+            - "6121:6121" # char server
+            - "6900:6900" # login server
+        volumes:
+            - ".:/rathena" # mount directory inside container
+            - "./tools/docker_db.conf:/rathena/conf/import/inter_conf.txt" # load db connection
+        tty: true
+        stdin_open: true
+        build:
+            context: .
+            dockerfile: Dockerfile
+        depends_on:
+            - db
+
+volumes:
+    rathenadb:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,6 +24,7 @@ services:
         volumes:
             - ".:/rathena" # mount directory inside container
             - "./tools/docker_db.conf:/rathena/conf/import/inter_conf.txt" # load db connection
+        init: true # helps with signal forwarding and process reaping
         tty: true
         stdin_open: true
         build:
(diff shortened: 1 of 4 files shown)


Normynator:
I like the idea. Docker is awesome.
I think a little guide is the only thing which is missing.
What you wrote in your PR description would already be enough, just as part of the repo so people can find it easily.

vstumpf:
I think you'll have the same issue I had, stopping the server will leave zombies. If you do an ./athena-start stop, it will fail to clean up the processes. 

This means you'll have to bring down your compose cluster and bring it back up to recompile and restart.

Latiosu:
> I like the idea. Docker is awesome.
> I think a little guide is the only thing which is missing.
> What you wrote in your PR description would already be enough, just as part of the repo so people can find it easily.

Agreed, I think some documentation is needed too. Any recommendations for where to put it?
I was thinking `CONTRIBUTING.md` was the most "developer-specific" documentation place that exists at the moment, but I feel there might be somewhere better.

Latiosu:
> I think you'll have the same issue I had, stopping the server will leave zombies. If you do an ./athena-start stop, it will fail to clean up the processes.
> 
> This means you'll have to bring down your compose cluster and bring it back up to recompile and restart.

I haven't yet done any detailed testing with zombie processes but I can see your concern.
Although, restarting the containers with `docker-compose down && docker-compose up` won't erase the compiled files or your changes since I've set up a bind mount.

A developer is able to run `./athena-start start` right away after they restarted their containers since the binaries are saved to disk and re-mounted into the container.
image

Latiosu:
I've enabled Tini to help reap processes in the dev container and added documentation for how to use the development environment in `CONTRIBUTING.md`

Here is preview of how it looks:
image

lighta:
Awesome, nice work. I was doing a little docker myself to check rA when I was inform of your PR.
I'll check this and will try to make it integrate asap =)

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+            - "3306:3306" # allow DB connections from host
+        volumes:
+            - "rathenadb:/var/lib/mysql" # save database to local disk
+            - "./sql-files/:/docker-entrypoint-initdb.d" # initialize db with ./sql-files
+        environment:
+            MYSQL_ROOT_PASSWORD: ragnarok
+            MYSQL_DATABASE: ragnarok
+            MYSQL_USER: ragnarok
+            MYSQL_PASSWORD: ragnarok
+    server:
```
lighta, line 17: I think I would have split server across 3 images to be a bit more flexible

Latiosu: That works too, I see the pros here being STDOUT logs are more distinctly separated and being able to specify different resource limits per server. Cons that I can think of is perhaps slightly more overhead of having more containers running.

secretdataz:
superseded by #5572

Issue #4881, opened 2020-05-02
13.1 Guards display twice a message window OnTouch and get frozen


Artuvazro:
* **rAthena Hash**: 2e4ec1d

* **Client Date**: 2018-06-21

* **Server Mode**: pre-renewal

* **Description of Issue**: 
  * Result: Guard NPCs in lhz_b01 have a duplicate with an OnTouch label and a x,y area defined that makes the OnTouch trigger twice. When ep13_ryu > 19 the OnTouch triggered twice makes the message window appear twice, the second one without any NEXT button, so the NPC is frozen and the player has to foce-close the client.
  * How to Reproduce: Try to walk **between** two of the guards located at lhz_b01, for example, the ones at lhz_b01 147,252.

Daegaladh:
This will be fixed with #4725

Atemo:
Fixed in https://github.com/rathena/rathena/commit/7efd696af7da0f46bf291a3ed25509dbb9e164d2

PR #4882, opened 2020-05-02
Fixes Issue #4881 : Duplicated OnTouch tags freeze NPC messages in 13.1 Guard NPCs

Artuvazro:
Removes unnnecessary trigger areas from duplicated NPC that caused wrong UI behavior.

* **Addressed Issue(s)**: #4881

* **Server Mode**: pre-renewal and renewal

* **Description of Pull Request**: Removes unnnecessary trigger areas from duplicated NPC that caused wrong UI behavior.

### DIFF
--- a/npc/quests/quests_13_1.txt
+++ b/npc/quests/quests_13_1.txt
@@ -1132,33 +1132,33 @@ function	script	Guard_13_1	{
 	end;
 }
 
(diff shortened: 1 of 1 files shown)


Daegaladh:
This needs to be fixed by fixing OnTouch behavior instead.

Issue #4875, opened 2020-05-01
SJ_STARSTANCE, SJ_LUNARSTANCE doesnt give bonus status


qwerty7vp:
* **rAthena Hash**: 4829f3775

* **Client Date**: 20180919Ragexe

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Both Stances dont give bonus status
  * Expected Result: Should give bonus status correctly
  * How to Reproduce: Create Star Emperor and use those stances
  * Official Information:
  

Lunar Stance
![image](https://user-images.githubusercontent.com/64515036/80779206-1fe96980-8b95-11ea-98b9-345834e0cf75.png)

Star Stance
![image](https://user-images.githubusercontent.com/64515036/80779282-650d9b80-8b95-11ea-9551-abd09a1ce4e4.png)

* **Modifications that may affect results**: Nothing
`

LordWhiplash:
I've looked at this yesterday, and noticed that too, thanks for reporting.
Not sure, but i think the problem of this stances are in status.cpp

![Screenshot_20200502-025005](https://user-images.githubusercontent.com/62680611/80856549-79977580-8c21-11ea-8133-a9185c364c67.png)
(Star Stance is in a section about preRE aspd, and i dont know if it is really suposed to be "-=" instead of "+=")



![Screenshot_20200502-024856](https://user-images.githubusercontent.com/62680611/80856541-4c4ac780-8c21-11ea-9435-bfb580f43655.png)
(I think it need to be here)

About the lunar stance, seems that the defined val2 is divided for 100, i mean, when lunar stance is level 3, val2 = 5, and divided by 100, becoming 0,05% of your Max hp, that would increase your HP in 8 if your MaxHP in the screenshot is clean 16723 (strange, since it did not increase at all)

![Screenshot_20200502-025518](https://user-images.githubusercontent.com/62680611/80856949-ac8f3880-8c24-11ea-98b5-b2734cb0b279.png)
(Val2 definition)


![Screenshot_20200502-025452](https://user-images.githubusercontent.com/62680611/80857042-553d9800-8c25-11ea-9b17-11ba91a1c815.png)
(Bonus HP definition)

Latiosu:
@LordWhiplash You were on the right track.
SJ_STARSTANCE wasn't applying attack speed effect in renewal since there wasn't a condition for it in `status_calc_aspd`.
[...]

PR #4884, opened 2020-05-03
Fix bonuses for SJ_STARSTANCE and SJ_LUNARSTANCE

Latiosu:
* **Addressed Issue(s)**: #4875
* **Server Mode**: Both
* **Description of Pull Request**: Corrects Max HP calculation for SJ_LUNARSTANCE and missing ASPD bonus for SJ_STARSTANCE (renewal only).

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -3398,7 +3398,7 @@ static int status_get_hpbonus(struct block_list *bl, enum e_status_bonus type) {
 				bonus += 30;
 #endif
 			if(sc->data[SC_LUNARSTANCE])
-				bonus += bonus * sc->data[SC_LUNARSTANCE]->val2 / 100;
+				bonus += sc->data[SC_LUNARSTANCE]->val2;
 
 			//Decreasing
 			if(sc->data[SC_VENOMBLEED])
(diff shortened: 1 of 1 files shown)


Latiosu:
**Local Development Testing:**
---
**SJ_STARSTANCE Before**
SJ_STARSTANCE-before
**SJ_STARSTANCE After**
SJ_STARSTANCE-after

---
**SJ_LUNARSTANCE Off**
SJ_LUNARSTANCE-off

**SJ_LUNARSTANCE On**
SJ_LUNARSTANCE-on

src/map/status.cpp
```diff
@@ -7549,6 +7549,8 @@ static short status_calc_aspd(struct block_list *bl, struct status_change *sc, b
 			bonus += sc->data[SC_INCREASEAGI]->val1;
 		if (sc->data[SC_NIBELUNGEN] && sc->data[SC_NIBELUNGEN]->val2 == RINGNBL_ASPDRATE)
 			bonus += 20;
+		if (sc->data[SC_STARSTANCE])
```
ecdarreola, line 7552: Just confirming if its appropriate to add this bonus same with pre-renewal (status_calc_aspd) even it doesn't have its class implemented there. For the sake of others who are recompiling them at pre-re.

Latiosu: I believe there already is a condition in place for Pre-renewal but I haven't tested it personally.
See `status_calc_aspd_rate`:
https://github.com/rathena/rathena/blob/132180325435847a9470f82cb795525d03be1b67/src/map/status.cpp#L7742-L7743

ecdarreola: Ah yeah. overlooked at it I thought both pre-re and re are not having its bonuses. I stand corrected.

Issue #4880, opened 2020-05-02
MO_ExtremityFist doesn't double damage (Renewal)


ecdarreola:
* **rAthena Hash**: 2e4ec1d

* **Client Date**: 20180621

* **Server Mode**: Renewal

* **Description of Issue**: MO_ExtremityFist doesn't apply skillratio += 100 if sphere count is more than 5.
  * Result: Same damage observed if sphere count is >|< 5 
  * Expected Result: At sphere count > 5 it should double damage.
  * How to Reproduce: Compile renewal server -> allstats 130 -> use skill asura first without rising dragon to baseline damage -> use rising dragon and asura
  * Official Information: 
  

* **Modifications that may affect results**: None
  
  

I also have tried modifying the code https://github.com/rathena/rathena/blob/2e4ec1d1f656daaad832033c31a8b6939cb8d266/src/map/battle.cpp#L3793

to: `if(sd->spiritball > 5)` and same result observed.

Not quite sure what does this do `skillratio = min(500000,skillratio); //We stop at roughly 50k SP for overflow protection`

qwerty7vp:
> Not quite sure what does this do `skillratio = min(500000,skillratio); //We stop at roughly 50k SP for overflow protection`

as it said it capped MO_EXTREMITYFIST
for maximum at 50k SP (500k% skill ratio)

ecdarreola:
How does your test do? Does damage doubles if sphere count is > 5 ?

LordWhiplash:
Hi, i've just tested, and you're right, bonus is not right...
The problem is related to this:
![image](https://user-images.githubusercontent.com/62680611/80909268-475b4600-8cfd-11ea-947e-1a937dff1563.png)

I mean, imagine you have 2k SP, this would mean your skill ratio would be = 100 * (7 + 2000/10), this is 20700% ATK, to be honest 20800% since the base 100% is included... when you add 100% for 6 or more spheres, it goes to 20900%, so, it seems no damage increase...

The bonus should be "skillratio += 100 + 100 * (7 + sstatus->sp / 10)", because "Increases damage by 100% when there are 6 spirit sphere or more remains." is meant to DOUBLE the skill ratio, not add 100. Of course it needs to be discuted, it is about text interpretation.

I've applied this changes to test, and seems right now.
![image](https://user-images.githubusercontent.com/62680611/80909471-1a0f9780-8cff-11ea-8c96-f4cd29d74b2f.png)

[...]

PR #4886, opened 2020-05-03
Fixes MO_EXTREMITYFIST

LordWhiplash:
Fixes MO_EXTREMITYFIST double damage with more than 5 spheres

* **Addressed Issue(s)**: #4880

* **Server Mode**: Both

* **Description of Pull Request**: MO_EXTREMITYFIST is suposed to do double damage when user has 6 or more spirit spheres

Thanks to @ecdarreola

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3791,7 +3791,7 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 		case MO_EXTREMITYFIST:
 #ifdef RENEWAL
 			if (wd->miscflag&1)
-				skillratio += 100; // More than 5 spirit balls active
+				skillratio += 100 + 100 * (7 + sstatus->sp / 10); // More than 5 spirit balls active
 #endif
 			skillratio += 100 * (7 + sstatus->sp / 10);
 			skillratio = min(500000,skillratio); //We stop at roughly 50k SP for overflow protection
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -3791,7 +3791,7 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 		case MO_EXTREMITYFIST:
 #ifdef RENEWAL
 			if (wd->miscflag&1)
-				skillratio += 100; // More than 5 spirit balls active
+				skillratio += 100 + 100 * (7 + sstatus->sp / 10); // More than 5 spirit balls active
```
ecdarreola, line 3794: This is not the proper way to do it. I figured it how, however I don't know how to create a PR so I'll comment it here. Apologies for it.

Changed the code to this one.
`		case MO_EXTREMITYFIST:
			skillratio += 100 * (7 + sstatus->sp / 10);
#ifdef RENEWAL
			if (wd->miscflag&1)
				skillratio *= 2; // Doubles the damage by 100%
#endif			
			skillratio = min(500000,skillratio); //We stop at roughly 50k SP for overflow protection
			break;`

LordWhiplash: I believe both codes are doing the same thing... I may change it tomorow... In your way is more elegant :)

PR #4887, opened 2020-05-03
Fixed itemshop currency trade restrictions check

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Itemshop currency items are ment to be consumed in the shop, so there's no point on check for trade restrictions in this case.


aleos89:
Rather than removing the src flag, we should just adjust the `items.conf::allow_bound_sell` config to `0x2`.

Daegaladh:
> Rather than removing the check we should just adjust the `items.conf::allow_bound_sell` config to `0x2`.

That makes no sense, that part of the code is inside of `else if (!item->bound) {` so it would never apply to bound items.

Issue #4901, opened 2020-05-06
mob chat not displaying


iraciz:
* **rAthena Hash**: 

* **Client Date**: 20180620

* **Server Mode**: pre-renewal

* **Description of Issue**: mobs are not displaying the chat associated to mob_chat_db
  * Result: not sending the chat in their associated skill
  * Expected Result: Monster talk with the color text after casting the skill with the mob_chat line_id
  * How to Reproduce: Summon any of this monsters: 1916,  1929,  2022
  * Official Information: they should talk.

  

* **Modifications that may affect results**: tested in a new clean rathena version

Kreustoo:
Ok, I have the same problem, and while testing things, the ONLY changment I had was that line of useless code.
And SM just started talking (1916).
![wtf](https://user-images.githubusercontent.com/14162419/81226100-97325980-8fea-11ea-8dab-b7a1564333ea.JPG)

If I removes it, he doesn't talk anymore, it makes no sense to me but maybe it may help someone.
Or just if someone can tell me they have the same behaviour.

iraciz:
> Ok, I have the same problem, and while testing things, the ONLY changment I had was that line of useless code.
> And SM just started talking (1916).
> ![wtf](https://user-images.githubusercontent.com/14162419/81226100-97325980-8fea-11ea-8dab-b7a1564333ea.JPG)
> 
> If I removes it, he doesn't talk anymore, it makes no sense to me but maybe it may help someone.
> Or just if someone can tell me they have the same behaviour.

Thank you Sir, I made that change it the problem was fixed.
![mobchatfix](https://user-images.githubusercontent.com/15825408/81237485-64c23580-8fc5-11ea-9422-9ec07f5f3e7c.png)

Now my question is? WHY WAS REMOVED THAT LINE?

Daegaladh:
Please, don't close the issue until is fixed

PR #4908, opened 2020-05-07
Fixed monster messages not loading

Daegaladh:
* **Addressed Issue(s)**:  #4901 

* **Server Mode**: Both

* **Description of Pull Request**: Fixes monster messages not loading because mob_skill_db parser requires mob_chat_db to be parsed before.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -5525,10 +5525,9 @@ static void mob_load(void)
 	};
 
 	// First we parse all the possible monsters to add additional data in the second loop
-	if( db_use_sqldbs ){
+	if( db_use_sqldbs )
 		mob_read_sqldb();
-		mob_read_sqlskilldb();
-	}else{
+	else
 		for(int i = 0; i < ARRAYLENGTH(dbsubpath); i++){
 			int n2 = strlen(db_path)+strlen(DBPATH)+strlen(dbsubpath[i])+1;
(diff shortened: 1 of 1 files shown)


Kreustoo:
Sorry to be that guy, but did you try that modification with use_sql_db: yes in inter_athena.conf?

I did it :
test
And the modification only works with no sql, if you use sql, I still don't have chat. But yeah, without use the sql db it works.

To be clear :
--compile with your modification--

=> use_sql_db: no
=> launch server => SM is talking

=> close the server
=> use_sql_db: yes
=> launch serve => SM is muted

And as I said in the issue my weird modification:
wtf
It works in both sql and no sql on my side

It makes no sense to me.

Daegaladh:
@Kreustoo try the lastest fix, it should work now for both modes

Kreustoo:
Oh right moving **mob_read_sqlskilldb** to the same level as **mob_readskilldb** and putting the parsing of **chat_db** before it seems to work.

_But it stills make no sense to me why it needs that while in earlier version of rathena it worked ^^._

Daegaladh:
> _But it stills make no sense to me why it needs that while in earlier version of rathena it worked ^^._
Because mob_skill_db now checks if the index on mob_chat_db exists before asigning it to a skill, so the latest needs to be parsed first.

PR #4914, opened 2020-05-08
mapduplicate script command

sader1992:

* **Server Mode**: both

mapduplicate script command allows to duplicate maps in the server without the need to duplicate it in the client
this command will make things much easier for custom systems and event scripts
the idea is when you want to create a new event or a new pvp map or system, for example instead of uploading the script with the related map files or the "how to duplicate a map in your client"
this will allow scripters/server owners to skip this steps.

**Old idea to store the map names:**
> the reason I use `$map_duplicate_list$` global variable to save the maps name is that `$@` variable get cleared on `@reloadscript` command, the user can save the map names on his own or can check the list of duplicated maps from the  global variable
> 
> still to do is clearing the `$map_duplicate_list$`  on server restart, if there is an alternative to the global variable idea, it would be better!
> for example , we could create an empty vector to save the map names to it , and create a command script that give us the output of that vector , i think this would be better, will check it out, however i am up for suggestions.

and hopefully soon a duplicate command for npcs would be the best compo with this command.

test script:
```cs
prontera,153,180,0	script	mapduplicate_test	444,{
	if(select("create new duplicate to prontera:names list") == 2){
		if(!getarraysize($mapduplicate$)){
			mes "no maps on the list yet";
			end;
		}
		mes "select a map to warp to it!";
		.@s = select(implode($mapduplicate$, ":")) -1;
		warp $mapduplicate$[.@s],155,183;
		end;
	}
	$mapduplicate$[getarraysize($mapduplicate$)] = mapduplicate("prontera");
	debugmes $mapduplicate$[getarraysize($mapduplicate$)];
end;

OnInterIfInitOnce:
	deletearray $mapduplicate$[0],getarraysize($mapduplicate$);
end;
}
```

talk to the npc , create a new copy of prontera, check the map lists , click on the map and it will teleport you to it.

IRL Example:
```cs
//npc that have all the maps duplicate and load once at the server start.
-	script	eventmaps_duplicates	-1,{
OnInterIfInitOnce:
	//you can put all the map duplicates here.
	$some_event_map$ = mapduplicate("prontera");
end;
}

//some random event that uses the duplicated map
prontera,150,180,0	script	some_event	444,{
	mes "whatto the even map ?";
	if(select("yes:no") == 2)
		end;
	warp $some_event_map$,150,180; //this will warp you to the map event.
end;
}
```

https://github.com/rathena/rathena/pull/4914/commits/95a5872b07208b8f2cf9c5745dcc55325cafd372
With this update , the duplicated maps uses alphabet instead of numbers at the start
with that now instance maps not the same as the duplicated maps
also for maps with long name `a~z` have more than `1~9`
for example if i use prontera i can do from `1#prontera` to `99#prontera` means 99
but with alphabet i can do `a#prontera` to `zz#prontera` means ‭676‬
image

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3116,15 +3116,22 @@ map ID doesn't exist.
 
 *mapduplicate("<map name>")
 
-This command allows you to duplicate a map using npc,
-Without adding the map files to the client , it work the same as
-The map duplication in the instance system,
+This command allows you to duplicate a map using script,
+Without adding a copy of the map files to the client ,
+It work the same as map duplication in the instance system,
 The duplicate map will be have no map data (aka npcs and mapflags),
-Duplicate Maps are removed on the server restart.
-'$map_duplicate_list$' Global Variable will have a list of all the
-map duplicate names.
+Duplicated Maps are removed on the server restart.
 
 Returns the new map name if successed , and Empty string if failed.
+
+Example:
+	//warp to the duplicated map.
+	warp $mymap$,150,180;
+	end;
+
+	OnInterIfInitOnce:
+		//duplicate prontera once on the server start.
+		$mymap$ = mapduplicate("prontera");
 
 ---------------------------------------
 
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3114,6 +3114,12 @@ map ID doesn't exist.
 
 ---------------------------------------
 
+*mapduplicate("<map name>")
+
+Returns the new map name if successed , and Empty string if failed.
+
+---------------------------------------
+
 *getgmlevel({<char_id>})
 
 This function will return the (GM) level associated with the player group to which
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -3116,6 +3116,14 @@ map ID doesn't exist.
 
 *mapduplicate("<map name>")
 
+This command allows you to duplicate a map using npc,
+Without adding the map files to the client , it work the same as
+The map duplication in the instance system,
+The duplicate map will be have no map data (aka npcs and mapflags),
+Duplicate Maps are removed on the server restart.
+'$map_duplicate_list$' Global Variable will have a list of all the
+map duplicate names.
+
 Returns the new map name if successed , and Empty string if failed.
 
 ---------------------------------------
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -2716,6 +2716,7 @@ int map_duplicate(int src_m)
 		return -1;
 	}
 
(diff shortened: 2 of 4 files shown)


Atemo:
I'm fine with the idea but I would rather use and update the current instance system for example with

- allow infinite instance timer
- a command to change the timer of a running instance eventually
- a command to retrieve the instance id by type eventually
- a command to keep (or not) the instance on reloadscript
- a command to create dummy instance data, the data in instance_db.yml for dynamical script, eventually

NPC script example
--------------------------------
```
  - Id: 100
    Name: prontera instance
    TimeLimit: -1
    Enter:
      Map: prontera
      X: 155
      Y: 183
```

```
prontera,153,180,0	script	mapduplicate_test	444,{
	switch( select( "Create new duplicate to prontera", "Names list" ) ) {
	case 1:
		debugmes "Instance ID: " + instance_create("prontera instance", IM_NONE);
		end;
	case 2:
		.@size = instance_list(IM_NONE, "prontera");	// save the instance IDs in .@instance_list and return the size list
		if (.@size == 0) {
			mes "No maps on the list yet";
			end;
		}
		mes "Select a map to warp to it!";
		for ( .@i = 0; .@i < .@size; ++.@i )
			.@menu$ += instance_mapname("prontera", .@instance_list[.@i]) + ":";
		.@s = select(.@menu$) -1;
		if (instance_enter("prontera instance",155,183,getcharid(0),.@instance_list[.@s]) != IE_OK) {
			mes "The map doesn't exist anymore.";
		end;
	}
}
```

mazvi:
Can npc also duplicate on duplicated map?

sader1992:
there is no npc duplicate command in rathena yet , you can use the one by zell in the forum of just prepare your npcs before and unitwarp them to the map
the idea of the command to duplicate the map only without anydata , however i still thinking about what Atemo mentioned , maybe we can add a an option to copy the npc data or not , same goes for mapflag

mazvi:
Okay only map without data also great.. but mapflag i think must be cannot save location.. because if map destroy the player will be stuck maybe..

sader1992:
iI think the server already handle that the same way as the instance system , but  I will check it out.

zellkennedy:
> Okay only map without data also great.. but mapflag i think must be cannot save location.. because if map destroy the player will be stuck maybe..

Players in a invalid map when loggin will be warped to prontera

mimishi:
i suggest to have remove duplicate map command on this

sader1992:
I need this for references , a new pr open for the same thing but different way
https://github.com/rathena/rathena/pull/5112

Issue #4934, opened 2020-05-11
@useskill 0 is a crash


Stingor:
* **rAthena Hash**: last
* **Description of Issue**: be dumb and try @useskill 0 10 Stingor on you
* **Modifications that may affect results**: 

```
 	if (skill_id < 1 || skill_lv < 1) {
 		clif_displaymessage(fd, msg_txt(sd,198)); // This skill number doesn't exist.
 		return -1;
 	}
```

PR #4935, opened 2020-05-11
Fixed crash caused by invalid skill id

CairoLee:
Thanks to @Stingor

* **Addressed Issue(s)**: #4934

* **Server Mode**: Both

* **Description of Pull Request**: Make some potential crash. Thanks to @Stingor

### DIFF
--- a/src/map/atcommand.cpp
+++ b/src/map/atcommand.cpp
@@ -5835,6 +5835,14 @@ ACMD_FUNC(useskill)
 		return -1;
 	}
 
+	if (!skill_id) {
+		clif_displaymessage(fd, msg_txt(sd, 198)); // This skill number doesn't exist.
+		return -1;
+	}
+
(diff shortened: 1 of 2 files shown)


CairoLee:
I make a mistake in commit information.
If it is finally approved to merge this Pull Request, it is recommended manager use the square method to merge and remove the issue number in the commit message.

src/map/atcommand.cpp
```diff
@@ -5835,6 +5835,14 @@ ACMD_FUNC(useskill)
 		return -1;
 	}
 
+	if (!skill_id) {
```
Stingor, line 5838: hi maybe put this also to avoid error in console :D
[Error]: Skill '65535' is undefined! skill.cpp:191::skill_check 

`if (!skill_id && skill_id < MAX_SKILL_ID) {`

Issue #4954, opened 2020-05-14
PR_SUFFRAGIUM should be self-targeted screenwide AoE that affects caster


flamefury:
* **rAthena Hash**: 0a0b05e6a1801050d1b49d7993140a2dd7a7b05e

* **Client Date**: Doesn't matter

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Suffragium is a targeted skill that only affects one player and can't be used on yourself.
  * Expected Result: Suffragium is a self-targeted AoE that affects all allied members within range. Testing on kRO Sakray shows a range of 18 cells away (37x37). If ally is within visible range (29x29), skill animation appears on them and they perform their cast animation. If they are outside visible range, no animations occur on ally.
  * How to Reproduce: Try Suffragium.
  * Official Information: https://www.divine-pride.net/database/skill/67
iRO and kRO descriptions both say target is self and effect occurs for user and party members.

* **Modifications that may affect results**: None

flamefury:
Animation when out of sight of Priest will be dealt with in #4975

PR #4955, opened 2020-05-14
Suffragium should be self AoE for full party

flamefury:
* **Addressed Issue(s)**:  #4954 
* **Server Mode**: Renewal
Make Suffragium be a screenwide AoE that affects the entire party, including self.

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -2381,16 +2381,13 @@ Body:
     Description: Suffragium
     MaxLevel: 3
     Type: Magic
-    TargetType: Support
+    TargetType: Self
     DamageFlags:
       NoDamage: true
       Splash: true
-    Flags:
-      NoTargetSelf: true
-    Range: 9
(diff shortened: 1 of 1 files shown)


aleos89:
Is it a client bug or something about the animation, or is this really how Gravity intended it to be?

flamefury:
> Is it a client bug or something about the animation, or is this really how Gravity intended it to be? If it's really the feature of the skill then we need to add a range check in source and just not send `clif_skill_nodamage()` and just start the status instead.

I've no idea if this is intended or not on Gravity's part, but I will say this is true of every skill that can affect a target beyond visible range, not just Suffragium. Damage or buff occurs, but no flinch animation or buff animation occurs on the target. On official clients, it's been like that as long as I can remember.

Might warrant another issue since this now goes beyond Suffragium's functions.

flamefury:
Animation issue will be dealt with in #4975. This can be merged, if it all looks good.

PR #4966, opened 2020-05-17
Corrects Hell's Plant behavior

### LINKED ISSUE #4959: GN_HELL_PLANT
* **rAthena Hash**: 7004fe3

* **Client Date**: 20180621

* **Server Mode**: Renewal

* **Description of Issue**: Hell Plant skill still being removed upon map change.
  * Result: Upon map change hell plant skill unit is being removed however status icon remains on the player.
  * Expected Result: Unit should still be attached on the player?
  * How to Reproduce: Cast hell plant to map A with mobs to see damage and move to different map.
  * Official Information: 
It was discussed here https://github.com/rathena/rathena/pull/4813#issuecomment-623208057
and was changed in https://github.com/rathena/rathena/pull/4813/commits/62822e1949c29b3ac74f9cc7c4c497622029a768
  

* **Modifications that may affect results**: Several PR but should not affect GN_HELL_PLANT.

aleos89:
* **Addressed Issue(s)**: Fixes #4959

* **Server Mode**: Renewal

  * Skill is no longer a unit skill but a status.
  * Adjusts the attack rate timer to official.
  * Skill does not proc cards.
  * Skill does not reflect damage.
Thanks to @Tokeiburu and @ecdarreola!


ecdarreola:
Timer is messed up when changing maps. It becomes negative value.

aleos89:
The animation won’t show now because it requires a newer client to display properly.

ecdarreola:
> The animation won’t show now because it requires a newer client to display properly.

This also applies with the timer?

attackjom:
According to KRO skill description

Hell plant damage formula should be  (100xskill_lv) instead (500xskill_lv) ?

ecdarreola:
Timer goes negative upon changing maps.

Haydrich:
image
while trying hell plant i got this , this is two effect of the skill, is that official ?
and after changing map the skill disappear but the buff icon still works

Issue #4972, opened 2020-05-17
Bonus damage added 4 times


SeravySensei:
https://github.com/rathena/rathena/blob/7004fe3d95610a8311628a53e84201dd8f6f8863/src/map/battle.cpp#L5703

This line adds the bonus damage (example : Fallen Empire on mob lv 100 = +5000 damage) to all ATK stats, as well as to damage in the previous line.
However damage is later calculated as the sum of the 4 different ATK stats for player characters.
As a result, Fallen Empire deals 20000 additional damage instead of the expected 5000. This bug should affect all skills that use battle_calc_skill_constant_addition.

Note that ATKRATE is safe to use this way because percentage additions give the same result regardless of applying individually or to the total sum, but additive bonus does not work that way.

Suggested solution : Add damage to Mastery ATK only in this line, or postpone adding the damage until after the sum of the four ATK stats is made.

Akkarinage:
Why is this issue not following the issue template? The devs put the template there for a reason - removing it wastes people's time trying to find the information in your issue (which in this case isn't even there).

SeravySensei:
Nevermind then, sorry for trying to help.

PR #4974, opened 2020-05-17
Fix bonus damage

SeravySensei:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/4972

* **Server Mode**: 

Renewal. I think Pre-re is unaffected but didn't test.

 Adds the final bonus damage to the total damage directly instead of to the ATK values.


src/map/battle.cpp
```diff
@@ -5697,10 +5697,9 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 		ATK_RATE(wd.damage, wd.damage2, ratio);
 		RE_ALLATK_RATE(&wd, ratio);
 
-		ratio = battle_calc_skill_constant_addition(&wd, src, target, skill_id, skill_lv); // other skill bonuses
+		int bonus_damage = battle_calc_skill_constant_addition(&wd, src, target, skill_id, skill_lv); // other skill bonuses
```
ecdarreola, line 5700: 'initializing': conversion from 'int64' to 'int', possible loss of data

PR #4988, opened 2020-05-20
Fix Reveberation Damage Multihit

attackjom:
* **Addressed Issue(s)**: -

* **Server Mode**: Renewal

 fix attack of Reveberation from multihit to split damage

Level 10 = 2200% atk split to 10 hit.


aleos89:
Moved to #5013.

PR #5004, opened 2020-05-24
add cutins for Friendship Quest

iraciz:
I made this quest in Iro Chaos server 2 years ago, and all the cutins are enabled, 
Lhz friendship quest related cutins, exist since forever in data.grf and was never applied here. 15 years or more without noticing this.

* **Addressed Issue(s)**: 

* **Server Mode**: 


Atemo:
How did you get the cutins? they are missing on rathena but the cutins displayed are differents on my aegis

iraciz:
Here
https://www.youtube.com/watch?v=025ypOHNduA
and here
ect

Atemo:
I would prefer to use the cutin name from the aegis file, do have it?

Atemo:
Thanks for the base !

PR #5006, opened 2020-05-24
Removed account gender

Lemongrass3110:

* **Server Mode**: Both

Adds an upgrade script to change all existing entries to the respective gender in their account.
Cleanup of a few strange design decisions in the character server code.
Removed SEX_ACCOUNT.


src/char/char.cpp
```diff
@@ -1410,32 +1384,53 @@ int char_make_new_char_sql(struct char_session_data* sd, char* name_, int str, i
...
-#if PACKETVER >= 20120307
+	// Check inputs from the client - never trust it!
+
+	// Check the slot
+	if( slot < 0 || slot >= sd->char_slots ){
+		return -4; // invalid slot
+	}
+
+	// Check gender
 #if PACKETVER >= 20151001
```
Daegaladh, line 1395: This packetver check needs to be removed, since now all clients use the switch to work.

PR #5015, opened 2020-05-26
Monster Spawn Update

attackjom:
* **Addressed Issue(s)**: -

* **Server Mode**: Renewal

- KRO has a patch that update map-spawn  of monster, so old Aegis data may incorrect now.
- Update spawning of monster by Navigation's mob data
- If monster has add/remove from maps , I will recheck from KRO sakray ingame.
- If ingame-monster spawn are not same as Navigation monster spawn , I will use old data from rathena.

Some monster had relocate spawn point in KRO and not same as rathena
ex. Elder willow, Coco in north map of Orc village, Dewata dungeon.

This will correct that.


renniw:
Thank you ja ^^

Issue #4734, opened 2020-03-26
Modified Temporal Boots for Soul Reaper & Star Emperor 


reunite-ro:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/652d77eb0287a13fca25d47eb5317de1b538fe87

* **Client Date**: 20180620

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Star Emperor and Soul Reaper cannot wear Modified Temporal Boots
  * Expected Result: from kRO translation, seems like its included with SE and SR too.
 
장착 : 상위 건너계열, 상위 닌자 계열, 상위 권성 계열, 상위 소울링커 계열

  * How to Reproduce:  Try to equip with SE and SR job.

reunite-ro:
For now, from itemdb :

For Modified using job: 0x60000000 (kagerou/oboro+rebellion)
Class:63 (Which include all classes)

For Temporal (normal) using job: 0xFFFFFFFF (All classes)
Class:56 (Which includes all 3rd classes)

in current itemdb, Star Emperor and Soul Reaper can use normal temporal which means they 'behave' like 3rd job classes.

Source : https://github.com/rathena/rathena/wiki/Custom-Items

reunite-ro:
Since Super Novice Expanded also considered as 3rd job, currently (in rAthena) they can wear temporal too.

Just additional information since altenate styles also enabled for this class but currently they didnt have altenate style.

https://github.com/rathena/rathena/issues/4787

reunite-ro:
And also, I think we need to update wiki/docs too regarding new jobs in rathena that added recently for example Soul Reaper/Star Emperor since their job values also missing in it.

https://github.com/rathena/rathena/wiki/Custom-Items

Lemongrass3110:
Should be fixed since item db is YAML now.

Atemo:
The Jobs field for Modified X Boots need to be updated.

mazvi:
now its YAML :D

mazvi:
need update i got translate from kRO https://raw.githubusercontent.com/llchrisll/ROenglishRE/master/Client%20Support/Renewal/2020/System/itemInfo_EN.lua

the translate from kRO 22110 must be used on Star Emperor and Soul Reaper

https://www.divine-pride.net/database/item/22110/

reunite-ro:
There is no structure for StarEmperor and SoulReaper yet right now.

PR #4335, opened 2019-09-03
Converted item database to YAML

aleos89:
* **Addressed Issue(s)**: Fixes #1365

* **Server Mode**: Pre-renewal and Renewal

* **Description of Pull Request**: 
  * Combines item_buyingstore, item_delay, item_flag, item_nouse, item_stack, and item_trade databases into one.
  * General cleanups and optimizations.
  * Includes CSV2YAML conversion tool.
  * Disabled the SQL item database.
Thanks to @Lemongrass3110!

TODO:
- [x] Cleanup csv2yaml tool to parse new item method.
- [x] Fix csv2yaml header output.
- [x] Supports SQL to YAML.
- [x] Look into optimizing MSVS Debug mode to perform faster (compile in release for faster testing).

PR #5022, opened 2020-05-28
Fixed Super Novice can be used Temporal Boots

qwerty7vp:
* **Addressed Issue(s)**: some part of #4734 

* **Server Mode**: Renewal

* **Description of Pull Request**: Fixed Temporal boots jobs required
since Super Novice considered as 3rd Job so they can use these items which they shouldn't


reunite-ro:
just make sure the issue are not closed till SE and SR problems fixed.

Atemo:
Sure, the issue can/will be fixed after https://github.com/rathena/rathena/pull/4335/commits/5c238ebddd82f59764b026072f059e46f92786b3

Atemo:
Thanks

PR #5025, opened 2020-05-29
Whikebain Ears should Ignore 100% MDEF

qwerty7vp:
* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

* **Description of Pull Request**: Fixed item bonus : Ignore MDEF from 25% to 100%
use kRO divine-pride as reference


attackjom:
I think this should follow old data because it follow from Aegis data.  and KRO description didn't mention about How about value of ignoring MDEF.

qwerty7vp:
hmm if it's 25% it should write in description right? (e.g. iRO)
and afaik if it doesnt write how much is it
usually it's 100% or you know how much is it in kRO ?
and how can I know which one should follow old data or not

LordWhiplash:
This item is quite controversal, since kRO/cRO/euRO/idRO/twRO/thRO description seems to say it ignore MDEF completely, while iRO one say specificaly it ignores 25% MDEF, And then, the jRO/bRO item not even mention ignoring MDEF, but adds +3 crit and -50% DEF for the user...
For me makes sense if it ignores the full MDEF for 5 seconds, all other itens i've found specified the % rate, and the one that do not specified it is 100% ignore, it is High Wizard Card (http://kafra.kr/#!/en/KRO/itemdetail/4365)
Remembering that the changes aplied to the nom slotted version should be applied on the sloted one too (ID 19134 for kRO, and ID 5986 for jRO, the jRO version of sloted is not in rathena item_db)

aleos89:
```
	event OnActiveMWeapon:				
		ShowEffect "" EF_STEAL			
		SetIgnoreMdefClass 0 100			
		SetIgnoreMdefClass 1 100			
	return				
```
This is from 15.2.

Issue #5019, opened 2020-05-27
Renamed Pets don't have an Egg with "Beloved" prefix


alisonrag:
* **rAthena Hash**:  Latest (https://github.com/rathena/rathena/commit/7efd696af7da0f46bf291a3ed25509dbb9e164d2)

* **Client Date**:  RagexeRE 2018-04-18a

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Renamed Pets don't have an Egg with "Beloved" prefix
  * Expected Result: Renamed Pets when they back to the Egg, the egg must have the "Beloved" prefix
  * How to Reproduce: 
1 - Create a pet with  `petegg` or capture a pet
2 - Hatch the egg
3 - Rename the pet
4 - Return pet to the egg

  * Official Information:
Unofficial: https://www.novaragnarok.com/wiki/Pet_System#Overview
Official: https://browiki.org/wiki/Mascotes#Alterar_Nome

* **Modifications that may affect results**: 
Need to check if emulator is saving 1 in card[3] when the pet is renamed

Generated a "Beloved" egg using the following code:
in `clif.cpp` function `clif_addcards`:
```
if (item->card[0] == CARD0_PET) { //pet eggs
        buf->card[0] = 1;
        buf->card[1] = 2;
        buf->card[2] = 0;
        buf->card[3] = 1; //Pet renamed flag.
        return;
    }
```

using openkore as base (data parsed from bRO and iRO): 
`buf->card[0] ` must be equals to 1 when EGG
`buf->card[1] + buf->card[2]` must be equals to the PET unique ID (pet table)
`buf->card[3] ` must be equal to 1 if pet is renamed

print of expected result:
![image](https://user-images.githubusercontent.com/10372732/83060254-89905280-a031-11ea-84ad-1ff44be08161.png)

LordWhiplash:
I can confirm this.

![image](https://user-images.githubusercontent.com/62680611/83083284-24058b80-a05c-11ea-956f-f32336a26ff5.png)
(Pet is renamed)

![image](https://user-images.githubusercontent.com/62680611/83083316-38e21f00-a05c-11ea-85d4-dfe0977c513f.png)
(No beloved prefix on egg)

SeravySensei:
I've also noticed the client displays "intimacy : unrecognized" on the item description. Probably related?

Lemongrass3110:
> 
> 
[...]

PR #5040, opened 2020-06-03
Added and fixed some pet features

Lemongrass3110:
* **Addressed Issue(s)**: #5019

* **Server Mode**: Both

* **Description of Pull Request**: Fixes missing "Beloved" flagging and adds support for showing the pet intimacy in the item description.

Thanks to @alisonrag


admkakaroto:
Lemon, has a very old bug with the rename of pets, after renaming a pet the option to put a new name remains active, even after you have applied the new name, it is not possible to change the name again but the option remains active, this it's pretty silly, but is it possible to fix it?

Lemongrass3110:
> 
> 
> Lemon, has a very old bug with the rename of pets, after renaming a pet the option to put a new name remains active, even after you have applied the new name, it is not possible to change the name again but the option remains active, this it's pretty silly, but is it possible to fix it?
> 
> 

@admkakaroto this is how it is on official servers too sadly and cannot be fixed.

Issue #4975, opened 2020-05-17
Priest self-AoE buff ranges are off and animations shouldn't play on allies when ally is out of sight


flamefury:
* **rAthena Hash**: 85b04cd72a1ce05d58b8de14ea5462d048688f73
* **Client Date**: 2018-04-18
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: Range of Magnificat and Gloria appear too short. Also, when Suffragium and Impositio Manus is used, allies within range of the skills but outside of the viewing range shouldn't play their skill or cast animations. rAthena test one cell out of range: https://streamable.com/weezsr
  * Expected Result: Magnificat and Gloria should reach up to 18 cells away (37x37, same as Impositio Manus and Suffragium). Also, no animations should play on allies outside of the range.
  * How to Reproduce: Set `area_size` in `conf/battle/client.conf` to 17 or lower, have a character outside of visible range, and use the aforementioned skills.
  * Official Information:
kRO Sakray test of skills: https://streamable.com/95msez
Magnificat, Gloria, Impositio Manus and Suffragium all trigger even though the AB can't be seen, and the Warlock performs no animations upon receiving the buffs.
Oddly, this is not true for any of the Arch Bishop buff skills, as Clementia, Canto Candidus, etc. all perform animations on the ally who is affected by it, even if the AB cannot be seen on screen. This may potentially mean the Priest skills are bugged on Sakray.

* **Modifications that may affect results**: 
area_size in client.conf was set to 14 to match official.

flamefury:
Angelus also suffers from the same issues.

PR #5056, opened 2020-06-06
Resolve Aco/Priest self-AoE skill ranges and animation behaviour for partymates

flamefury:
* **Addressed Issue(s)**: #4975
* **Server Mode**: Renewal

Sets AoE size for Angelus, Gloria, Magnificat, Impositio Manus to 18 to match kRO Sakray tests.
Fixes Impositio Manus SP consumption per level to match kRO Sakray.

Check if party member is within visible range before sending `clif_skill_nodamage`, otherwise, just do an `sc_start` instead for Angelus, Gloria, Magnificat, Impositio Manus and Suffragium.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -7657,14 +7657,25 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		}
 		status_damage(src, src, sstatus->max_hp,0,0,1, skill_id);
 		break;
-
 	case AL_ANGELUS:
 #ifdef RENEWAL
-	case MC_LOUD:
 	case PR_SUFFRAGIUM:
+	case PR_IMPOSITIO:
 #endif
 	case PR_MAGNIFICAT:
 	case PR_GLORIA:
+		if (sd == NULL || sd->status.party_id == 0 || (flag & 1))
+			if (check_distance_bl(src, bl, AREA_SIZE))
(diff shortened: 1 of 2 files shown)


src/map/skill.cpp
```diff
@@ -7657,14 +7657,25 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
-
 	case AL_ANGELUS:
 #ifdef RENEWAL
-	case MC_LOUD:
 	case PR_SUFFRAGIUM:
+	case PR_IMPOSITIO:
 #endif
 	case PR_MAGNIFICAT:
 	case PR_GLORIA:
+		if (sd == NULL || sd->status.party_id == 0 || (flag & 1))
```
aleos89, line 7667: I would just add curly brackets here for the main if.

src/map/skill.cpp
```diff
@@ -7657,14 +7657,25 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			if (check_distance_bl(src, bl, AREA_SIZE))
+				clif_skill_nodamage(bl, bl, skill_id, skill_lv, sc_start(src, bl, type, 100, skill_lv, skill_get_time(skill_id, skill_lv)));
```
mrjnumber1, line 7669: would be better to write the sc_start once (even though it's unlikely to change)

PR #5070, opened 2020-06-08
Cleaned up chlogif_parse functions

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Cleaned up unused or potentially invalid passing sd's in chlogif_parse functions

### DIFF
--- a/src/char/char_logif.cpp
+++ b/src/char/char_logif.cpp
@@ -179,11 +179,11 @@ void chlogif_prepsend_global_accreg(void) {
 }
 
 void chlogif_send_global_accreg(const char *key, unsigned int index, int64 int_value, const char* string_value, bool is_string) {
-	int nlen = WFIFOW(login_fd, 2);
-	size_t len;
-
 	if (!chlogif_isconnected())
 		return;
+
+	int nlen = WFIFOW(login_fd, 2);
+	size_t len;
 
 	len = strlen(key)+1;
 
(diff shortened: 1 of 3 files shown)


src/char/char_logif.cpp
```diff
@@ -295,7 +295,8 @@ int chlogif_parse_ackconnect(int fd, struct char_session_data* sd){
 	return 1;
 }
 
-int chlogif_parse_ackaccreq(int fd, struct char_session_data* sd){
+int chlogif_parse_ackaccreq(int fd){
+	struct char_session_data* sd = NULL;
```
aleos89, line 299: This define can be moved down to below the `RFIFOSKIP(fd,21);` line.

src/char/char_logif.cpp
```diff
@@ -332,8 +333,9 @@ int chlogif_parse_ackaccreq(int fd, struct char_session_data* sd){
  * Receive account data from login-server
  * AH 0x2717 <aid>.L <email>.40B <expiration_time>.L <group_id>.B <birthdate>.11B <pincode>.5B <pincode_change>.L <isvip>.B <char_vip>.B <char_billing>.B
  **/
-int chlogif_parse_reqaccdata(int fd, struct char_session_data* sd){
+int chlogif_parse_reqaccdata(int fd){
 	int u_fd; //user fd
+	struct char_session_data* sd = NULL;
```
aleos89, line 338: Same for this, it can be defined after the `RFIFOREST(fd) < 75` check

src/char/char_logif.cpp
```diff
@@ -566,7 +568,7 @@ int chlogif_parse_askkick(int fd, struct char_session_data* sd){
 			}
 			else
 			{// Manual kick from char server.
-				struct char_session_data *tsd;
+				struct char_session_data *tsd = NULL;
```
aleos89, line 571: This doesn't have to be initialized to null because we do direct assignments in the ARR_FIND with no pre or post checks.

PR #5093, opened 2020-06-13
Web token feature

secretdataz:
* **Addressed Issue(s)**: 
* **Server Mode**: Irrelevant
The login server will now send randomized authentication token with AC_ACCEPT_LOGIN3 packet for use with various web services recent clients use.
Please note that this implementation does not use a secure random string generator to generate the token.


cahya1992:
just want to know this pull request have related with new function for emblem guild?

LolyAll:
Let's face it, this PR is useless.

Lemongrass3110:
> 
> 
> Let's face it, this PR is useless.

@LolyAll any reason for a comment like this?
This is required for some of the 2020 client features. So personally I would not consider it useless.

attackjom:
> Let's face it, this PR is useless.

without this  you can't use guild emblem in 2020 clients.

makuexile:
still can't use guild emblem after update my database using 20200401 client.

secretdataz:
Guild emblem requires an external REST service. This PR is just a part that rAthena need to handle.

Issue #5155, opened 2020-06-28
WOE Controller Reward Issue


mazvi:
* **rAthena Hash**: 005a2c4

* **Client Date**: 20180620

* **Server Mode**: RENEWAL

* **Description of Issue**: 

use option 1 | 2 | 8 

  * Result:  Offline players also get prizes,
  * Expected Result:  offline players shouldn't get prizes, and duplicate IP players only get 1.
  * How to Reproduce: 
  * Official Information:
  

* **Modifications that may affect results**:

Atemo:
```
//  [2] Mail all rewards.
//      - If not set, players receive items in their inventory.
//      - Only ONE item can be sent via mail for PACKETVER < 20150513 while later clients are limited to MAIL_MAX_ITEM (5).
//      - Note that offline players do NOT receive rewards.
```
The last line should be read
```
//      - If not set, note that offline players do NOT receive rewards.
```

>and duplicate IP players only get 1

It's not the behaviour ? `Members in a guild with the same IP address are not rewarded` = 1 reward per IP

mazvi:
@Atemo sorry i wrong to write issue 

this is the issue and i was edit my post

  * Result:  Offline players also get prizes,
  * Expected Result:  offline players shouldn't get prizes, and duplicate IP players only get 1.

Atemo:
`Offline players also get prizes` it was expected.
However I will suggest an new option

mazvi:
This is new issue on latest woe update. If using option 1 2 8 16 if online player on teleport or warp loading screen this player not get reward

PR #5198, opened 2020-07-09
Custom woe controller script update

Atemo:
* **Addressed Issue(s)**: #5155 

* **Server Mode**: -

Added the setting '[16] Mail online player only.' to only reward player online by mail.
Updated the docs.


mazvi:
Option 16 can restrict duplicated ip?

Atemo:
Option 16 can be used with option 8 (Duplicate IP check), if it's what you mean.

Issue #5196, opened 2020-07-09
Clan System - Crashes


ecdarreola:
* **rAthena Hash**: 7457e32

* **Client Date**: 20200401 or 20200304

* **Server Mode**: Renewal

* **Description of Issue**: when I use clan chat it appears twice on the chat.
  * Result: see image below.
![image](https://user-images.githubusercontent.com/48939874/87055228-a629c800-c236-11ea-9d17-cbd1364b2638.png)

  * Expected Result: only one line of the message should appear.
  * How to Reproduce: Join clan -> use /cl <message> -> observe chat box/log.
As for the random crash, I'm not able to fully reproduce it but here's how I crashed the server using windows debug.
Created a new character -> spam /cl test -> crash server.
  * Official Information: Not sure but afaik message should only appear once.
  

* **Modifications that may affect results**: #5192 #5141 #3640 
  
  

Not sure if this is related to my other report #5194

ecdarreola:
Additional findings, if you enter a map thru portal the clan line message adds + 1. (see image below)

![image](https://user-images.githubusercontent.com/48939874/87055762-58618f80-c237-11ea-9b9d-759e19fafdf3.png)

Badarosk0:
Have you made any changes to the src? It looks like it's normal for me.
![screenRock Beta000](https://user-images.githubusercontent.com/13624462/87067424-663cf380-c1ea-11ea-81e6-ab9a853eab09.jpg)

ecdarreola:
@Badarosk0 no changes done to the source except for the PR mentioned above.

Haydrich:
i can confirm it
How to Reproduce :
1- @warp prt_in 41 112
2- join any clan
3- 
![image](https://user-images.githubusercontent.com/54556365/87074767-a36fa880-c21f-11ea-8cf0-8c15e41382ab.png)
allow clan message appear in the chat
4- /cl HI
![image](https://user-images.githubusercontent.com/54556365/87074798-ad91a700-c21f-11ea-8cbd-8cc2b419db6e.png)

ecdarreola:
Additional findings: Clan basic info -> number of members keeps increasing upon changing map.

PR #5202, opened 2020-07-11
Fixed a few possible crashes

Lemongrass3110:
* **Addressed Issue(s)**: #5196

* **Server Mode**: Both

Adds a few missing session checks that might lead to a crash if the session index is negative.

Thanks to @ecdarreola


ecdarreola:
Works perfect. No crashes for 2 days now. Thank you!

Issue #5203, opened 2020-07-11
Clan Online Count


ecdarreola:
* **rAthena Hash**: d3d864d

* **Client Date**: 20200401 and 20200304

* **Server Mode**: Renewal

* **Description of Issue**: Clan count increments every after map change when viewing clan info using (ctrl+g)
  * Result: Number of members of clan can reach 100 or 1000+ in count if you keep warping to different map.
  * Expected Result: Should not increase
  * How to Reproduce: Join any clan and from the building where you have join just keep entering the portal and notice the number of members go up. (see attached yt link https://youtu.be/Ijc0Ow1egT4)
  * Official Information:
  

* **Modifications that may affect results**: None

PR #5204, opened 2020-07-11
Fixed some issue with clif_loadConfirm

Lemongrass3110:
* **Addressed Issue(s)**: #5203, #5278 and #5376

* **Server Mode**: Both

Fixed an invalid preprocessor check.
Added a missing check if a player is already in a clan, because the function is called on each map load on newer clients.

Thanks to @ecdarreola

### DIFF
--- a/src/map/clan.cpp
+++ b/src/map/clan.cpp
@@ -111,12 +111,16 @@ int clan_getNextFreeMemberIndex( struct clan* clan ){
 }
 
 void clan_member_joined( struct map_session_data* sd ){
-	struct clan* clan;
-	int index;
-
 	nullpo_retv(sd);
 
-	clan = clan_search(sd->status.clan_id);
+	if( sd->clan != nullptr ){
+		clif_clan_basicinfo( sd );
+		clif_clan_onlinecount( sd->clan );
+		return;
+	}
+
+	struct clan* clan = clan_search(sd->status.clan_id);
+	int index;
 
 	nullpo_retv(clan);
 
@@ -129,9 +133,6 @@ void clan_member_joined( struct map_session_data* sd ){
 
 		intif_clan_member_joined(clan->id);
 		clif_clan_onlinecount(clan);
-
-		if (clan->instance_id > 0)
-			instance_reqinfo(sd, clan->instance_id);
 	}
 }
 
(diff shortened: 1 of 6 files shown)


src/map/pc.cpp
```diff
@@ -1801,7 +1801,7 @@ void pc_reg_received(struct map_session_data *sd)
 		}
 	}
 
-#if PACKETVER_MAIN_NUM < 20190403 || PACKETVER_RE_NUM < 20190320 || PACKETVER_ZERO_NUM < 20190410
+#if !( PACKETVER_MAIN_NUM >= 20190403 || PACKETVER_RE_NUM >= 20190320 || PACKETVER_ZERO_NUM >= 20190410 )
```
ecdarreola, line 1804: This line prevents guild info to show on newer clients. It will load once but upon character relog you will not be able to show guild info.

ecdarreola:
Working great for the clan count now showing correct numbers.

Lemongrass3110:
This still has some issues that need to be fixed. Seems like moving the calls over to loadConfirm created some problems, which were not triggered because of the invalid ifdef so far...

admkakaroto:
Here a Little bug:

https://youtu.be/BssNQjeV3Lg

Sometimes Guild Window opens, but after relogging with the character, the window does not open again for him.

Badarosk0:
> Here a Little bug:
> 
> https://youtu.be/BssNQjeV3Lg
> 
> Sometimes Guild Window opens, but after relogging with the character, the window does not open again for him.

I can confirm this bug.

vstumpf:
This probably will also fix #5376 and #5394, if duplicates keep getting added to the clan member list, the list will start pointing to bad pointers.

admkakaroto:
> This probably will also fix #5376 and #5394, if duplicates keep getting added to the clan member list, the list will start pointing to bad pointers.

Yes but, this pr are creating others bugs. So cannot be used.

src/map/clan.cpp
```diff
@@ -116,6 +116,12 @@ void clan_member_joined( struct map_session_data* sd ){
 
 	nullpo_retv(sd);
 
+	if( sd->clan != nullptr ){
```
lighta, line 119: is it ok to have this before the clan lookup ?
```
clan = clan_search(sd->status.clan_id);
nullpo_retv(clan);
```
it seem we are sending data to client which might be invalid, this look a bit incoherent for me

RagnaWay:
Update!

aleos89 (approved):

Issue #5189, opened 2020-07-07
MySQL function SHA2 does not exist


bgamez23:
* **rAthena Hash**: ff38cf70d2e6c73523f5809cbdc8fc8f71292383

* **Client Date**: 

* **Server Mode**: Pre-Renewal

* **Description of Issue**: i got this error [SQL]: DB error - FUNCTION ragnarok.SHA2 does not exist on centos but does not exist in windows server.
  * Result: [SQL]: DB error - FUNCTION ragnarok.SHA2 does not exist
[Debug]: at account.cpp:664 - UPDATE `login` SET `web_auth_token` = LEFT( SHA2( CONCAT( UUID(), RAND() ), 256 ), 16 ), `web_auth_token_enabled` = '1' WHERE `account_id` = '2000000'. it also happens on another account who login the game.

  * Expected Result: 
  * How to Reproduce: i don't know how to produce since it only appears when i login my character.
  * Official Information:
  

* **Modifications that may affect results**: pre-renewal setup server and vip enabled

secretdataz:
What's your CentOS version?

bgamez23:
Cent0S release 6.10 (Final)

bgamez23:
![image](https://user-images.githubusercontent.com/1854368/86787138-5b257e80-c097-11ea-8ee2-9c2d6eec2ddd.png)

aleos89:
What SQL product are you using and it's version?

bgamez23:
![image](https://user-images.githubusercontent.com/1854368/86928176-ca1dd880-c166-11ea-96ed-6edff02f1f78.png)

aleos89:
You need to update to 5.5.6 or later.

LiamKarlMitchell:
What about doing the crypt in code rather than the DB.
I've been curious to implement bcrypt

Lemongrass3110:
> What about doing the crypt in code rather than the DB.
> I've been curious to implement bcrypt

@LiamKarlMitchell he can simply leave out the hashing or change the sql command. It was just added because there were security concerns for only using UUID. Basically doing it in code would drag some big dependencies into rAthena, which we want to prevent if we can.

Badarosk0:
> * **rAthena Hash**: [ff38cf7](https://github.com/rathena/rathena/commit/ff38cf70d2e6c73523f5809cbdc8fc8f71292383)
> 
> * **Client Date**:
> 
> * **Server Mode**: Pre-Renewal
> 
[...]

PR #5205, opened 2020-07-11
Improved web authentication token generation

Lemongrass3110:
* **Addressed Issue(s)**: #5189

* **Server Mode**: Both

Improved the web authentication token generation by adding checks for the existence of the SHA2 or MD5 hashing function.
If none of both is available a warning is issued and no hashing function will be used.
Additionally we issue a warning that tells the user to consider upgrading his MySQL version.

Thanks to @bgamez23 and @Badarosk0


Lemongrass3110:
> 
> 
> ./configure --enable-packetver=20180620 --enable-64bit --with-MYSQL_LIBS=/usr/lib64/libmysqlclient.so
> 
> How if i use config like that? Are safe using this commit?

@mazvi your configure flags do not have any affect on this.

Issue #5058, opened 2020-06-07
delete_timer error : function mismatch 00E251C0


admkakaroto:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/8efd61f9b27dd5163d8f71f55f2025470f1254b3
* **Client Date**: 20200401
* **Server Mode**: RE
* **Description of Issue**: If you die while using the 'Rolling Cuter' skill, this error message is displayed on the map_server.
  * How to Reproduce: Whenever you die using the 'Rolling Cuter' skill, this message will appear on the map_server.

```
[Error]: delete_timer error : function mismatch 00E251C0(unknown timer function) != 00E8CE00(unknown timer function)
[Error]: delete_timer error : function mismatch 00E251C0(unknown timer function) != 00E8CE00(unknown timer function)
[Error]: delete_timer error : function mismatch 00E251C0(unknown timer function) != 00E8CE00(unknown timer function)
[Error]: delete_timer error : function mismatch 00E251C0(unknown timer function) != 00E8CE00(unknown timer function)
[Error]: delete_timer error : function mismatch 00EE6CD0(unit_walktoxy_timer) != 00E8CE00(unknown timer function)
[Error]: delete_timer error : function mismatch 00DAC6C0(npc_timerevent) != 00E8CE00(unknown timer function)
[Error]: delete_timer error : function mismatch 00EE6CD0(unit_walktoxy_timer) != 00E8CE00(unknown timer function)
[Error]: delete_timer error : function mismatch 00E251C0(unknown timer function) != 00E8CE00(unknown timer function)
```

aleos89:
I'm unable to reproduce this. Are there any extra steps? Maybe a video would help.

admkakaroto:
> I'm unable to reproduce this. Are there any extra steps? Maybe a video would help.

go to any map that has monsters or even pvp, put the 'Rolling Cross' skill on the hotkey, hold the hotkey button and wait for some monsters to kill you while you are using the skill.

admkakaroto:
Here, test video: https://www.youtube.com/watch?v=SMBY-I7p-To

qwerty7vp:
![asdqf](https://user-images.githubusercontent.com/64515036/84459469-baa78e80-ac91-11ea-964f-a473b583da60.gif)

I'm unable to reproduce this too

admkakaroto:
This have no sense, i will redownload and test using a clean rA, just a min.

admkakaroto:
[...]

PR #5206, opened 2020-07-11
Cancel skill usage keeping on death

Lemongrass3110:
* **Addressed Issue(s)**: #5058

* **Server Mode**: Both

Skill usage keeping will be cancelled properly on death now.
Additionally added the function to the known timer functions, so that the correct name will be displayed in future errors.

Thanks to @admkakaroto and @qwerty7vp

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -8323,6 +8323,11 @@ int pc_dead(struct map_session_data *sd,struct block_list *src)
 			duel_reject(sd->duel_invite, sd);
 	}
 
+	if( sd->skill_keep_using.tid != INVALID_TIMER ){
+		delete_timer( sd->skill_keep_using.tid, skill_keep_using );
+		sd->skill_keep_using.tid = INVALID_TIMER;
+	}
+
 	pc_close_npc(sd,2); //close npc if we were using one
 
 	/* e.g. not killed thru pc_damage */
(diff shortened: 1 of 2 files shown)


admkakaroto:
@Lemongrass3110 the problem can still happen if your SP runs out while performing the skill. To reproduce is very simple, hold the Rolling Cuter's hotkey and keep attacking a monster until your SP ends, keep holding the Hotkey.

Masao87:
Please review this issue again, this keeps on happening and now crashes the map-server as well including a big server rollback @Lemongrass3110 .

unknown

Lemongrass3110:
@Masao87 has already been fixed by @Daegaladh in 2d164cb

Masao87:
@Lemongrass3110 We had this fix by @Daegaladh already added to our code at the time this error is occuring on our server with the latest rA hash.

So I'm not sure if his fix doesn't work to 100% or this initial implementation of yours, but somewhere something is still not working it seems and still causing this issue to happen as you can see on the screen I attached above.

admkakaroto:
> @Masao87 has already been fixed by @Daegaladh in [2d164cb](https://github.com/rathena/rathena/commit/2d164cb8c89268cb57e658a97658441fe8990758)

this still happens if you are holding the hotkey when your SP reaches its end.

Issue #5216, opened 2020-07-14
Non-Instance maps's name changed when instancing them


inhyositsu:
* **rAthena Hash**: Latest

* **Client Date**: 2018

* **Server Mode**: Both

* **Description of Issue**: 
Source maps which aren't in the Instance System(ones without '@' in the map name), its name will be changed mistakenly and forever until server ends when creating new Instance Map for them.

  * Result: 
  * Expected Result: Source map's name shouldn't be changed at all
  * How to Reproduce: create an Instance that contains a non-instance map an enter them.
  * Official Information:
  

* **Modifications that may affect results**:

PR #5217, opened 2020-07-14
Fix the problem that changing source map's name when instancing new o…

inhyositsu:
…ne which map name's length is longer than 8

* **Addressed Issue(s)**: #5216

* **Server Mode**: 
Both

[Issue]
Source maps which aren't in the Instance System(ones without '@' in the map name), its name will be changed mistakenly and forever until server ends when creating new Instance Map for them.

### DIFF
--- a/src/map/instance.cpp
+++ b/src/map/instance.cpp
@@ -632,6 +632,28 @@ int instance_create(int owner_id, const char *name, e_instance_mode mode) {
 }
 
 /**
+ * Generating map name for instances
+ * @param instance_id: Instance ID to add map to
+ * @param dst_name: pointer to put the new name to
+ * @param src_name: source map's name
+ */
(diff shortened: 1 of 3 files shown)


aleos89:
Not sure I fully understand here. I don't see any resolution into the issue you stated aside from removing duplicate code.

inhyositsu:
Actually the two codes in the original version are not the same, so i didn't just remove the duplicating ones.

aleos89:
Can you explain or show me where they aren't the same? I'm just trying to review it and it looks like nothing changed to me.

Lemongrass3110:
Also please let us know the (old) line where the real issue was. I do not see any modification of any source map as well.

inhyositsu:
I'm sorry I just found this issue has been fixed in git 15c954 and my version is older than that and I didn't aware that this bug has been corrected. So there's no need to pull this request. Thank you.

PR #5236, opened 2020-07-21
Fixes NPC_RUN behavior

### LINKED ISSUE #3622: NPC_RUN don`t work
* **rAthena Hash**: 84b4f09

* **Client Date**: 

* **Server Mode**: Pre-Re/Re

* **Description of Issue**: monsters when using the skill (NPC_RUN) does not run away from the target.
For example monster Antonio (id 1247).
  * Result: When skill used it only show HELP emotion and don`t run away from target.
  * Expected Result: When monsters use skill NPC_RUN they must run away from the target (players)
  * How to Reproduce: 
  * Official Information: Skill must work like at that video from kRo (Skill id: 354)
[YouTube link](https://youtu.be/Z9NTpwdCmy8) - Halloween Event 2018

  

### LINKED ISSUE #5220: Sphere Mine not moving, issue with NPC_RUN
* **rAthena Hash**: c52b3b784a7f958050e4f3186b001382151bd648 (not the latest, but at least parts of the code are still part of the latest revision code as well)

* **Client Date**: 2018-06-21

* **Description of Issue**: 
  * Result:  Sphere Mine does not move upon being hit. It simply stays in its current position, shouts /hlp and explodes.
  * Expected Result:  When hitting the Sphere Mine, it should start moving in the direction at which you hit it, for the amount of cells equal to the skill level used.
  * How to Reproduce: 

Spawn a Sphere Mine, hit it and see. There are a couple of reasons for this that we found through debugging.
NPC_RUN calls `unit_escape` which calls `unit_walktoxy(bl, bl->x + dist*dirx[dir], bl->y + dist*diry[dir], 0)`. This means the Flag in unit_walktoxy is set to 0 while it should be set to 2 for `&2: Force walking (override can_move)`. This was done since commit 81ca299d913305da56be82bbe49a02a0c12c273d.

In addition, according to commit 6ef71063cbc98b61fcae8eb05cd9a22e6b87be9c it should walk the amount of cells that's equal to the skill level of NPC_RUN used, not some random amount of cells like called in `unit_escape`. Not entirely sure about this, but most sources seem to point this way. Herc also does it based on skill level: https://github.com/HerculesWS/Hercules/blob/stable/src/map/skill.c#L8464

A few other notes: when we set the flag to 2 properly and when your ASPD is high the moment you hit the Sphere Mine, it starts to walk the moment you start the attack, but when your attack lands, the walking of the Sphere Mine stops. When your ASPD is low however, even when you hit the Sphere Mine, it will continue walking as intended.
Additionally, when it explodes, the drops of the Sphere Mine appear on the original spawn location, not on the location at which it exploded.

Some of this could also be related to issue https://github.com/rathena/rathena/issues/3622

  * Official Information: Pretty much every resource about the skill, as well as the in-game skill description.
  

* **Modifications that may affect results**:  Quite a lot, but none relevant to this part of the code.

aleos89:
* **Addressed Issue(s)**: Fixes #3622 and fixes #5220

  * Monsters now properly escape from the attacker.
  * NPC_RUN should determine distance based on skill level > 1.
  * Resolves Sphere Mine when summoned by an Alchemist not escaping.
Thanks to @Balferian, @Indigo000, and @Daegaladh!


Indigo000:
Tested both NPC_RUN on monsters as well as Sphere Mine. Both seem to work properly on my end.

Daegaladh:
got an error on a monster using NPC_RUN
```[Error]: unit_walk_timer mismatch 140 != 153```

Indigo000:
> 
> 
> got an error on a monster using NPC_RUN
> `[Error]: unit_walk_timer mismatch 140 != 153`

Was about to post this. Just got these as well. A whole lot of them in fact.

Indigo000:
No more timer mismatches on my side after the latest update.

Issue #5064, opened 2020-06-08
Round Trip caused mobs to stuck after killed


qwerty7vp:
* **rAthena Hash**: d87ac21

* **Client Date**: 20200401

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: 
Aggressive mobs stuck after killed with Round Trip, I tried with Desperado many times and mobs wont stuck
  * Expected Result: 
Mobs shouldnt stuck after killed
  * How to Reproduce: 
Kill Aggressive mobs with Round Trip ( I also got additional info from Joam )
I found it mostly happens with Aggressive mobs

![image](https://user-images.githubusercontent.com/64515036/83988420-5a50de00-a96d-11ea-9f4d-708a4b74b9b0.png)

Round Trip :

![dodo01](https://user-images.githubusercontent.com/64515036/83988627-00044d00-a96e-11ea-97af-b346b44e2bb6.gif)

Desperado :

![dodo02](https://user-images.githubusercontent.com/64515036/83988797-989acd00-a96e-11ea-9baa-ec6fe37ed427.gif)

  * Official Information:
  

* **Modifications that may affect results**:

Daegaladh:
Before the packet update this bug rarely happened, but after that, it happens a lot :/

Hope8888:
20200304 Client
@aleos89 After this modification, this bug will not happen, for me

![image](https://user-images.githubusercontent.com/19237837/88615772-2a000300-d0c5-11ea-8b01-7929b7912461.png)

Lemongrass3110:
Related issue: HerculesWS/Hercules#2109

Jammo13:
Another note here, the more people close to each other or in a party on a single map the more frequently mobs get stuck on map. Tested with 8 people in party and only need about 5 mobs around the party that are being killed and one of them always gets stuck.

PR #5263, opened 2020-07-30
Fixed dead monsters stuck after killed

Daegaladh:
*Thanks to @Hope8888

* **Addressed Issue(s)**: #5064 

* **Server Mode**: Both

* **Description of Pull Request**:  Avoids monsters moving after dead so they don't get stuck on screen


Daegaladh:
Still happens on some cases (when the monster dies while you're moving and the monsters gets momentarily out of sight range), we'll need to dig deeper and find the ultimate cause

saya9200:
I just tested this. Before applying the pull request, it happened very very often. Now it doesnt happen at all to me.

Daegaladh:
> I just tested this. Before applying the pull request, it happened very very often. Now it doesnt happen at all to me.

Unfortunately it causes some undesirable side effects (monsters stopping and other weird stuff), we need to check this thoroughly

lighta:
You could try to reduce this check only for mob and not al the BL, just to see if that reduce the side effect.

Daegaladh:
> 
> 
> You could try to reduce this check only for mob and not al the BL, just to see if that reduce the side effect.

They still don't move correctly :/

KrokusPokus:
Not sure if it makes any difference, but I have been using a similar check in clif_move() for a  while now and haven't noticed anything weird. Can you confirm that this causes the same side effects for you? (And what's the best way to trigger the side effects?)

```
void clif_move(struct unit_data *ud)
{
	unsigned char buf[16];
	struct view_data* vd;
	struct block_list* bl = ud->bl;
	struct status_change *sc = NULL;

/*TEST*/	struct status_data *status;
/*TEST*/	status = status_get_status_data(bl);
/*TEST*/	if (status->hp == 0)
/*TEST*/		return;

	vd = status_get_viewdata(bl);
...

```

attackjom:
this fix from Hercules https://github.com/HerculesWS/Hercules/commit/c04d79112575bc4e756231f8b447378dbab951ad

can solve the issue "when the monster dies while you're moving and the monsters gets momentarily out of sight range".

Playtester (approved):

Daraen1 (approved):
It seems to be good for me, work fine ;)

GG!

aleos89 (approved):

PR #5267, opened 2020-07-30
Fixes Gate of Hell behavior

### LINKED ISSUE #3259: Gate of Hell
* **rAthena Hash**: f196304
* **Client Date**: Any
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: Gate of Hell's damage doesn't match kRO.
  * Expected Result: Matches kRO.
  * How to Reproduce: 
  * Official Information: The tests below were done on kRO Sakray server.

### SP consumption issue
* On kRO, Gate of Hell requires 30% of your SP to be casted on **any skill level**. On rAthena, it requires 20% + 20 SP and it varies depending of your skill level.
* On kRO, the skill does not consume a fixed SP value, it consumes 11% ~ 20% of your SP and nothing else (even though it does indeed display Sp : 30 on the skill bar, that's probably meant as the 30% SP requirement). On rAthena, 11 ~ 20 SP is consumed when casting the skill and then a 11% ~ 20% SP is consumed.

### SP damage bonus
* The formula currently used is ```ATK_ADD(wd.damage, wd.damage2, (sstatus->sp * (1 + skill_lv * 2 / 10)) + 10 * status_get_lv(src));```. This formula only works for skill_lv 5 and 10:
```
1 + 1 * 2 / 10 = 1 (should be 1.2)
1 + 2 * 2 / 10 = 1 (should be 1.4)
1 + 3 * 2 / 10 = 1 (should be 1.6)
1 + 4 * 2 / 10 = 1 (should be 1.8)
1 + 5 * 2 / 10 = 2
...
```
... you get the idea! This is not intended and I tested it on kRO as well.

### Long range modifiers
* Gate of Hell's damage is not reduced by long range modifiers (long_attack_def_rate).

### The HP/SP damage part
* On kRO, this damage is not affected by card modifiers (but it is reduced by def). Currently on rAthena, the damage is combined to wd.damage and then a battle_calc_cardfix is applied on the whole thing.

### Card modifiers
(The tests below were done without a weapon to remove the variance.)
image

image image

This one is tad more complex. Card fixes are applied before the damage parts are combined (so before `wd.damage = wd.statusAtk + wd.weaponAtk + wd.equipAtk + wd.masteryAtk;`) because each one is treated differently.
* The statusAtk and the masteryAtk are not affected by elemental modifers (such as Deviling). They have the 'NK_NO_ELEFIX' flag. Currently on rAthena, the modifiers are applied on the whole damage instead.
* The weaponAtk and the equipAtk are affected by elemental modifiers (which is the current behavior on rAthena). These tests were not shown on screenshots though.

I believe that covers most of it! I do have plenty of screenshots (or could take more), but there would be too many to post. If one's required, I don't mind adding them though.

aleos89:
* **Addressed Issue(s)**: Fixes #3259

* **Server Mode**: Renewal

  * Fixes SP damage bonus for levels 1-4 being too low caused by a rounding issue.
  * Gate of Hell is no longer reduced by long range modifiers.
  * Fixes the HP/SP damage for statusAtk and masteryAtk to properly apply the NK_IGNOREELEMENT flag when checking target players. This subsequently applies to all skills.
Thanks to @Tokeiburu and @exneval!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -723,7 +723,7 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 				//It was discovered that ranged defense also counts vs magic! [Skotlex]
 				if( flag&BF_SHORT )
 					cardfix = cardfix * (100 - tsd->bonus.near_attack_def_rate) / 100;
-				else
+				else if (!nk[NK_IGNORELONGCARD])
 					cardfix = cardfix * (100 - tsd->bonus.long_attack_def_rate) / 100;
 #endif
 				cardfix = cardfix * (100 - tsd->bonus.magic_def_rate) / 100;
@@ -929,7 +929,7 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 				}
 				if( flag&BF_SHORT )
 					cardfix = cardfix * (100 - tsd->bonus.near_attack_def_rate) / 100;
-				else if (battle_getcurrentskill(src) != SR_GATEOFHELL)	// BF_LONG (there's no other choice)
+				else if (!nk[NK_IGNORELONGCARD])	// BF_LONG (there's no other choice)
 					cardfix = cardfix * (100 - tsd->bonus.long_attack_def_rate) / 100;
 				if( tsd->sc.data[SC_DEF_RATE] )
 					cardfix = cardfix * (100 - tsd->sc.data[SC_DEF_RATE]->val1) / 100;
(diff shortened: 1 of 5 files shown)


src/map/battle.cpp
```diff
@@ -929,7 +929,7 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 				}
 				if( flag&BF_SHORT )
 					cardfix = cardfix * (100 - tsd->bonus.near_attack_def_rate) / 100;
-				else	// BF_LONG (there's no other choice)
+				else if (battle_getcurrentskill(src) != SR_GATEOFHELL)	// BF_LONG (there's no other choice)
```
lighta, line 932: I don't like the notion of skill_check in here, why do we need this ?
If anything I'd rather have a BF_IGNORE_LONG, but honestly I don't really get it.

sawadaza (commented):
I want to use it in skill_db.txt help me

PR #5280, opened 2020-08-03
Corrects status icon timer after relog

### LINKED ISSUE #4984: Relogin make Negative timer on status associated status_change_timer.
* **rAthena Hash**:  https://github.com/rathena/rathena/commit/d9f533165ea731a3b624d87b8bf95aa671b140fb

* **Client Date**: 2018-06-21ragexeRE

* **Server Mode**: Renewal (Maybe Pre-re too ?)

* **Description of Issue**: 
- When relogin , skill that have effect (every A second. -> do B) will have negatives times left in status icon.
  * Result: 
- Those statuses will have negatives time left when relogin.
  * Expected Result: 
- It should show properly time left of statuses.
  * How to Reproduce: 
Already Test.
- Poisoning weapon + Toxin , Death Hurt
- HelpAngel
- Small / Medium Life potion
- Soul Unity

  * Official Information:
  
Test in KRO
- I ever use Poisoning Weapon + Toxin  it show timer left properly (After GX rebalance)

aleos89:
* **Addressed Issue(s)**: Fixes #4984

  * Resolves status icon timers not properly showing the difference of remaining/total time when a player relogs.
  * Fixes status icon timers going into negative values.
Thanks to @attackjom!

### DIFF
--- a/sql-files/main.sql
+++ b/sql-files/main.sql
@@ -910,6 +910,7 @@ CREATE TABLE IF NOT EXISTS `sc_data` (
   `account_id` int(11) unsigned NOT NULL,
   `char_id` int(11) unsigned NOT NULL,
   `type` smallint(11) unsigned NOT NULL,
+  `tick_total` bigint(20) NOT NULL,
   `tick` bigint(20) NOT NULL,
   `val1` int(11) NOT NULL default '0',
   `val2` int(11) NOT NULL default '0',
--- a/src/char/char_mapif.cpp
+++ b/src/char/char_mapif.cpp
@@ -279,7 +279,7 @@ int chmapif_parse_askscdata(int fd){
 		int aid, cid;
 		aid = RFIFOL(fd,2);
 		cid = RFIFOL(fd,6);
-		if( SQL_ERROR == Sql_Query(sql_handle, "SELECT type, tick, val1, val2, val3, val4 from `%s` WHERE `account_id` = '%d' AND `char_id`='%d'",
+		if( SQL_ERROR == Sql_Query(sql_handle, "SELECT type, tick, tick_total, val1, val2, val3, val4 from `%s` WHERE `account_id` = '%d' AND `char_id`='%d'",
 			schema_config.scdata_db, aid, cid) )
 		{
 			Sql_ShowDebug(sql_handle);
@@ -299,10 +299,11 @@ int chmapif_parse_askscdata(int fd){
 			{
 				Sql_GetData(sql_handle, 0, &data, NULL); scdata.type = atoi(data);
 				Sql_GetData(sql_handle, 1, &data, NULL); scdata.tick = strtoll( data, nullptr, 10 );
-				Sql_GetData(sql_handle, 2, &data, NULL); scdata.val1 = atoi(data);
-				Sql_GetData(sql_handle, 3, &data, NULL); scdata.val2 = atoi(data);
-				Sql_GetData(sql_handle, 4, &data, NULL); scdata.val3 = atoi(data);
-				Sql_GetData(sql_handle, 5, &data, NULL); scdata.val4 = atoi(data);
+				Sql_GetData(sql_handle, 2, &data, NULL); scdata.tick_total = strtoll(data, nullptr, 10);
+				Sql_GetData(sql_handle, 3, &data, NULL); scdata.val1 = atoi(data);
+				Sql_GetData(sql_handle, 4, &data, NULL); scdata.val2 = atoi(data);
+				Sql_GetData(sql_handle, 5, &data, NULL); scdata.val3 = atoi(data);
+				Sql_GetData(sql_handle, 6, &data, NULL); scdata.val4 = atoi(data);
 				memcpy(WFIFOP(fd, 14+count*sizeof(struct status_change_data)), &scdata, sizeof(struct status_change_data));
 			}
 			if (count >= 50)
@@ -942,14 +943,14 @@ int chmapif_parse_save_scdata(int fd){
 			int i;
 
 			StringBuf_Init(&buf);
-			StringBuf_Printf(&buf, "INSERT INTO `%s` (`account_id`, `char_id`, `type`, `tick`, `val1`, `val2`, `val3`, `val4`) VALUES ", schema_config.scdata_db);
+			StringBuf_Printf(&buf, "INSERT INTO `%s` (`account_id`, `char_id`, `type`, `tick`, `tick_total`, `val1`, `val2`, `val3`, `val4`) VALUES ", schema_config.scdata_db);
 			for( i = 0; i < count; ++i )
 			{
 				memcpy (&data, RFIFOP(fd, 14+i*sizeof(struct status_change_data)), sizeof(struct status_change_data));
 				if( i > 0 )
 					StringBuf_AppendStr(&buf, ", ");
-				StringBuf_Printf(&buf, "('%d','%d','%hu','%" PRtf "','%ld','%ld','%ld','%ld')", aid, cid,
-					data.type, data.tick, data.val1, data.val2, data.val3, data.val4);
+				StringBuf_Printf(&buf, "('%d','%d','%hu','%" PRtf "','%" PRtf "','%ld','%ld','%ld','%ld')", aid, cid,
+					data.type, data.tick, data.tick_total, data.val1, data.val2, data.val3, data.val4);
 			}
 			if( SQL_ERROR == Sql_QueryStr(sql_handle, StringBuf_Value(&buf)) )
 				Sql_ShowDebug(sql_handle);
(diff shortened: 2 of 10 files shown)


src/char/char_mapif.cpp
```diff
@@ -299,10 +299,11 @@ int chmapif_parse_askscdata(int fd){
...
 				Sql_GetData(sql_handle, 1, &data, NULL); scdata.tick = strtoll( data, nullptr, 10 );
-				Sql_GetData(sql_handle, 2, &data, NULL); scdata.val1 = atoi(data);
-				Sql_GetData(sql_handle, 3, &data, NULL); scdata.val2 = atoi(data);
-				Sql_GetData(sql_handle, 4, &data, NULL); scdata.val3 = atoi(data);
-				Sql_GetData(sql_handle, 5, &data, NULL); scdata.val4 = atoi(data);
+				Sql_GetData(sql_handle, 2, &data, NULL); scdata.tick_total = strtoll(data, nullptr, 10);
+				Sql_GetData(sql_handle, 3, &data, NULL); scdata.val1 = atoi(data);
+				Sql_GetData(sql_handle, 4, &data, NULL); scdata.val2 = atoi(data);
+				Sql_GetData(sql_handle, 5, &data, NULL); scdata.val3 = atoi(data);
+				Sql_GetData(sql_handle, 6, &data, NULL); scdata.val4 = atoi(data);
```
Lemongrass3110, line 306: Move it to index 6 instead. It does not hurt us and will not break as much if someone has a customization.

src/map/clif.cpp
```diff
@@ -6126,97 +6126,79 @@ void clif_cooking_list( struct map_session_data *sd, int trigger, uint16 skill_i
...
+		p.PacketType = sc_notickType;
+	p.index = type;
+	p.AID = bl->id;
+	p.state = (uint8)flag;
 
-		WBUFL(buf,9) = client_tick(tick);
-		WBUFL(buf,13) = val1;
-		WBUFL(buf,17) = val2;
-		WBUFL(buf,21) = val3;
+#if PACKETVER >= 20120618
```
Lemongrass3110, line 6175: Move this into the ifdef below so we do not have to double check the setting

src/map/clif.cpp
```diff
@@ -6126,97 +6126,79 @@ void clif_cooking_list( struct map_session_data *sd, int trigger, uint16 skill_i
...
-		WBUFL(buf,21) = val2;
-		WBUFL(buf,25) = val3;
-	}
-#elif PACKETVER >= 20090121
-	if (flag && battle_config.display_status_timers) {
-		if (tick <= 0)
-			tick = 9999; // this is indeed what official servers do
+		p.PacketType = status_changeType;
+	} else
+		p.PacketType = sc_notickType;
```
Lemongrass3110, line 6170: This uses a dangerous assumption and should be changed.
It reintreprets the "newer" packet structure as the old one.
Additionally since we use the sizeof operator on the structure it will be the old packet id with the new packet size and could cause problems.

src/map/clif.cpp
```diff
@@ -6126,97 +6126,79 @@ void clif_cooking_list( struct map_session_data *sd, int trigger, uint16 skill_i
...
-
-	nullpo_retv(bl);
-
-	sd = BL_CAST(BL_PC, bl);
-
-	if (!(status_type2relevant_bl_types(type)&bl->type)) // only send status changes that actually matter to the client
-		return;
-
-	clif_status_change_sub(bl, bl->id, type, flag, tick, val1, val2, val3, ((sd ? (pc_isinvisible(sd) ? SELF : AREA) : AREA_WOS)));
+void clif_status_change(struct block_list *bl, int type, int flag, t_tick tick_total, int val1, int val2, int val3) {
```
Lemongrass3110, line 6200: Could be moved to the header file as a define for this convenience function.

src/map/clif.cpp
```diff
@@ -6267,13 +6249,13 @@ void clif_efst_status_change_sub(struct block_list *tbl, struct block_list *bl,
 #if PACKETVER > 20120418
 			clif_efst_status_change(tbl, bl->id, AREA_WOS, StatusIconChangeTable[type], tick, sc_display[i]->val1, sc_display[i]->val2, sc_display[i]->val3);
 #else
-			clif_status_change_sub(tbl, bl->id, StatusIconChangeTable[type], 1, tick, sc_display[i]->val1, sc_display[i]->val2, sc_display[i]->val3, AREA_WOS);
+			clif_status_change_sub(tbl, StatusIconChangeTable[type], 1, tick, tick, sc_display[i]->val1, sc_display[i]->val2, sc_display[i]->val3, AREA_WOS);
```
Lemongrass3110, line 6252: You lost the information which bl will have the status by removing the bl->id parameter?

src/map/clif.cpp
```diff
@@ -6267,13 +6249,13 @@ void clif_efst_status_change_sub(struct block_list *tbl, struct block_list *bl,
...
 #endif
 		}else{
 #if PACKETVER > 20120418
 			clif_efst_status_change(tbl, bl->id, target, StatusIconChangeTable[type], tick, sc_display[i]->val1, sc_display[i]->val2, sc_display[i]->val3);
 #else
-			clif_status_change_sub(tbl, bl->id, StatusIconChangeTable[type], 1, tick, sc_display[i]->val1, sc_display[i]->val2, sc_display[i]->val3, target);
+			clif_status_change_sub(tbl, StatusIconChangeTable[type], 1, tick, tick, sc_display[i]->val1, sc_display[i]->val2, sc_display[i]->val3, target);
```
Lemongrass3110, line 6258: Same as above?

src/map/clif.cpp
```diff
@@ -6126,97 +6126,79 @@ void clif_cooking_list( struct map_session_data *sd, int trigger, uint16 skill_i
...
+	if (battle_config.display_status_timers > 0) {
+		p.Left = client_tick(tick);
+		p.val1 = val1;
+		p.val2 = val2;
+		p.val3 = val3;
 	}
 #endif
-	clif_send(buf, packet_len(WBUFW(buf,0)), bl, target_type);
+
+	clif_send(&p, sizeof(p), bl, (sd && sd->status.option & OPTION_INVISIBLE) ? SELF : AREA);
```
Lemongrass3110, line 6188: Why not use pc_isinvisible here?
Also you lost the tbl vs bl aspect here I think.

src/map/status.cpp
```diff
@@ -12757,6 +12765,24 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
...
+ * @param val1: Depends on type of status change
+ * @param val2: Depends on type of status change
+ * @param val3: Depends on type of status change
+ * @param val4: Depends on type of status change
+ * @param tick: Remaining duration
+ * @param flag: Value which determines what parts to calculate. See e_status_change_start_flags
+ * @return adjusted duration based on flag values
+ */
+int status_change_start(struct block_list *src, struct block_list *bl, enum sc_type type, int rate, int val1, int val2, int val3, int val4, t_tick tick, unsigned char flag) {
+	return status_change_start_sub(src, bl, type, rate, val1, val2, val3, val4, 0, tick, flag);
```
Lemongrass3110, line 12783: Could be me, but this looks like you always start the SC with 0 as remaining duration now?

sql-files/main.sql
```diff
@@ -910,6 +910,7 @@ CREATE TABLE IF NOT EXISTS `sc_data` (
   `account_id` int(11) unsigned NOT NULL,
   `char_id` int(11) unsigned NOT NULL,
   `type` smallint(11) unsigned NOT NULL,
+  `tick_total` bigint(20) NOT NULL,
```
lighta, line 913: I'm not really fan of tick_total namming.
I mean a tick is a timepoint, timepoint_total doesn't make much sense in my head.
tick_duration would have been better imo. We want to define an interval of tick for the status ? ok

now does the remaining `tick`, will present a timepoint of the timeout for the skill, or will it be a duration as well ?
where tick < tick_duration

src/map/chrif.cpp
```diff
@@ -1330,6 +1330,7 @@ int chrif_save_scdata(struct map_session_data *sd) { //parses the sc_data of the
 				data.tick = 0; //Negative tick does not necessarily mean that sc has expired
 		} else
 			data.tick = INFINITE_TICK; //Infinite duration
```
lighta, line 1332: you shouldn't have `tick > tick_total`, only data.tick_total should be set as INFINITE_TICK

src/map/status.cpp
lighta: for me he reassign the remaining in `status_change_start`  so it's fine

eppc0330:
SCs with ```tick_time``` like ```M_LIFE_POTION``` still have same problem.

Pokye:
@aleos89 update please <3

PR #5292, opened 2020-08-05
Fixes guild skill delay call on map change

### LINKED ISSUE #5278: guild_block_skill
* **rAthena Hash**: 

* **Client Date**:

 Not aplicable

* **Server Mode**: 

Not aplicable

* **Description of Issue**: 
  * Result:  The function guild_block_skill is called when refresh client
  * Expected Result:  guild_block_skill called only once time, at login
  * How to Reproduce:  change map or refresh client
  * Official Information: Not aplicable
  
Commit 4944 include the function "clif_loadConfirm", that call 'guild_member_joined', and 'guild_member_joined' call 'guild_block_skill'
  
  
Not aplicable?

aleos89:
* **Addressed Issue(s)**: Fixes #5278

  * Fixes guild masters triggering the guild skill delay call on each map change.
Thanks to @mplaten!


src/map/guild.cpp
```diff
@@ -802,7 +802,7 @@ void guild_member_joined(struct map_session_data *sd) {
 		guild_request_info(sd->status.guild_id);
 		return;
 	}
-	if (strcmp(sd->status.name,g->master) == 0) {	// set the Guild Master flag
+	if (sd->state.connect_new == 1 && strcmp(sd->status.name,g->master) == 0) {	// set the Guild Master flag
```
Lemongrass3110, line 805: This is probably related to the problem with the clan counter, because this function is called too often now.

Lemongrass3110: See #5204

lighta (approved):

src/map/guild.cpp
Triedge: Uhm... this bug return!!!

Triedge: Or more specifically, it seems that this fix was not applied. At least the change history in guild.cpp says it was never implemented.

Lemongrass3110:
Closing this in favor of #5204

PR #5310, opened 2020-08-10
Fixes renewal Brandish Spear range

### LINKED ISSUE #5251: Brandish Spear Change status from melee to only range
rAthena Hash:  latest

 Which mode does your server use:  Renewal

Description of Issue:  Brandish Spear skill seems to have a dual effect. Melee and Range attacks at the same time. This mod was on May 2020.

  * Result: Brandish Spear is blocked by safety wall and peneuma, both at the same time or separately.
  * Expected Result: Only pneuma should block it. Safety wall instead should NOT block it. Brandish is considered only as a range skill not melee anymore.

  * Official Information:  https://irowiki.org/wiki/Brandish_Spear

Please consider that this is for iRO Trasendence. For testing purposes the idea is to reproduce on my private server. Any information to how to change will help me a lot.

aleos89:
* **Addressed Issue(s)**: Fixes #5251

* **Server Mode**: Renewal

  * Range should be BF_LONG only and not stacked with BF_SHORT.
  * Cleans up battle_range_type checks.
Thanks to @PipeDeveloper!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -2278,13 +2278,22 @@ static int battle_range_type(struct block_list *src, struct block_list *target, 
 	if (skill_get_inf2(skill_id, INF2_ISTRAP))
 		return BF_SHORT;
 
-	// When monsters use Arrow Shower or Bomb, it is always short range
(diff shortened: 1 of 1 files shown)


PipeDeveloper:
Tested, works perfect.

PR #5322, opened 2020-08-13
Small walkpath optimizations

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Just some optimizations

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -79,24 +79,25 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
 int unit_walktoxy_sub(struct block_list *bl)
 {
 	int i;
-	struct walkpath_data wpd;
-	struct unit_data *ud = NULL;
+	struct walkpath_data wpd = {0};
 
 	nullpo_retr(1, bl);
-	ud = unit_bl2ud(bl);
-	if(ud == NULL) return 0;
+
+	struct unit_data *ud = unit_bl2ud(bl);
+	if (ud == nullptr)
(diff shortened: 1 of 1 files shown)


src/map/unit.cpp
```diff
@@ -717,15 +718,17 @@ int unit_walktoxy( struct block_list *bl, short x, short y, unsigned char flag)
...
 		return 0;
 
+	if (bl->type != BL_NPC) {
+
 #ifdef OFFICIAL_WALKPATH
-	if( !path_search_long(NULL, bl->m, bl->x, bl->y, x, y, CELL_CHKNOPASS) // Check if there is an obstacle between
-		&& wpd.path_len > 14	// Official number of walkable cells is 14 if and only if there is an obstacle between. [malufett]
-		&& (bl->type != BL_NPC) ) // If type is a NPC, please disregard.
-			return 0;
+		if( !path_search_long(NULL, bl->m, bl->x, bl->y, x, y, CELL_CHKNOPASS) // Check if there is an obstacle between
```
mattthewaz, line 724: Hi, random passerby here, if you're going for optimization, you might want to chain these ifs together into a single if:
if (not an npc **AND** (there is an obstacle **OR** the path exceeds the maximum length))
This could technically be faster than nested ifs.

Daegaladh: Right, thanks for pointing it out!

aleos89 (approved):

aleos89 (approved):

PR #5335, opened 2020-08-17
Corrects some details about new Alchemist Potions

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Renewal

  * Fixes the Concentrated Red, Blue, and Golden Syrup Potions to only require 5 Empty Potion Bottles instead of 10.
  * Fixes Golden X to return 1 damage instead of 0 on reflect.
Thanks to @Badarosk0 and Sigma!


Badarosk0:
```Fixes Golden X to return 1 damage instead of 0 on reflect.```

this same reflect problem occurs when this PR is applied. Perhaps it is something similar.

PR #5342, opened 2020-08-19
Inventory list script command update

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

This is a suggestion to expand the functionality of the script command `getinventorylist` to others tables (cart, storage, guild storage) to retrieve easily the item data in the table.
The news commands are using the same variables name to store the datas.

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -873,11 +873,11 @@ bool pc_can_give_bounded_items(struct map_session_data *sd)
  * @param index: Item inventory index
  * @return True if the item can be traded or false otherwise
  */
-bool pc_can_trade_item(map_session_data *sd, int index) {
-	if (sd && index >= 0) {
-		return (sd->inventory.u.items_inventory[index].expire_time == 0 &&
-			(sd->inventory.u.items_inventory[index].bound == 0 || pc_can_give_bounded_items(sd)) &&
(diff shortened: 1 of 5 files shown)


src/map/pc.cpp
```diff
@@ -873,11 +873,11 @@ bool pc_can_give_bounded_items(struct map_session_data *sd)
  * @param index: Item inventory index
  * @return True if the item can be traded or false otherwise
  */
-bool pc_can_trade_item(map_session_data *sd, int index) {
-	if (sd && index >= 0) {
-		return (sd->inventory.u.items_inventory[index].expire_time == 0 &&
-			(sd->inventory.u.items_inventory[index].bound == 0 || pc_can_give_bounded_items(sd)) &&
-			itemdb_cantrade(&sd->inventory.u.items_inventory[index], pc_get_group_level(sd), pc_get_group_level(sd))
+bool pc_can_trade_item(map_session_data *sd, struct item *item) {
```
lighta, line 876: can we use ref instead ?
bool pc_can_trade_item(map_session_data& sd, struct item& item)
Pointer was only leftover of C, but since we have this converted in C++ I don't see reason to keep this like this.
We could then remove the if(sd && item)

lighta (approved):
As usual please provide basic unit_test in npc/test/*
Those also serve as a basic sample.

PR #5344, opened 2020-08-20
Updates the Magic Book NPC

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Renewal

  * Updates the Magic Book NPC, Lea, to the new item IDs.
  * Lea no longer requires items for Magic Books.

### DIFF
--- a/npc/re/quests/magic_books.txt
+++ b/npc/re/quests/magic_books.txt
@@ -15,6 +15,7 @@
 //= 1.1a Updated NPC name to "Master Velofos". [Euphy]
 //= 1.1b Fixed the rand part in "Mysterious Documents" to match
 //=	 Aegis & fixed a bracket issue. [Capuche]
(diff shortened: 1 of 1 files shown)


Surefirer:
I need to change the type of nameid in ragnarok_logs -> picklog to int and length set to 6, otherwise when I receive a book from npc, it will throw an error. Original setting is smallint(5)

aleos89:
Sounds like you didn't run [a SQL upgrade](https://github.com/rathena/rathena/blob/master/sql-files/upgrades/upgrade_20200808_logs.sql).

Surefirer:
oops, my bad:)

mazvi:
Need add packver checker for old client

Latiosu (approved):

Atemo (approved):

PR #5388, opened 2020-09-08
Session check standarization

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

### DIFF
--- a/src/char/char_logif.cpp
+++ b/src/char/char_logif.cpp
@@ -757,7 +757,7 @@ int chlogif_parse(int fd) {
 }
 
 TIMER_FUNC(chlogif_check_connect_logserver){
-	if (session_isValid(login_fd))
+	if (chlogif_isconnected())
 		return 0;
 
 	ShowInfo("Attempt to connect to login-server...\n");
--- a/src/char/char_logif.cpp
+++ b/src/char/char_logif.cpp
@@ -839,7 +839,7 @@ void chlogif_on_ready(void)
 	chlogif_send_acc_tologin(INVALID_TIMER, gettick(), 0, 0);
 
 	// if no map-server already connected, display a message...
-	ARR_FIND( 0, ARRAYLENGTH(map_server), i, session_isValid(map_server[i].fd) && !map_server[i].empty() );
+	ARR_FIND( 0, ARRAYLENGTH(map_server), i, session_isValid(map_server[i].fd) && !map_server[i].map.empty() );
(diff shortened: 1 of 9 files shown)


secretdataz:
Please flag both relevant functions as inline while you're at it.

src/char/char_logif.cpp
```diff
@@ -757,7 +757,7 @@ int chlogif_parse(int fd) {
 }
 
 TIMER_FUNC(chlogif_check_connect_logserver){
-	if (login_fd > 0 && session[login_fd] != NULL)
+	if (session_isValid(login_fd))
```
Lemongrass3110, line 760: Replace with chlogif_isconnected()

Lemongrass3110, line 760:
```suggestion
	if (chlogif_isconnected())
```

src/char/char_mapif.cpp
```diff
@@ -72,7 +72,7 @@ int chmapif_sendallwos(int sfd, unsigned char *buf, unsigned int len){
  * @return : the number of map-serv the packet was sent to (O|1)
  */
 int chmapif_send(int fd, unsigned char *buf, unsigned int len){
-	if (fd >= 0) {
+	if (session_isValid(fd)) {
```
Lemongrass3110, line 75: You missed two cases above. In chmapif_sendall and chmapif_sendallwos. Please also change them.

src/map/clif.cpp
```diff
@@ -9321,7 +9321,7 @@ void clif_GM_kick(struct map_session_data *sd, struct map_session_data *tsd)
 	if (sd == NULL)
 		tsd->state.keepshop = true;
 
-	if (fd > 0)
+	if (session_isValid(fd))
```
Lemongrass3110, line 9324: Missing clif_authfail_fd

Lemongrass3110 (changes requested):
I agree with @secretdataz that it would be a good improvement to inline these functions while we are it.

Looks good in general only missed a few references that I found by searching for "fd" variable references. Might be that we inherited some other creative variable names though. But we can always fix these later on.

Beside my in-code-comments I saw that you missed two files completely: intif.cpp and script.cpp
Please check them for references too.

aleos89:
I think what @secretdataz and @Lemongrass3110 commented on are good as well.

aleos89 (approved):

Lemongrass3110 (approved):

PR #5422, opened 2020-09-28
Fixes a potential crash when adding new maps

### LINKED ISSUE #5396: crashfix: adding non-existing map to maps_athena
* **rAthena Hash**: 4bac43435

* **Client Date**:  irrelevant

* **Server Mode**: irrelevant

* **Description of Issue**: 
map-server crash when you add non-existing map (in cache) into name in maps_athena.conf
map-server crash with the error: 
`Exception thrown: read access violation.
info was 0x3CCFFF6. occurred`
map.cpp:3565
```c++
p += sizeof(struct map_cache_map_info) + info->len;
```

* **How to Reproduce**: 
Add random map name in conf/maps_athena.conf
Please do not add map that already exists in map_cache.dat

This may be a small thing, but it would be best if there is a "check" to see if the map name actually exists in the map_cache.dat
Removing the non-existing map name could solve the issue, but the crash does not make any sense for most people.

My work around is to add a check for "empty map name"
find map.cpp:3558
``
```c++
for(i = 0; i < header->map_count; i++) {
		info = (struct map_cache_map_info *)p;

		if( strcmp(m->name, info->name) == 0 )
			break; // Map found

		// Jump to next entry..
		p += sizeof(struct map_cache_map_info) + info->len;
	}
```
replace map.cpp:3558
```c++
for(i = 0; i < header->map_count; i++) {
		info = (struct map_cache_map_info *)p;

		// name exists in maps_athena.conf but not in map_cache.dat -> crash
		if (info->name != NULL && info->name[0] == '\0')
			continue;

		if( strcmp(m->name, info->name) == 0 )
			break; // Map found

		// Jump to next entry..
		p += sizeof(struct map_cache_map_info) + info->len;
	}
```

aleos89:
* **Addressed Issue(s)**: Fixes #5396

  * Fixes a potential crash on start up when adding a map to maps_athena.conf and the map has not yet been added to map_cache.dat.
Thanks to @nubspixel!

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -3559,7 +3559,7 @@ int map_readfromcache(struct map_data *m, char *buffer, char *decode_buffer)
 		info = (struct map_cache_map_info *)p;
 
 		// name exists in maps_athena.conf but not in map_cache.dat
-		if (info->name != nullptr && info->name[0] == '\0')
+		if (info->name == nullptr || info->name[0] == '\0')
 			continue;
 		if( strcmp(m->name, info->name) == 0 )
 			break; // Map found
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -3558,6 +3558,9 @@ int map_readfromcache(struct map_data *m, char *buffer, char *decode_buffer)
 	for(i = 0; i < header->map_count; i++) {
 		info = (struct map_cache_map_info *)p;
 
+		// name exists in maps_athena.conf but not in map_cache.dat
+		if (info->name != nullptr && info->name[0] == '\0')
+			continue;
 		if( strcmp(m->name, info->name) == 0 )
 			break; // Map found
 


src/map/map.cpp
```diff
@@ -3558,6 +3558,9 @@ int map_readfromcache(struct map_data *m, char *buffer, char *decode_buffer)
 	for(i = 0; i < header->map_count; i++) {
 		info = (struct map_cache_map_info *)p;
 
+		// name exists in maps_athena.conf but not in map_cache.dat
+		if (info->name != nullptr && info->name[0] == '\0')
```
lighta, line 3562: wouldn't be simplier to have this check inverted ?
`if(info->name == nullptr || info->name[0] == '\0')`
I mean, while we are checking for nullptr we might as well quit if it's the case. The previous check would have allow a nullptr.

lighta (approved):

PR #5423, opened 2020-09-28
Fixes Magnetic field behavior

### LINKED ISSUE #5401: Mechanic - Madogear Skills Issues.
* **rAthena Hash**: 

* **Client Date**: 

 20200304

* **Server Mode**: 

 renewal

* **Description of Issue**: 
  * Result:  Mechanics skill [Magnetic field] does NOT snare anyone or anything at all but it does lock you in place and drains your SP heavily., Mechanics Skill [Hovering] only avoids traps (excluding Thorn trap for some reason but ankle snare is fine ) but it doesnt avoid any other ground base skills, tested with Pillar of fire, heavens drive and a few others. and it only works if the item catalyst is EQUIPPED, which is wrong, it should only have to be in the inventory.
  * Expected Result:  Magnetic field should snare anyone not in hover state in a 5x5 area around the caster, Hover skill should avoid traps and all other ground based skills while in hover state., also the  catalyst item for Hover skill has to be equipped and not just in your inventory, which is wrong.
  * How to Reproduce:  Download latest rA server, make a mechanic, equip madogear, give yourself all skills and items and test and you'll get the same results as pointed outa bove.
  * Official Information: https://www.divine-pride.net/database/skill/2263 , https://irowiki.org/wiki/Hover, https://www.divine-pride.net/database/skill/2272/magnetic-field, , https://irowiki.org/wiki/Magnetic_Field
  

   None
   None

aleos89:
* **Addressed Issue(s)**: Fixes #5401

* **Server Mode**: Renewal

  * Minor optimizations to core skill behavior.
  * Adjusts the status timer to be stored in val4.
  * Impose a unit lock before doing status damage.
  * Check for status state before attempting to reapply the status timer.
Thanks to @LotusRO!

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -5913,10 +5913,6 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 			map_foreachinallrange(skill_area_sub, bl, skill_get_splash(skill_id, skill_lv), splash_target(src), src, skill_id, skill_lv, tick, flag|BCT_ENEMY|SD_SPLASH|1, skill_castend_damage_id);
 		}
 		break;
-
-	case NC_MAGNETICFIELD:
-		sc_start2(src,bl,SC_MAGNETICFIELD,100,skill_lv,src->id,skill_get_time(skill_id,skill_lv));
-		break;
 	case SC_FATALMENACE:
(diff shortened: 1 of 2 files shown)


LotusRO:
Thank you for fixing this! @aleos89

PR #5473, opened 2020-10-22
Fixes item bonuses on restricted maps

### LINKED ISSUE #693: Official MDef bonus
Official MDef bonus must like (renewal) ATK:MATK in item_db part. Defined MATK in item_db column instead as `bonus bMAtk,n;` is allowing the MAtk is still calculated even in restricted area for the item.

So does the DEF:MDEF, currently, by using MDEF as `bonus bMDef,n;`, restrcited item doesn't give MDef it supposed to give.

Here's The WOE TE items that gives MDef, of course the item has other effect but Mdef bonus is still calculated.

TE WoE Coat
> MDef +15 (normal effect)
> (Only on WOE TE area) MaxHP +250, MaxSP +150

*This image below, TE WoE Coat wore **outside** castle (non WoE TE area)*
screenfree thor gratis 966

*This image below, TE WoE Coat wore **inside** castle (**on** WoE TE area)*
screenfree thor gratis 967

From idRO.

aleos89:
* **Addressed Issue(s)**: Fixes #693

  * Items that give a bonus of STR, AGI, VIT, INT DEX, LUK, MDEF, MaxHP, or MaxSP can now be nullified if the item is restricted based on the mapflag.
Thanks to @cydh!

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -3030,6 +3030,23 @@ static void pc_bonus_addvanish(std::vector<s_vanish_bonus> &bonus, int16 rate, i
 	bonus.push_back(entry);
 }
 
+/**
+ * Checks if a bonus is restricted from a map
+ * @param sd: Player data
+ * @return True if the item bonus is granted or false otherwise
+ */
+static bool pc_bonus_restricted(map_session_data *sd) {
+	nullpo_retr(false, sd);
+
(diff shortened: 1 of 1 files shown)


aleos89:
Closing in favor of @Atemo's PR which is a much better suited fix!

PR #5494, opened 2020-10-28
Converts the random option databases to YAML

### LINKED ISSUE #3084: Random option group database rework
* **rAthena Hash**: 10782d83e
* **Client Date**: Irrelevant
* **Server Mode**: Both
* **Description of Issue**: 
    The random option group database (item_randomopt_group.txt) is impractical to use in scenarios with multiple, random randopt_value and multiple option slots because you have to list all the permutations of the random options.
*  ** To-do list**:
- [ ] Support unweighted randopt_value range instead of specific value per entry
- [ ] Support per-slot random option specification instead of listing whole combination in an entry
Additional suggestions are welcome.
- [ ] Use YAML format instead of CSV

aleos89:
* **Addressed Issue(s)**: Fixes #3084

* **Server Mode**: Renewal

  * Creating random option groups now further reflects official behavior.
  * Group bonuses can now be created without having to create multiple permutations of the same group.
  * Specific slot options can now be defined which guarantee a bonus.
  * Non-guaranteed options can now be defined along with a maximum amount.
  * Random option value can now be defined as a range instead of a static value.
  * Adds conversion of official random option groups.
  * Includes CSV2YAML conversion tool.


src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+	if (this->nodeExists(node, "Option")) {
+		std::string name;
 
-		lines++;
+		if (!this->asString(node, "Option", name))
+			return 0;
 
-		if (line[0] == '/' && line[1] == '/') // Ignore comments
-			continue;
+		if (random_option_db.option_search_name(name)) {
```
Lemongrass3110, line 2307: Should only trigger, if the found id != the current id, otherwise it could be someone redefined the name in the import file.

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+	return false;
+}
+
+/**
+ * Return the constant value of the given random option.
+ * @param name: Random option name
+ * @param id: Random option ID
+ * @return True on success or false on failure
+ */
+bool RandomOptionDatabase::option_get_id(std::string name, uint16 *id) {
```
Lemongrass3110, line 2357: Make id a reference, this way it cannot be nullptr and assignment does not require dereferencing.

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+		randopt->id = id;
+	}
+
+	if (this->nodeExists(node, "Group")) {
+		std::string name;
+
+		if (!this->asString(node, "Group", name))
+			return 0;
+
+		if (random_option_group.option_search_name(name)) {
```
Lemongrass3110, line 2400: Only trigger this, if the found id is not the same as the current one, otherwise it might be someone has redefined the group name in import.

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
-		if (str[1][strlen(str[1]) - 1] != '}') {
-			ShowError("itemdb_read_randomopt(#2): Invalid format (Script column) in line %d of \"%s\", skipping.\n", lines, path);
-			continue;
+
+		randopt->name = name;
+	}
+
+	for (const YAML::Node &slotNode : node["Slots"]) {
+		if (randopt->slot.size() >= MAX_ITEM_RDM_OPT) {
+			this->invalidWarning(slotNode, "Reached maximum of %d Random Option group options.\n", MAX_ITEM_RDM_OPT);
```
Lemongrass3110, line 2410:
```suggestion
			this->invalidWarning(slotNode, "Reached maximum of %d Random Option group options. Skipping the remaining slots...\n", MAX_ITEM_RDM_OPT);
```

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
 		}
-		else {
-			int id = -1;
-			struct s_random_opt_data *data;
-			struct script_code *code;
 
-			str[0] = trim(str[0]);
-			if (ISDIGIT(str[0][0])) {
-				id = atoi(str[0]);
+		int16 slot;
```
Lemongrass3110, line 2414: If it is [1-x[ make it uint16 maybe?

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+
+		if (!this->asInt16(slotNode, "Slot", slot))
+			return 0;
+
+		if (slot < 1 || slot > MAX_ITEM_RDM_OPT) {
+			this->invalidWarning(slotNode["Slot"], "Invalid Random Opton Slot number %d given, must be between 1~%d, skipping.\n", slot, MAX_ITEM_RDM_OPT);
```
Lemongrass3110, line 2420: int16 should be %h or uint16 %hu

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
-		g->id = id;
-		g->total = 0;
-		g->entries = NULL;
-		uidb_put(itemdb_randomopt_group, g->id, g);
+	if (this->nodeExists(node, "MaxRandom")) {
+		uint16 max;
+
+		if (!this->asUInt16(node, "MaxRandom", max))
+			return 0;
+
```
Lemongrass3110, line 2524: Should be limited by MAX_ITEM_RDM_OPT?

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+		randopt->max_random = max;
+	} else {
+		if (!exists)
+			randopt->max_random = 0;
 	}
 
-	RECREATE(g->entries, struct s_random_opt_group_entry, g->total + rate);
+	if (this->nodeExists(node, "Random")) {
+		if (randopt->slot.size() >= MAX_ITEM_RDM_OPT) {
+			this->invalidWarning(node["Random"], "No Random options can be added because there are no Random Option slots available.\n", MAX_ITEM_RDM_OPT);
```
Lemongrass3110, line 2533: Unused argument?

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+			return 0;
```
Lemongrass3110, line 2534: Maybe break and only skip, like you did above?

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+		}
+
+		for (const YAML::Node &randomNode : node["Random"]) {
+			std::shared_ptr<s_random_opt_group_entry> entry;
```
Lemongrass3110, line 2538: If I am not mistaken this code and the code above do the same thing and should be put into a sub function maybe?

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+}
+
+RandomOptionGroupDatabase random_option_group;
+
+/**
+ * Check if the given random option group name exists.
+ * @param name: Random option name
+ * @return True on success or false on failure
+ */
+bool RandomOptionGroupDatabase::option_search_name(std::string name) {
```
Lemongrass3110, line 2641: Rename to option_exists?

src/map/itemdb.cpp
```diff
@@ -2274,194 +2272,396 @@ bool itemdb_isNoEquip(struct item_data *id, uint16 m) {
...
+	return false;
+}
+
+/**
+ * Return the constant value of the given random option group.
+ * @param name: Random option group name
+ * @param id: Random option group ID
+ * @return True on success or false on failure
+ */
+bool RandomOptionGroupDatabase::option_get_id(std::string name, uint16 *id) {
```
Lemongrass3110, line 2656: Change to reference?

src/map/itemdb.hpp
```diff
@@ -919,28 +920,65 @@ struct item_data
...
 struct s_random_opt_group_entry {
-	struct s_item_randomoption option[MAX_ITEM_RDM_OPT];
+	uint16 id;
+	int16 min_value, max_value;
+	int8 param;
+	uint16 chance;
 };
 
 /// Struct for Random Option Group
 struct s_random_opt_group {
```
Lemongrass3110, line 942: Plural name for the vectors?

src/map/mob.cpp
```diff
@@ -2098,19 +2099,51 @@ static TIMER_FUNC(mob_ai_hard){
...
+			item->option[i].id = options[idx]->id;
+			item->option[i].value = rnd_value(options[idx]->min_value, options[idx]->max_value);
+			item->option[i].param = options[idx]->param;
+		}
+
+		// Apply Random options (if available)
+		if (group->max_random > 0) {
+			std::vector<std::shared_ptr<s_random_opt_group_entry>> options = group->random_option;
+
+			for (size_t i = 0; i < group->max_random; i++) {
```
Lemongrass3110, line 2132: I think the inner and outer loop should be switched a little maybe or merged?

Pseudo code:
```
for (size_t i = 0; i < min( group->max_random, MAX_ITEM_RDM_OPT ); i++) {
	// If it already has a option  - skip it
	if (item->option[j].id > 0)
		continue;
	// Apply random logic
}
```

Random logic from here and above could be put into a sub function again.

db/re/item_randomopt_group.yml
```diff
@@ -0,0 +1,7391 @@
...
+#       Param           Parameter value. (Default: 0)
+#       Chance          Chance applied specifically to this Item Random Option (1 = 0.01%, 10000 = 100%). (Default: 0)
+###########################################################################
+
+Header:
+  Type: RANDOM_OPTION_GROUP
+  Version: 1
+
+Body:
+  - Id: 0
```
Lemongrass3110, line 49: Should be removed if it does not any effect?

Lemongrass3110 (changes requested):

Lemongrass3110 (approved):

ijped27:
Is there any possible way to revert this commit or can i clone this and revert this and how

Issue #5516, opened 2020-11-03
Doram can't equip many equipments


kaninhot004:
* **rAthena Hash**: 

d15860342

* **Client Date**: 

20200304

* **Server Mode**: 

Renewal

* **Description of Issue**: 
  * Result: Doram cannot equip simple hat and other equipments.
  * Expected Result: Hat and other equipments should equipable.
  * How to Reproduce: Create new characters >> @ item hat >> Equip
  * Official Information: Hat and other equipments are equipable.

* **Modifications that may affect results**: 
None

PR #5518, opened 2020-11-03
Fixed item restrictions for "All"

Lemongrass3110:
* **Addressed Issue(s)**:  #5516

* **Server Mode**: Renewal

Since Doram's are 16 and MAPID_ALL was defined with only the first 16 bits were covered by 0xFFFF, the job restriction failed if Summoner was not explicitly allowed.

Thanks to @kaninhot004


aleos89 (approved):

vstumpf (approved):
LGTM

Issue #5083, opened 2020-06-11
Invisible Items due Capacity Limits


Everade:
* **rAthena Hash**: f481cc9

* **Client Date**: 2018-02-21bRagexeRE

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Taking advantage of the official  **vip_storage_increase: 300** leads to players having more items in their storage than its default max capacity, once the vip time has run out. While this isn't a problem by itself, the main complains from players has always been that their **items vanish**.

**Here's an example:**
Default max Storage Capacity is **500**
VIP Storage Capacity increase is **300**

Players now stack up 750 items, after that, the VIP time runs out.
So they eventualy end up having **750/500** items, and are being forced to clean their storage or re-gain vip status to be able to store further items. alright ~

But... at this point, **_the client does only display a total of 500 items_**. the other 250 are invisible! The only way to re-gain these missing items, is by removing items, so they eventualy start popping up again. Each time you take one stack out, one of the invisible items pops back up.

  * Expected Result: All items should be visible, even if you're above storage capacity. However storing or stacking further items should no longer be allowed.
  * How to Reproduce: Have more than your max capacity of item inside your storage, and check if all items are actualy visible.
  * Official Information: Unknown

* **Modifications that may affect results**: 
vip_storage_increase: 300

PR #5521, opened 2020-11-04
Fixed invisible items after VIP expiration

Lemongrass3110:
* **Addressed Issue(s)**: #5083

* **Server Mode**: Both

Fixed the bug and while being at it fixed the refresh code for storages as well.

Thanks to @Everade and @mazvi


mazvi:
eg when the user has used 600 items

vip runs out, then the user can see these 600 items.

when the user takes 1 item, the slot appears 599,

can the user not add it back when the vip runs out in that condition?

or the user must re-login first.

Lemongrass3110:
> 
> 
> eg when the user has used 600 items
> 
> vip runs out, then the user can see these 600 items.
> 
> when the user takes 1 item, the slot appears 599,
> 
> can the user not add it back when the vip runs out in that condition?
> 
> or the user must re-login first.
> 
> And will "Ragnarok Official" also appear all if the VIP runs out? I didn't check this, I'm just worried about a potential storage bug if it is implemented.

@mazvi I can tell you from playing euRO years ago that it works like this. You will be able to get items out of the storage (see all of them), but you cannot put anything into the storage anymore, because you are over your allowed limit of course.

aleos89 (approved):

PR #5523, opened 2020-11-05
Questinfo correction

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

The bubble from `questinfo` script command wasn't always displayed when several `questinfo` were defined in a npc.

Additional note: the `questinfo` is no longer automatically duplicated from the original npc inside an instance, it now need to be defined under `OnInstanceInit`.

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -4342,11 +4342,6 @@ void map_remove_questinfo(int m, struct npc_data *nd) {
 	nullpo_retv(nd);
 	nullpo_retv(mapdata);
 
-	for (const auto &it : nd->qi_data) {
-		if (it.condition)
-			script_free_code(it.condition);
-	}
-
 	util::vector_erase_if_exists(mapdata->qi_npc, nd->bl.id);
 	nd->qi_data.clear();
 }
@@ -4360,10 +4355,6 @@ static void map_free_questinfo(struct map_data *mapdata) {
 		if (!nd || nd->qi_data.empty())
 			continue;
 
-		for (const auto &qi : nd->qi_data) {
-			if (qi.condition)
-				script_free_code(qi.condition);
-		}
 		nd->qi_data.clear();
 	}
 
@@ -4879,8 +4870,7 @@ void do_final(void){
 	mapit_free(iter);
 
 	for (int i = 0; i < map_num; i++) {
-		struct map_data *mapdata = map_getmapdata(i);
-		map_free_questinfo(mapdata);
+		map_free_questinfo(map_getmapdata(i));
 	}
 
 	/* prepares npcs for a faster shutdown process */
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -4878,6 +4878,11 @@ void do_final(void){
 		map_quit(sd);
 	mapit_free(iter);
 
+	for (int i = 0; i < map_num; i++) {
+		struct map_data *mapdata = map_getmapdata(i);
+		map_free_questinfo(mapdata);
+	}
+
 	/* prepares npcs for a faster shutdown process */
 	do_clear_npc();
 
@@ -4888,8 +4893,6 @@ void do_final(void){
 		ShowStatus("Cleaning up maps [%d/%d]: %s..." CL_CLL "\r", i++, map_num, mapdata->name);
 		map_foreachinmap(cleanup_sub, i, BL_ALL);
 		channel_delete(mapdata->channel,false);
-
-		map_free_questinfo(mapdata);
 	}
 	ShowStatus("Cleaned up %d maps." CL_CLL "\n", map_num);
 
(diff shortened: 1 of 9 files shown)


lighta:
"Additional note: the questinfo is no longer automatically duplicated from the original npc inside an instance, it now need to be defined under OnInstanceInit."
Did we check if this won't break some existing script ? How can user detect if they script need fixing ? Can we have a small parser in tools for this ?

Atemo:
>How can user detect if they script need fixing ? Can we have a small parser in tools for this ?

A parser for this seems overkill but we can add additional notes in the docs or look deeper in the code to keep the old behaviour maybe

src/map/map.cpp
```diff
@@ -4875,6 +4878,11 @@ void do_final(void){
 		map_quit(sd);
 	mapit_free(iter);
 
+	for (int i = 0; i < map_num; i++) {
+		struct map_data *mapdata = map_getmapdata(i);
+		map_free_questinfo(mapdata);
```
aleos89, line 4883:
```suggestion
		map_free_questinfo(map_getmapdata(i));
```

src/map/pc.cpp
```diff
@@ -13287,28 +13260,45 @@ void pc_show_questinfo(struct map_session_data *sd) {
...
+		for (auto &qi : nd->qi_data) {
+			if (!qi.condition || achievement_check_condition(qi.condition, sd)) {
 				show = true;
-			else
-				show = false;
+				// Check if need to be displayed
+				if (!sd->qi_display[i].is_active || qi.icon != sd->qi_display[i].icon || qi.color != sd->qi_display[i].color) {
+					sd->qi_display[i].is_active = true;
+					sd->qi_display[i].icon = static_cast<e_questinfo_types>(qi.icon);
+					sd->qi_display[i].color = static_cast<e_questinfo_markcolor>(qi.color);
```
aleos89, line 13283: Do these need to be casted if they are of the same type?

src/map/pc.cpp
```diff
@@ -13287,28 +13260,45 @@ void pc_show_questinfo(struct map_session_data *sd) {
...
+				}
+				break;
+			}
+		}
+		if (show == false) {
+			// Check if need to be hide
+			if (sd->qi_display[i].is_active) {
+				sd->qi_display[i].is_active = false;
+				sd->qi_display[i].icon = static_cast<e_questinfo_types>(QTYPE_NONE);
+				sd->qi_display[i].color = static_cast<e_questinfo_markcolor>(QMARK_NONE);
```
aleos89, line 13294: Same question, these shouldn't need to be casted if they are of the same type.

src/map/pc.cpp
```diff
@@ -13333,10 +13323,10 @@ void pc_show_questinfo_reinit(struct map_session_data *sd) {
 	struct map_data *mapdata = map_getmapdata(sd->bl.m);
 	nullpo_retv(mapdata);
 
-	if (mapdata->qi_data.empty())
+	if (mapdata->qi_npc.empty())
 		return;
 
-	CREATE(sd->qi_display, bool, (sd->qi_count = mapdata->qi_data.size()));
+	CREATE(sd->qi_display, struct s_qi_display, (sd->qi_count = mapdata->qi_npc.size()));
```
aleos89, line 13329: Might be best to drop CREATE in favor of a vector.

src/map/pc.hpp
```diff
@@ -676,7 +682,7 @@ struct map_session_data {
 	std::vector<int> cloaked_npc;
 
 	/* ShowEvent Data Cache flags from map */
-	bool *qi_display;
+	struct s_qi_display *qi_display;
```
aleos89, line 685:
```suggestion
	std::vector<std::shared_ptr<s_qi_display>> qi_display;
```

This step will require a bit of logic cleanup `pc_show_questinfo_reinit`.

src/map/pc.hpp
```diff
@@ -676,7 +682,7 @@ struct map_session_data {
...
 	int qi_count;
```
aleos89, line 686:
```suggestion
```

src/map/unit.cpp
```diff
@@ -3399,7 +3399,7 @@ int unit_free(struct block_list *bl, clr_type clrtype)
 
 			if (sd->qi_display) {
 				aFree(sd->qi_display);
-				sd->qi_display = NULL;
+				sd->qi_display = nullptr;
 			}
 			sd->qi_count = 0;
```
aleos89, line 3404: This whole block can be removed once utilizing shared_ptr/vector.

aleos89 (changes requested):
Overall looks good! Just a few notes to use some C++ features.

Lemongrass3110 (approved):

aleos89 (approved):

PR #5536, opened 2020-11-10
fix an initialization bug of combo_idx in pc_checkcombo

inhyositsu:

* **Server Mode**: Both

* **Description of Pull Request**: the original code won't initial correctly in debian environment, and will make combo bonus doesn't work


aleos89:
Thanks for the PR, but this could actually be fully removed. I went ahead and make a separate PR because the one you created here was made in your master branch, which would be dangerous if I made a change. Make sure for any future PR you create a separate branch so that it won't break your master! Thanks again!

Issue #5368, opened 2020-08-29
MVP Drop Rate Like Normal Monster


mazvi:
* **rAthena Hash**: 1e43c6c

* **Client Date**: 20180620

* **Server Mode**: RENEWAL

* **Description of Issue**: 
  * Result: Drop Rate Like Normal Monster/Mini Boss
  * Expected Result:  The drop rate should refer to the MVP
  * How to Reproduce:  @mi 3124 This monster will use the normal monster / miniboss droprate instead of MVP
  * Official Information:
  

* **Modifications that may affect results**: 
  
  

after trying to kill him no MVP logo came out. on divine pride should be MVP https://www.divine-pride.net/database/search?q=charleston+3

Thanks to @Stolao for the discussion

The problem is because this monster doesn't have an MVP drop, so it's mistaken for a boss. 

PS: i was added MVP drop 607,526, and 501 but not work.. The drop item still like a boss monster/normal and not like MVP

3124,CHARLESTON3,Charleston 3,Charleston 3,145,23671401,1,3132000,2473000,1,1500,1000,200,90,208,106,205,206,228,105,10,12,2,6,60,0x620B695,200,676,2016,672,0,607,5000,526,5000,501,5000,16031,200,28101,200,16033,100,21011,200,22043,400,22044,400,6752,5000,0,0,0,0,0,0

aleos89:
The MVP icon is determined based on the monster mode `MD_MVP`.

mazvi:
So what the solution? What the hex code for this?

aleos89:
You can see the details for monster modes in [doc/mob_db_mode_list](https://github.com/rathena/rathena/blob/master/doc/mob_db_mode_list.txt). It requires `0x80000` to be MD_MVP.

mazvi:
up :D

mazvi:
According to commit 7992e0891bc05fd92a9ea4c474548d3d7d72e880

I checked on https://raw.githubusercontent.com/rathena/rathena/master/db/re/mob_db.yml

check charleston3

do we need change Class: Boss to MVP?

tag to: @aleos89

aleos89:
No, MVP is a mode, not a class. You can see a list of the diff types in `doc/mob_db_mode_list.txt`.

mazvi:
Charleston 3 is MVP

attackjom:
https://youtu.be/BErG5xvkhL4

This video show it didn't have gold skeleton logo also show HP bar.

[...]

PR #5559, opened 2020-11-17
Cleanup of MVP mode checking

Lemongrass3110:
* **Addressed Issue(s)**: #5368

* **Server Mode**: Both

Adjusted checks to correctly use the mode MD_MVP for checking if a monster is a MVP or not.
Fixed some MVP exp related issues like a missing cap for exp_bonus_max_attacker or still giving out 1 exp point where it should not.
Additionally only sending the MVP exp message to the client if you actually got exp from the monster.
Removed yet another memset.
Changed the calculation type to t_exp and capped values sent to the client depending on the packet version.

Thanks to @mazvi and @aleos89


src/map/clif.cpp
```diff
@@ -1169,7 +1169,7 @@ static void clif_set_unit_idle( struct block_list* bl, bool walking, send_target
 		p.maxHP = -1;
 		p.HP = -1;
 	}
-	p.isBoss = ( bl->type == BL_MOB && ( ( ( TBL_MOB *)bl )->db->mexp > 0 ) ) ? 1 : 0;
+	p.isBoss = ( bl->type == BL_MOB ) ? status_has_mode( &( ( mob_data *)bl )->db->status, MD_MVP ) : false;
```
attackjom, line 1172: Is this value must be 2 for MVP monster (Golden skull logo)
and 1 for Miniboss (Silver skull logo) ?

Lemongrass3110: Fixed in fe559fe

PR #5561, opened 2020-11-17
Removes a useless nullification loop

aleos89:
* **Addressed Issue(s)**: N/A

  * The vector is a new instantiation so it will be free and clear from the start.
  * Takeover of #5536.
Thanks to @inhyositsu!

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -10250,11 +10250,6 @@ static int pc_checkcombo(struct map_session_data *sd, item_data *data) {
 
 		combo_idx.reserve(nb_itemCombo);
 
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -10250,11 +10250,6 @@ static int pc_checkcombo(struct map_session_data *sd, item_data *data) {
 
 		combo_idx.reserve(nb_itemCombo);
 
-		// Zero out temporary combo array
-		for (auto &tmp_combo : combo_idx) {
```
lighta, line 10254: add a constructor in s_itemchk, so it will be more safe and strickly equivalent.
bt

PR #5572, opened 2020-11-25
Latiosu feat/docker

lighta:
* **Server Mode**: Both
Adds an easy way for developers new to rAthena to get a local development environment up and running with only a single pre-requisite (Docker Desktop).

**Instructions for use:**
1. `docker-compose up -d` to spin up database (automatically imports `/sql-files/*.sql` on first load)
2. `docker exec -it rathena bash` to connect to lightweight linux container with all rathena dependencies (plus gdb)
3. Devs can now run usual commands (`./configure`, `make`, etc ...) or start server (`./athena-start`, `gdb map-server`, etc ...)
4. `docker-compose down` to free resources when done

**Demonstration:** https://terminalizer.com/view/c05443023904

**Additional Information:**
- Database is persisted to local disk so no need to recreate database every time.
- First time the database volume is created, all `*.sql` files are imported into the database. This does not happen on future start ups unless the volume has been deleted (devs can delete volume with `docker-compose down --volumes`)
- Devs can inspect state of containers with `docker-compose ps` and modify the setup as they see fit (Official Docs: https://docs.docker.com/compose/compose-file/)


lighta:
@Latiosu update with minor changes

Normynator (approved):

Normynator:
lgtm!

Latiosu (approved):

secretdataz (approved):

Issue #5577, opened 2020-11-27
script can not remove target by setunitdata UMOB_TARGETID value to 0


javirst:
* **rAthena Hash**: 
629c7b5

* **Client Date**: 
any

* **Server Mode**: 
any

* **Description of Issue**: 
  * Result: 
`setunitdata .@mob, UMOB_TARGETID, 0;`
will show error
`buildin_setunitdata: Error in finding target for BL_MOB!`
  * Expected Result: 
unlock the target instead
  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

PR #5578, opened 2020-11-27
Unlock target when setunitdata TARGETID 0

javirst:
* **Addressed Issue(s)**: 
#5577

* **Server Mode**: 
Both

Make the mob unlock current target when UMOB_TARGETID are set to 0
```c
setunitdata .@mobid, UMOB_TARGETID, 0;
```

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8327,7 +8327,7 @@ Parameter (indexes) for NPCs are:
 		- *_ADELAY: see doc/mob_db.txt
 		- *_DMOTION: see doc/mob_db.txt
 		- *_BODY2: enable (1) the alternate display, or disable (0)
-		- *_TARGETID: when set to 0 the unit will release the target
+		- *_TARGETID: when set to 0 the unit will release the target and stop attacking
 
 		- UMOB_AI: none (0); attack (1); marine sphere (2); flora (3); zanzou (4); legion (5); faw (6)
 		- UMOB_SCOPTION: see the 'Variables' section at the top of this document
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8327,6 +8327,7 @@ Parameter (indexes) for NPCs are:
 		- *_ADELAY: see doc/mob_db.txt
 		- *_DMOTION: see doc/mob_db.txt
 		- *_BODY2: enable (1) the alternate display, or disable (0)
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -18319,6 +18319,10 @@ BUILDIN_FUNC(setunitdata)
 			case UMOB_ADELAY: md->base_status->adelay = (short)value; calc_status = true; break;
 			case UMOB_DMOTION: md->base_status->dmotion = (short)value; calc_status = true; break;
 			case UMOB_TARGETID: {
+				if (value==0) {
+					mob_unlocktarget(md,0);
```
lighta, line 18323: is 0 as tick really ok ? @Lemongrass3110, @aleos89 
wouldn't be better to use `gettick()`

lighta (changes requested):
look ok otherwise

lighta (approved):

src/map/script.cpp
aleos89: I would say for this unique case when sending 0 you’re specifically telling the script engine you want to release the target. This would be a custom case though that we can define.

aleos89 (approved):

src/map/script.cpp
javirst: compiled both, and both of them works.

Lemongrass3110 (approved):

src/map/script.cpp
aleos89: How much of an offset? Something marginal like <1 second? Also, might be worth adding a note in the script command documentation that sending 0 for these params will release the target!

javirst: it's less than a second, so i think its fine using `gettick()`
I have added the script commands doc, but by logic if getunitdata umob_targetid return 0 if there is no target then setunitdata should do the same.

doc/script_commands.txt
```diff
@@ -8327,6 +8327,7 @@ Parameter (indexes) for NPCs are:
 		- *_ADELAY: see doc/mob_db.txt
 		- *_DMOTION: see doc/mob_db.txt
 		- *_BODY2: enable (1) the alternate display, or disable (0)
+		- *_TARGETID: when set to 0 the unit will release the target
```
Lemongrass3110, line 8330:
```suggestion
		- *_TARGETID: when set to 0 the unit will release the target and stop attacking
```

Issue #5576, opened 2020-11-27
WE_CALLPARTNER fails if Child is autotrading


Radio-Gagaa:
* **rAthena Hash**: b143241222470419e26f47ba919226ca8c19b007

* **Client Date**: 2018-06-21

* **Server Mode**: Pre-Renewal

* **Description of Issue**: 
  * Result:  When you are married and have both a spouse and a child and try to use WE_CALLPARTNER to call your spouse while your child is autotrading, the skill will fail. Looks like this was introduced in 7a9aa7fbcfe40080c1b743137fc68aebb596e632 due to it falling through into skill code it should not fall through to.
  * Expected Result:  Whether the child is online, autotrading or whatever should not have any impact on the usage of WE_CALLPARTNER.
  * How to Reproduce:  Be married and have a child. Put child in autotrade. Try to use WE_CALLPARTNER. It will fail.
  * Official Information:  None.
  

* **Modifications that may affect results**:

lighta:
This is due to some missing break in `int skill_castend_id(int tid, unsigned int tick, int id, intptr_t data)`

PR #5579, opened 2020-11-28
Fix WE_CALLPARTNER, WE_CALLPARENT and WE_CALLBABY falltrough check

lighta:
#5576 

* **Server Mode**: 
Both

Fix a fallthrough of skill requirement on WEDDING skill.
Simplify clif_callpartner() code.


src/map/clif.cpp
```diff
@@ -9225,26 +9225,14 @@ void clif_wedding_effect(struct block_list *bl)
...
 /// Notifies the client of the name of the partner character (ZC_COUPLENAME).
 /// 01e6 <partner name>.24B
-void clif_callpartner(struct map_session_data *sd)
+void clif_callpartner(struct map_session_data& sd, struct map_session_data& p_sd)
 {
-	unsigned char buf[26];
-
-	nullpo_retv(sd);
-
+	unsigned char buf[26]={0};
```
aleos89, line 9230: Might be best to utilize the packet structure since we are modifying this function.

aleos89 (commented):
I would adjust the clif function to utilize the packet structure, other than that everything else looks good!

aleos89 (approved):

src/map/skill.cpp
```diff
@@ -11812,8 +11812,16 @@ TIMER_FUNC(skill_castend_id){
 		//These should become skill_castend_pos
 		switch (ud->skill_id) {
 			case WE_CALLPARTNER:
-				if (sd)
-					clif_callpartner(sd);
+				if (sd) {
+					struct map_session_data *p_sd = pc_get_partner(sd);
+					if (!p_sd || p_sd->state.autotrade) {
```
Lemongrass3110, line 11817: Compared to WE_CALLPARENT and WE_CALLBABY we now fail the skill even if the partner is not online.
In the other two skills we only fail if the other player is online and in autotrade state.
This might explain why clif_callpartner originally supported calls without a target player session too.

src/map/clif.cpp
```diff
@@ -9501,26 +9501,24 @@ void clif_wedding_effect(struct block_list *bl)
...
-	WBUFW(buf,0) = 0x1e6;
+	p.packetType = HEADER_ZC_COUPLENAME;
 
-	if( sd->status.partner_id ) {
-		const char *p = map_charid2nick(sd->status.partner_id);
-		struct map_session_data *p_sd = pc_get_partner(sd);
-		if( p != NULL && p_sd != NULL && !p_sd->state.autotrade )
-			safestrncpy(WBUFCP(buf,2), p, NAME_LENGTH);
+	if( sd.status.partner_id ) {
+		const char *p_name = map_charid2nick(sd.status.partner_id);
```
Lemongrass3110, line 9511:
```suggestion
```

src/map/clif.cpp
```diff
@@ -9501,26 +9501,24 @@ void clif_wedding_effect(struct block_list *bl)
...
+		struct map_session_data *p_sd = pc_get_partner(&sd);
+		if (p_name != nullptr && p_sd != nullptr && !p_sd->state.autotrade)
```
Lemongrass3110, line 9513:
```suggestion
		if (p_sd != nullptr && !p_sd->state.autotrade)
```

src/map/clif.cpp
```diff
@@ -9501,26 +9501,24 @@ void clif_wedding_effect(struct block_list *bl)
...
+			safestrncpy(p.name, p_name, NAME_LENGTH);
```
Lemongrass3110, line 9514:
```suggestion
			safestrncpy(p.name, p_sd->status.name, NAME_LENGTH);
```

Lemongrass3110 (approved):
Still questioning why we handle WE_CALLPARTNER differently as the two other skills when it comes to (inofficial) autotrade. But rest looks good now.

Issue #5489, opened 2020-10-26
Renewal KN_BOWLINGBASH after 2018 rebalance patch needs full rework


humanwizzard:
* **rAthena Hash**: ad3f48f9ded76de3c25093ae8b70f0ddeab8caaa

* **Client Date**: 20200401

* **Server Mode**: Renewal

* **Description of Issue**: KN_BOWLINGBASH still uses gutter line system
  * Result: on some cells player not shouting skill name and not doing any damage but pushes monsters
  * Expected Result: skill should push monsters 5 cells and do damage depend on monster count that get hit
  * How to Reproduce: use the skill on monsters in open location (ex. Orc Village)
  * Official Information:in 2018 rebalance patch kRO removed gutter line system from the skill https://irowiki.org/wiki/Bowling_Bash
https://www.youtube.com/watch?v=Q0dAN1cM7ME please watch this video before comment
  

* **Modifications that may affect results**:

PR #5598, opened 2020-12-03
Fixes Bowling Bash behavior

aleos89:
* **Addressed Issue(s)**: #5489

* **Server Mode**: Renewal

  * Removes the gutterline feature from Bowling Bash in renewal mode.
  * Skill will now behave as a normal 5x5 AoE.
  * Fixes knockback getting removed during battle calculation.
Thanks to @humanwizzard!

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -5078,6 +5078,9 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 	case MC_CARTREVOLUTION:
 	case NPC_SPLASHATTACK:
 		flag |= SD_PREAMBLE; // a fake packet will be sent for the first target to be hit
+#ifdef RENEWAL
+	case KN_BOWLINGBASH:
+#endif
 	case AS_SPLASHER:
 	case HT_BLITZBEAT:
 	case AC_SHOWER:
@@ -5272,7 +5275,9 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 			skill_attack(skill_get_type(skill_id), src, src, bl, skill_id, skill_lv, tick, flag);
 		break;
 
+#ifndef RENEWAL
 	case KN_BOWLINGBASH:
+#endif
 	case MS_BOWLINGBASH:
 		{
 			int min_x,max_x,min_y,max_y,i,c,dir,tx,ty;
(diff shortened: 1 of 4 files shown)


humanwizzard:
hit count increase depending on the number of targets with 2h sword: not working = always 2 hits 

knockback should be:
lv 1 ~ 2 = 1
lv 3 ~ 4 = 2
lv 5 ~ 6 = 3
lv 7 ~ 8 = 4
lv 9 ~ 10 = 5

humanwizzard:
knockback works strange now:
1 ~ 3 targets = knockback works
4 ~ 7 targets = no knockback
8 ~ 11 targets = knockback works
12 ~ 15 targets = no knockback
and so on

humanwizzard:
knockback works fine now but hit count got broken:
4 ~ 5 targets = 2 hit
6 ~ 7 targets = 3 hit

humanwizzard:
knockback works 2 times (2 ~ 10 cells) on all except one target and hit count works only on that target

src/map/skill.cpp
```diff
@@ -5272,7 +5272,20 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 			skill_attack(skill_get_type(skill_id), src, src, bl, skill_id, skill_lv, tick, flag);
 		break;
 
+#ifdef RENEWAL
 	case KN_BOWLINGBASH:
+		if (flag & 1) {
+			skill_attack(skill_get_type(skill_id), src, src, bl, skill_id, skill_lv, tick, (skill_area_temp[0]) > 0 ? SD_ANIMATION | skill_area_temp[0] : skill_area_temp[0]);
+			skill_blown(src, bl, skill_get_blewcount(skill_id, skill_lv), -1, BLOWN_NONE);
+			skill_area_temp[0] = 0;
```
Daegaladh, line 5280: This line makes the recursive invocation not working properly

Daegaladh (changes requested):

src/map/skill.cpp
Daegaladh, line 5280: This line makes recursive invocation not work properly

Daegaladh (changes requested):

Daegaladh (approved):

PR #5604, opened 2020-12-04
Clean up to the item buy/sell checks

aleos89:
* **Addressed Issue(s)**: N/A

  * An item's sell value will now always be calculated whether the item exists or not. This resolves adjustment of an item's buy value during the import process.
  * Moves a couple checks to the loadingFinished function.
Thanks to @Daegaladh!


Lemongrass3110 (approved):

src/map/itemdb.cpp
```diff
@@ -157,15 +148,7 @@ uint64 ItemDatabase::parseBodyNode(const YAML::Node &node) {
 		item->value_buy = buy;
```
Atemo, line 148:
```suggestion
		item->value_buy = buy;
		item->value_sell = 0;
```
Define `buy` should 'remove' the current `sell`

src/map/itemdb.cpp
```diff
@@ -177,13 +160,9 @@ uint64 ItemDatabase::parseBodyNode(const YAML::Node &node) {
 
 		item->value_sell = sell;
```
Atemo, line 161:
```suggestion
		item->value_sell = sell;
		item->value_buy = 0;
```
Define `sell` should 'remove' the current `buy`

Atemo (approved):

src/map/itemdb.cpp
Lemongrass3110: @Atemo only if the entry already exists, correct? Applies to both of your comments.

Atemo: Well yes you're right the values are re-assigned when the entry doesn't exist but it doesn't change the result it's still correct

PR #5624, opened 2020-12-14
Implements asynchronous query_sql script commnads.

inhyositsu:

* **Server Mode**: Both

* **Description of Pull Request**: Implements asynchronous query_sql script commnads.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -8868,11 +8868,19 @@ Example:
 
 *query_sql("your MySQL query"{, <array variable>{, <array variable>{, ...}}});
 *query_logsql("your MySQL query"{, <array variable>{, <array variable>{, ...}}});
+*query_sql_async("your MySQL query"{, <array variable>{, <array variable>{, ...}}});
+*query_logsql_async("your MySQL query"{, <array variable>{, <array variable>{, ...}}});
 
 Executes an SQL query. A 'select' query can fill array variables with up to 2 billion rows of
 values, and will return the number of rows (i.e. array size) or -1 on failure.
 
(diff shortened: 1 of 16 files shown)


inhyositsu:
I have been using these code in my own server for several days, it really helps me to solve query blocking issue. Here's for those who need it.

Lemongrass3110:
Well this certainly looks promising!

But could you explain how the threads (futures) join back into the main thread?
You seem to know how threading works and so I assume that you also know the possible problems that arise from it (modifying data of another thread, calling functions from another thread that assume single threadedness, not even going to start talking about deadlocks or anything similar now).
We have discussed this PR in the staff's discord channel today and noone could clearly say how or if you do this cleanly.

Thank you for taking your time!

inhyositsu:
First of all, sorry for lack of comments in my code, this is the first I push my code in this large scale.

I add do_future() in the server's main loop(core.cpp)
that's where and how the main thread get data from the async-sql thread.
the main issue about multi-threading on all Athena-Like emulators is data racing,
so i add mutex(es) in the JobQueue template class (jobqueue.hpp) to prevent it.
Both main thread and aysnc-sql thread are using the same JobQueue object to access the sharing data.

I've been using my code on my own server for couple of weeks and it seems to be fine.

lukasrmattos:
I'm using this for some internal tests. Received a crash when @reloadscript.

image

Query:

`
SELECT `id`,  `mobid`, `status`, `x`, `y`, `dir`, `dir_type`, `scrpt`
  FROM `house_data` 
WHERE `char_id` = '150000'   
    AND `status` = '1'  
`

Call stack:

image

image

image

Issue #5547, opened 2020-11-16
update items Flags: DropEffect


idk-whoami:
* **rAthena Hash**: 9348ea441d0adae38ca2b4b10aa53db63ff915dd

* **Client Date**: 20200401

* **Server Mode**: re

* **Description of Issue**: 
rdata.grf 6 color
rAthena 5 color
![Capture](https://user-images.githubusercontent.com/60462533/99214789-aef16280-2803-11eb-9a15-76ede128e87e.PNG)

![บันทึก_2020_11_16_11_53_13_135](https://user-images.githubusercontent.com/60462533/99214395-9e8cb800-2802-11eb-890a-5eb92bf78fba.gif)

![บันทึก_2020_11_16_11_53_33_767](https://user-images.githubusercontent.com/60462533/99214464-c9770c00-2802-11eb-9211-299b32cd2907.gif)

![บันทึก_2020_11_16_11_54_05_823](https://user-images.githubusercontent.com/60462533/99214528-f3c8c980-2802-11eb-950e-a3987f36c29d.gif)

setting
`  - Id: 501
    AegisName: Red_Potion
    Name: Red Potion
    Type: Healing
    Buy: 10
    Weight: 70
    Flags:
      DropEffect: WHITE_PILLAR
    Script: |
      itemheal rand(45,65),0;
  - Id: 502
    AegisName: Orange_Potion
    Name: Orange Potion
    Type: Healing
    Buy: 50
    Weight: 100
    Flags:
      DropEffect: BLUE_PILLAR
    Script: |
      itemheal rand(105,145),0;
  - Id: 503
    AegisName: Yellow_Potion
    Name: Yellow Potion
    Type: Healing
    Buy: 180
    Weight: 130
    Flags:
      DropEffect: YELLOW_PILLAR
    Script: |
      itemheal rand(175,235),0;
  - Id: 504
    AegisName: White_Potion
    Name: White Potion
    Type: Healing
    Buy: 1200
    Weight: 150
    Flags:
      DropEffect: PURPLE_PILLAR
    Script: |
      itemheal rand(325,405),0;
  - Id: 505
    AegisName: Blue_Potion
    Name: Blue Potion
    Type: Healing
    Buy: 5000
    Weight: 150
    Flags:
      DropEffect: ORANGE_PILLAR
    Script: |
      itemheal 0,rand(40,60);
  - Id: 506
    AegisName: Green_Potion
    Name: Green Potion
    Type: Healing
    Buy: 40
[...]

PR #5647, opened 2020-12-22
RED_PILLAR to cards

Atemo:
* **Addressed Issue(s)**: ?

* **Server Mode**: renewal

Added RED_PILLAR drop effect to cards with location on renewal.


idk-whoami:
i think Darklord Essence series IDs 4908 to 4925 contained in boxes
https://www.divine-pride.net/database/item/22567

Balferian:
@Atemo cards should be use Pink_pillar effect.
As i know it can be used only from client side (iteminfo).
Red pillar and Pink pillar are different.
https://github.com/rathena/rathena/issues/5547

Daegaladh:
> 
> 
> @Atemo cards should be use Pink_pillar effect.
> As i know pink pillar can be used only from client side (iteminfo).
> Red pillar and Pink pillar are different.
> #5547

I think Gravity changed all cards to use RED_PILLAR in latest updates

aleos89:
Correct, it is now the red pillar:
image

Lemongrass3110 (approved):

Lemongrass3110:
Should be:
`Dropeffect: Client`

The actual effect id is defined inside the LUA file. The positive thing about this is, that we can parse the items that have it from there. :-)

Issue #5649, opened 2020-12-26
Memory Leaks


mplaten:
* **rAthena Hash**: dfa1df7

* **Client Date**: 20200401

* **Server Mode**: boths

* **Description of Issue**: After load a npc with @loadnpc and make interaction, when i use @mapexit, i receive memory leaks error
  * Result: 
  * Expected Result: 
  * How to Reproduce:  load any npc with @loadnpc [patch/to/npc.ext], interact with npc and use @mapexit
  * Official Information: 
  

* **Modifications that may affect results**: Not aplicable. New download

Akkarinage:
Do you think it would be helpful if you gave us the error? Is it more expansive than "Memory leaks"? Are you able to run with gdb for additional catchable output?

mplaten:
> Do you think it would be helpful if you gave us the error? Is it more expansive than "Memory leaks"? Are you able to run with gdb for additional catchable output?

![image](https://user-images.githubusercontent.com/10901983/103283619-d23d2f00-49b7-11eb-8803-c23689aec330.png)

gen1x8:
Memory leak also occur when using this command script maprespawnguildid (script in woe_controller.txt)

create npc script:

> payon,146,162,3	script	test	635,{
> 	mes "You Click me";
> 	set .@maptes$, "prt_gld";
> 	maprespawnguildid .@maptes$,0,3;
> end;
> }

then shutdown rathena.:
in the console there is some warning.

> [31/Dec 01:09][Warning]: db_vdestroy: Database is still in use, 1 lock(s) left. Continuing database destruction.
> Database allocated at map.cpp:5199
> [31/Dec 01:09][Status]: Close Map DB Connection....
> [31/Dec 01:09][Status]: Close Log DB Connection....
> [31/Dec 01:09][Status]: Finished.
> [31/Dec 01:09][Warning]: Memory leak detected at ERS 'db.cpp::db_iterator_ers', 1 objects not freed.
> 
> Memory manager: Memory leaks found at 2020/12/31 01h09m54s (Git Hash dfa1df7de65906f3e70ec6cccc20d6a508bb5ea4).
> 0001 : map.cpp line 2518 size 16 address 0x0x7fcc866a0cdc
> [31/Dec 01:09][Warning]: Memory manager: Memory leaks found and fixed.

mplaten:
> Memory leak also occur when using this command script maprespawnguildid (script in woe_controller.txt)
> 
> create npc script:
> 
[...]

PR #5667, opened 2021-01-03
Fixed memory leak when using loadnpc or reloadnpcfile command

CairoLee:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/5649

* **Server Mode**: Both

Here is small change to fix memory leak when using loadnpc or reloadnpcfile command.


src/map/npc.cpp
```diff
@@ -581,6 +581,7 @@ int npc_event_doall_path( const char* event_name, const char* path ){
 
 	ShowStatus( "Event '" CL_WHITE "%s" CL_RESET "' executed with '" CL_WHITE "%d" CL_RESET "' NPCs.\n", event_name, count );
 
+	if (iter) mapit_free(iter);
```
Lemongrass3110, line 584:
```suggestion
	mapit_free(iter);
```

Lemongrass3110 (approved):

PR #5668, opened 2021-01-03
[Suggestion] Item DB format for Script field

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

Small formatting change of the item DB yaml files to be more human readable.

- [x] Code review
- [ ] Any suggestion ?


Lemongrass3110 (commented):
Looks good to me at first sight!

Only things I came accross were old checks where one if should be inside another, but that is not a problem of this PR. Like if refine > 5 and then another if with refine > 8 should always be inside the other.

Anything special you want us to check?

Atemo:
I was thinking to get rid of the newlines and double spaces when converting to SQL to not update the SQL files if the script format only is changed.

>Only things I came accross were old checks where one if should be inside another, but that is not a problem of this PR. Like if refine > 5 and then another if with refine > 8 should always be inside the other.

You're right, in another PR eventually.

> Anything special you want us to check?

Just need to check if nothing has been broken, but I will do it.

Lemongrass3110:
> 
> 
> I was thinking to get rid of the newlines and double spaces when converting to SQL to not update the SQL files if the script format only is changed.

You could do that temporary locally to compare it. But for the final version it is not too bad to have it.
Should not happen a lot after your master merge anyway.

PR #5719, opened 2021-01-25
Fixes Golden X potion reflect effect

### LINKED ISSUE #5651: Reflect damage should displayed only against reflecting status enemy
* **rAthena Hash**: #5335

* **Client Date**: 20200603

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: 
    During SC_REF_T_POTION turned on, a character always gets 1 damage while attacking, though an enemy does not have any reflecting ability.

  * Expected Result: 
     SC_REF_T_POTION should display reflect damage only against reflecting status enemy.
 
  * How to Reproduce: 
   Maybe add struct block_list *target on int64 battle_calc_return_damage...? But I dont know how. Error occurs when I did.
  * Official Information: 
   
    https://youtu.be/xfaceJtbq40
    
    See 3:49 on video.
    It shows that character gets 1 reflect damage only when Amdarais use Reflect Shield.
  

aleos89:
* **Addressed Issue(s)**: Fixes #5651

* **Server Mode**: Renewal

  * Resolves an issue with Golden X potion causing 1 reflect damage even if the target doesn't have any reflect abilities active.
Thanks to @eppc0330!


eppc0330:
Thanks!

Lemongrass3110 (approved):

src/map/battle.cpp
```diff
@@ -7363,21 +7360,20 @@ int64 battle_calc_return_damage(struct block_list* bl, struct block_list *src, i
...
-			if (rdamage < 1) rdamage = 1;
+			rdamage = i64max(rdamage, 1);
 		}
 	}
 
 	if (sd && sd->bonus.reduce_damage_return != 0) {
 		rdamage -= rdamage * sd->bonus.reduce_damage_return / 100;
-		if (rdamage < 1)
-			rdamage = 1;
+		rdamage = i64max(rdamage, 1);
```
Lemongrass3110, line 7376: Maybe revert this because of #5722

Lemongrass3110 (approved):

PR #5721, opened 2021-01-25
setnpcdisplay potential crash

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

Fixed a potential crash in setnpcdisplay script command + floating npc.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -16855,7 +16855,7 @@ BUILDIN_FUNC(setnpcdisplay)
 	}
 
 	// update npc
-	if( newname && nd->bl.m != -1 )
+	if( newname )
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -16855,7 +16855,7 @@ BUILDIN_FUNC(setnpcdisplay)
 	}
 
 	// update npc
-	if( newname )
+	if( newname && nd->bl.m != -1 )
```
Lemongrass3110, line 16858: Please do not do this here. Do it inside the npc_setdisplayname function please, just in case it is called from somewhere else in the future.

Lemongrass3110 (changes requested):

Lemongrass3110 (approved):

PR #5757, opened 2021-02-25
Removed duplicates RaceGroups mob_db

idk-whoami:
* **Addressed Issue(s)**: 

* **Server Mode**: re


aleos89:
Do you mind applying the fix to the cause too, so that we can fix this in one fell swoop?

Change the header output from:
```cpp
	std::string class_ = "Normal";
	bool header = false;

	for (const auto &race2 : mob_race2) {
		for (const auto &mobit : race2.second) {
			if (mobit == mob_id) {
				// These two races are now Class types
				if (race2.first == RC2_GUARDIAN) {
					class_ = "Guardian";
					continue;
				} else if (race2.first == RC2_BATTLEFIELD) {
					class_ = "Battlefield";
					continue;
				}

				if (!header) {
					body << YAML::Key << "RaceGroups";
					body << YAML::BeginMap;
				}

				body << YAML::Key << name2Upper(constant_lookup(race2.first, "RC2_") + 4) << YAML::Value << "true";

				if (!header) {
					body << YAML::EndMap;
					header = true;
				}
			}
		}
	}

	if (fields[24]) {
		int ele = strtol(fields[24], nullptr, 10);
```

To:
```cpp
	std::string class_ = "Normal";
	bool header = false;

	for (const auto &race2 : mob_race2) {
		for (const auto &mobit : race2.second) {
			if (mobit == mob_id) {
				// These two races are now Class types
				if (race2.first == RC2_GUARDIAN) {
					class_ = "Guardian";
					continue;
				} else if (race2.first == RC2_BATTLEFIELD) {
					class_ = "Battlefield";
					continue;
				}

				if (!header) {
					body << YAML::Key << "RaceGroups";
					body << YAML::BeginMap;
					header = true;
				}

				body << YAML::Key << name2Upper(constant_lookup(race2.first, "RC2_") + 4) << YAML::Value << "true";

			}
		}
	}

	if (header)
		body << YAML::EndMap;

	if (fields[24]) {
		int ele = strtol(fields[24], nullptr, 10);
```

Thanks!

idk-whoami:
Thanks to @aleos89 !
Would you like to add fixed change the header output to this #5757 ?

aleos89:
Yeah, you can go ahead and apply it to this PR.

aleos89 (approved):

PR #5783, opened 2021-03-15
Updated the stats of monsters according to aegis and Divine pride

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: re

Thanks to @Lemongrass3110 !

### DIFF
--- a/db/re/mob_db.yml
+++ b/db/re/mob_db.yml
@@ -82,11 +82,11 @@ Body:
     AegisName: SCORPION
     Name: Scorpion
     Level: 16
-    Hp: 153
-    BaseExp: 108
-    JobExp: 81
+    Hp: 136
+    BaseExp: 169
+    JobExp: 115
     Attack: 33
-    Attack2: 7
+    Attack2: 25
     Defense: 16
     MagicDefense: 5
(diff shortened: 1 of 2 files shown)


eppc0330:
I think mobs' attack is all wrong...
For attack part, perhaps previous version was better.

aleos89:
A source change is required along with the database change for monster attack. The database now contains a "difference" value from ATK1 to ATK2 instead of two direct values.

Lemongrass3110:
@eppc0330 sorry my bad. Used the wrong ATK value. Now it should be better again.

eppc0330:
> @eppc0330 sorry my bad. Used the wrong ATK value. Now it should be better again.

Thanks!!

Lemongrass3110:
Hmm should we merge it?
Otherwise we will not get any feedback anyway. :-)

Needs another SQL sync before that though.

aleos89 (approved):
🔥

Masao87:
Looks great! Thanks for all the hard work you guys put into this :)

Hopefully there will be similar updates with drop rate corrections of certain items and sell / buy values in the item db :D

Issue #5804, opened 2021-03-29
GC_POISONREACT missing skill bonus


eppc0330:
* **rAthena Hash**: master

* **Client Date**: 20210107

* **Server Mode**: Renewal

* **Description of Issue**: 
 1. Counter attack, of SC_POISONREACT when hit by poison element monster, should have chance to attack twice when character's learning TF_DOUBLE, with same rate of TF_DOUBLE.

 2. Calculation of sc->data[SC_POISONREACT]->val2 is wrong. Number of TF_POISON counters should be 1/1/2/2/3/3/4/4/5/5.

  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: http://kafra.kr/#!/ko/KRO/skillinfo/139
  Sorry it's not English but this is the latast skill info.
  

* **Modifications that may affect results**:

PR #5805, opened 2021-03-29
Fix GC_POISONREACT missing skill bonus

eppc0330:
* **Addressed Issue(s)**: #5804 

* **Server Mode**: Renewal

Poison React now have chance to attack twice, with same rate of Double Attack, when character is learning Double Attack.

Number of counters is changed to 1/1/2/2/3/3/4/4/5/5.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3755,6 +3755,7 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
 			if( tsc && tsc->data[SC_JYUMONJIKIRI] )
 				wd->div_ = wd->div_ * -1;// needs more info
 			break;
+#ifdef RENEWAL
 		case AS_POISONREACT:
 			skill_lv = pc_checkskill(sd, TF_DOUBLE);
 			if (skill_lv > 0) {
@@ -3763,6 +3764,7 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
 				}
 			}
 		break;
+#endif
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/map/battle.cpp
```diff
@@ -3755,6 +3755,14 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
 			if( tsc && tsc->data[SC_JYUMONJIKIRI] )
 				wd->div_ = wd->div_ * -1;// needs more info
 			break;
+		case AS_POISONREACT:
```
aleos89, line 3758:
```suggestion
#ifdef RENEWAL
		case AS_POISONREACT:
```

src/map/battle.cpp
```diff
@@ -3755,6 +3755,14 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
...
+			skill_lv = pc_checkskill(sd, TF_DOUBLE);
+			if (skill_lv > 0) {
+				if(rnd()%100 < (7 * skill_lv)) {
+					wd->div_++;
+				}
+			}
+		break;
```
aleos89, line 3765:
```suggestion
		break;
#endif
```

src/map/status.cpp
```diff
@@ -10293,7 +10293,7 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			}
 			break;
 		case SC_POISONREACT:
-			val2=(val1+1)/2 + val1/10; // Number of counters [Skotlex]
+			val2=(val1 - ((val1-1) % 1 - 1)) / 2; // Number of counters [Skotlex]
```
aleos89, line 10296:
```suggestion
#ifdef RENEWAL
			val2=(val1 - ((val1-1) % 1 - 1)) / 2;
#else
			val2=(val1+1)/2 + val1/10; // Number of counters [Skotlex]
#endif
```

aleos89 (changes requested):
A couple small changes to keep a separation between renewal and pre-renewal.

aleos89 (approved):

Issue #5713, opened 2021-01-24
The Sign Quest "Stone Delivery" Part - Stone does not get recognized


Erukanu:
* **rAthena Hash**: Up-to-date (fresh)

* **Client Date**: N/A

* **Server Mode**: Pre-Renewal

* **Description of Issue**: The  "Char name" Stone does not get recognized by the Knight NPC in Yuno
  * Result: Impossible to continue the Quest 
  * Expected Result: Stone gets recognized, and you get the receipt from the Knight 
  * How to Reproduce: Simply do the Quest until you reach that point.
  * Official Information: https://irowiki.org/wiki/Sign_Quest 
  

* **Modifications that may affect results**:  in Line 6827 
change 
set .@sign3, .@signid & 65535; 
into
set .@sign3, .@signid;

This fixes the problem since the Steward insert the real char id into the card slot 2 of the Stone item

gengstergile:
yeap, i also got this issue, i will try your solution and check back

Erukanu:
> yeap, i also got this issue, i will try your solution and check back

Any news? ^^


Also can anyone of the higher ups confirm this ?^^

KrokusPokus:
I can confirm this bug. It is caused by line 6827 in "npc\quests\the_sign_quest.txt":
`set .@sign3, .@signid & 65535;`

Reason for causing the problems:
Variable .@sign3 in this case is supposed to contain the char_id which will be handed over to the countitem2 function.
A long time ago, char_id used to be of "long" type. Nowadays, char_id is of type "uint32". Since the script copies only the lower 16 bits of the char_id, the countitem2 function doesn't recognize the customized item as belonging to this character. Which of course leads to the NPC claiming it's the wrong item...

Simplest fix: Change line to
`set .@sign3, .@signid;`

Erukanu:
> I can also confirm this bug. As mentioned above, it's caused by line 6827 in "npc\quests\the_sign_quest.txt":
> `set .@sign3, .@signid & 65535;`
> 
> Reason for causing the problems:
> Variable .@SIGN3 in this case is supposed to contain the char_id which will be handed over to the countitem2 function.
> A long time ago, char_id used to be of "long" type. Nowadays, char_id is of type "uint32". Since the script copies only the lower 16 bits of the char_id, the countitem2 function doesn't recognize the customized item as belonging to this character. Which of course leads to the NPC claiming it's the wrong item...

[...]

PR #5807, opened 2021-03-29
Fix getnameditem command

Balferian:
* **Addressed Issue(s)**:  https://github.com/rathena/rathena/issues/5713
* **Server Mode**: -
* **Description of Pull Request**: Fix for getnameditem command.
The command must assign a character name to the item as it saw in command description.
Some quests cannot be completed without assigned item.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -7834,8 +7834,8 @@ BUILDIN_FUNC(getnameditem)
 	item_tmp.amount=1;
 	item_tmp.identify=1;
 	item_tmp.card[0]=CARD0_CREATE; //we don't use 255! because for example SIGNED WEAPON shouldn't get TOP10 BS Fame bonus [Lupus]
-	item_tmp.card[2]=tsd->status.char_id & 65535;
-	item_tmp.card[3]=tsd->status.char_id >> 16;
+	item_tmp.card[2]=GetWord(tsd->status.char_id,0);
+	item_tmp.card[3]=GetWord(tsd->status.char_id,1);
 	if(pc_additem(sd,&item_tmp,1,LOG_TYPE_SCRIPT)) {
 		script_pushint(st,0);
 		return SCRIPT_CMD_SUCCESS;	//Failed to add item, we will not drop if they don't fit
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -7834,7 +7834,7 @@ BUILDIN_FUNC(getnameditem)
 	item_tmp.amount=1;
 	item_tmp.identify=1;
 	item_tmp.card[0]=CARD0_CREATE; //we don't use 255! because for example SIGNED WEAPON shouldn't get TOP10 BS Fame bonus [Lupus]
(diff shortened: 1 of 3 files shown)


KrokusPokus:
This will fix the bug for chars who aren't stuck already, but chars stuck for example in the Sign Quest will stay that way because they can't reaquire a new item 7049 with fixed (truncated) char_id in card2 member.

Maybe it would be better to keep using the untruncated char_id (instead of splitting into a high and low part) and change scripts accordingly (item struct's "card" member is t_itemid which is uint32). Afaik, the Sign Quest's line 6827 is the only place where a char_id split into two 16 bit parts is assumed, so we might possibly just store the full value in getnameditem like we do now and plop the full value back into countitem2 and remove the parts relating to the high part/low part separation.

Balferian:
> This will fix the bug for chars who aren't stuck already, but chars stuck for example in the Sign Quest will stay that way because they can't reaquire a new item 7049 with fixed (truncated) char_id in card2 member.
> 
> Maybe it would be better to keep using the untruncated char_id (instead of splitting into a high and low part) and change scripts accordingly (item struct's "card" member is t_itemid which is uint32). Afaik, the Sign Quest's line 6827 is the only place where a char_id split into two 16 bit parts is assumed, so we might possibly just store the full value in getnameditem like we do now and plop the full value back into countitem2.

To fix bugs with items apply this query:
`UPDATE inventory SET card0 = 254, card2 = char_id & 65535, card3 = char_id >> 16 WHERE nameid = 7049;`

src/map/script.cpp
```diff
@@ -7834,7 +7834,7 @@ BUILDIN_FUNC(getnameditem)
 	item_tmp.amount=1;
 	item_tmp.identify=1;
 	item_tmp.card[0]=CARD0_CREATE; //we don't use 255! because for example SIGNED WEAPON shouldn't get TOP10 BS Fame bonus [Lupus]
-	item_tmp.card[2]=tsd->status.char_id;
+	item_tmp.card[2]=tsd->status.char_id & 65535;
 	item_tmp.card[3]=tsd->status.char_id >> 16;
```
Lemongrass3110, line 7838:
```suggestion
	item_tmp.card[2]=GetWord(tsd->status.char_id,0);
	item_tmp.card[3]=GetWord(tsd->status.char_id,1);
```

Lemongrass3110 (changes requested):
Please also fix:
atcommand.cpp::getring

Additionally please add the following upgrade script into the sql upgrades:
```
UPDATE `inventory` SET `card2` = `card2` & 65535 WHERE `card0` = 254 OR `card0` = 255;
UPDATE `cart_inventory` SET `card2` = `card2` & 65535 WHERE `card0` = 254 OR `card0` = 255;
UPDATE `storage` SET `card2` = `card2` & 65535 WHERE `card0` = 254 OR `card0` = 255;
UPDATE `guild_storage` SET `card2` = `card2` & 65535 WHERE `card0` = 254 OR `card0` = 255;
```

Lemongrass3110 (approved):

PR #5809, opened 2021-03-30
Corrections in card_seller.txt

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

Corrections in `card_seller.txt`
--------------------------------------------------------------------------------------
* The script only checks the slot 10 but the card can be in any slot.
* When not defined `mode_mvp` field is supposed to be NULL in `mob_db_re.sql`.

### DIFF
--- a/npc/custom/card_seller.txt
+++ b/npc/custom/card_seller.txt
@@ -42,16 +42,16 @@ OnInit:
 	}
 	freeloop 1;
 	
-	.@string$ = "( `name_aegis` IN ( SELECT DISTINCT `drop1_item` FROM `" + .@mob_db$ + "` WHERE `drop1_nosteal` = 1 AND `mode_mvp` IS NULL ) " +
-		"OR `name_aegis` IN ( SELECT DISTINCT `drop2_item` FROM `" + .@mob_db$ + "` WHERE `drop2_nosteal` = 1 AND `mode_mvp` IS NULL ) " +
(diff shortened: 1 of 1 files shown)


npc/custom/card_seller.txt
```diff
@@ -40,10 +40,22 @@ OnInit:
 		.@mob_db$  = "mob_db";
 		.@item_db$ = "item_db";
 	}
-	freeloop 1;	
-	.@total = query_sql( "SELECT DISTINCT LEFT( `name_english`, 1 ) AS alphabets FROM `"+ .@item_db$ +"` WHERE `type` = 'Card' AND `name_aegis` IN ( SELECT DISTINCT `drop10_item` FROM `"+ .@mob_db$ +"` WHERE `mode_mvp` = 0 ) ORDER BY alphabets;", .@alphabet$ );
+	freeloop 1;
+	
+	.@string$ = "( `name_aegis` IN ( SELECT DISTINCT `drop1_item` FROM `" + .@mob_db$ + "` WHERE `drop1_nosteal` = 1 AND `mode_mvp` IS NULL ) " +
```
Lemongrass3110, line 45: Why the nosteal condition?

As for MVP I would suggest doing it with `COALESCE(mode_mvp,0) = 0` instead, just in case someone does not only put NULL or 1 into it.

Atemo: The nosteal is a leftover from a old query

Lemongrass3110 (approved):

PR #5811, opened 2021-03-31
Add support for new cart design

Balferian:
* **Addressed Issue(s)**:  -
* **Server Mode**:  -
Added support for new cart design.
image

### DIFF
--- a/src/common/mmo.hpp
+++ b/src/common/mmo.hpp
@@ -1082,7 +1082,11 @@ struct clan{
 	#warning "Cart won't be able to be saved for relog"
 	#endif
 #if PACKETVER >= 20150826
-	#define MAX_CARTS 12		// used for 3 new cart design
+	#if PACKETVER >= 20191106
+		#define MAX_CARTS 13		// used for new cart design
+	#else
+		#define MAX_CARTS 12		// used for 3 new cart design
+	#endif
 #else
 	#define MAX_CARTS 9
 #endif
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -12388,6 +12388,9 @@ void clif_parse_ChangeCart(int fd,struct map_session_data *sd)
 
 	if( 
 #ifdef NEW_CARTS
+#if PACKETVER >= 20191106
+		(type == 13 && sd->status.base_level > 100) ||
+#endif
 		(type == 9 && sd->status.base_level > 130) ||
 		(type == 8 && sd->status.base_level > 120) ||
 		(type == 7 && sd->status.base_level > 110) ||


Lemongrass3110:
Is this really available in changecart?
All newer carts have been added to MC_CARTDECORATE and it would automatically be available if you increase MAX_CART as this is respected in clif_parse_SelectCart...

Balferian:
> Is this really available in changecart?
> All newer carts have been added to MC_CARTDECORATE and it would automatically be available if you increase MAX_CART as this is respected in clif_parse_SelectCart...

> if you increase MAX_CART

Are you sure?
Without it cart don`t changed.

> #if PACKETVER >= 20191106
> 		(type == 13 && sd->status.base_level > 100) ||
> #endif

Also new cart appears in the list of MC_CHANGECART when player reach 100 level.
image

Lemongrass3110:
It was only a question from my side. I did not get to test it on kRO already.
Is this cart available for genetic too? Or restricted to mechanic?
Based on the look I think it would look strange on genetic... But who knows.

Well however as you pointed out correctly, then we need to change the clif_parse_SelectCart code though...
And limit it to cart type 10-12.

Balferian:
> It was only a question from my side. I did not get to test it on kRO already.
> Is this cart available for genetic too? Or restricted to mechanic?
> Based on the look I think it would look strange on genetic... But who knows.
> 
> Well however as you pointed out correctly, then we need to change the clif_parse_SelectCart code though...
> And limit it to cart type 10-12.

Yes. genetic can choose it too.
Check this: Video from kRo: https://youtu.be/wjxx2M-EgPA?t=248

src/common/mmo.hpp
```diff
@@ -1082,7 +1082,11 @@ struct clan{
 	#warning "Cart won't be able to be saved for relog"
 	#endif
 #if PACKETVER >= 20150826
```
Lemongrass3110, line 1084:
```suggestion
#if PACKETVER >= 20191106
	#define MAX_CARTS 13		// used for another new cart design
#elif PACKETVER >= 20150826
```

src/common/mmo.hpp
```diff
@@ -1082,7 +1082,11 @@ struct clan{
...
-	#define MAX_CARTS 12		// used for 3 new cart design
+	#if PACKETVER >= 20191106
+		#define MAX_CARTS 13		// used for new cart design
+	#else
+		#define MAX_CARTS 12		// used for 3 new cart design
+	#endif
```
Lemongrass3110, line 1089:
```suggestion
	#define MAX_CARTS 12		// used for 3 new cart design
```

Lemongrass3110 (approved):

admkakaroto:
a1
Using Baby Genetic, client 2021-01-07aRagexeRE, this showing like the image.

Duble new cart in select window and showing more than 1 at the some time in game e.e

Balferian:
> a1
> Using Baby Genetic, client 2021-01-07aRagexeRE, this showing like the image.
> 
> Duble new cart in select window and showing more than 1 at the some time in game e.e

Look like it is client issue.
Tested on 2020-09-06b:
image

Lemongrass3110:
Just wanted to say. This has to be a clientside issue.
Because the list of available carts is not sent from the server to the client.

Issue #5766, opened 2021-03-02
Oddities between rAthena and aegis with gear swapping and status effects granted by the items


mrjnumber1:
(Aleos said it was OK that I'm not using the template here because of how similar all of these reports are. Sorry if this messes up any bots or annoys you)

On aegis, statuses are granted when gears are put on and then removed. This causes a few behavior discrepancies with rAthena and odd behaviors in-game for players.

### SC_End used incorrectly in UnequipScript for certain items
When the player is wearing a card that grants these skills, uses the skill, and removes the item, the server will disable the status. It turns out, this is a hardcoded check only for MG_SIGHT, TF_HIDING, and AS_CLOAKING 
```
case FUNC_DISABLESKILL:
  m_pcSkill->DeleteSkillFromItemServer((WORD)reqMsg->par1, reqMsg->par2);
  m_pcSkill->NotifyAllSkillInfo()
  if (reqMsg->par1 == SKID_MG_SIGHT)
    ResetEffectState(EFFECTSTATE_SIGHT);
  else if (reqMsg->par1 == SKID_TF_HIDING)
    ResetEffectState(EFFECTSTATE_BURROW);
  else if (reqMsg->par1 == SKID_AS_CLOAKING)
    ResetEffectState(EFFECTSTATE_HIDING);
  return;
```
**Expected Behavior:** User keeps EFST_AUTOGUARD and EFFECTSTATE_RUWACH and EFST_ANGELUS+EFFECTSTATE_ANGELUS when removing Clock_Card/Satellite_Hairband/Bell_Ribbon https://streamable.com/uim50y
**Behavior:** User loses these effects. There may be other SC_Ends used in the item DB, which are incorrect, that I did not find. Any item that grants MG_SIGHT/TF_HIDING/AS_CLOAKING must always disable them on switching off (this seems to be correct in our DB at least)

### bonus2 bSpeedRate,25;
This applies to items such as Brood_Axe, Sleipnir, Moonlight_Flower_Card, etc
**Expected Behavior**: Receive EFST_MOVHASTE_INFINITY val1 25 and get the cute AGI up effect on the client 
**Behavior**: We just get a silent speed buff
(Interestingly, when the user wears Brood_Axe + Sleipnir, the efst is removed ... but the user still has the speed boost. Well, whatever. This means rathena has this weird interaction proper woot!)

### Maya Purple Card / Box_of_Sunlight interactions
**Expected Behavior:**
1. Put on Maya_Puple_Card hat
[...]

PR #5816, opened 2021-03-31
Adds SC_MOVEHASTE_INFINITY status

aleos89:
* **Addressed Issue(s)**: Partial fix for #5766.

  * Removes bSpeedRate,25 bonuses for several items which should be applied via a status change.
Thanks to @mrjnumber1!


eppc0330:
I have applied this SC earlier and I've got some issues.

First, suggestion: 
```
if(getstatus(SC_MOVEHASTE_INFINITY)==0) sc_start SC_MOVEHASTE_INFINITY,INFINITE_TICK,25;
```

Without status-on checking, client displays speedup text too much, like belows:

ALL EXPERIMENT SCREENSHOTS ARE FROM MY RATHENA SERVER.
1. Everytime when take on/off equipments.
movhaste01

2. Everytime transform's on/off.
movhaste02

3. On log-in.
When equip one movhaste_infinity item.
movhastelogin

When equip two movhaste_infinity item.
movhastelogin01
 
And so on.


Also Rathena's something code, sorry I don't know exactly what it is, is different from official.

With more than two MOVHASTE_INFINITY items equipped, when take off one of them EFST is turned off - even though speed bonus of MOVHASTE_INFINITY is still working - until equip something or relogin. Check below:
movhaste03

eppc0330:
Simple and nice stacking check!

eppc0330:
I need we need something on/off check

As I comented before, changed script still has same problem.

CASE 1: Client displays speedup text too much.

Everytime when take on/off equipments.
movhaste01

Everytime transform's on/off.
movhaste02

On log-in.
When equip one movhaste_infinity item.
movhastelogin

When equip two movhaste_infinity item.
movhastelogin01

And so on.

This part. Not fixed yet.

eppc0330:
Sorry, still texting too much message after [fb9a524](https://github.com/rathena/rathena/pull/5816/commits/fb9a52416f7fb261aa89a0233f52984f2f43220b)

eppc0330:
\Im not sure this will help but in KRo when equipping movhaste item,
speedup message shown before equipped message

movhasteofficial02

White line is speed message. Yellow line is equip on/off message

aleos89:
Thanks for the image, but correct me if I'm wrong since I'm not at a PC, but in rAthena we equip the item and then do a status calculation so that's why the equip and bonus messages are flipped.

eppc0330:
@aleos89 Yes I think you are right. The problem is doing with the reason you explained.

PR #5822, opened 2021-04-01
Fixes monster Sharp Shooting damage formula

### LINKED ISSUE #5769: Mob casted SN_SHARPSHOOTING should be a lot stronger
* **rAthena Hash**: latest
* **Client Date**: x
* **Server Mode**: both!
* **Description of Issue**: 
  * Result: Mobs are getting the same damage output on SN_SHARPSHOOTING as players
  * Expected Result: Do proper damage!
  * How to Reproduce: summon a Cecil and get wrecked, I suppose
  * Official Information: Mob casted SN_SHARPSHOOTING damage is `140 * skLevel (140-700%)` as long as it hits in its 8path splash. When it uses the fallback single target it, the original ratio `(250-450%)` is used. Recently the Ratio for RE was updated, so this will require further testing that's unavailable to me, sorry. The best solution for this I can think of is to give mobs the 8path splash, and flag when it's a single skill_attack to switch to the player/merc ratio. For Pre-RE this can be fixed right away but for Renewal we need more info..

* **Modifications that may affect results**: x

aleos89:
* **Addressed Issue(s)**: Fixes #5769

  * Adjusts the formula for Sharp Shooting if a monster uses the skill.
Thanks to @mrjnumber1!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -4109,8 +4109,8 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 		case SN_SHARPSHOOTING:
 		case MA_SHARPSHOOTING:
 			if (src->type == BL_MOB) { // TODO: Did these formulas change in the renewal balancing?
-				if (battle_config.skill_eightpath_algorithm == 1)
-					skillratio += 40 * skill_lv;
+				if (wd->miscflag & 1) // Splash damage bonus
+					skillratio += -100 + 140 * skill_lv;
 				else
 					skillratio += 100 + 50 * skill_lv;
 			} else {
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -4107,20 +4107,21 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 #endif
 			break;
 		case SN_SHARPSHOOTING:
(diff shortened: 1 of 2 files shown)


Lemongrass3110 (approved):

src/map/battle.cpp
```diff
@@ -4108,12 +4108,19 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 			break;
 		case SN_SHARPSHOOTING:
 		case MA_SHARPSHOOTING:
+			if (src->type == BL_MOB) { // TODO: Did these formulas change in the renewal balancing?
+				if (battle_config.skill_eightpath_algorithm == 1)
+					skillratio += 40 * skill_lv;
+				else
+					skillratio += 100 + 50 * skill_lv;
```
vstumpf, line 4115: I'm not sure this is right, I think @mrjnumber1 was saying the damage increases if the target is in an eight-path direction from the source.

If the source is at (0,0), a target at (4,4) will take extra damage, and a target at (7, 3) will take less damage.

aleos89: Good point! I didn’t read it like that!

Lemongrass3110 (approved):
Seeing all the ifs for skill_id == SN_SHARPSHOOTING I would suggest thinking about creating its own case and duplicate the code maybe...

PR #5827, opened 2021-04-02
Fixes an example for script command makeitem3

### LINKED ISSUE #5796: Slight oversight in script_commands.txt makeitem3 documentation example
The example currently says: 
```
Example to get Crimson Weapon with Ghost property:
	// 0.5% chance to get +0 Valkyrie Shield [1]
	// with Neutral Resistance +10% and 5% damage reduction from Demi-Human or Player
	// when Valkyrie Randgris killed
```
But the code says:
https://github.com/rathena/rathena/blob/c8567f31ec6cb06aae627c762acaa32f9b94a7ec/doc/script_commands.txt#L4898

However:
1000-950 = 50
50/1000 = 5/100 = 5%

Which is different from the 0.5% stated in the text from the example. 
The actual drop chance is still slightly lower because rand(0,1000) is inclusive of both limits.

Suggested change:
`if (killedrid == 1751 && rand(1,10000) > 950) { // Valkyrie Randgris`

Since the range 1-10000 follows current rAthena conventions for percentages in other places

Edit: I've started forking and making a pull request but I think that for an one-liner it's not really worth it.

aleos89:
* **Addressed Issue(s)**: Fixes #5796

  * Fixes a small typo in the documentation for script command makeitem3.
Thanks to @jaBote!

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -4895,7 +4895,7 @@ Example to get Crimson Weapon with Ghost property:
 	// with Neutral Resistance +10% and 5% damage reduction from Demi-Human or Player
 	// when Valkyrie Randgris killed
 	OnNPCKillEvent:
-		if (killedrid == 1751 && rand(0,10000) > 950) { // Valkyrie Randgris
+		if (killedrid == 1751 && rand(0,10000) > 9950) { // Valkyrie Randgris
 			getmapxy(.@map$,.@x,.@y,BL_PC);
 			setarray .@OptID[0],RDMOPT_ATTR_TOLERACE_NOTHING,RDMOPT_RACE_TOLERACE_HUMAN;
 			setarray .@OptVal[0],10,5;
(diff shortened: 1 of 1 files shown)


doc/script_commands.txt
```diff
@@ -4895,7 +4895,7 @@ Example to get Crimson Weapon with Ghost property:
 	// with Neutral Resistance +10% and 5% damage reduction from Demi-Human or Player
 	// when Valkyrie Randgris killed
 	OnNPCKillEvent:
-		if (killedrid == 1751 && rand(0,1000) > 950) { // Valkyrie Randgris
+		if (killedrid == 1751 && rand(0,10000) > 950) { // Valkyrie Randgris
```
vstumpf, line 4898: Shouldn't this be
```suggestion
		if (killedrid == 1751 && rand(0,10000) > 9950) { // Valkyrie Randgris
```

Atemo (approved):

Issue #5836, opened 2021-04-06
TK_POWER shouldn't count actual player


Indigo000:
* **rAthena Hash**: fc8e63a2af306de18b9bd0c749429a39f49a7ea9

* **Client Date**: 2018-06-21

* **Server Mode**: Both

* **Description of Issue**: TK_POWER currently includes the current player in the count while it shouldn't. There has been a check that was intended to prevent this (e7219585edc70153f8d029a67fc53ec42cb15a43) but all this does is prevent a solo party from being included. If there's more than one player in a party, then all of them get included.
  * Result:  In a party of more than 1, all members are included in TK_POWER/Kihop. It should include the Kihop user.
  * Expected Result:  A party of 2 members should give only a 10% ATK boost instead of the current 20%.
  * How to Reproduce:  Get a party of more than 1 and check the damage.
  * Official Information:  Not a great source https://irowiki.org/classic/Kihop but also commit e7219585edc70153f8d029a67fc53ec42cb15a43 which seems to indicate that the user should not be included.
  

* **Modifications that may affect results**:

Tydus1:
Just reminding to close this issue since a fix commit is already pushed into master branch.

Daegaladh:
Fixed in #5838

PR #5838, opened 2021-04-06
Fix TK_POWER Attack Bonus

Tydus1:
* **Addressed Issue(s)**: #5836 

* **Server Mode**: Both

Multiplier value based on party members count doesn't include own player

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3632,7 +3632,8 @@ static void battle_calc_skill_base_damage(struct Damage* wd, struct block_list *
 				}
 #endif
 				if(sd->status.party_id && (skill=pc_checkskill(sd,TK_POWER)) > 0) {
-					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Tydus1][Inkfish] 
+					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Inkfish]
+						// Reduce count by one (self) [Tydus1]
 						i -= 1; 
 						ATK_ADDRATE(wd->damage, wd->damage2, 2*skill*i);
 						RE_ALLATK_ADDRATE(wd, 2*skill*i);
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3632,7 +3632,7 @@ static void battle_calc_skill_base_damage(struct Damage* wd, struct block_list *
 				}
 #endif
 				if(sd->status.party_id && (skill=pc_checkskill(sd,TK_POWER)) > 0) {
-					i = max(party_foreachsamemap(party_sub_count, sd, 0) - 1, 0); // Bonus Attack multiplier based on party members count doesn't include own player [Tydus1] [Inkfish]
+					i = party_foreachsamemap(party_sub_count, sd, 0) - 1; // Bonus Attack multiplier based on party members count doesn't include own player [Tydus1] [Inkfish]
 					if( i > 0 ) { 
 						ATK_ADDRATE(wd->damage, wd->damage2, 2*skill*i);
 						RE_ALLATK_ADDRATE(wd, 2*skill*i);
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -3632,7 +3632,8 @@ static void battle_calc_skill_base_damage(struct Damage* wd, struct block_list *
 				}
 #endif
 				if(sd->status.party_id && (skill=pc_checkskill(sd,TK_POWER)) > 0) {
-					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Inkfish]
+					i = party_foreachsamemap(party_sub_count, sd, 0) - 1; // Bonus Attack multiplier based on party members count doesn't include own player [Tydus1] [Inkfish]
```
NeutralDev, line 3635: This actually has potential to break. If somehow **i** was ever equal 0, which normally wouldn't happen in this case, but you never know! You would be subtracting 1 from 0 making it -1. Under normal circumstances wouldn't be a big deal for the below if statement. However this case is different because we're returning an int from **party_foreachsamemap** to a **uint16 i**, which makes it unsigned, therefore it would underflow and equal 65535 instead of -1. I want to recommend the following change so this underflow can't possibly happen and would always cap at 0 before going negative:

`i = max(party_foreachsamemap(party_sub_count, sd, 0) - 1, 0); // Bonus Attack multiplier based on party members count doesn't include own player [Tydus1] [Inkfish]`

Tydus1: Done! Thanks for the review @NeutralDev

src/map/battle.cpp
```diff
@@ -3632,7 +3632,8 @@ static void battle_calc_skill_base_damage(struct Damage* wd, struct block_list *
 				}
 #endif
 				if(sd->status.party_id && (skill=pc_checkskill(sd,TK_POWER)) > 0) {
-					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Inkfish]
+					i = max(party_foreachsamemap(party_sub_count, sd, 0) - 1, 0); // Bonus Attack multiplier based on party members count doesn't include own player [Tydus1] [Inkfish]
+					if( i > 0 ) {
```
Lemongrass3110, line 3636: I think this is overcomplicating things.
Just keep it the old way and reduce the variable i inside the if, because there it is always > 1 in the current code.

Lemongrass3110, line 3636:
```suggestion
					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Inkfish]
						// Bonus Attack multiplier based on party members count doesn't include own player [Tydus1]
						i -= 1; 
```

Tydus1:
It was a good observation by @NeutralDev in my opnion, although so far it's not possible to get a negative value from decrement ( since party_foreachsamemap will always return >1 ). Well, suggestions are there! I appreciate the reviews!

src/map/battle.cpp
```diff
@@ -3632,7 +3632,8 @@ static void battle_calc_skill_base_damage(struct Damage* wd, struct block_list *
 				}
 #endif
 				if(sd->status.party_id && (skill=pc_checkskill(sd,TK_POWER)) > 0) {
-					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Inkfish]
+					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Tydus1][Inkfish] 
+						i -= 1;
```
Lemongrass3110, line 3636:
```suggestion
					if( (i = party_foreachsamemap(party_sub_count, sd, 0)) > 1 ) { // exclude the player himself [Inkfish]
						// Reduce count by one (self) [Tydus1]
						i -= 1; 
```

Lemongrass3110 (approved):

aleos89 (approved):

PR #5860, opened 2021-04-14
Added missing etc item according to KRO description

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal


RagnaWay:
Also add **23899**, it is being used in Attendance kRO.
If you have a question about item **25791**, which you also added, you don't need any function in that item to use it?

Atemo:
23899 is a consumable, I only add the etc. item here.
25791 is an etc item, this type of item doesn't support a script. The item is probably exchanged to the npc.

PR #5901, opened 2021-04-27
Add new jobs supports to Modified Temporal Boots

inhyositsu:
* **Addressed Issue(s)**: Issue #4734 

* **Server Mode**: Renewal

* **Description of Pull Request**: Fix the issue described.


reunite-ro:
does we need structure for itemdb related sql table also ?

inhyositsu:
> does we need structure for itemdb related sql table also ?

For sure!
I didn't notice about the sql itemdb cause i'm not using this at all

aleos89:
You're also missing the addition of the two new columns on itemdb.cpp line 2232.

Lemongrass3110:
@inhyositsu sadly it is not as easy as this. Athena internally uses multiple masks to administrate which jobs can use which items.

First mask: Which basejob can use it.
Second mask: Which 2-1 job can use it.
Third mask: Which 2-2 job can use it.

And now to the mask that is creating a problem with this.
Fourth mask: Which "classes" can use it. (Better known as the "Classes" entry of the database, which determines if normal jobs, upper jobs, baby jobs, third jobs, etc. can use an item)

Here is the problem now:
You want an item to be available to only 3rd jobs (therefore you would have to set Classes of All_Third to true.
But this actually means that anything that is not a 3rd job, cannot use this item.

Figured out the problem already?
If not let me continue: Since we need KagerouOboro and Rebellion, which are both considered 2-1 jobs of Class "normal", but also Star Emperor and Soul Reaper, which are considered as 3rd class version of the 2-1 jobs Star Gladiator and Soul Linker, we cannot support this at the same time with the current system. Since if you add the limitation to 3rd class you will block KagerouOboro and Rebellion, but on the other hand if you remove the limitation to 3rd class you will allow it for "normal" Star Gladiators and Soul Linkers as well.

I guess you might understand now, why we did not want to rewrite the whole system for a hand full of items.

Lemongrass3110:
Since @inhyositsu reacted with a thumbs up I guess he understood my points.
I am closing this for now.
If you really want to work on this feel free to make another pull request.

Thank you for your contribution anyway! Please do not let this demotivate you. :-)

Issue #5941, opened 2021-05-15
script command getexp/getexp2 not support int64


Angelic234:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/58f2d2173f604296f0f72d2eb92afda879c3cfa1

* **Client Date**: ANY

* **Server Mode**: BOTH

* **Description of Issue**: 
  * Result: 
  * Expected Result: 
  * How to Reproduce: 
easy test npc
```
prontera,163,189,3	script	testexp	1_M_KNIGHTMASTER,{
	.@test1 = 7000000000;
	.@test2 = 2000000000;
	mes "" + .@test1;
	mes "" + .@test2;
	getexp .@test1,.@test2;
	getexp2 .@test1,.@test2;
	end;
}
```
  * Official Information: 
  

* **Modifications that may affect results**:

PR #5942, opened 2021-05-15
Fixed int64 support for some exp commands

Lemongrass3110:
* **Addressed Issue(s)**: #5941

* **Server Mode**: Both

Added int64 support to some missed commands.

Thanks to @Angelic234

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -10492,23 +10492,23 @@ BUILDIN_FUNC(makepet)
 BUILDIN_FUNC(getexp)
 {
 	TBL_PC* sd;
-	int base=0,job=0;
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -10492,23 +10492,23 @@ BUILDIN_FUNC(makepet)
 BUILDIN_FUNC(getexp)
 {
 	TBL_PC* sd;
-	int base=0,job=0;
+	int64 base=0,job=0;
```
aleos89, line 10495: Guess these could be defined below with the parse or remove the assignment here.

aleos89 (approved):

aleos89 (approved):

PR #5949, opened 2021-05-18
Remove delayed consumption when switching types

### LINKED ISSUE #5937: Item Import Cannot Overwrite Type
* **rAthena Hash**: 16b4cbf

* **Client Date**: 20200401

* **Server Mode**: RENEWAL

* **Description of Issue**: 
  * Result: Not Overwrite
  * Expected Result: Overwrite
  * How to Reproduce: 

Using this on import

```
  - Id: 12599
    AegisName: Kurma
    Name: Kurma
    Type: Healing
    Buy: 20
    Weight: 100
    Script: |
      callfunc "F_RamadhanBuff";
```

The item will still be Type: Delay Consume instead of Healing. It should be Healing. 

  * Official Information: 
  

aleos89:
* **Addressed Issue(s)**: Fixes #5937

  * Removed the delayed consumption flag when switching item types.
Thanks to @mazvi!

### DIFF
--- a/src/map/itemdb.cpp
+++ b/src/map/itemdb.cpp
@@ -103,8 +103,7 @@ uint64 ItemDatabase::parseBodyNode(const YAML::Node &node) {
 			constant = IT_USABLE;
 			item->flag.delay_consume |= DELAYCONSUME_TEMP;
 		} else {
-			if (exists && item->flag.delay_consume & DELAYCONSUME_TEMP) // Remove delayed consumption flag if switching types
-				item->flag.delay_consume &= ~DELAYCONSUME_TEMP;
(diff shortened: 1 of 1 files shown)


mazvi:
My problem is not only about delay, but the item doesn't decrease when eaten. because iam not using itemskill.

Original file: DelayConsume
Import file: Healing

Script: not itemskill 

item doesn't decrease when eaten

Lemongrass3110:
That is what the delay flag is doing...
Have you tested the PR before commenting?

Lemongrass3110 (approved):

Daegaladh (approved):

Issue #5971, opened 2021-05-25
Delete/Reset Item Drop Index Entries


Everade:
**The Problem**
The mob_db.yml doesn't seem to support any kind of **Drop removal** _from an existing index_.
That means the only way to get rid of an actual item drop is by modifying the original mob_db.yml, which of course is an undesirable outcome.

**Describe the solution you'd like**
a.) Remove single drop by defining the index number you would like to remove.
     For example with an empty entry:
     `    Drops:
      - Item:
        Rate:
        Index: 0`

     or with a single command like
     `- Index: Reset 1`

b.) Remove the entire index for the selected mob id, then continue by defining your own drops (or not).
     For example:
     `    Drops:
      - Index: Reset All
      #Optional continue here with your own index
      - Item: Apple
         Rate: 10`

**Additional context**
It would make it much more flexible to customize existing drops without ever touching the original database.

**Some common examples:**
Monster has 10 items in its index, but you would like to remove 3 of them entirely.
Monster has 6 item in its index and its your goal to remove them all.
Monster has 5 items in its index and you're planning to replace 2 and remove 1 entirely.

PR #5974, opened 2021-05-25
Added more support for clearing

Lemongrass3110:
* **Addressed Issue(s)**: #5971

* **Server Mode**: Both

Support clearing for all database keys.
Additionally added support for dropping monster drops.
If you find more nodes that are missing support please report them.
Thanks to @Everade


src/common/database.cpp
```diff
@@ -150,7 +150,11 @@ void YamlDatabase::parse( const YAML::Node& rootNode ){
 		const char* fileName = this->currentFile.c_str();
 
 		for( const YAML::Node &node : bodyNode ){
-			count += this->parseBodyNode( node );
+			if( this->nodeExists( node, "Remove" ) ){
```
Atemo, line 153: The doc is missing for the user to understand how to use it (maybe add it for each yml doc)

Everade:
There's still an issue on triggering a full clear the MvpDrops.
The total index of 0-2 doesn't seem to clear the index **gaps**, so in this example below only 1 of the 3 declared items will actualy show up ingame.

`  - Id: 3621
    Name: Pet Child
    MvpDrops:
      - Clear: true
      - Item: Old_Blue_Box
        Rate: 50
      - Item: Old_Violet_Box
        Rate: 50
      - Item: Yggdrasilberry
        Rate: 500
    Drops:
      - Clear: true
      - Item: Ultra_Light_MShield
        Rate: 10
      - Item: Treasure_Box
        Rate: 100
      - Item: Old_Violet_Box
        Rate: 40
      - Item: Old_Blue_Box
        Rate: 40
      - Item: Yggdrasilberry
        Rate: 300
      - Item: Carnium
        Rate: 50
      - Item: Bradium
        Rate: 50
        StealProtected: true`

Everade:
Missing yaml2sql converter update.
Currently doesn't support `- Remove`

Issue #5906, opened 2021-04-28
Acid Demonstration should NOT proc Strip Armor against mobs


Tydus1:
* **rAthena Hash**: Latest rAthena

* **Client Date**: 20210401

* **Server Mode**: Pre-Renewal

* **Description of Issue**: 
  * Result: Acid Demonstration is inflicting Strip Armor when used on a mob. It results in a lowered damage after the status.
  * Expected Result: Acid Demonstration should NOT inflict Strip Armor when used against mobs.
  * How to Reproduce: Create a Creator and bomb Golden Acidus(ID: 1713). When Strip Armor procs you will notice AD will deal less damage.
  * Note: AD damage calculation looks fine since strip armor on mobs reduce their vit and then the AD damage. The point is AD should not proc it by itself (against mobs).
  * Official Information: The profs provided here were obtained from a discursion on OriginRO discord (Thanks to Xellie). The images below are a supposedly test on Aegis. (Would be nice if anyone with Aegis access could confirm)
![screenXellieRO020](https://user-images.githubusercontent.com/16859064/116485926-d6dcfb80-a862-11eb-8a79-2633c81d105e.jpg)
![screenXellieRO021](https://user-images.githubusercontent.com/16859064/116485934-dba1af80-a862-11eb-8fcc-d913d3e65440.jpg)

  

* **Modifications that may affect results**:

Xelliekins:
Massive apologies for the huge delay: 

https://gitlab.com/Xellie/aegis-verification/-/issues/9

In summary:
Armor break seems to simply not work on mobs. 

Skills tested

- Acid Terror
- Tarot Card of Fate (Chariot)
- Head Crusher

Item
 - Mailbreaker


**No changes in AD damage were observed.**

PR #5978, opened 2021-05-27
Fix CR_ACIDDEMONSTRATION proccing strip status on mobs

Tydus1:
* **Addressed Issue(s)**: #5906
* **Description of Pull Request**: Added a check to prevent skill_break_equip function to be called when Acid Demonstration and AM_ACIDTERROR are used against mobs.
* This solves the adressed issue. However, analyzing the code, skills like Tarot Card of Fate (Gypsy/Clown), Head Crush (Lord Knight), some 3rd job skills... have the same effect of breaking enemy equipments and they all call skill_break_equip so... maybe all those skills should NOT inflict strip status when used against mobs? As I don't have all this information, i'm just pull requesting a fix for Acid Demonstration & Acid Terror (I assume they have the same behavior...)


aleos89 (approved):

src/map/skill.cpp
```diff
@@ -1628,6 +1629,9 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
 		break;
 
 	case CR_ACIDDEMONSTRATION:
+		if (!dstmd)	{
```
RagnaWay, line 1632: wouldn't it be better to remove this TAB and replace it with a SPACE

aleos89, line 1632:
```suggestion
		if (!dstmd) {
```

aleos89 (approved):

Lemongrass3110:
This is mixed between pre-renewal and renewal right now.
Is it intended for both or just one of the modes?

Tydus1:
Fixed in hash:https://github.com/rathena/rathena/commit/038c1778e18aa52dfd62d57a49d06f1055422e04

Issue #5981, opened 2021-05-31
Amatsu Dungeon Entrance Quest Log Bugs + Fix


KrokusPokus:
* **rAthena Hash**: f2b0c94

* **Server Mode**: any

* **Description of Issue**: 
Two missing quest log related problems + 1 typo.
There are two guards the player can talk to to start the quest. Both will `set event_amatsu,1;` but only one of them updates the quest log with `setquest 8131;`
Also, the player can end up at `set event_amatsu,4;` both with `setquest 8132;` or with `setquest 8133;` depending on if you talk to NPC Kouji 1 or 2 times before moving on to the next step. Unfortunately, the script only does a `changequest 8133,8134;`, so in case the quest log was at 8132, it will become stuck.

* **Fix**: 
``` diff
diff --git "a/C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\TortoiseGit\\quests_amatsu-f2b0c94.000.txt" "b/D:\\Coding\\Ragnarok\\GIT_rAthena\\npc\\quests\\quests_amatsu.txt"
index 3f65cb52f..2467e2839 100644
--- "a/C:\\Users\\ADMINI~1\\AppData\\Local\\Temp\\TortoiseGit\\quests_amatsu-f2b0c94.000.txt"
+++ "b/D:\\Coding\\Ragnarok\\GIT_rAthena\\npc\\quests\\quests_amatsu.txt"
@@ -1410,6 +1410,7 @@ amatsu,171,174,4	script	Gate Soldier#ama6	767,{
 		mes " ";
 		next;
 		set event_amatsu,1;
+		setquest 8131;
 		mes "[Gate Soldier]";
 		mes "She has been sick for months.";
 		mes "That is why she is staying in here.";
@@ -1836,7 +1837,7 @@ ama_in02,200,176,4	script	Lord of Palace#ama	768,{
 	else if (event_amatsu == 5) {
 		if (countitem(1022) > 0) {
 			mes "[Ishida Yoshinaga]";
-			mes "I've hear the great news!";
+			mes "I've heard the great news!";
 			mes "My mother seems to have gotten better. What was her disease?";
 			next;
 			mes "[Ishida Yoshinaga]";
@@ -2177,7 +2178,10 @@ ama_in01,180,173,3	script	Kitsune Mask#ama	762,{
 		mes "Good. Show that spiritual";
 		mes "energy to the fox several times.";
 		mes "Sooner or later, you'll be successful.";
[...]

PR #5985, opened 2021-06-01
Corrected the quests log in quests_amatsu.txt

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/5981

* **Server Mode**: both

Thanks to @KrokusPokus !

### DIFF
--- a/npc/quests/quests_amatsu.txt
+++ b/npc/quests/quests_amatsu.txt
@@ -2178,9 +2178,10 @@ ama_in01,180,173,3	script	Kitsune Mask#ama	762,{
 		mes "Good. Show that spiritual";
 		mes "energy to the fox several times.";
 		mes "Sooner or later, you'll be successful.";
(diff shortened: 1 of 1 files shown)


KrokusPokus:
In you version
```
-		changequest 8133,8134;
+		if (isbegin_quest(8133) == 1)
+			erasequest 8133;
+		setquest 8134;
```
I think it doesn't handle the case that the player could have either quest 8132 or 8133 when he gets to that part of the script, depending on if he talked to the Kouji NPC one or two times. In the former case quest tip 8132 would never be removed. That's why I proposed
```
-		changequest 8133,8134;
+		if (isbegin_quest(8132) == 1)
+			changequest 8132,8134;
+		else if (isbegin_quest(8133) == 1)
+			changequest 8133,8134;
```
to either go from tip 8132->8134 or 8133->8134.

Atemo:
You're right thanks!

PR #5997, opened 2021-06-03
YAML loading optimization

secretdataz:
* **Addressed Issue(s)**: -

* **Server Mode**: both

This PR replaces [yaml-cpp](https://github.com/jbeder/yaml-cpp) as a YAML document parser with [rapidyaml](https://github.com/biojppm/rapidyaml) library.  
An optimization was also made to `itemdb_searchname1` function which reduced its complexity to constant.  
Item database parsing function also has reduced complexity to linear complexity.
This results in reduced loading time in debug mode to a fifth compared to current master branch (0dd318e).  
Loading time comparison is available in the table below. Loading time is measured in milliseconds.  
| Function | Master | Master with itemdb optimization | This PR |
|---|---|---|---|
| do_init_itemdb (debug) | 154323 | 75269 | 24282 |
| do_init (debug) | 268201 | N/A | 53118 |
| do_init (release) | 40562 | N/A | 20644 |

## Known issues & caveats

- Reporting invalid node or invalid script will mark the line and column as 1
- The optimization made `itemdb_searchname1` case-sensitive.
- Only compiles on MSBuild / Visual Studio atm. CMakeLists.txt needs some modification to build.
- SQL DB loading is untested

## Building
* Install [vcpkg](https://vcpkg.io/en/index.html)
* Run `./vcpkg install ryml:x86-windows-static` or `./vcpkg install ryml:x64-windows-static` inside the directory you installed vcpkg.
* On an elevated terminal, run `vcpkg integrate install`
* Build with MSBuild or Visual Studio normally

thanks to @vstumpf for his contribution on the CMake update.

### Feedbacks and suggestions are welcome.


biojppm:
@aleos89 just a quick note regarding the commit above saying yaml-cpp is more robust than rapidyaml. I'm curious to hear about your difficulties emitting with rapidyaml. Feel free to open an issue at https://github.com/rapidyaml/issues to discuss.

aleos89:
@biojppm So what we use the emitter for is to load up some old CSV files to convert to YAML. After I converted all of our formatting from yaml-cpp to rapidyaml I ran into several issues with the way the CSV content is stored into NodeRef. yaml-cpp is directly assignable of any type: string, char, int, etc. When messing with rapidyaml I could only get my work to compile via c4::to_csubstr() which of course was converting to base64 and when I dumped it to the file I was getting `Key: <hash>`.
  
Here's an example of what I was playing around with:
```cpp
static bool instance_readdb_sub(char* str[], int columns, int current) {
	body |= ryml::MAP;
	body["Id"] = c4::to_csubstr(str[0]);
	body["Name"] = c4::to_csubstr(str[1]);
	if (atoi(str[2]) != 3600)
		body["TimeLimit"] = c4::to_csubstr(str[2]);
	if (atoi(str[3]) != 300)
		body["IdleTimeOut"] = c4::to_csubstr(str[3]);

	ryml::NodeRef enter = body["Enter"];
	enter |= ryml::MAP;

	enter["Map"] = c4::to_csubstr(str[4]);
	enter["X"] = c4::to_csubstr(str[5]);
	enter["Y"] = c4::to_csubstr(str[6]);

	if (columns > 7) {
		ryml::NodeRef map = body["AdditionalMaps"];
		map |= ryml::MAP;

		for (int i = 7; i < columns; i++) {
			if (!strlen(str[i]))
				continue;

			if (strcmpi(str[4], str[i]) == 0)
				continue;

			map[c4::to_csubstr(str[i])] = "true";
		}
	}

	return true;
}
```

Unless I'm completely missing something from the quickstart.cpp, I couldn't find a way for us to dump our content properly. Also, is there a way to just create a NodeRef without a tree? I also had to create an empty `parse_in_place` call to generate a tree but I have no use for the tree since all I want to do is fill up a NodeRef variable and then dump that variable to a file. Thanks!

PR #6010, opened 2021-06-08
Converted statpoint.txt to YAML

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

  * Converts the Statpoint Database into YAML.


db/import-tmpl/statpoint.yml
```diff
@@ -0,0 +1,31 @@
...
+#
+###########################################################################
+# Statpoint Database
+###########################################################################
+#
+# Statpoint Settings
+#
+###########################################################################
+# - Level             BaseLevel required.
+#   Point             Total of the statpoint given from BaseLevel 1 to 'Level'.
```
aleos89, line 26:
```suggestion
#   Point             Total status points given from BaseLevel 1 to 'Level'.
```

src/map/pc.cpp
```diff
@@ -12328,36 +12332,56 @@ static bool pc_readdb_job_noenter_map(char *str[], int columns, int current) {
...
-				stat=0;
-			if (i > MAX_LEVEL)
-				break;
-			statp[i]=stat;
-			i++;
-			entries++;
+
+void PlayerStatPointDatabase::clear() {
+	this->statpoint_table.clear();
+}
```
aleos89, line 12338: This can be removed, a trick is provided in pc.hpp.

src/map/pc.cpp
```diff
@@ -12328,36 +12332,56 @@ static bool pc_readdb_job_noenter_map(char *str[], int columns, int current) {
...
+		return 0;
+	}
+
+	this->statpoint_table[level] = point;
+
+	return 1;
+}
+
+
+// generate the remaining parts of the db if necessary
```
aleos89, line 12375:
```suggestion
/**
 * Generate the remaining parts of the db if necessary.
 */
```

aleos89 (commented):
Should `Point` be `Points` in the table? Other than that, just some minor things!

src/map/pc.cpp
```diff
@@ -12328,36 +12332,56 @@ static bool pc_readdb_job_noenter_map(char *str[], int columns, int current) {
...
+void PlayerStatPointDatabase::loadingFinished() {
+	this->statpoint_table[0] = 45; // seed value
+
+	for (uint16 i = 1; i <= MAX_LEVEL; i++) {
+		if (util::umap_find(this->statpoint_table, i) == nullptr) {
+			this->statpoint_table[i] = this->statpoint_table[i-1] + this->pc_gets_status_point((i-1), false);
+			ShowError("Missing statpoint for Level %d in statpoint.yml\n", i);
```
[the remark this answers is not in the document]
Atemo: It will always print the last file read (the import file) since the check is after loading but I believe it's fine

src/map/pc.cpp
```diff
@@ -12328,36 +12341,52 @@ static bool pc_readdb_job_noenter_map(char *str[], int columns, int current) {
...
+	return 1;
+}
+
+/**
+ * Generate the remaining parts of the db if necessary.
+ */
+void PlayerStatPointDatabase::loadingFinished() {
+	this->statpoint_table[0] = 45; // seed value
+
+	for (uint16 i = 1; i <= MAX_LEVEL; i++) {
```
Lemongrass3110, line 12384: Maybe add a check here for the battle config (status point table vs algorithm) and initialize the table here if the algorithm is used.

src/map/pc.cpp
```diff
@@ -7974,13 +7987,13 @@ int pc_resetstate(struct map_session_data* sd)
 	if (battle_config.use_statpoint_table)
 	{	// New statpoint table used here - Dexity
 		if (sd->status.base_level > MAX_LEVEL)
-		{	//statp[] goes out of bounds, can't reset!
+		{	//out of bounds, can't reset!
```
Lemongrass3110, line 7990: Would it not be better to simply cap the player's level to MAX_LEVEL here?
Or rather to the maximum level depending on the job?

src/map/pc.cpp
```diff
@@ -7595,14 +7595,27 @@ static int pc_setstat(struct map_session_data* sd, int type, int val)
...
-	if (battle_config.use_statpoint_table) //Use values from "db/statpoint.txt"
-		return (statp[level+1] - statp[level]);
-	else //Default increase
+/**
+ * Calculates the number of status points PC gets when leveling up (from level to level+1)
+ * @param level: Player base level.
+ * @param table: Use table value or formula.
+ * @return Status points at specific base level.
+ */
+uint32 PlayerStatPointDatabase::pc_gets_status_point(uint16 level, bool table) {
```
Lemongrass3110, line 7613: Try to remove the table parameter and always use the level difference as return value.

src/map/pc.cpp
```diff
@@ -12328,36 +12341,52 @@ static bool pc_readdb_job_noenter_map(char *str[], int columns, int current) {
...
+	if (!this->asUInt32(node, "Points", point))
+		return 0;
+
+	if (level == 0) {
+		this->invalidWarning(node["Level"], "The minimum level is 1.\n");
+		return 0;
+	}
+
+	if (level > MAX_LEVEL) {
+		// this->invalidWarning(node["Level"], "Level %d exceeds maximum BaseLevel %d, skipping.\n", level, MAX_LEVEL);
```
Lemongrass3110, line 12369: Why do we comment this?

src/map/pc.hpp
```diff
@@ -1117,6 +1117,26 @@ class AttendanceDatabase : public TypesafeYamlDatabase<uint32, s_attendance_peri
...
+class PlayerStatPointDatabase : public YamlDatabase {
+private:
+	std::unordered_map<uint16, uint32> statpoint_table;
+
+public:
+	PlayerStatPointDatabase() : YamlDatabase("STATPOINT_DB", 1) {
+
+	}
+
+	void clear() { };
```
Lemongrass3110, line 1129: This should be calling statpoint_table.clear() otherwise it is useless.

src/map/pc.cpp
```diff
@@ -12328,36 +12341,52 @@ static bool pc_readdb_job_noenter_map(char *str[], int columns, int current) {
...
+	this->statpoint_table[level] = point;
+
+	return 1;
+}
+
+/**
+ * Generate the remaining parts of the db if necessary.
+ */
+void PlayerStatPointDatabase::loadingFinished() {
+	this->statpoint_table[0] = 45; // seed value
```
Lemongrass3110, line 12382: We should double check this...
And maybe put it into inter server config. Because in the char server it is hardcoded to 48.
I also had to hardcode it to 48 in 1803324.
It is strange enough that the character gets created with 48 status points and then the table says he should have 45 points...

src/map/pc.cpp
Lemongrass3110: If they go beyond the default maximum I would propose to move them into the import file (maybe commented, otherwise we run into the same problem).

src/map/pc.cpp
Lemongrass3110: Yeah just let the algorithm start at level 2 then and everything is fine if we put the 48 into a config.

src/map/pc.cpp
Atemo: I have commented them in their respective files as the Points per level for renewal/pre-renewal are differents

src/map/pc.cpp
```diff
@@ -12422,14 +12445,8 @@ void pc_readdb(void) {
...
 
-	// generate the remaining parts of the db if necessary
-	k = battle_config.use_statpoint_table; //save setting
-	battle_config.use_statpoint_table = 0; //temporarily disable to force pc_gets_status_point use default values
-	statp[0] = 45; // seed value
-	for (; s <= MAX_LEVEL; s++)
-		statp[s] = statp[s-1] + pc_gets_status_point(s-1);
-	battle_config.use_statpoint_table = k; //restore setting
-	
+	statpoint_db.reload();
```
Lemongrass3110, line 12448: @Atemo I did not really understand why you changed from .load() to .reload() here?

Lemongrass3110 (approved):
Except the one comment everything else is fine for me now.

src/map/pc.hpp
Lemongrass3110: Oh. Just noticed this is still here even thought you marked it as resolved. :-(

src/map/pc.cpp
Atemo: https://github.com/rathena/rathena/pull/6010#discussion_r662849464
I thought you meant to call `statpoint_table.clear()`, `reload()` call `clear()`. I misunderstood ?

Lemongrass3110 (approved):

PR #6011, opened 2021-06-09
Converted mob_boss to yaml

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

  * Converts the Monster summon group Database into YAML. (mob_boss.txt, mob_branch.txt, mob_classchange.txt, mob_poring.txt, mob_mission.txt, mob_pouch.txt, mob_random_db.txt)


src/map/mob.cpp
```diff
@@ -5463,59 +5451,115 @@ uint64 MobAvailDatabase::parseBodyNode(const YAML::Node &node) {
...
+			std::shared_ptr<s_mob_db> mob = mobdb_search_aegisname(mob_name.c_str());
+
+			if (mob == nullptr) {
+				this->invalidWarning(mobit["Mob"], "Unknown mob %s.\n", mob_name.c_str());
+				continue;
+			}
+
+			uint32 rate;
+
+			if (!this->asUInt32(mobit, "Rate", rate))
```
Lemongrass3110, line 5531: I think it might be good to change the way we store "Rate" here.
Do we really need it to be that detailed?
I think our UInt16Rate would be sufficient. But you might have to use UInt16 for the "remove" feature.

src/tool/csv2yaml.cpp
```diff
@@ -3497,3 +3531,74 @@ static bool read_homunculus_expdb(const char* file) {
 	ShowStatus("Done reading '" CL_WHITE "%d" CL_RESET "' entries in '" CL_WHITE "%s" CL_RESET "'.\n", count, file);
 	return true;
 }
+// Copied and adjusted from mob.cpp
+static bool mob_readdb_group(char* str[], int columns, int current) {
+	std::string group_name = str[0];
+	uint16 mob_id = static_cast<uint16>(strtol(str[1], nullptr, 10));
+	bool is_default = mob_id == 0;
+
+	if (group_name.rfind("MOBG_", 0) == 0)
```
Lemongrass3110, line 3540: rfind is the last found entry.
Maybe better use strncasecmp on str[0]
Also probably give an error, if the name does not start with this and do not convert it.

src/map/mob.cpp
```diff
@@ -507,42 +495,42 @@ struct mob_data* mob_spawn_dataset(struct spawn_data *data)
...
+	} while ((
 		mob == nullptr ||
 		mob_is_clone(mob_id) ||
 		(flag&RMF_DB_RATE && (entry->rate < 1000000 && entry->rate <= rnd() % 1000000)) ||
 		(flag&RMF_CHECK_MOB_LV && lv < mob->lv) ||
 		(flag&RMF_MOB_NOT_BOSS && status_has_mode(&mob->status,MD_STATUSIMMUNE) ) ||
 		(flag&RMF_MOB_NOT_SPAWN && !mob_has_spawn(mob_id)) ||
 		(flag&RMF_MOB_NOT_PLANT && status_has_mode(&mob->status,MD_IGNOREMELEE|MD_IGNOREMAGIC|MD_IGNORERANGED|MD_IGNOREMISC) )
-	) && (i++) < MAX_MOB_DB && msummon->count > 1);
+	) && (i++) < MAX_MOB_DB && summon->list.size() > 0);
```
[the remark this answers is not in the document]
Lemongrass3110: But since it is also checked above already, we could remove it?

src/map/mob.cpp
Atemo: The tables are using the values from aegis, for example
```
      - Mob: HORN
        Rate: 357142
```
need int32 or I'm missing something?

src/map/mob.cpp
```diff
@@ -507,42 +495,42 @@ struct mob_data* mob_spawn_dataset(struct spawn_data *data)
...
+	uint32 i = 0;
+	uint16 mob_id = 0;
+	std::shared_ptr<s_randomsummon_entry> entry = nullptr;
 
 	do {
-		rand = rnd()%msummon->count;
-		entry = &msummon->list[rand];
+		entry = util::vector_random(summon->list);
 		mob_id = entry->mob_id;
 		mob = mob_db.find(mob_id);
```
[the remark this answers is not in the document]
Atemo: Is that necessary? I thought it returns a value from `summon->list` and `summon->list` can't be empty because of the check above

src/map/mob.cpp
Lemongrass3110: Haha I did not check the entries. Guess we have to stick with stupid AEGIS like logic then.

src/map/mob.cpp
```diff
@@ -507,42 +495,42 @@ struct mob_data* mob_spawn_dataset(struct spawn_data *data)
...
+	} while ((
 		mob == nullptr ||
 		mob_is_clone(mob_id) ||
 		(flag&RMF_DB_RATE && (entry->rate < 1000000 && entry->rate <= rnd() % 1000000)) ||
 		(flag&RMF_CHECK_MOB_LV && lv < mob->lv) ||
 		(flag&RMF_MOB_NOT_BOSS && status_has_mode(&mob->status,MD_STATUSIMMUNE) ) ||
 		(flag&RMF_MOB_NOT_SPAWN && !mob_has_spawn(mob_id)) ||
 		(flag&RMF_MOB_NOT_PLANT && status_has_mode(&mob->status,MD_IGNOREMELEE|MD_IGNOREMAGIC|MD_IGNORERANGED|MD_IGNOREMISC) )
-	) && (i++) < MAX_MOB_DB && msummon->count > 1);
+	) && (i++) < MAX_MOB_DB);
```
Lemongrass3110, line 529: I do not know why someone used the MAX_MOB_DB constant here, but we should probably replace this.
I think in the other cases I did something like .size() * 3
I think the best would be to replace this whole chaos with a normal for loop.

Most likely like this (I did not compile this to try):
```
for( size_t i = 0, max = summon->list.size() * 3; i < max; i++ ){
	std::shared_ptr<s_randomsummon_entry> entry = util::umap_random( summon->list );
	std::shared_ptr<s_mob_db> mob = mob_db.find( entry->mob_id );

	if( 	mob == nullptr ||
		mob_is_clone( entry->mob_id ) ||
		(flag&RMF_DB_RATE && (entry->rate < 1000000 && entry->rate <= rnd() % 1000000)) ||
		(flag&RMF_CHECK_MOB_LV && lv < mob->lv) ||
		(flag&RMF_MOB_NOT_BOSS && status_has_mode(&mob->status,MD_STATUSIMMUNE) ) ||
		(flag&RMF_MOB_NOT_SPAWN && !mob_has_spawn( entry->mob_id )) ||
		(flag&RMF_MOB_NOT_PLANT && status_has_mode(&mob->status,MD_IGNOREMELEE|MD_IGNOREMAGIC|MD_IGNORERANGED|MD_IGNOREMISC) )
	){
		continue;
	}

	return entry->mob_id;
}

return summon->default_mob_id;
```

Atemo: Should we check if
```
std::shared_ptr<s_mob_db> mob = mob_db.find( summon->default_mob_id );
```
can be nullptr as well?

Lemongrass3110: You can add it for safety if you want. :-)

PR #6032, opened 2021-06-17
Converted item group database to YAML

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

  * Converts the item_bluebox, item_cardalbum, item_giftbox, item_group_db, item_misc,  item_package and item_violetbox txt into YAML.

---------------------------------------------------------------------------------------------------------------------
### Additional note

  * Currently the `SubGroup` can be removed using `Clear` on import. Should we add an index for each entry to be able to edit them vs readability?
Example:
```
  - Group: ACCESORY
    Contain:
      - SubGroup: 1
        List:
          - Id: 0
            Item: Ring
            Rate: 10
          - Id: 1
            Item: Earring
            Rate: 11
```


db/item_group_db.yml
```diff
@@ -0,0 +1,51 @@
...
+#       List:                  List of item(s) for the SubGroup.
+#        - Item                Available item that will be obtained from this item group. Requires the AegisName of the item.
+#          Rate                Probability to get the item. Not a percentage value!
+#          Amount              Amount of item that will be obtained (Default: 1).
+#          Duration            Makes the item a rental item, which will be expire in the given amount of minutes. Not intended for use with stackable items (Default: 0).
+#          Announced           If player obtained this item, it will be broadcast to the server (true/false. Default: false).
+#          GUID                Makes the given item(s) with Unique ID (true/false. Default: false).
+#          Named               Inscribes the item with the obtainer's name (true/false. Default: false).
+#          Bound               Binds the obtained item. CHAR, ACCOUNT, GUILD, PARTY or NONE (Default: NONE).
+#       Clear                  Remove the given SubGroup (Optional).
```
aleos89, line 38:
```suggestion
# - Group                      Group name of the item.
#   Contain:
#     - SubGroup:              SubGroup number associated with the list of item. See doc/item_group.txt for more information.
#       List:                  List of item(s) for the SubGroup.
#        - Item                AegisName of item that will be obtained from this item group.
#          Rate                Probability to get the item. See doc/item_group.txt for more information. (Default: 0)
#          Amount              Amount of item that will be obtained. (Default: 1)
#          Duration            Makes the item a rental item which will expire within the given amount in minutes. Not intended for use with stackable items. (Default: 0)
#          Announced           If player obtains this item it will be broadcasted to the server. (Default: false)
#          GUID                Makes the given item(s) have a Unique ID. (Default: false)
#          Named               Inscribes the item with the obtainer's name. (Default: false)
#          Bound               Binds the obtained item. (Default: None)
#       Clear                  Remove the given SubGroup. (Optional)
```

doc/item_group.txt
```diff
@@ -63,94 +69,80 @@ Amount: Amount of item that will be obtained.
 
 ---------------------------------------
 
-Random: Set this to '0' and the item will always be obtained ("must" item).
-		Random value will allocates where the item will be stored at random group.
-		(max. random group is 4, defined as MAX_ITEMGROUP_RANDGROUP in 'src/map/itemdb.cpp'.)
+SubGroup: Set this to '0' and the item will always be obtained ("must" item).
+		  SubGroup value will allocates where the item will be stored at random group.
```
aleos89, line 73:
```suggestion
SubGroup: Setting this to '0' makes the item always obtainable ("must" item).
		  SubGroup value will allocate where the item will be stored at random group.
```

aleos89 (commented):
I couldn't update the header in "db/re/item_group_db.yml" since the file was too large to view. Looks great overall!

db/import-tmpl/item_group_db.yml
```diff
@@ -0,0 +1,43 @@
...
+#       List:                  List of item(s) for the SubGroup.
+#        - Item                Available item that will be obtained from this item group. Requires the AegisName of the item.
+#          Rate                Probability to get the item. Not a percentage value!
+#          Amount              Amount of item that will be obtained (Default: 1).
+#          Duration            Makes the item a rental item, which will be expire in the given amount of minutes. Not intended for use with stackable items (Default: 0).
+#          Announced           If player obtained this item, it will be broadcast to the server (true/false. Default: false).
+#          GUID                Makes the given item(s) with Unique ID (true/false. Default: false).
+#          Named               Inscribes the item with the obtainer's name (true/false. Default: false).
+#          Bound               Binds the obtained item. CHAR, ACCOUNT, GUILD, PARTY or NONE (Default: NONE).
+#       Clear                  Remove the given SubGroup (Optional).
```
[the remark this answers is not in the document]
Atemo: Why not let `'IG_' is appended to the name when the file is read.`, the user need this information, no?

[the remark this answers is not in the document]
aleos89: The thought was that we haven't done this for other databases, so it's extra info. If it's important enough we can add it to the doc/ file if need be.

aleos89 (approved):
Looks good to me. Can have @Lemongrass3110 final John Hancock.

src/map/itemdb.cpp
```diff
@@ -1086,34 +1087,28 @@ s_item_combo *itemdb_combo_exists(uint32 combo_id) {
...
  */
-bool itemdb_group_item_exists(unsigned short group_id, t_itemid nameid)
+bool itemdb_group_item_exists(uint16 group_id, t_itemid nameid)
 {
-	struct s_item_group_db *group = (struct s_item_group_db *)uidb_get(itemdb_group, group_id);
-	unsigned short i, j;
+	std::shared_ptr<s_item_group_db> group = itemdb_group.find(group_id);
 
-	if (!group)
+	if (group == nullptr || group->random.empty())
```
Lemongrass3110, line 1100:
```suggestion
	if (group == nullptr)
```
Not necessary since the for loop below will cancel early in this case.

src/map/itemdb.cpp
```diff
@@ -1122,16 +1117,18 @@ bool itemdb_group_item_exists(unsigned short group_id, t_itemid nameid)
...
 		return -1;
 
-	for (int i = 0; i < MAX_ITEMGROUP_RANDGROUP; i++) {
-		for (int j = 0; j < group->random[i].data_qty; j++) {
-			int16 item_position = pc_search_inventory(sd, group->random[i].data[j].nameid);
+	for (const auto &random : group->random) {
+		if (random.second->data.empty())
+			continue;
```
Lemongrass3110, line 1129:
```suggestion
```
Not necessary since the for loop below will cancel early in this case.

src/map/itemdb.cpp
```diff
@@ -1197,38 +1194,40 @@ int itemdb_searchname_array(struct item_data** data, int size, const char *str)
 	return count;
 }
 
+std::shared_ptr<s_item_group_entry> itemdb_get_random_itemsubgroup(std::shared_ptr<s_item_group_random> random) {
```
Lemongrass3110, line 1197: Turn this into a function of s_item_group_random maybe?

src/map/itemdb.cpp
```diff
@@ -1536,133 +1534,205 @@ char itemdb_isidentified(t_itemid nameid) {
...
+				} else {
+					entry->bound = BOUND_NONE;
+				}
+
+				std::shared_ptr<s_item_group_random> random = util::umap_find(group->random, subgroup);
+
+				if (random == nullptr) {
+					random = std::make_shared<s_item_group_random>();
+					group->random[subgroup] = random;
+				}
```
Lemongrass3110, line 1716: All this should be moved above otherwise the import will not work correctly.
Additionally it is searched on each iteration instead of once only.

src/map/itemdb.cpp
```diff
@@ -1536,133 +1534,205 @@ char itemdb_isidentified(t_itemid nameid) {
...
-	random->data_qty += prob;
-	return true;
+	return 1;
+}
+
+void ItemGroupDatabase::loadingFinished() {
+	for (const auto &group : *this) {
+		for (const auto &random : group.second->random) {
+			for (const auto &it : random.second->data) {
+				random.second->total_rate += it->rate;
```
Lemongrass3110, line 1732: Is it just me or is the = 0 initialization missing?

src/map/pc.cpp
```diff
@@ -13225,16 +13225,18 @@ short pc_get_itemgroup_bonus(struct map_session_data* sd, t_itemid nameid) {
...
-		if (!group_id || !(group = itemdb_group_exists(group_id)))
+		if (group_id == 0 || group == nullptr || group->random.empty())
 			continue;
-		
-		for (i = 0; i < group->random[0].data_qty; i++) {
-			if (group->random[0].data[i].nameid == nameid) {
-				bonus += it.val;
-				break;
+
+		if (group->random.count(0) > 0 && !group->random[0]->data.empty()) {
```
Lemongrass3110, line 13234: Is this really intended to only apply to "must" items?
Should we issue a warning on empty here?
I do not think server admins understand this.
Based on documentation I would assume all subgroups...
`bonus2 bAddItemGroupHealRate,ig,n;	Increases HP recovered by n% for items of item group ig`

db/re/item_package.txt
```diff
@@ -892,23 +892,23 @@ IG_Xmas_Pet_Scroll,732,150,10,1,0,0,1	// 10x Crystal_Jewel__
...
-IG_Lucky_Scroll08,4169,1,1,0,0,0,1	// 1x Dark_Illusion_Card
-IG_Lucky_Scroll08,5215,2,1,0,0,0,1	// 1x Angelic_Chain_
-IG_Lucky_Scroll08,5216,3,1,0,0,0,1	// 1x Satanic_Chain_
-IG_Lucky_Scroll08,5217,2,1,0,0,0,1	// 1x Magestic_Goat_TW
-IG_Lucky_Scroll08,5218,3,1,0,0,0,1	// 1x Bunny_Band_
-IG_Lucky_Scroll08,5219,5,1,0,0,0,1	// 1x Drooping_Kitty_
-IG_Lucky_Scroll08,5077,65,1,0,0,0,1	// 1x Tulip_Hairpin
-IG_Lucky_Scroll08,12901,60,1,0,0,0,1	// 1x Insurance_Package
-IG_Lucky_Scroll08,610,110,15,0,0,0,1	// 15x Leaf_Of_Yggdrasil
+IG_Lucky_Scroll08,12211,220,5,1,0,0,1	// 5x Kafra_Card
```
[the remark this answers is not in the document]
Atemo: I temporary left the txt files in case we need to run csv2yaml again

src/map/pc.cpp
Atemo: Ha yeah the old random[0] is now random[1] that part is a mistake from me

src/map/itemdb.cpp
Atemo: Hmm what do you mean?

src/map/pc.cpp
Lemongrass3110: But do you know why we only take the first subgroup into account here?
I think we should loop the subgroups or give better documentation :(

db/re/item_package.txt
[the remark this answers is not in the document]
Lemongrass3110: Ok then I understand it

src/map/itemdb.hpp
```diff
@@ -827,28 +827,27 @@ struct s_item_combo {
...
+	uint8 bound; /// Makes the item as bound item (according to bound type)
 };
 
 /// Struct of random group
 struct s_item_group_random
 {
-	struct s_item_group_entry *data; /// Random group entry
-	unsigned short data_qty; /// Number of item in random group
+	std::vector<std::shared_ptr<s_item_group_entry>> data; /// Random group entry
+	uint32 total_rate;
```
Lemongrass3110, line 843:
```suggestion
	uint32 total_rate;
	
	std::shared_ptr<s_item_group_entry> get_random_itemsubgroup();
```

src/map/itemdb.cpp
Lemongrass3110: https://github.com/rathena/rathena/pull/6032#discussion_r666507757

src/map/pc.cpp
Atemo: I looped all the subgroup in the last commit https://github.com/rathena/rathena/pull/6032/commits/a92c48da8c0e511520eae43800cc9bb8b092a0b5#diff-f0da4401634371005af9888ebaff866f7126cf02aee33b9824dea381a7065e0cR13232
I guess it only checked in 1 subgroup because the groups used for the `bAddItemGroupHealRate` command are only here to group certain items. Theses groups are never used to give a random item, so there is only 1 subgroup for them

Lemongrass3110: Nice one :)

src/map/itemdb.cpp
```diff
@@ -1536,133 +1530,207 @@ char itemdb_isidentified(t_itemid nameid) {
...
+						this->invalidWarning(listit["Group"], "Invalid Bound %s.\n", bound_name.c_str());
+						continue;
+					}
+
+					entry->bound = static_cast<uint8>(bound);
+				} else {
+					entry->bound = BOUND_NONE;
+				}
+
+				random->data.push_back(entry);
```
Lemongrass3110, line 1714: Should we use vector here or make item id the key on this layer (for import etc) and turn this into a umap?

Atemo: Some groups are using the same item ID in the subgroup (ex: guid + stackable items).
```
  - Group: SAGITTARIUS_SCR_BOX
    Contain:
      - SubGroup: 0
        List:
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
          - Item: Sagittarius_Scroll
            GUID: true
```
The ID can't be used as key for them

Lemongrass3110: Uhm. This does not make any sense...
This should be:
```
  - Group: SAGITTARIUS_SCR_BOX
    Contain:
      - SubGroup: 0
        List:
          - Item: Sagittarius_Scroll
            Amount: 10
            GUID: true
```
And GUID parameter should be respected in source correctly.

Lemongrass3110: It only makes sense if it has different rates and is in there multiple times. I am not sure if such an entry exists?

Atemo: > 
> 
> It only makes sense if it has different rates and is in there multiple times. I am not sure if such an entry exists?

I didn't find an entry with these specificities.
For reference aegis separates the stackable items with guid
```
		{name = "must", item = "Lucky_Clip_Box", cnt = 1, hour = 0, onair = false, guid = true, probability= 1};
		{name = "must", item = "Lucky_Clip_Box", cnt = 1, hour = 0, onair = false, guid = true, probability= 1};
```

src/map/itemdb.cpp
```diff
@@ -1197,49 +1190,51 @@ int itemdb_searchname_array(struct item_data** data, int size, const char *str)
...
-		qty = group->random[sub_group-1].data_qty;
+	if (group == nullptr) {
+		ShowError("get_random_entry: Invalid group id %hu\n", group_id);
+		return nullptr;
 	}
-	if (!qty) {
-		ShowError("itemdb_get_randgroupitem: No item entries for group id %d and sub group %d\n", group_id, sub_group);
-		return NULL;
+
+	if (group->random.empty() || group->random.count(sub_group) == 0) {
```
Lemongrass3110, line 1221: Maybe split this into two error messages so that the user know what he messed up.

src/map/itemdb.cpp
```diff
@@ -1267,15 +1260,17 @@ static void itemdb_pc_get_itemgroup_sub(struct map_session_data *sd, bool identi
 		tmp.card[3] = GetWord(sd->status.char_id, 1);
 	}
 
-	if (!itemdb_isstackable(data->nameid))
-		get_amt = 1;
-	else
+	uint16 get_amt = 0;
+
+	if (itemdb_isstackable(data->nameid) && data->GUID == 0)
```
Lemongrass3110, line 1265: This should actually take the "GUID" parameter of the item db into account as well not only the "GUID" from item group db.

Lemongrass3110 (approved):
If you are done and have remove the leftover files in db (guess no further conversion is needed). It is good for me.

Everade:
@Atemo 
Are const.txt definitions no longer possible for groups?
I'm getting "Invalid group EVENT_BOX" error.

const.txt example:
`IG_Event_Box	8882`

Doesn't seem to be recognized any longer and loading any custom item_group fails.

Atemo:
It's because of IG_MAX check
https://github.com/rathena/rathena/pull/6146/commits/6a15314890bfc0af6290bb703c1ef0123054a9be

Issue #6055, opened 2021-06-25
mob_avail messing up the quest_db mob target entry.


k1ngJ:
* **rAthena Hash**: af0331588c4afd8b4c581cd08dd3f38771035fc2

* **Client Date**: 2021-01-07a

* **Server Mode**: pre-re

* **Description of Issue**: 
  * Result:  Monster in mob_avail db was not being registered as target mob in quest db, instead it registers the target(sprite) mob.
  * Expected Result:  It should register the original mob id.
  * How to Reproduce:  Put a mob in mob_avail db, and in quest db, put the original mob id as target.
  * Official Information: N/A
  

* **Modifications that may affect results**:  N/A

k1ngJ:
Works totally fine now, thank you.

PR #6056, opened 2021-06-25
Add another entry in view_data for original mob id.

k1ngJ:
Fixes the bug where the original id is not being loaded on quest db if its in mob_avail db.

* **Addressed Issue(s)**: #6055

* **Server Mode**: Pre-re

* **Description of Pull Request**: Follow up to the issue mentioned above.

Now it will check if it has an original mob id instead of a fake(sprite) id and load it to quest db instead.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -5193,7 +5193,7 @@ uint64 MobAvailDatabase::parseBodyNode(const YAML::Node &node) {
 
 			constant = sprite_mob->vd.class_;
 		}
-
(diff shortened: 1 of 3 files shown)


Lemongrass3110:
It might be better to add the field to s_mob_db finally.
It has been annoying a few times already...

Lemongrass3110:
Closing this in favor of #6069
Please test it and give us feedback if this resolves your issue.

Thank you for your contribution anyway!

PR #6066, opened 2021-06-29
Converted produce_db and skill_changematerial_db into YAML

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

  * Converts the produce_db and skill_changematerial_db Tables files into YAML.

---------------------------------------------------------------------------------------------------
### Notes:
`skill_changematerial_db.yml` and `skill_produce_db.yml` are both part of the produce system. `skill_changematerial_db.yml` just have more field which don't work with other itemlv, so the files has been separated (same system from the old txt version).

- The txt files should be removed before merging to master !


Gardosen:
small question/suggestion about this,

the file is called skill_produce_db.yml so it is self speaking that this file is about producing an item(weapon).
why do we use  the node element name 

#    - Produced                Item produced.

? two things i wonder about
#1 Produced sounds like this will definitely happen, but as there is a rate, it means it can also fail.
#2 as i already mentioned, it is the skill_produce_db.yml, so to make it easier readable for a human i think it would make sense same as on other files, to name this just "Item".

sticking to the old notation seems logical, but in this case i never understood why this element was called Produced.

i will review the source code a bit more later :)

Kind regards
Gardosen

Atemo:
The name of the node can be corrected to be more pertinent ofc. Thanks for the feedback !

src/map/skill.cpp
```diff
@@ -22672,48 +22658,229 @@ static bool skill_parse_row_nocastdb(char* split[], int columns, int current)
...
+
+/**
+ * Reads and parses an entry from the produce_db.
+ * @param node: YAML node containing the entry.
+ * @return count of successfully parsed rows
+ */
+uint64 SkillProduceDatabase::parseBodyNode(const YAML::Node &node) {
+	uint16 itemlv;
+
+	if (!this->asUInt16(node, "ItemLV", itemlv))
```
aleos89, line 22720:
```suggestion
	if (!this->asUInt16(node, "ItemLevel", itemlv))
```

src/map/skill.cpp
```diff
@@ -22672,48 +22658,229 @@ static bool skill_parse_row_nocastdb(char* split[], int columns, int current)
...
+
+			nameid = item->nameid;
+
+			if (produce->data.erase(nameid) == 0)
+				this->invalidWarning(subit["Clear"], "Item %s was not defined.\n", produced_name.c_str());
+
+			continue;
+		}
+
+		if (!this->asString(subit, "Produced", produced_name))
```
aleos89, line 22759:
```suggestion
		if (!this->asString(subit, "Product", produced_name))
```

src/map/skill.cpp
```diff
@@ -22672,48 +22658,229 @@ static bool skill_parse_row_nocastdb(char* split[], int columns, int current)
...
+			return 0;
+
+		struct item_data *item = itemdb_search_aegisname(produced_name.c_str());
+
+		if (item == nullptr) {
+			this->invalidWarning(subit["Produced"], "Item %s does not exist.\n", produced_name.c_str());
```
aleos89, line 22765:
```suggestion
			this->invalidWarning(subit["Product"], "Item %s does not exist.\n", produced_name.c_str());
```

src/map/skill.cpp
```diff
@@ -22672,48 +22658,229 @@ static bool skill_parse_row_nocastdb(char* split[], int columns, int current)
...
+		bool id_exists = produce->data.count(nameid) != 0;
+		std::shared_ptr<s_skill_produce_db_entry> entry;
+
+		if (id_exists)
+			entry = produce->data[nameid];
+		else {
+			if (this->total_id >= MAX_SKILL_PRODUCE_DB) {
+				this->invalidWarning(subit["Produced"], "Maximum db entries reached. Increase MAX_SKILL_PRODUCE_DB.\n");
+				return 0;
+			}
```
aleos89, line 22780:
```suggestion
```
The MAX_SKILL_PRODUCE_DB is no longer needed since we are using dynamic storage.

src/map/skill.cpp
```diff
@@ -22672,48 +22658,229 @@ static bool skill_parse_row_nocastdb(char* split[], int columns, int current)
...
+			entry = std::make_shared<s_skill_produce_db_entry>();
+			entry->nameid = nameid;
+		}
+
+		entry->itemlv = static_cast<uint8>(itemlv);
+
+		if (this->nodeExists(subit, "BaseRate")) {	// note: BaseRate is only used for skill changematerial (itemlv 26)
```
aleos89, line 22787: Should this section be wrapped with `if (itemlv == 26)` to prevent any values being set if it's not needed or do we not care?

src/map/skill.cpp
```diff
@@ -22672,48 +22658,229 @@ static bool skill_parse_row_nocastdb(char* split[], int columns, int current)
...
+						return 0;
+
+					if (rate == 0) {
+						if (entry->qty.erase(amount) == 0)
+							this->invalidWarning(Quantityit["Rate"], "Amount %hu was not defined.\n", amount);
+						continue;
+					}
+					if (rate > 1000) {
+						this->invalidWarning(Quantityit["Rate"], "Rate %hu can't be highter than 1000.\n", rate);
+						return 0;
```
aleos89, line 22822:
```suggestion
						this->invalidWarning(Quantityit["Rate"], "Rate %hu can't be higher than 1000, capping.\n", rate);
						rate = 1000;
```

src/map/skill.cpp
Atemo: You're right, but I kept MAX_SKILL_PRODUCE_DB for now because it requires to update `clif.cpp` 
https://github.com/rathena/rathena/pull/6066/files?file-filters%5B%5D=.cpp#diff-e095af7d5069bf7fbe9926283903389918f6afe47ea6bae5a4880503b59c7d2cR6028

db/import-tmpl/produce_db.yml
```diff
@@ -0,0 +1,49 @@
...
+#
+# Item Produce Settings
+#
+###########################################################################
+# - ItemLevel                  Number which determines what kind of a crafting window will pop-up.
+#   Recipe:
+#    - Product                 AegisName of the produced item.
+#      Skill:                  Skill requirement. (Default: null)
+#        Name                  Skill name required.
+#        Level                 Skill level required.
```
Lemongrass3110, line 30:
```suggestion
#      SkillName               Skill name required. (Default: null)
#      SkillLevel              Skill level required. (Default: 1)
```

db/import-tmpl/produce_db.yml
```diff
@@ -0,0 +1,49 @@
...
+#      Consumed:               List of items consumed to produce the Product.
+#        - Item                AegisName of the consumed item.
+#          Amount              Amount required.
+#          Clear               Remove the item with the given AegisName from Consumed. (Optional)
+#      NotConsumed:            List of items not consumed to produce the Product. (Default: null)
+#        - Item                AegisName of the unconsumed item.
+#          Clear               Remove the item with the given AegisName from NotConsumed. (Optional)
```
Lemongrass3110, line 37: Why dont we combine these two and make "Consumed" a flag (Default: true?) ?

db/import-tmpl/produce_db.yml
```diff
@@ -0,0 +1,49 @@
...
+#
+#                              Additional fields for Change Material (ItemLevel: 26):
```
Lemongrass3110, line 39:
```suggestion
###########################################################################
# Additional fields for Change Material (ItemLevel: 26):
```

Lemongrass3110: Is there a reason why we limit it to only level 26?

Lemongrass3110 (commented):
Just some input/ideas on the design.

We might have to merge this before #6437

db/import-tmpl/produce_db.yml
Atemo: I found it clearer to separate the 2 because Amount is required for Consumed and is not required for NotConsumed than add an exception in the description

db/import-tmpl/produce_db.yml
Atemo: Well it was designed like that xD BaseRate is only used for skill change material. Maybe we could extend it later, for now we could just stick with the convertion txt to yaml

db/import-tmpl/produce_db.yml
```diff
@@ -0,0 +1,49 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89, line 2:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

db/pre-re/produce_db.yml
```diff
@@ -0,0 +1,54 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89, line 2:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

db/pre-re/skill_changematerial_db.yml
```diff
@@ -0,0 +1,45 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89, line 2:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

db/pre-re/skill_produce_db.yml
```diff
@@ -0,0 +1,3009 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

db/produce_db.yml
```diff
@@ -0,0 +1,57 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89, line 2:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

db/re/produce_db.yml
```diff
@@ -0,0 +1,54 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89, line 2:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

db/re/skill_changematerial_db.yml
```diff
@@ -0,0 +1,911 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89, line 2:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

db/re/skill_produce_db.yml
```diff
@@ -0,0 +1,2535 @@
+# This file is a part of rAthena.
+#   Copyright(C) 2021 rAthena Development Team
```
aleos89, line 2:
```suggestion
#   Copyright(C) 2022 rAthena Development Team
```

aleos89 (commented):
Year updates!

db/import-tmpl/produce_db.yml
aleos89: Another option would be:
```yaml
  Recipe:
    Catalysts:
      - Item
        Amount
        Consumed    Whether the item is consumed or not. (Default: true)
        Clear
```

PR #6067, opened 2021-06-30
Items removed from RE that were in the PRE-RE folder

senaczyk:
ID: 16030, 16031 and 16032


Lemongrass3110:
They are in there for skill tree reasons.
This is why CI failed for your PR.

PR #6068, opened 2021-07-01
Update some heal items to kRO

kyurejl:


Atemo (commented):
Do you have the aegis item datas ?

If your changes are based on KRO description, for me they are incorrects and we should keep the old values. For example `Orange_Potion` https://www.divine-pride.net/database/item/502/

`빨간허브와 노란허브를 잘 갈아 만든 체력 회복제 HP 를 105 가량 회복 시켜 준다.`
The description can be translated to "... restores about 105 HP", which mean the value is random and not a fixed value.

PR #6069, opened 2021-07-01
Added ID field to mob_db structure

Lemongrass3110:
* **Addressed Issue(s)**: #6055

* **Server Mode**: Both

Added an ID field to the mob_db structure for in source usage in various places.

Thanks to @k1ngJ

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -4283,7 +4283,8 @@ uint64 MobDatabase::parseBodyNode(const YAML::Node &node) {
 			return 0;
 
 		mob = std::make_shared<s_mob_db>();
-		mob->vd.class_ = static_cast<uint16>(mob_id);
+		mob->id = mob_id;
+		mob->vd.class_ = static_cast<uint16>(mob->id);
 	}
 
 	if (this->nodeExists(node, "AegisName")) {
@@ -5191,7 +5192,7 @@ uint64 MobAvailDatabase::parseBodyNode(const YAML::Node &node) {
 				return 0;
 			}
 
-			constant = sprite_mob->vd.class_;
+			constant = sprite_mob->id;
 		}
 
 		mob->vd.class_ = (unsigned short)constant;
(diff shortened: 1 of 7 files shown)


src/map/mob.cpp
```diff
@@ -5500,7 +5501,8 @@ uint64 MobSummonDatabase::parseBodyNode(const YAML::Node &node) {
 			this->invalidWarning(node["Default"], "Unknown mob %s.\n", mob_name.c_str());
 			return 0;
 		}
-		summon->default_mob_id = mob->vd.class_;
+
```
Gardosen, line 5504: i am not sure if we care for code style in this matter but we should avoid unnecessary empty lines.

src/map/mob.cpp
```diff
@@ -4283,7 +4283,8 @@ uint64 MobDatabase::parseBodyNode(const YAML::Node &node) {
 			return 0;
 
 		mob = std::make_shared<s_mob_db>();
-		mob->vd.class_ = static_cast<uint16>(mob_id);
+		mob->id = mob_id;
+		mob->vd.class_ = static_cast<uint16>(mob->id);
```
Gardosen, line 4287: we should add a secury check for mob_id > UINT16_MAX here.

you have set mob->id to uint32 which is the same as mob_id. but vd.class_ is a uint16. if there is ever the case where mob_id exceeds UINT16_MAX, mob->id would be fine, but vd.class_ would be truncated without any notice.

Gardosen (commented):
looks good so far. i added a comment which should be looked at i think.

src/map/mob.cpp
Lemongrass3110: Hey!
Thank you for your (valid) input.
I did not bother to add/adjust any checks that were not there before.
I agree with you that code-wise it looks bad to downcast it like this, but for the time being we should be out of danger, because we can assume `#define MAX_MOB_DB2 31999`. Once this limit is lifted from Gravity clientside, this will be a whole different story though (just like with item ids).

PR #6075, opened 2021-07-05
Better battle.cpp format

kyurejl:
Beautiful formatting in battle.cpp, no other changes.
Examples:
all "if(" now is "if ("
"if( bl->type == BL_SKILL )" now is "if (bl->type == BL_SKILL)"
"if (damage > 0 )" now is "if (damage > 0)"

### DIFF
--- a/db/re/item_db_usable.yml
+++ b/db/re/item_db_usable.yml
@@ -101,7 +101,7 @@ Body:
     Buy: 10
     Weight: 70
     Script: |
-      itemheal rand(45,65),0;
+      itemheal 45,0;
   - Id: 502
     AegisName: Orange_Potion
     Name: Orange Potion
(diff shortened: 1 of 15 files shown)


Lemongrass3110:
Beautiful is a personal opinion for code style, which is why we did not agree on a repository wide standard so far.
If we really pull this off we will need to check how to integrate it into commit hooks or whatever, just doing it once is not an option here I guess. Also only applying it to a single file seems like a bad idea.

To also bring in my personal opinion:
Spaces instead of tabs is a no-go for me. :-P

sader1992:
spaces instead of tabs, this is javascript's way which I always hated it xD, in my opinion this is very ugly :P

the way i see it, this will only put too much work on the existing servers when they  are updated, and break some diff files.

PR #6077, opened 2021-07-08
Converted castle_db.txt to YAML

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

  * Converts the Guild Castles Database into YAML.


src/common/mmo.hpp
```diff
@@ -721,9 +722,9 @@ struct guild {
 
 struct guild_castle {
 	int castle_id;
-	int mapindex;
-	char castle_name[NAME_LENGTH];
-	char castle_event[EVENT_NAME_LENGTH];
+	uint16 mapindex;
+	std::string castle_name,
+		castle_event;
```
Lemongrass3110, line 727: I am not sure if this works.
This struct is used in inter-server communication as well. I am not sure if std::string is getting serialized and deserialized properly.
Did you test this?

src/map/guild.cpp
```diff
@@ -269,25 +270,72 @@ bool guild_check_skill_require( struct guild *g, uint16 id ){
...
+		gc->castle_id = castle_id;
 	}
+
+	if (this->nodeExists(node, "Map")) {
+		std::string map_name;
+
+		if (!this->asString(node, "Map", map_name))
+			return 0;
+
+		uint16 mapindex = mapindex_name2id(map_name.c_str());
```
Lemongrass3110, line 300: By using this function missing maps are already getting reported. For a silent call you need to call it with mapindex_name2idx(mapname, nullptr)

Lemongrass3110: Also this function returns the mapid and not the mapindex. So variable naming is bad or maybe even wrong logic?

src/map/guild.cpp
```diff
@@ -269,25 +270,72 @@ bool guild_check_skill_require( struct guild *g, uint16 id ){
...
+		gc->castle_name = castle_name;
+	}
+
+	if (this->nodeExists(node, "NPC")) {
+		std::string npc_name;
+
+		if (!this->asString(node, "NPC", npc_name))
+			return 0;
+
+		if (npc_name.size() > EVENT_NAME_LENGTH) {
```
Lemongrass3110, line 326: I think what you want to check is NPC_NAME_LENGTH?
Also the error message below should be changed.
Unless you already want to build the full event name here and you forgot to append `::OnGuildBreak` here.

src/map/guild.cpp
```diff
@@ -2066,20 +2096,17 @@ int guild_break(struct map_session_data *sd,char *name) {
  * from char-server.
  */
 void guild_castle_map_init(void) {
-	int num = db_size(castle_db);
```
Lemongrass3110, line 2069: Please change this whole dirty function...
C++ iterate over it and build a vector with the ids that need to be requested.
Thank you!

src/map/guild.hpp
```diff
@@ -119,4 +116,19 @@ void guild_retrieveitembound(uint32 char_id,uint32 account_id,int guild_id);
...
+class CastleDatabase : public TypesafeYamlDatabase <int32, guild_castle> {
+public:
+	CastleDatabase() : TypesafeYamlDatabase("CASTLE_DB", 1) {
+
+	}
+
+	const std::string getDefaultLocation();
+	uint64 parseBodyNode(const YAML::Node &node);
+
+	std::shared_ptr<guild_castle> guild_mapname2gc(const char* mapname);
```
Lemongrass3110, line 128:
```suggestion
	std::shared_ptr<guild_castle> mapname2gc(const char* mapname);
```

src/map/guild.hpp
```diff
@@ -119,4 +116,19 @@ void guild_retrieveitembound(uint32 char_id,uint32 account_id,int guild_id);
...
+	std::shared_ptr<guild_castle> guild_mapindex2gc(int16 mapindex);
```
Lemongrass3110, line 129:
```suggestion
	std::shared_ptr<guild_castle> mapindex2gc(int16 mapindex);
```

src/map/mob.cpp
```diff
@@ -827,7 +827,7 @@ int mob_spawn_guardian(const char* mapname, int16 x, int16 y, const char* mobnam
 	struct mob_data *md=nullptr;
 	struct spawn_data data;
 	struct guild *g=nullptr;
-	struct guild_castle *gc;
+	std::shared_ptr<guild_castle> gc = nullptr;
```
Lemongrass3110, line 830: Move declaration down instead?

Lemongrass3110:
Btw I know it does not make sense for you now, but can you please put the castle db into pre-re and re (duplicate) when you are done?
I will introduce a V2 for renewal later on. ;-)

src/common/mmo.hpp
Atemo: castle_name and castle_event are not in int_guild.cpp nor in the guild_castle table (handled by int_guild.cpp)
```
CREATE TABLE IF NOT EXISTS `guild_castle` (
  `castle_id` int(11) unsigned NOT NULL default '0',
  `guild_id` int(11) unsigned NOT NULL default '0',
  `economy` int(11) unsigned NOT NULL default '0',
  `defense` int(11) unsigned NOT NULL default '0',
  `triggerE` int(11) unsigned NOT NULL default '0',
  `triggerD` int(11) unsigned NOT NULL default '0',
  `nextTime` int(11) unsigned NOT NULL default '0',
  `payTime` int(11) unsigned NOT NULL default '0',
  `createTime` int(11) unsigned NOT NULL default '0',
  `visibleC` int(11) unsigned NOT NULL default '0',
  `visibleG0` int(11) unsigned NOT NULL default '0',
  `visibleG1` int(11) unsigned NOT NULL default '0',
  `visibleG2` int(11) unsigned NOT NULL default '0',
  `visibleG3` int(11) unsigned NOT NULL default '0',
  `visibleG4` int(11) unsigned NOT NULL default '0',
  `visibleG5` int(11) unsigned NOT NULL default '0',
  `visibleG6` int(11) unsigned NOT NULL default '0',
  `visibleG7` int(11) unsigned NOT NULL default '0',
  PRIMARY KEY  (`castle_id`),
  KEY `guild_id` (`guild_id`)
) ENGINE=MyISAM;
```
so there are only used for the mapserver.
I made some tests on local, I have no issue yet

src/map/guild.cpp
Atemo: Ha true! I forgot half of the string

src/common/mmo.hpp
Lemongrass3110: Talking about this:
```
int intif_parse_GuildCastleDataLoad(int fd)
{
	return guild_castledataloadack(RFIFOW(fd,2), (struct guild_castle *)RFIFOP(fd,4));
}
```

Atemo: I just reverted the changes in mmo

src/map/guild.cpp
Atemo: The function is using some packets sent to charserver and I have some issue with that. I can't fix the issues on my local. Can you commit the changes when you have time?

Lemongrass3110: Done! :)

Lemongrass3110 (approved):

PR #6143, opened 2021-07-27
Libevent and Windows GUI

phit666:
None, this is a new feature.

Both

Adding libevent library (libevent.org) as socket backend will benefit the server with its iterface simplicity (fewer codes) and performance (tested and stable). It will use the best backend socket interface in linux (EPOLL) and IOCP in windows.

Now the windows GUI interface made it possible to remove the signal handler (usually causing deadlock) and to add menu option like reloading scripts/dbs, this is a lot quicker than typing command ingame to admin tasks.


src/config/packets.h
```diff
@@ -7,18 +7,21 @@
...
  **/
-
+//2008_09_10a
 #ifndef PACKETVER
-	#define PACKETVER 20151104
+	#define PACKETVER 20141022 //20131223 //20101116//20100730//20151104
 #endif
 
+#define NO_PACKET_OBFUSCATION
+
```
Gardosen, line 16: this entire file looks weird. it is completely different to what exists inside the master branch. merging this into master would change the file entirely.

src/config/renewal.h
```diff
@@ -4,7 +4,7 @@
 #define _CONFIG_RENEWAL_H_
 
 //quick option to disable all renewal option, used by ./configure
-//#define PRERE
```
Gardosen, line 7: this should be reverted for the final PR to be merged.

src/custom/defines_pre.h
```diff
@@ -7,6 +7,9 @@
  * rAthena configuration file (http://rathena.org)
  * For detailed guidance on these check http://rathena.org/wiki/SRC/config/
  **/
+#define PACKET_OBFUSCATION_KEY1 0x631c511c
+#define PACKET_OBFUSCATION_KEY2 0x111c111c
+#define PACKET_OBFUSCATION_KEY3 0x111c111c
```
Gardosen, line 12: this should be reverted for the final PR to be merged.

nothing which should be added to the rathena repo.

src/map/mob.c
```diff
@@ -2581,6 +2597,8 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
 				pc_damage_log_clear(tmpsd[i],md->bl.id);
 		}
 
+		//ShowInfo("Debug, TP 3 pnum:%d.\n", pnum);
+
```
Gardosen, line 2601: same for all of these ShowInfo. They should all be removed.

src/map/map.h
```diff
@@ -49,7 +49,7 @@ void map_msg_reload(void);
 #define NATURAL_HEAL_INTERVAL 500
 #define MIN_FLOORITEM 2
 #define MAX_FLOORITEM START_ACCOUNT_NUM
-#define MAX_LEVEL 175
+#define MAX_LEVEL 255 //175
```
Gardosen, line 52: this should be reverted for the final PR to be merged.

src/map/clif.c
```diff
@@ -924,10 +924,14 @@ static int clif_setlevel_sub(int lv) {
 		lv = battle_config.max_lv;
 	}
 
+	//ShowInfo("clif_setlevel_sub, %d\n", lv);
+
```
Gardosen, line 928: this should be reverted for the final PR to be merged.

src/map/clif.c
```diff
@@ -924,10 +924,14 @@ static int clif_setlevel_sub(int lv) {
...
 	return lv;
 }
 
 static int clif_setlevel(struct block_list* bl) {
+
+
```
Gardosen, line 934: unnecessary lines

Gardosen (changes requested):
there is a lot which neeeds to be cleaned up. 

i have not checked the source code in general as i do not work on windows. but the general structure needs to be reworked and cleaned up before this should hit the master branch.

PR #6154, opened 2021-08-01
Libevent

phit666:
* **Addressed Issue(s)**: Provide an option in Windows and other Unix OS to choose the best socket backend available.

* **Server Mode**: Both Pre-Renewal and Renewal

* **Description of Pull Request**:  Windows is still using SELECT backend, which is not really efficient compare to linux's EPOLL available now, so through libevent Windows can now use IOCP as backend. Libevent (libevent.org) can actually use the best available socket backend in the OS like EPOLL in linux.


Lemongrass3110:
Could you explain the benefits of IOCP on our single threaded environment?

phit666:
> Could you explain the benefits of IOCP on our single threaded environment?

Well it will actually spawn 8 workers (current code has hint of 4 cores, can be removed to let auto detection of cores), so not really single threaded anymore.

Lemongrass3110:
rAthena itself is single threaded that is why I am asking.
I understand the concept of it and that you hardcoded a 4 core hint right now.
But I do not really see the benefit of making the network component multithreaded when the rest of the servers are not.
That was why I was asking what your intention behind this was.

phit666:
It will receive and process data in multithreaded  environment, you will see the  parsing of data is not in the do_socket call anymore, it will parse the data instantly once the call back for read is called.

hnomkeng:
now using 1 single core ?

hnomkeng:
image
@phi

PR #6201, opened 2021-08-18
Fixed changesex not changing characters sex

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Fixes changesex not changing characters sex at all


mazvi:
On latest ragexe, sex is based on character not account

Daegaladh:
> On latest ragexe, sex(gender) is based on character not account

I know, this fix only affects the changesex command (not changecharsex), which should change the sex setting of the account and set it to each of the characters.
Before the fix, it didn't change the sex setting characters at all (so it only worked for exes older than 2015), but changed the class for gender based classes, resulting in errors.

Lemongrass3110 (approved):

PR #6233, opened 2021-09-03
Suggestion: NPC name as parameter in enablenpc

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

This is a suggestion to make the required NPC name parameter optional (the default name will be the npc attached to the script) for the commands 
- enablenpc
- disablenpc
- hideonnpc
- hideoffnpc
- cloakonnpc
- cloakoffnpc
- isnpccloaked

### DIFF
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -262,7 +262,7 @@ bool npc_enable_target(npc_data* nd, uint32 char_id, int flag)
 		map_session_data *sd = map_charid2sd(char_id);
 	
 		if (!sd) {
-			ShowError("npc_enable: Attempted to %s a NPC on an invalid target %d.\n", (flag & 8) ? "show" : "hide", char_id);
+			ShowError("npc_enable: Attempted to %s a NPC '%s' on an invalid target %d.\n", (flag & 8) ? "show" : "hide", nd->name, char_id);
 			return false;
 		}
 
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -254,20 +254,15 @@ static int npc_cloaked_sub(struct block_list *bl, va_list ap)
 /*==========================================
  * Disable / Enable NPC
  *------------------------------------------*/
-bool npc_enable_target(const char* name, uint32 char_id, int flag)
+bool npc_enable_target(npc_data* nd, uint32 char_id, int flag)
 {
-	struct npc_data* nd = npc_name2id(name);
-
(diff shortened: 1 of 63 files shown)


src/map/npc.cpp
```diff
@@ -254,20 +254,15 @@ static int npc_cloaked_sub(struct block_list *bl, va_list ap)
...
-		return false;
-	}
+	nullpo_ret(nd);
 
 	if (char_id > 0 && (flag & 24)) {
 		map_session_data *sd = map_charid2sd(char_id);
 	
 		if (!sd) {
-			ShowError("npc_enable: Attempted to %s a NPC '%s' on an invalid target %d.\n", (flag & 8) ? "show" : "hide", name, char_id);
+			ShowError("npc_enable: Attempted to %s a NPC on an invalid target %d.\n", (flag & 8) ? "show" : "hide", char_id);
```
aleos89, line 265: Can this be made to `nd->name` so we don't lose this bit of info as to which NPC is at fault?

src/map/script.cpp
```diff
@@ -11647,40 +11647,38 @@ BUILDIN_FUNC(getareadropitem)
  *------------------------------------------*/
 BUILDIN_FUNC(enablenpc)
 {
-	const char *str = script_getstr(st,2);
-	if (npc_enable(str,1))
-		return SCRIPT_CMD_SUCCESS;
-
-	return SCRIPT_CMD_FAILURE;
-}
+	npc_data* nd = nullptr;
```
aleos89, line 11650:
```suggestion
	npc_data* nd;
```
Doesn't need an assignment since we don't do any pre-checks and it gets it's assignment a few lines below.

src/map/script.cpp
```diff
@@ -11647,40 +11647,38 @@ BUILDIN_FUNC(getareadropitem)
...
+	int flag = 0;
```
aleos89, line 11651: Guess the assignment can stay here since your checks below doesn't have a final "else". Is this the default value you want to hold (IE: disablenpc) if nothing is matched?

aleos89 (approved):
Looks good overall! Just a few questions is all I had.

src/map/script.cpp
Atemo: >Is this the default value you want to hold (IE: disablenpc) if nothing is matched?

Yes it's the idea

PR #6253, opened 2021-09-13
Update char.cpp

gamingmagic:
fixing starting item error on equip switch. cant produce starting item on new character.

* **Addressed Issue(s)**:  

* **Server Mode**:  Both

### DIFF
--- a/src/char/char.cpp
+++ b/src/char/char.cpp
@@ -1507,7 +1507,7 @@ int char_make_new_char( struct char_session_data* sd, char* name_, int str, int 
 	char_id = (int)Sql_LastInsertId(sql_handle);
 	//Give the char the default items
 	for (k = 0; k <= MAX_STARTITEM && tmp_start_items[k].nameid != 0; k++) {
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
Why do you think this is necessary?
The column is defined as not null with a default value of 0.
So if the column is not specified it should use that default value during the insert.

gamingmagic:
i just do minor adjustment on my side.

i got the below error

(09/13/2021 23:19:39) [ SQL Error ] : DB error - Field 'equip_switch' doesn't have a default value
(09/13/2021 23:19:39) [ Debug ] : at char.cpp:1511 - INSERT INTO `inventory` (`char_id`,`nameid`, `amount`, `equip`, `identify`) VALUES ('5580702', '70016', '1', '256', '1')
(09/13/2021 23:19:39) [ SQL Error ] : DB error - Field 'equip_switch' doesn't have a default value

Lemongrass3110:
Well indeed on rAthena master it has a default value.
Please check your table definition.

gamingmagic:
ah i understand now. lol thanks

PR #6263, opened 2021-09-20
Updates Poisonous Smoke behavior

### LINKED ISSUE #6199: Poisonous Smoke doesn't affect enemies at all.
* **rAthena Hash**: 
f0542cacb2ea49023ebc074f52067e858f354576

* **Client Date**: 
2021-01-07aRagexeRE

* **Server Mode**: 
Renewal

* **Description of Issue**: 
  * Result: 
  The Guillotine Cross Skill - **Poisonous Smoke** - doesn't affect players in PvP at all. I've tested it many times but never poisoned anyone.

  **It's also weird because after standing around for a very long time, i just suddenly get 100 damage out of the wild.
  I was able to reproduce this, so it wasn't a one time event.**
  I'm not sure if that's related to the skill, or if it's another bug. See for yourself at the end of the video.

https://user-images.githubusercontent.com/12261601/129773962-891fd2f6-377d-4cd8-a814-a02a265791b2.mp4

  * Expected Result:
   50% chance to poison enemies when they're standing within the 5x5 aoe smoke.
  
  * How to Reproduce: 
  **Cast the skill on an enemy. (I've been testing in PvP on players.)
  **Wait for the effect to take effect. (Won't happen)
  **Also wait for a very long time after the poison skill cast to see the 100dmg popping up.
 

  * Official Information: 
  https://www.divine-pride.net/database/skill/2032

  None

aleos89:
* **Addressed Issue(s)**: Fixes #6199

* **Server Mode**: Renewal

  * Adjusts the interval to every 2 seconds.
  * Adjusts the success rate to 50%.
  * Fixes the Poison Weapon effect not applying the 'enemy' status to those inside of Poisonous Smoke.
Thanks to @Everade!

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -19571,7 +19571,7 @@ Body:
     Unit:
       Id: Poisonsmoke
       Layout: 2
-      Interval: 1000
+      Interval: 2000
       Target: Enemy
       Flag:
         NoOverlap: true
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -14583,8 +14583,8 @@ int skill_unit_onplace_timer(struct skill_unit *unit, struct block_list *bl, t_t
 			break;
 
 		case UNT_POISONSMOKE:
-			if( battle_check_target(ss,bl,BCT_ENEMY) > 0 && !(tsc && tsc->data[sg->val2]) && rnd()%100 < 20 )
-				sc_start(ss,bl,(sc_type)sg->val2,100,sg->val3,skill_get_time2(GC_POISONINGWEAPON, 1));
+			if( battle_check_target(ss,bl,BCT_ENEMY) > 0 && !(tsc && tsc->data[sg->val2]) && rnd()%100 < 50 )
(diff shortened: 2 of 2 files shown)


Everade:
**Pyrexia Poison Status**
 *The affected player sees ways more damage on the visual indicator than actualy applied. I'm only getting 100 dmg, yet i'm seing damage displayed in the range of 10-30k. However as an outstanding player i see 100dmg indicators as expected.
 Not sure if this is intended as part of the "Confusion State"?! However i still see wrong damage after curing my blind and confusion state using a "Green Potion".
 image

**Venom Bleed Status**
 * MaxHP decreased by 15% should only be applied for 15seconds, not 5 minutes.

Other than that, it seems to work properly now, thanks @aleos89 !

aleos89:
Sorry, my comment about the previous commit for Venom Bleed cancelling skills on each hit is actually in reference to Pyrexia!

As for Pyrexia's damage issue, it's purely a client bug. When the EFST is active the clif_damage packet is broken. You can notice also if you start physically attacking someone with the EFST active, they will start seeing large damage output.

For Pyrexia and the blind status. Unless we can find a source that states/shows it's curable or not, I'll leave it as it is now. We have the comment in the code saying it is curable, so someone had tested/found it before.

PR #6281, opened 2021-10-01
Implemented new bonus for official Sarah Card behavior. Fixes #6189

### LINKED ISSUE #6189: Sarah card
help me
sarah card, 
when taking damage that exceeds 40% of the character's Max HP, the user only takes 40% of the MaxHP damage.
 Fixed damage attacks cannot reduce damage.  Against this card's abilities.

secretdataz:
* **Addressed Issue(s)**: #6189 

* **Server Mode**: Renewal

* **Description of Pull Request**: Implemented bAbsorbDmgMaxHp2 bonus which reduces the damage taken by the player to n% of their max HP if the damage they would take exceeds n% of their max HP.   Changed Sarah Card to use the new behavior instead of old bonus.


laziem:
missing `dmg_ori` before the check

laziem:
I got question about the fixed damage part.

"고정 데미지를 주는 공격은 피해를 감소 시킬 수 없다."
"Cannot reduce attacks that deal fixed damage."

Should skill like Cluster Bomb and Nova Explosion will ignore Sarah Card?

PR #6282, opened 2021-10-01
Remove Maximum Power-Thrust buff on weapon unequip. Fixes #6277

### LINKED ISSUE #6277: Maximum Power-Thrust incorrect function
wiki info = The effect of this skill wears off when logging out or changing any piece of gear. (https://irowiki.org/wiki/Maximum_Power-Thrust)

but the buff is not removed when changing equipment (character select clear the buff)

i try use it
    UnEquipScript: |
        sc_end SC_OVERTHRUSTMAX;

but the buff doesn't even go out by adding it to equipment and removing it from the character.

I'm trying to implement the orc's headdress and combos that enable the use of the skill in conjunction with the card but the skill's effect is removed if the equipment is removed.
the skill works but the effect is not removed when something or the equipment that enables it is removed.

I tested a device with similar functions (Garden_of_Eden 19272) and  UnEquipScript: | work fine

secretdataz:
* **Addressed Issue(s)**: #6277

* **Server Mode**: Both

* **Description of Pull Request**:  Remove Maximum Power-Thrust buff on weapon unequip.

* ** To-do: 

- [x] Confirm if this PR should affect pre-renewal

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -10989,7 +10989,9 @@ bool pc_unequipitem(struct map_session_data *sd, int n, int flag) {
 		skill_enchant_elemental_end(&sd->bl, SC_NONE);
 		status_change_end(&sd->bl, SC_FEARBREEZE, INVALID_TIMER);
 		status_change_end(&sd->bl, SC_EXEEDBREAK, INVALID_TIMER);
+#ifdef RENEWAL
 		status_change_end(&sd->bl, SC_MAXOVERTHRUST, INVALID_TIMER);
+#endif
 	}
 
 	// On armor change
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -10989,6 +10989,7 @@ bool pc_unequipitem(struct map_session_data *sd, int n, int flag) {
 		skill_enchant_elemental_end(&sd->bl, SC_NONE);
 		status_change_end(&sd->bl, SC_FEARBREEZE, INVALID_TIMER);
 		status_change_end(&sd->bl, SC_EXEEDBREAK, INVALID_TIMER);
+		status_change_end(&sd->bl, SC_MAXOVERTHRUST, INVALID_TIMER);
```
Lemongrass3110, line 10992:
```suggestion
#ifdef RENEWAL
		status_change_end(&sd->bl, SC_MAXOVERTHRUST, INVALID_TIMER);
#endif
```

Lemongrass3110: Thanks to @Playtester

Lemongrass3110 (approved):

aleos89 (approved):
Approved after Lemon's fix!

Issue #6312, opened 2021-10-12
npcwalkto/unitwalkto missing walk animation


Everade:
* **rAthena Hash**: 
7a0233388a7ca82bd9ce626901cadb526f6e3908

* **Client Date**: 
2021-01-07aRagexeRE

* **Server Mode**: 
Renewal

* **Description of Issue**: 
  * Result:
   npcwalkto and unitwalkto do not trigger the walking animation on monster sprites while walking.
  * Expected Result:
   Sprites that feature a walking animation should be animated accordingly.
  * How to Reproduce: 
```
prontera,157,182,4	script	Walk Test	1771,{

OnTimer3000:
		getmapxy .map$,.@currentX,.@currentY,BL_NPC;
		getfreecell(.map$,.@x,.@y,.@currentX,.@currentY,5,5);
		unitwalk getnpcid(0), .@x, .@y;

OnInit:
	stopnpctimer;
	initnpctimer;
	end;
}
```
  * Official Information: 
  This is an inofficial rAthena command as far as i'm aware of.

* **Modifications that may affect results**: 
  Custom walk test script.

PR #6325, opened 2021-10-19
Fixed walking NPCs with monster sprites

Lemongrass3110:
* **Addressed Issue(s)**: #6312

* **Server Mode**: Both

NPCs that use monster sprites also need to be respawned as monsters once they start walking.

Thanks to @Everade


Lemongrass3110:
> 
> 
> It seem walking-npc in Kro is 0xC type

@attackjom I tried 0xC. It seems to be fine for monster sprites, but not for player ones. :-(
But at least it removes the attack cursor if we use that one for monsters.

Issue #6324, opened 2021-10-19
Map-Server: Program received signal SIGSEGV


mazvi:
* **rAthena Hash**: d04e03e

* **Client Date**: 20200401

* **Server Mode**: RE

* **Description of Issue**: Map-Server: Program received signal SIGSEGV
  * Result: 

**Map Server Window**
```
Program received signal SIGSEGV, Segmentation fault.
0x000000000063c9d1 in skill_onskillusage (sd=sd@entry=0xb9ad0a0, bl=<optimized out>,
    skill_id=skill_id@entry=357, tick=<optimized out>) at skill.cpp:2333
2333                            if (rnd()%1000 >= it->rate)
Missing separate debuginfos, use: debuginfo-install MariaDB-shared-10.2.36-1.el7.centos.x86_64 glibc-2.17-317.el7.x86_64 keyutils-libs-1.5.8-3.el7.x86_64 krb5-libs-1.15.1-50.el7.x86_64 libcom_err-1.42.9-19.el7.x86_64 libgcc-4.8.5-44.el7.x86_64 libselinux-2.5-15.el7.x86_64 libstdc++-4.8.5-44.el7.x86_64 openssl-libs-1.0.2k-21.el7_9.x86_64 pcre-8.32-17.el7.x86_64 zlib-1.2.7-18.el7.x86_64
(gdb)
```
**GDB BT FULL**
```
(gdb) bt full
#0  0x000000000063c9d1 in skill_onskillusage (sd=sd@entry=0xb9ad0a0, bl=<optimized out>,
    skill_id=skill_id@entry=357, tick=<optimized out>) at skill.cpp:2333
        tbl = <optimized out>
#1  0x000000000063d564 in skill_castend_nodamage_id (src=<optimized out>, bl=<optimized out>,
    skill_id=<optimized out>, skill_lv=<optimized out>, tick=<optimized out>, flag=<optimized out>)
    at skill.cpp:11505
        sd = <optimized out>
        dstsd = <optimized out>
        md = <optimized out>
        dstmd = <optimized out>
        hd = <optimized out>
        mer = <optimized out>
        sstatus = <optimized out>
        tstatus = <optimized out>
        tsc = <optimized out>
        tsce = <optimized out>
        i = <optimized out>
        type = <optimized out>
        __FUNCTION__ = "skill_castend_nodamage_id"
#2  0x0000000000657423 in skill_castend_id (tid=<optimized out>, tick=tick@entry=7493187336,
[...]

PR #6327, opened 2021-10-20
Added null pointer checks for autobonus

secretdataz:
* **Addressed Issue(s)**: #6324

* **Server Mode**: Both


aleos89:
Should we safely just apply this to all autobonus loops?

Issue #5416, opened 2020-09-26
Tomb of Honor: Monster skills don't exist.


saya9200:
* **rAthena Hash**: cac6121f0406e8f25faf333bc8cedcb4c9e81db6

* **Client Date**: 2020

* **Server Mode**: re 

  * Official Information:
  I sadly don't know the names of the skills they use but they use a loooooooooooooot of skills on kRO videos that they dont use at all over here. They are looking like 3rd job skills.

An example would be ignition break from seyren. We dont have that skill in our skill database. or any of them.
I feel like we got a very "easy" version of tomb of honor compared to kRO

aleos89:
Fixed in [424e05d](https://github.com/rathena/rathena/commit/424e05d8f748ebbf88b7a0c5bf167bffb52bac5b).

PR #6343, opened 2021-10-29
Implemented some NPC skills

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/5416

* **Server Mode**: renewal

Implemented a bunch of NPC skills (renewal only).
The skills in mob_skill_db are uncommented.

NPC_CHEAL
NPC_MILLENNIUMSHIELD
NPC_STORMGUST2
NPC_ARROWSTORM
NPC_DRAGONBREATH
NPC_WIDEWEB
NPC_REVERBERATION
NPC_COMET
NPC_VENOMIMPRESS
NPC_SR_CURSEDCIRCLE
NPC_DARKPIERCING
NPC_FATALMENACE
NPC_PSYCHIC_WAVE
NPC_DRAGONBREATH
NPC_MAGMA_ERUPTION
NPC_RAYOFGENESIS
NPC_LEX_AETERNA
NPC_MANDRAGORA 
 NPC_DANCINGBLADE
Uncommented NPC_DARKPIERCING
NPC_PULSESTRIKE2


db/re/skill_db.yml
```diff
@@ -16920,6 +16933,31 @@ Body:
...
+    Type: Magic
+    TargetType: Ground
+    Flags:
+      IsNpc: true
+      ShowScale: true
+    Range: 9
+    Hit: Single
+    HitCount: 1
+    Element: Water
+    Knockback: 2
```
attackjom, line 16949: based on playing in KRO this skill doesn't have knock back. (use by Ominous Freezer)

Rayvakarian:
Finally!

Michieru:
Nice job, from what I know there is also some few old NPC skills missing:
id 2539	NPC_ASSASSINCROSS
id 2540 	NPC_DISSONANCE
id 2541	NPC_UGLYDANCE
Duno if you plan to add them, they are use for biolab 3 mobs

src/map/skill.cpp
```diff
@@ -1394,6 +1399,13 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
 #endif
 		break;
 
+	case NPC_STORMGUST2:
+		if (skill_lv < 3)
+			sc_start(src,bl,SC_FREEZE,10-((skill_lv-1)*(skill_lv+1)),skill_lv,skill_get_time2(skill_id,skill_lv));	// 10;7;3 % chance
```
Lemongrass3110, line 1404: Could we make that more readable?
I kinda dont understand why a formula like this is necessary for 2 values... :(

src/map/battle.cpp
```diff
@@ -4415,7 +4415,16 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
...
 		case NPC_ARROWSTORM:
-			skillratio += 900 + 80 * skill_lv;
+			if (skill_lv > 4)
+				skillratio += 1900;
+			else
+				skillratio += 900;
+			break;
+		case NPC_DRAGONBREATH:
+			if (skill_lv > 5)
+				skillratio += 500 + 500 * (skill_lv - 5);
```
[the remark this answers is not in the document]
Atemo: Yes it's correct
Level 1-5 is using fire element
Level 6-10 is using water element
The skill is like a combination of RK_DRAGONBREATH and RK_DRAGONBREATH_WATER

src/map/skill.cpp
Atemo: Alright (it's a clumsy formula for the 10;7;3 % chance per skill level)

src/map/battle.cpp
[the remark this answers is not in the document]
Lemongrass3110: Ok can we just add a small comment to make that clear?

src/map/skill.cpp
Lemongrass3110: Just make it a short if with the 3 values or if/else if/else then it will be more readable I guess.

Lemongrass3110 (approved):
Looks good except the really small comments I left that are mainly for readability reasons.

PR #6344, opened 2021-10-30
Improve docker-compose tool

Danil0v3s:
* **Addressed Issue(s)**: 

The current docker-compose solution is not pretty ideal, instead I propose using separate services to each server and one additional service as a helper to build the binaries when needed

* **Server Mode**: 

Both

This solution would help keeping track of the entire history of logs of each server, keeping track of memory usage and things like that.


alycecil:
This is great, having multiple docker images in compose is the "right" way!

Danil0v3s:
Haven't given much attention to this because when using the services in this way, when they send their IP back to the client it's within the docker loopback... unless official clients have a diffing to ignore that IP and always connect to the IP specified on the clientinfo, this PR wouldn't work

tools/docker/docker-compose.yml
```diff
@@ -14,12 +14,27 @@ services:
             MYSQL_DATABASE: ragnarok
             MYSQL_USER: ragnarok
             MYSQL_PASSWORD: ragnarok
-    server:
+
+    builder:
```
alycecil, line 18: What does `builder` do; is it just to check that the image starts?

alycecil: Or was it to build the project for the other images but missing its `command:`

tools/docker/docker-compose.yml
```diff
@@ -35,5 +50,45 @@ services:
         depends_on:
             - db
 
+    char:
+        image: "rathena:local"
+        container_name: "rathena-char"
+        command: "-c ./char-server"
```
alycecil, line 56: these should probably be using the absolute path just for clarity's sake 
`command: "-c /rathena/char-server"`

tools/docker/docker-compose.yml
```diff
@@ -35,5 +50,45 @@ services:
...
+        build:
+            context: .
+            dockerfile: Dockerfile
+        depends_on:
+            - login
+
+    map:
+        image: "rathena:local"
+        container_name: "rathena-map"
+        command: "-c ./map-server"
```
alycecil, line 76: these should probably be using the absolute path just for clarity's sake 
`command: "-c /rathena/map-server"`

tools/docker/docker-compose.yml
```diff
@@ -14,12 +14,27 @@ services:
...
+        stdin_open: true
+        build:
+            context: .
+            dockerfile: Dockerfile
+    
+    login:
         image: "rathena:local"
-        container_name: "rathena"
+        container_name: "rathena-login"
+        command: "-c ./login-server"
```
alycecil, line 36: these should probably be using the absolute path just for clarity's sake
`command: "-c /rathena/login-server"`

tools/docker/docker-compose.yml
Danil0v3s: When it's the first time you use rAthena you won't be able to shell into any other image to build the project so I created this image to just be there in case you need to build...

Do you know of any better way of tackling this issue?

Danil0v3s: Also I didn't want to have the command there to build because then we would build everytime we start the containers

alycecil: A couple steps to have the builder build when we want it to, have a command so we wont get the daemon pid 1 nonsense 

Add a file `builder.sh` in folder `tools/docker`, run `chmod 755 builder.sh` then make its body:
```

if [ ! -f /rathena/login-server ]; then
  export runBuild=1;
elif [ ! -f /rathena/char-server ]; then
  export runBuild=1;
elif [ ! -f /rathena/map-server ]; then
  export runBuild=1;
else
  export runBuild=0;
fi

if [ "${runBuild}" -eq "1" ]; then
  ### checking that ./configure has ran by looking for make file
  if [! -f /rathena/make ]; then
    echo "Error : ./configure has not been ran successfully. Please start the docker image alone and run the command /rathena/configure on that image.";
    echo "Resolution : run the following command \"docker-compose run builder /rathena/configure\"";
    exit 1;
  fi

  make clean server;
fi
```

make builder in the docker-compose file have a `command:` line of `command: "/rathena/tools/docker/builder.sh"`
Making lines 18-21 resemble the following:
```
    builder:
        image: "rathena:local"
        container_name: "builder"
        command: "/rathena/tools/docker/builder.sh"
        ...
```        

----

Should probably also update the readme for docker to let the others know how to configure it

Namely, with updating the ```### How to setup a local development environment :computer:```'s line 
```
2. `docker exec -it rathena bash` to connect to dev container
```

to 
```
1. `docker-compose run builder bash` to connect to a dev container in order to run aditional build or shell scripts within the linux context.
```

Maybe also with some extra lines like,
```
1. All commands expect you to have change directory to this directory, `$projectRoot/tools/docker`.
1. Run the configure to setup your make files for the docker _alpine_ image `docker-compose run builder /rathena/configure`
1. If you want the builder to build your project on each start change line 8 from
\```bash
  export runBuild=0;
\```
to
\```bash
  export runBuild=1;
\```

```

and

alycecil: Also I noticed the the -c's on the other three rathena images in the cluster that the `-c` was causing issues on both a Windows 10 and a Mac-Intel - Big Sur and had to remove them, have not tried on the debian image i have yet

Danil0v3s: Odd thing is I'm using Ubuntu as my Windows11 subsystem and couldn't get it running without the -C.

Also I've made some changes to the build.sh, instead of asking to run configure beforehand, I've added a env variable to the compose file and we always run the configure if we don't have the make file. What do you think @alycecil ?

alycecil: Sounds great!

Danil0v3s:
@alycecil I think I'm done with the suggestions! I've tested this on a Windows and a Intel Mac and they run just fine

tools/docker/docker-compose.yml
```diff
@@ -14,12 +14,30 @@ services:
...
+            - "./asset/char_conf.txt:/rathena/conf/import/char_conf.txt"   #localdev login-char relation
+            - "./asset/map_conf.txt:/rathena/conf/import/map_conf.txt"     #localdev char-map relation
+        init: true # helps with signal forwarding and process reaping
+        tty: true
+        stdin_open: true
+        build:
+            context: .
+            dockerfile: Dockerfile
+        environment:
+            BUILDER_CONFIGURE: "--enable-packetver=20200401"
```
Akkarinage, line 34: Is there a reason you're defining an old packetver?

Danil0v3s: Not really, I've updated with the one defined in master branch

tools/docker/docker-compose.yml
```diff
@@ -34,6 +52,44 @@ services:
             dockerfile: Dockerfile
         depends_on:
             - db
+    char:
+        image: "rathena:local"
+        container_name: "rathena-char"
+        command: sh -c "/bin/wait-for db:3306 -- /rathena/char-server"
```
alycecil, line 58: Oh thats awesome!

tools/docker/builder.sh
```diff
@@ -0,0 +1,21 @@
+if [ ! -f /rathena/login-server ]; then
```
alycecil, line 1: This looks great

alycecil (approved):
LGTM, Thank you so much. I love this PR already.

secretdataz (approved):

PR #6356, opened 2021-11-12
Minor fixes and cleanups to the mob database

aleos89:
* **Addressed Issue(s)**: N/A

  * Made some minor fixes and cleanups to the mob database documentation and parser.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -4322,7 +4322,7 @@ uint64 MobDatabase::parseBodyNode(const YAML::Node &node) {
 		if (!this->asString(node, "AegisName", name))
 			return 0;
 
-		if (name.length() > NAME_LENGTH) {
+		if (name.size() > NAME_LENGTH) {
 			this->invalidWarning(node["AegisName"], "AegisName \"%s\" exceeds maximum of %d characters, capping...\n", name.c_str(), NAME_LENGTH - 1);
 		}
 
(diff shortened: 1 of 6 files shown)


Atemo (approved):

src/map/mob.cpp
```diff
@@ -4322,7 +4322,7 @@ uint64 MobDatabase::parseBodyNode(const YAML::Node &node) {
 		if (!this->asString(node, "AegisName", name))
 			return 0;
 
-		if (name.length() > NAME_LENGTH) {
+		if (name.size() > NAME_LENGTH) {
```
Lemongrass3110, line 4325: "Both string::size and string::length are synonyms and return the same value."
https://www.cplusplus.com/reference/string/string/size/

Is there a reason why we change it? Or just to always use size?

src/map/mob.cpp
```diff
@@ -4367,6 +4367,11 @@ uint64 MobDatabase::parseBodyNode(const YAML::Node &node) {
 		if (!this->asUInt16(node, "Level", level))
 			return 0;
 
+		if (level > MAX_LEVEL) {
+			this->invalidWarning(node["Level"], "Level %d exceeds MAX_LEVEL, capping to %d.\n", level, MAX_LEVEL);
```
Lemongrass3110, line 4371:
```suggestion
			this->invalidWarning(node["Level"], "Level %hu exceeds MAX_LEVEL, capping to %hu.\n", level, MAX_LEVEL);
```

Lemongrass3110 (approved):
Only a general question, but rest looks good.

src/map/mob.cpp
aleos89: Yeah I think I used length at the time as I was typing and thinking differently. Since we use size elsewhere I was just trying to keep it harmonized!

PR #6363, opened 2021-11-18
Implemented news cards and enchants

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal

Implemented news cards and enchants according to KRO description.

### DIFF
--- a/db/re/item_db_equip.yml
+++ b/db/re/item_db_equip.yml
@@ -81335,52 +81335,6 @@ Body:
       bonus bMaxSP,100;
       bonus2 bSPDrainRate,30,3+(.@r/3);
       bonus2 bHPDrainRate,10,1+(.@r/4);
-  - Id: 19221
-    AegisName: Golden_Angel_Band_TH
-    Name: Golden Angel Hairband
-    Type: Armor
-    Weight: 10
-    Defense: 10
-    Locations:
-      Head_Top: true
-    ArmorLevel: 1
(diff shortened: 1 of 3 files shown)


idk-whoami:
some item additional bonus 2% for enchantgrade B or A
E.g. IDs 310675, 310676

Atemo:
Yeah and I didn't know if the previous enchantment level should work (they work), for example if the grade is B the bonus in grade D should be triggered too

attackjom:
> Yeah and I didn't know if the previous enchantment level should work (they work), for example if the grade is B the bonus in grade C.. should be triggered too

Yes. It increase by addition.
if you have gradeA you will get all the bonus in previous.

db/re/item_db_equip.yml
```diff
@@ -83146,6 +83100,52 @@ Body:
...
+         }
+         if (readparam(bInt) >= 70) {
+            bonus bMatk,5;
+            if (.@r >= 7)
+               bonus bMatk,10;
+         }
+         if (readparam(bVit) >= 70) {
+            bonus bMaxHPrate,2;
+            bonus bMaxSPrate,2;
+            if (.@r >= 7)
```
Litro, line 83130: is this intended or missing bracket?

Atemo: You're right the bracket seems missing !
https://github.com/rathena/rathena/pulls

PR #6369, opened 2021-11-22
Suggestion: Remove SQL files

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

Takeover of https://github.com/rathena/rathena/pull/6339

rAthena support YAML and SQL. Since there is a tool to convert YAML file to SQL (yaml2sql) for me there is no point in us (developper) to maintain 2 DBs.
I suggest to let the user create the SQL files if he choose to use the SQL DB (which is not the default) to lessen the burden of maintainers / active users who submit PRs.

I'm open to other suggestions.


Lemongrass3110:
I just noticed that the requirments for this should be:

1. Item Cash DB in YAML (#6299)
2. Mob Skill DB in YAML (no PR yet)

Otherwise we will only be able to partially drop this and have to maintain the remaining databases.

As discussed in Discord with @aleos89 and you there should be an explanation in [sql-files/README.md](https://github.com/rathena/rathena/tree/master/sql-files#sql-imports) and the CI scripts need to be adjusted.

Pokye:
I think it's much better to remove the SQL leaving only the YAML anyway, but how would FluxCP work without them?

Issue #6353, opened 2021-11-11
set/get instance variable commands


Everade:
**Is your request related to a missing feature?**
Yes, we're lacking a command that allows us to set an instance variable from outside an instance.
For example if you want to pass over instance settings from the instance entry npc into the instance upon creation.

**Describe the solution you'd like**
```
setinstancevar <instance_id>,<variable>,<expression>;
setinstancearray <instance_id>,<array name>[<first value>],<value>{,<value>...<value>};
```

Additionaly rename our current **get** command to reflect other getvar commands:
rename from getvariableofinstance
to getinstancevar

**Describe alternatives you've considered**
Or even better, incorporate it into all the set, setarray, getvar and getarray commands rather than creating new ones.
End goal being, that we can get and set instance variables of any kind from outside ;)

Atemo:
>Or even better, incorporate it into all the set, setarray, getvar and getarray commands rather than creating new ones.

A player can have several instances running at the same time (type guild, party..), meaning an instance ID is required to change the value of a specific instance. In my opinion adding an extra argument (the instance ID) in the `set` etc.. commands seems worse than creating a new command - so I would rather not.

https://github.com/rathena/rathena/blob/7dcc75d82e0521e920f6bcac23e2f36f79fc3e35/doc/script_commands.txt#L9405

The instance variable of a player outside the instance can be changed via
```
	// This will set the 'var variable of the specific instance ID to 1.
	set getvariableofinstance('var, instance_id(IM_GUILD)), 1;
```
It's the same system as when you want to change the npc variable from another npc, I agree it's not user friendly

Everade:
In my opinion it's worse having 5 different commands that do basically the same, rather than incoorporating them into a single one. We've been stacking up on new commands over the years like crazy, and while many of them are same by design, in a lot of cases the naming scheme doesn't help in any way either.
So it's usualy impossible to actualy remember them without jumping back and forth into the documentation unless you script 24/7.

Good example is the one i've given above:
```
getvariableofinstance   <----- Really weird one
```
And let's be real, the txt documentation doesn't really do a good job in helping you find something either.
But that's a different story.

I'm just happy when i can realise my projects in one way or another.
So thanks for letting me know that there's actualy a way to set an instance variable from outside 👍🏻

Lemongrass3110:
I also share @Atemo's opinion and dont think that combining it all into one command would be a good thing.
But I totally agree that commands could be added if currently only very strange workarounds make things possible.

[...]

PR #6374, opened 2021-11-25
Implementation of setinstancevar

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/6353

* **Server Mode**: Both

Implementation of `setinstancevar` to simplify the data manipulation of instance variable outside the instance map.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -26179,7 +26179,7 @@ struct script_function buildin_func[] = {
 
 	BUILDIN_DEF(achievement_condition,"i"),
 	BUILDIN_DEF(getinstancevar,"ri"),
-	BUILDIN_DEF2(getinstancevar, "getvariableofinstance","ri"),
+	BUILDIN_DEF2_DEPRECATED(getinstancevar, "getvariableofinstance","ri", "2021-12-13"),
 	BUILDIN_DEF(convertpcinfo,"vi"),
 	BUILDIN_DEF(isnpccloaked, "??"),
 
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -25241,13 +25241,13 @@ BUILDIN_FUNC(achievement_condition){
 /// Returns a reference to a variable of the specific instance ID.
 /// Returns 0 if an error occurs.
 ///
-/// getvariableofinstance(<variable>, <instance ID>) -> <reference>
-BUILDIN_FUNC(getvariableofinstance)
+/// getinstancevar(<variable>, <instance ID>) -> <reference>
+BUILDIN_FUNC(getinstancevar)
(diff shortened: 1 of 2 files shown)


Everade:
Thanks a lot for your work, much appreciated!

Please don't forget about the array command as originaly suggested.
```
setinstancevar <instance_id>,<variable>,<value>;
setinstancearray <instance_id>,<array name>[<first value>],<value>{,<value>...<value>};

```
And renaming the command **getvariableofinstance** to **getinstancevar** would be great. That would help a lot.

Atemo:
The system would require 'setinstancearray '+ more command like 'deleteinstancearray'.. to fully support the array manipulation. I don't plan to add them, at least not yet.
I can add an alias if you really want to
`BUILDIN_DEF2(getvariableofinstance, "getinstancevar","ri"),`
for me it's the same thing, as long as it does the job

Lemongrass3110:
> I can add an alias if you really want to `BUILDIN_DEF2(getvariableofinstance, "getinstancevar","ri"),` for me it's the same thing, as long as it does the job

Maybe this would be good, because the new function has a new name schema. But maybe we can even rename the old from getvariableofinstance to getinstancevar and add it as BUILDIN_DEF2_DEPRECATED so people can change their scripts to the new name?

Everade:
Lovely, thanks a bunch!

src/map/script.cpp
```diff
@@ -26122,7 +26178,8 @@ struct script_function buildin_func[] = {
 	BUILDIN_DEF(camerainfo,"iii?"),
 
 	BUILDIN_DEF(achievement_condition,"i"),
-	BUILDIN_DEF(getvariableofinstance,"ri"),
+	BUILDIN_DEF(getinstancevar,"ri"),
+	BUILDIN_DEF2(getinstancevar, "getvariableofinstance","ri"),
```
Lemongrass3110, line 26182:
```suggestion
	BUILDIN_DEF2_DEPRECATED(getinstancevar, "getvariableofinstance","ri", "2021-12-05"),
```
Adjust to the date when you merge it. :-)

Lemongrass3110 (approved):
The references to getvariableofinstance in script_commands.txt need to be updated (rename of getvariableofinstance)

Dieby:
image

Has anyone else had this error?

Atemo:
Which error? I only see an explicit warning

Dieby:
> Which error? I only see an explicit warning

Oh sorry, I explained myself wrong, I would like to know how to remove the warning

vstumpf:
Use `getinstancevar` instead of `getvariableofinstance`

PR #6382, opened 2021-11-30
Fix SC_CHANGE not using max matk

Tydus1:
* Thanks to Hercules for the fix;

* **Description of Pull Request**: Fix SC_CHANGE (HLIF mental change skill) to use max MATK on pre-renewal;

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -2260,8 +2260,10 @@ static int64 battle_calc_base_damage(struct block_list *src, struct status_data 
 	sd = BL_CAST(BL_PC, src);
 
 	if (!sd) { //Mobs/Pets
-		if (sc != NULL && sc->data[SC_CHANGE] != NULL)
+#ifndef RENEWAL
+		if (sc != nullptr && sc->data[SC_CHANGE] != nullptr)
 			return status->matk_max; // [Aegis] simply uses raw max matk for base damage when Mental Charge active
+#endif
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -2260,6 +2260,8 @@ static int64 battle_calc_base_damage(struct block_list *src, struct status_data
 	sd = BL_CAST(BL_PC, src);
 
 	if (!sd) { //Mobs/Pets
+		if (sc != NULL && sc->data[SC_CHANGE] != NULL)
+			return status->matk_max; // [Aegis] simply uses raw max matk for base damage when Mental Charge active
```
Lemongrass3110, line 2264:
```suggestion
#ifndef RENEWAL
		if (sc != NULL && sc->data[SC_CHANGE] != NULL)
			return status->matk_max; // [Aegis] simply uses raw max matk for base damage when Mental Charge active
#endif
```

Lemongrass3110: Unlike in Hercules this function is called in Renewal in rAthena as well.
Therefore we have to add a preprocessor check.

Lemongrass3110 (changes requested):

Lemongrass3110:
Originally fixed by @skyleo in HerculesWS/Hercules#3078

Lemongrass3110 (approved):

Atemo (approved):

Issue #6371, opened 2021-11-24
Problem with DC_WINKCHARM behavior


Daraen1:
* **rAthena Hash**: 
[ae1135d](https://github.com/rathena/rathena/commit/ae1135dd2b554fa8564e01c6da1866c7c9eee94d)

* **Client Date**: 

20180418

* **Server Mode**: 

I used Pre-Renewal

* **Description of Issue**: 
When I cast "Charming Wink" skill, the monster follow me at my cell. If i move in another cell, the monster don't follow me while he is still under the spell.
When the effect of skill if ended, the monster don't move and don't attack.
If i recast the skill, the monster don't move.
You can see the behavior with this link. 

[Charming Wink behavior](https://www.youtube.com/watch?v=FoH0lq_yw58)

* **Modifications that may affect results**: 
i dunno, i can just say, in Hercules, the skill work correctly

Daraen1:
I think the correct behavior for the Charming Wink skill should be the one you see at this link: [Good Behavior for Charming Wink Skill](https://www.youtube.com/watch?v=v99JaqIDKu4)

Still awaiting bug confirmation status, I sincerely think the skill is currently broken

Tydus1:
I can confirm it on hash a205287307acd96edf76e77eb0976a71c7fffbb1 (latest so far).

After a quick investigation, seems like unit_walktoxy_sub in unit.cpp is looping itself when charmed monster reach player, causing the mob to getting stuck on same cell. 

Test example: Summon a High Orc and wink charm it with some distance. If you keep running from it everything will be good. But if the high orc reach you and then you start moving away, it will be stuck on same cell.

I've made another test adding the code below on unit.cpp line 85 just to keep mobs "moving" in case they get in a to_x  and to_y to the same cell they are:

if (bl->x == ud->to_x && bl->y == ud->to_y)
{
ud->to_x = ud->to_x;
ud->to_y = ud->to_y;
}

It solved the problem of mobs getting stuck but it's not the correct solution.

Daraen1:
The problem described by @Tydus1 is exactly the one I demonstrate in the video on the first link. The good behavior of the skill is on the second link :)
Thanks for confirming the bug, I thought I was the only one going crazy lmao

Tydus1:
Analysing unit_walktoxy_timer in unit.cpp i've found a weird behavior: 

[...]

PR #6384, opened 2021-12-01
Fix SC_WINKCHARM behavior

Tydus1:
* Adds a check to prevent func unit_walktoxy_timer to use DIR_CENTER direction (which has -1 value) on SC_WINKCHARMed mobs. Using this direction may cause dump values when accessing dirx or diry array;

* **Addressed Issue(s)**: #6371 

* **Server Mode**: Both

* **Description of Pull Request**: Adds a check to prevent func unit_walktoxy_timer to use DIR_CENTER direction (which has -1 value) on SC_WINKCHARMed mobs. Using this direction may cause dump values when accessing dirx or diry array;

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -139,8 +139,14 @@ int unit_walktoxy_sub(struct block_list *bl)
 		i = status_get_speed(bl)*MOVE_DIAGONAL_COST/MOVE_COST;
 	else
 		i = status_get_speed(bl);
-	if( i > 0)
+	if( i > 0 ){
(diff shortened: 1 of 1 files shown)


Daraen1:
I just tried your pull request and it looks like the skill is working fine now, video proof!
[Charming Wink Fix](https://www.youtube.com/watch?v=o5eW2dfaVz4)
good job @Tydus1

Daraen1 (approved):

aleos89 (approved):

Issue #6389, opened 2021-12-04
Outdated Enchant Check Function


Everade:
* **rAthena Hash**: 
4cd2ef747d8c1d38ee257800701c258fe3af621b

* **Client Date**: 
Not relevant

* **Server Mode**: 
Renewal

* **Description of Issue**: 
  * Result: The function "F_IsCharm" is only checking for the old original enchants despite the new additions that took place over the years.
  * Expected Result: Add all missing enchant ids
  * How to Reproduce: Use any NPC that utilizes the function while using any of the new enchants on your equipment. The check will fail.
  * Official Information: No idea how official servers check for enchants. The fact that we consider them as actual cards still bugs me since their initial implementation...

* **Modifications that may affect results**: 
None

Atemo:
>No idea how official servers check for enchants. The fact that we consider them as actual cards still bugs me since their initial implementation...

(For info) On aegis the item data has a field to differentiate a normal card from enchant card, our item_db doesn't have that (not yet)

aleos89:
I was going to suggest we could use SubType and add Card support to it.

PR #6390, opened 2021-12-04
Updated F_IsCharm function

Everade:
* **Addressed Issue(s)**:
Fixes https://github.com/rathena/rathena/issues/6389

* **Server Mode**:
Renewal

Updated "F_IsCharm" function to support all latest enchants.

> **Note:**
> This function is used in a variety of NPCs which need to differentiate cards with enchants.
> Please note that there are some gaps in between these ID ranges, but since we did the same with ID 4700 - 4999, i figured it 
> should be fine as long as KRO doesn't utilizes those IDs in the future.


Everade:
Sorry for the mess, doing a new fork.

Issue #6396, opened 2021-12-08
Bio Explosion skill supposed to reduce intimacy at 1, its not the case.


Daraen1:
* **rAthena Hash**: 
 
    Lasted version (version of 08 december 2021)

* **Client Date**: 

     2018 04 18

* **Server Mode**: 

     Pre-renewal

* **Description of Issue**: 
  I think the Bio Explosion skill has a problem. According to several databases, after using the skill, the privacy should be reduced to 1.
  This is currently not the case
 https://www.youtube.com/watch?v=QrFR1Oid-T0
 
Note: I don't really know if this is a bug, after asking the question about the rAthena discord several times, I didn't get an answer, so I took the initiative to post it as a problem.
If a staff member can add the label "need more information" that would be nice.

aleos89:
Fixed in [eff452f](https://github.com/rathena/rathena/commit/eff452fa3526250e9c64da55bb260b274add67d3).

PR #6400, opened 2021-12-08
Fixed intimacy reduction for Vanilmirth Self-destruction and S.B.R.44

Daegaladh:
* **Addressed Issue(s)**: #6396 

* **Server Mode**: Both

* **Description of Pull Request**: Fixes Vanilmirth Self-destruction not reducing the intimacy level to 1 when not hitting any enemies

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -2441,12 +2441,11 @@ int skill_counter_additional_effect (struct block_list* src, struct block_list *
 		sc_start(src,src,SC_BLIND,2*skill_lv,skill_lv,skill_get_time2(skill_id,skill_lv));
 		break;
 	case HFLI_SBR44:	//[orn]
-	case HVAN_EXPLOSION:
 		if(src->type == BL_HOM){
-			TBL_HOM *hd = (TBL_HOM*)src;
-			hd->homunculus.intimacy = (skill_id == HFLI_SBR44) ? 200 : 100; // hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION)
-			if (sd && hd->master)
-				clif_send_homdata(sd,SP_INTIMATE,hd->homunculus.intimacy / 100);
+			TBL_HOM *hd = BL_CAST(BL_HOM,src);
+			hd->homunculus.intimacy = 200; // hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION)
+			if (hd->master)
+				clif_send_homdata(hd->master,SP_INTIMATE,hd->homunculus.intimacy / 100);
 		}
 		break;
 	case CR_GRANDCROSS:
@@ -7711,6 +7710,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			return 1;
 		}
 		status_damage(src, src, sstatus->max_hp,0,0,1, skill_id);
+		if(skill_id == HVAN_EXPLOSION && src->type == BL_HOM) {
+			TBL_HOM *hd = BL_CAST(BL_HOM,src);
+			hd->homunculus.intimacy = 100; // hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION)
+			if (hd->master)
+				clif_send_homdata(hd->master,SP_INTIMATE,hd->homunculus.intimacy / 100);
+		}
 		break;
 	case AL_ANGELUS:
 #ifdef RENEWAL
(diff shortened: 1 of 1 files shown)


Daraen1:
@Daegaladh i confirm. The fix work fine! Thank you

src/map/skill.cpp
```diff
@@ -2441,12 +2441,11 @@ int skill_counter_additional_effect (struct block_list* src, struct block_list *
 		sc_start(src,src,SC_BLIND,2*skill_lv,skill_lv,skill_get_time2(skill_id,skill_lv));
 		break;
 	case HFLI_SBR44:	//[orn]
-	case HVAN_EXPLOSION:
 		if(src->type == BL_HOM){
-			TBL_HOM *hd = (TBL_HOM*)src;
-			hd->homunculus.intimacy = (skill_id == HFLI_SBR44) ? 200 : 100; // hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION)
+			TBL_HOM *hd = BL_CAST(BL_HOM,src);
```
Lemongrass3110, line 2445:
```suggestion
			struct homun_data *hd = (struct homun_data *)src;
```
Please use normal cast here.
The BL type was already checked above.

src/map/skill.cpp
```diff
@@ -2441,12 +2441,11 @@ int skill_counter_additional_effect (struct block_list* src, struct block_list *
...
+			hd->homunculus.intimacy = 200; // hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION)
```
Lemongrass3110, line 2446: I have no idea about the skill, but judging from the description I think that both skills should set the intimacy to hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION) instead of hardcoded 200/100.
Can we confirm that?

src/map/skill.cpp
```diff
@@ -7711,6 +7710,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			return 1;
 		}
 		status_damage(src, src, sstatus->max_hp,0,0,1, skill_id);
+		if(skill_id == HVAN_EXPLOSION && src->type == BL_HOM) {
+			TBL_HOM *hd = BL_CAST(BL_HOM,src);
```
Lemongrass3110, line 7714: Please use normal cast here.
The BL type was already checked above.
```suggestion
			struct homun_data *hd = (struct homun_data *)src;
```

src/map/skill.cpp
```diff
@@ -7711,6 +7710,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			hd->homunculus.intimacy = 100; // hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION)
```
Lemongrass3110, line 7715:
```suggestion
			hd->homunculus.intimacy = hom_intimacy_grade2intimacy(HOMGRADE_HATE_WITH_PASSION);
```
Here it should be possible to use this since 100 and the value behind the function are the same.

src/map/skill.cpp
Daegaladh: Can't find any reliable source, apparently that 200 only appears in irowiki, but aegis probably just sets the intimacy to hate with passion

Lemongrass3110 (approved):

src/map/skill.cpp
Daegaladh: @Lemongrass3110 I've been doing some research and found some old posts on iRO boards talking about S.B.R.44 reducing the intimacy by 100 * skill level (same as damage), but others say it just consumes all the intimacy, so no idea...

aleos89: Maybe @mrjnumber1 can help us here!

Daraen1:
Most wikis like iRo wiki , divine pride or rms say intimacy is reduced to 1. If most databases agree, I think they can be trusted.

[Divine Pride](https://www.divine-pride.net/database/skill/8016/)
[Rate My Server](https://ratemyserver.net/index.php?sk_name=bio+explosion&page=skill_db&sk_search=Search)
[iRO Wiki](https://irowiki.org/wiki/Self-Destruction)
[iRO Wiki Classic](https://irowiki.org/classic/Self-Destruction)

aleos89 (approved):

Daraen1:
i don't know where i'm wrong but after S.B.R.44 skill, my filir have 1 intimacy.

i applied the code that you just sent.

S B R 44

PR #6406, opened 2021-12-12
Increases MAX_LEVEL to 250

Atemo:
* **Addressed Issue(s)**: N/A ?

* **Server Mode**: 

Note:
- The max level for each unit type is still capped in `job_exp.yml` and in conf folder


src/map/map.hpp
```diff
@@ -52,7 +52,7 @@ void map_msg_reload(void);
 #define NATURAL_HEAL_INTERVAL 500
 #define MIN_FLOORITEM 2
 #define MAX_FLOORITEM START_ACCOUNT_NUM
-#define MAX_LEVEL 200
+#define MAX_LEVEL 250
```
Lemongrass3110, line 55:
```suggestion
#ifdef RENEWAL
	#define MAX_LEVEL 250
#else
	#define MAX_LEVEL 99
#endif
```

db/re/statpoint.yml
```diff
@@ -431,6 +431,106 @@ Body:
...
+    Points: 4045
+  - Level: 247
+    Points: 4045
+  - Level: 248
+    Points: 4045
+  - Level: 249
+    Points: 4045
+  - Level: 250
+    Points: 4045
   # - Level: 201
```
Lemongrass3110, line 534: Can you remove the old commented entries?

db/pre-re/statpoint.yml
```diff
@@ -431,6 +431,106 @@ Body:
     Points: 4503
   - Level: 200
     Points: 4545
+  - Level: 201
+    Points: 4045
```
Lemongrass3110, line 435: Why do we add those here?
Anyway: 4045 is smaller than level 200 with 4545

PR #6408, opened 2021-12-14
Converts the Mob Skill DB file into YAML

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

  * Converts the Mob Skill DB file into YAML.

-----------------------------------------------------------------------------

- [ ] Removes txt file before merge !
- [ ] Implementes `Val5`. Currently only 5 slaves can be spawn on rAthena, official monsters can spawn (at least) 6 monsters


src/tool/csv2yaml.cpp
```diff
@@ -4781,3 +4803,189 @@ static bool pc_readdb_skilltree_yaml(void) {
...
+						body << YAML::Key << "Cond2String" << YAML::Value << it.cond2_name;
+				}
+				for (const auto &valit : it.val) {
+					std::string val_name = "Val" + std::to_string(valit.first);
+
+					body << YAML::Key << val_name << YAML::Value << valit.second;
+				}
+
+				body << YAML::EndMap;
+			}
```
aleos89, line 4978:
```suggestion
		if (!it.cond1_name.empty() && it.cond1_name != "ALWAYS")
			body << YAML::Key << "Condition" << YAML::Value << it.cond1_name;

		body << YAML::Key << "ConditionValues";
		body << YAML::BeginMap;

		uint16 cond_index = 0;

		body << YAML::Key << "Index" << YAML::Value << cond_index;

		if (!it.cond2_name.empty()) { // Cond2 becomes Val0
			if (atoi(it.cond2_name.c_str()) > 0)
				body << YAML::Key << "Value" << YAML::Value << it.cond2_name;
			else
				body << YAML::Key << "Value" << YAML::Value << it.cond2_name;
		}

		for (const auto &valit : it.val) {
			cond_index++;
			body << YAML::Key << "Index" << YAML::Value << cond_index;
			body << YAML::Key << "Value" << YAML::Value << valit.second;
		}

		body << YAML::EndMap;
```

src/map/mob.cpp
```diff
@@ -5671,269 +5664,464 @@ uint64 MobChatDatabase::parseBodyNode(const YAML::Node &node) {
...
+	if (atoi(cond2.c_str()) > 0)
+		Condition["Cond2Value"] = str[++index];
 	else
-		ms->msg_id = 0;
+		Condition["Cond2String"] = str[++index];
+	Condition["Val0"] = str[++index];
+	Condition["Val1"] = str[++index];
+	Condition["Val2"] = str[++index];
+	Condition["Val3"] = str[++index];
+	Condition["Val4"] = str[++index];
```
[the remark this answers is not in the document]
aleos89: I'm also thinking if we can get rid of the hard coded limit for ConditionValues and just use an iterator.

src/map/mob.hpp
```diff
@@ -178,18 +178,36 @@ enum e_aegis_monsterclass : int8 {
...
-	enum MobSkillState state;
+	e_MobSkillState state;
 	uint16 skill_id,skill_lv;
 	short permillage;
 	int casttime,delay;
-	short cancel;
+	bool cancel;
 	short cond1,cond2;
 	short target;
 	int val[5];
```
aleos89, line 188:
```suggestion
	std::vector<int32> val;
```
Would be great to vectorize this so that we can just remove the limit. If I'm thinking too far ahead that's fine and it can stay as a static array (but maybe increase it to 10 or something).

aleos89 (commented):
My 2¢!

db/import-tmpl/mob_skill_db.yml
```diff
@@ -0,0 +1,56 @@
...
+#        Cond1              Type of condition (Default: Always).
+#        Cond2Sring         String value (no value by default).
+#        Cond2Value         Integer value (default 0).
+#        Val0               Special value 0 according to Condition/Name (default 0).
+#        Val1               Special value 1 according to Condition/Name (default 0).
+#        Val2               Special value 2 according to Condition/Name (default 0).
+#        Val3               Special value 3 according to Condition/Name (default 0).
+#        Val4               Special value 4 according to Condition/Name (default 0).
+#      Emotion              Emotion displayed on the monster after (? while casting) the cast (no emotion by default).
+#      Chat                 Chat ID to display the associated sentence defined in mob_chat_db.yml (default 0).
```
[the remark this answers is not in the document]
aleos89: I couldn't open the 2 main mob_skill_db files because they were too large to preview, so don't forget those!

src/tool/csv2yaml.cpp
```diff
@@ -4918,3 +4940,229 @@ static bool itemdb_read_combos(const char* file) {
...
+			if (ai == nullptr || strncasecmp(str[index], "0x", 2) != 0) {
+				ShowError( "Unknown Ai for mode %s.\n", str[index] );
+				return false;
+			}
+			entry.mob_ai = *ai;
+		}
+	}
+
+	if (*str[17]) {
+		char *constant = const_cast<char *>(constant_lookup(atoi(str[17]), "ET_"));
```
aleos89, line 5093:
```suggestion
		char *constant = const_cast<char *>(constant_lookup(atoi(str[17]), "ET_") + 3);
```

aleos89 (commented):
Another review!

src/map/mob.hpp
```diff
@@ -185,19 +185,81 @@ enum e_aegis_monsterclass : int8 {
...
 	int casttime,delay;
-	short cancel;
-	short cond1,cond2;
-	short target;
-	int val[5];
+	bool cancel;
+	e_mob_skill_target target;
+	short cond1, cond2, cond3;
+	int mob_mode;
+	int val[6];
```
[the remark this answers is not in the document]
aleos89: Gotcha! Yeah a umap works too. The size is super small so we shouldn't see any noticeable delay in searches.

src/map/skill.cpp
```diff
@@ -9499,36 +9499,25 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
-			if(!battle_config.npc_emotion_behavior && skill_id == NPC_EMOTION && md->db->skill[md->skill_idx]->val[1])
-				mob_unlocktarget(md,tick);
+			int mode_passive = (md->db->skill[md->skill_idx]->mob_mode & MD_AGGRESSIVE) ? 0 : MD_AGGRESSIVE;	// Remove aggressive mode when the new mob type is passive.
 
-			if(md->db->skill[md->skill_idx]->val[1] || md->db->skill[md->skill_idx]->val[2])
-				sc_start4(src,src, type, 100, skill_lv,
-					md->db->skill[md->skill_idx]->val[1],
-					md->db->skill[md->skill_idx]->val[2],
-					md->db->skill[md->skill_idx]->val[3],
+			if (md->db->skill[md->skill_idx]->mob_mode > -1)
```
Playtester, line 9516: Looks like you gave NPC_EMOTION_ON a new functionality. Can you please document your detailed tests on which base you are changing this and link it to this PR? I couldn't find an issue that's being linked to this.

If needed, I can help checking on pre-renewal to make sure this isn't a renewal-only change.

src/map/skill.cpp
```diff
@@ -9537,6 +9526,9 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		}
 		break;
 
+	case NPC_EMOTION:
+		break;
```
Playtester, line 9530: I don't think we can move the functionality out like this, certain things only should happen when this skill is used, like how I recently found out it should show "Go!" for angry monsters.

I'm also not sure if for example monsters should drop target if it's not this skill. Should be verified and documented. I always thought the difference between NPC_EMOTION and NPC_EMOTION_ON was that on the latter the monster doesn't drop the target.

Playtester (commented):
Some concerns on the NPC_EMOTION changes.

src/map/skill.cpp
Atemo: NPC_EMOTION_ON and NPC_EMOTION, after this commit, should work exactly as before. Both skills did nothing when no mode was provided. This commit just extends the functionality to set the AI of the mob when it casts another skill :
```
// Monster Name,    State,    Skill ID,    Level,    Probability,    Cast Time,    Delay Time,    NO_DISPEL,    AI Setting,    SEND_EMOTICON/SEND_CHAT,     Emotion #/Chat #,    IF,    Condition value
E_NYAIL_PIG    BERSERK_ST    AL_DECAGI    5    700    500    15000    NO_DISPEL    MONSTER_TYPE_17
E_SALAMANDER    RUSH_ST    WZ_METEOR    9    300    500    20000        MONSTER_TYPE_21
E_SALAMANDER    RUSH_ST    NPC_WIDESILENCE    5    500    700    30000        MONSTER_TYPE_21    SEND_EMOTICON    9
E_SALAMANDER    RUSH_ST    NPC_ALLHEAL    1    1000    10000    30000    NO_DISPEL    MONSTER_TYPE_21    SEND_EMOTICON    28    IF_HP    10
GRIM_REAPER_ANKOU    BERSERK_ST    NPC_COMET    3    100    10000    75000        MONSTER_TYPE_21
GRIM_REAPER_ANKOU    BERSERK_ST    NPC_DARKCROSS    10    50    2000    12000        MONSTER_TYPE_21
GRIM_REAPER_ANKOU    BERSERK_ST    NPC_DRAGONFEAR    5    100    3000    30000        MONSTER_TYPE_21
TENDRILRION    BERSERK_ST    PF_SPIDERWEB    1    300    0    10000        MONSTER_TYPE_04    SEND_EMOTICON    7
```

The lines are differents here because this commit no longer uses the val[] set when loading mob_skill_db :
```
// from mob_parse_row_mobskilldb (txt format)
if(ms->skill_id == NPC_EMOTION && mob_id > 0 &&
		ms->val[1] == mob->status.mode)
	{
		ms->val[1] = 0;
		ms->val[4] = 1; //request to return mode to normal.
	}
	if(ms->skill_id == NPC_EMOTION_ON && mob_id > 0 && ms->val[1])
	{	//Adds a mode to the mob.
		//Remove aggressive mode when the new mob type is passive.
		if (!(ms->val[1]&MD_AGGRESSIVE))
			ms->val[3] |= MD_AGGRESSIVE;
		ms->val[2] |= ms->val[1]; //Add the new mode.
		ms->val[1] = 0; //Do not "set" it.
```

Basically from what I understand from the current code, NPC_EMOTION 'sets' (or removes) a mode depending of the mode provided while NPC_EMOTION_ON 'adds' the mode provided to the current mode.

Playtester: I always thought the "adding" of a mode was just a custom feature, especially since Aegis doesn't even have those granular modes that we have. It didn't depend on NPC_EMOTION vs. NPC_EMOTION_ON, but rather on whether you set var2, var3 and/or var4.
In our default mob_skill_db, only var1 was ever used AFAIK.

We probably need more tests on NPC_EMOTION_ON.

EvilCloudHermit:
Sorry to just write in here like this, but it would be great if we could continue working here. Editing monster kills would be so much easier with this change.

vstumpf:
I got a question, why can't we add this to mob_db and just split the files, like how we have our pcdb?

Atemo:
It was a possibility, but is there any benefit to it?
The current structure is used to preserve old txt options which will not be possible with mob_db (a cons)
```
// Note: if a negative MobID is provided, the skill will be treated as 'global':
//	-1: added for all boss types.
//	-2: added for all normal types.
//	-3: added for all mobs.
```

Issue #6423, opened 2021-12-28
[ Error ] : buildin_getexp: Called with base and job exp 0


mazvi:
* **rAthena Hash**: 745731b

* **Client Date**: 20200401

* **Server Mode**: RENEWAL

* **Description of Issue**: 

This is error_log on map log

(12/28/2021 07:58:25) [ Error ] : buildin_getexp: Called with base and job exp 0.
(12/28/2021 07:58:25) [ Debug ] : Source (NPC): Commodities Dealer#pa08 at mora (127,112)
(12/28/2021 07:58:25) [ Debug ] : Source (NPC): Commodities Dealer#pa08 is located in: npc/re/quests/quests_mora.txt
(12/28/2021 07:58:29) [ Error ] : buildin_getexp: Called with base and job exp 0.
(12/28/2021 07:58:29) [ Debug ] : Source (NPC): General Goods Dealer#pa0 at mora (119,118)
(12/28/2021 07:58:29) [ Debug ] : Source (NPC): General Goods Dealer#pa0 is located in: npc/re/quests/quests_mora.txt

  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

PR #6424, opened 2021-12-28
Fixed getexp calls in quests_mora.txt

Lemongrass3110:
* **Addressed Issue(s)**: #6423

* **Server Mode**: Renewal

If a player reached job level 50 or 70 no exp would be given and the getexp call is useless.
Therefore it should be prevented.

Thanks to @mazvi

### DIFF
--- a/npc/re/quests/quests_mora.txt
+++ b/npc/re/quests/quests_mora.txt
@@ -4928,10 +4928,10 @@ L_CheckQuest:
 		delitem getarg(1),getarg(2);
 		erasequest getarg(0);
 		specialeffect2 EF_STEAL;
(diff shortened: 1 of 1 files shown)


Atemo:
How about just removing the showerror in script.cpp and just returning SCRIPT_CMD_SUCCESS, the message is more annoying than useful
https://github.com/rathena/rathena/blob/master/src/map/script.cpp#L10614

Lemongrass3110:
Its useful to find invalid/unintentional calls. Like if you use a wrong variable name and it auto-resolves to 0.

secretdataz (approved):

Atemo (approved):

PR #6429, opened 2021-12-29
Always set all item data in npc_sellist_sub

### LINKED ISSUE #6422: Can't determine Brewed White Slim Potions when trigger OnSellItem
* **rAthena Hash**: 

6ddb09cbb

* **Client Date**: 

2018-06-20e

* **Server Mode**: 

Pre-Renewal

* **Description of Issue**: 
  * Result: Can't determine if a user sell a Brewed Potion (Ex: ClientName's White Slim Potion) when using the **callshop** syntax.
  * Expected Result: Should be atleast see the Card ID slot 1 = 254
  * How to Reproduce: Create a script that has **callshop,2** (Sell only) then try to sell a brewed potion. If you check the results of Card1 to Card4 you won't be able to see any ID. But if you try to sell a Forged Weapon you can see an ID per Card slot.

secretdataz:
* **Addressed Issue(s)**: Fixes #6422 
* **Server Mode**: Both
* **Description of Pull Request**: As described in #6422, there is no way to obtain an item creator's char id when using script-attached shop.


secretdataz:
@ronsoriano any feedback on the code I provided?

Lemongrass3110:
Guess the real fix would be to remove the "if itemdb_isequip".
You can build the name from the cards just normal then?

aleos89:
Agreed, wouldn't just removing line 2170 check sort this? The rest of the values would just be 0/null for non-equip items.

aleos89 (approved):

Issue #6415, opened 2021-12-19
Missing support for unequip all button


admkakaroto:
New clients have a new small feature that allows you to remove all items equipped on the character by clicking a button, on Hercules it has this packet, but on RA not yet, I would like to request the addition here too.

Client date: 2021-11-17_Ragexe

Packet: CZ_REQ_TAKEOFF_EQUIP_ALL, 0x0bad

Window:
![a111](https://user-images.githubusercontent.com/6227618/146679763-dad64c11-37d4-4d37-8c86-68b0e34db2e3.png)

Warning:
`clif_parse: Received unsupported packet (packet 0x0bad, 2 bytes received), disconnecting session #3.
`
Result:
Disconect from the game.

Balferian:
It should unequip items only from first tab, without shadow equip / decor.
Tested on kRo.
![image](https://user-images.githubusercontent.com/16213511/146689707-3d6b832c-3624-4867-bba5-54e26465c80f.png)

Lemongrass3110:
This client is not officially supported yet ;-)

RagnaWay:
`pUnequipAllItems`
https://github.com/HerculesWS/Hercules/commit/01bbe6b5b3846275d094cea4dd7e1a0e868d3490#diff-754b8ca8fe2af31705c40781ffe8bc82f1c3f830670a0b571cb0d674faf58226R2022

PR #6431, opened 2021-12-30
Initial support for 2021-11-03RagexeRE

Lemongrass3110:
* **Addressed Issue(s)**: #6415

* **Server Mode**: Both

Adds support for the latest kRO Sakray clients.

### DIFF
--- a/src/char/char-server.vcxproj
+++ b/src/char/char-server.vcxproj
@@ -177,6 +177,7 @@
     <ClInclude Include="int_pet.hpp" />
     <ClInclude Include="int_quest.hpp" />
     <ClInclude Include="int_storage.hpp" />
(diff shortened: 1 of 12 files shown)


laziem:
`[Warning]: clif_parse: Received unsupported packet (packet 0x0b93, 12 bytes received)`
comes from setting

image

admkakaroto:
Is it possible for Lemon to add a lock so that the button does not work to disable everything while interacting with npc's? I believe it is safer to avoid exploits or bugs.

hnomkeng:
can share 2021-11-03RagexeRE client ?

Lemongrass3110:
This is not a support area. Please ask such things in Discord.
It has been shared there previously already.

Issue #6434, opened 2021-12-31
Pre-re map server crash when changing to 4th job


Shaktohh:
* **rAthena Hash**: 7556453dfcd21180566ab56a891f628caa75e3db

* **Client Date**: 20200401

* **Server Mode**: Pre-Renewal

* **Description of Issue**: Change job in game to a 4th job and the map server crash

[Error]: Server received crash signal! Attempting to save all online characters!

mundussan:
Hmmmm, are you trying to use 4th jobs in pre renewal mode? Or you mean after merging 4th jobs, your pre renewal got buggy?

kaninhot004:
> Hmmmm, are you trying to use 4th jobs in pre renewal mode? Or you mean after merging 4th jobs, your pre renewal got buggy?

As I read, he want to use class 4 on pre-renewal.

Shaktohh:
> Hmmmm, are you trying to use 4th jobs in pre renewal mode? Or you mean after merging 4th jobs, your pre renewal got buggy?

Yes i changed to a 4th jobs.
I guess if it shouldn't be allowed, there should be a safeguard.

mundussan:
Well, afaik you can only use 3rd and 4rth jobs in pre renewal using the atcommand `@job`
But i never got an server map crash, my job only got recognized as novice 99.

Shaktohh:
Here the full bt 
`Thread 1 "map-server" received signal SIGSEGV, Segmentation fault.
0x00005555557c6037 in pc_calc_skilltree_normalize_job (sd=0x555559b8a120) at /usr/include/c++/8/bits/shared_ptr_base.h:1018
1018          _M_get() const noexcept
bt full
#0  0x00005555557c6037 in pc_calc_skilltree_normalize_job (sd=0x555559b8a120) at /usr/include/c++/8/bits/shared_ptr_base.h:1018
        skill_point = 217
        novice_skills = 9
        c = 53508
        novice_job = std::shared_ptr<s_job_info> (use count 2, weak count 0) = {get() = <optimized out>}
#1  0x00005555557c6649 in pc_calc_skilltree (sd=sd@entry=0x555559b8a120) at /home/debian/rAthena/src/map/pc.cpp:2195
        job = <optimized out>
        class_ = <optimized out>
        job_id = <optimized out>
        __FUNCTION__ = "pc_calc_skilltree"
        flag = <optimized out>
        tree = <optimized out>
[...]

PR #6435, opened 2022-01-01
Disallow changing to 4th on pre-renewal

Lemongrass3110:
* **Addressed Issue(s)**: #6434

Disables changing to jobs that dont exist in the job database.
If a player logs in with a job that does not exist in the job database the job will be changed to Novice.

Thanks to @Shaktohh


mazvi:
How about skillpoint and statpoint? Need reset to like first time register?

aleos89 (approved):

Lemongrass3110:
> How about skillpoint and statpoint? Need reset to like first time register?

@mazvi I am only using existing logic here. Actually characters that reach this are broken and you will have to manually fix them inside your database.

Issue #6395, opened 2021-12-07
Pet Capture Rate always 100%


Everade:
* **rAthena Hash**: 
4a778815a344a5555209dc79fa80b8fff11fbad9

* **Client Date**: 
Not relevant

* **Server Mode**: 
Renewal

* **Description of Issue**: 
  * Result: Pet Capture rate is extremely off. (always 100%)
  * Expected Result: Pet Database to set pet capture rate accordingly as given in the doc. (10000 = 100%)
  * How to Reproduce: Try to tame any pet that has a very low capture rate.
  To test this even further, i've set the battle configuration pet_catch_rate: 10 down, yet it still seems to be at 100% instead of 0.05%.
  Only with pet_catch_rate: 1 i started to actualy fail sometimes, but the chance was still somewhere inbetween 10~50% instead of 0.005% for Hydra.
  * Official Information: Obviously wrong.

* **Modifications that may affect results**: 
reduced pet_catch_rate:  battle config for further rate reduction.

Lemongrass3110:
```
	pet_catch_rate = (pet->capture + (sd->status.base_level - md->level)*30 + sd->battle_status.luk*20)*(200 - get_percentage(md->status.hp, md->status.max_hp))/100;

	if(pet_catch_rate < 1)
		pet_catch_rate = 1;

	if(battle_config.pet_catch_rate != 100)
		pet_catch_rate = (pet_catch_rate*battle_config.pet_catch_rate)/100;
```

Everade:
Silly me, i had 15k luk from previous tests.
Thanks for the info.

Lemongrass3110:
Confirmed that our current formula is wrong.
Thanks to @mrjnumber1 for the official formula:
```
bool CPC::IsCaptureSuccess(CNPC* pTarget, int item)
{
    auto info = g_petDefaultValue->GetDefaultValue(pTarget->m_characterInfo.job);
    if (!info) return false;
    
    auto percent = info->iCapturePercent;
    int hp=0, maxhp=0;
    pTarget->GetParameterValue(VAR_HP, &hp);
    pTarget->GetParameterValue(VAR_MAXHP, &maxhp);
    percent += (1.0 - hp / maxhp) * percent;
    return (GetServerRandom(0, 20000)%100 < (int)percent);
}
```

PR #6439, opened 2022-01-03
Fixed pet capture rate

Lemongrass3110:
* **Addressed Issue(s)**: #6395

* **Server Mode**: Both

Fixed the pet capture rate formula.
Added some missing checks for distance to the target monster and if the target monster is hiding.

Thanks to @Everade and @mrjnumber1

### DIFF
--- a/conf/battle/pet.conf
+++ b/conf/battle/pet.conf
@@ -19,8 +19,8 @@ pet_legacy_formula: no
 // Rate for catching pets (Note 2)
 pet_catch_rate: 100
 
-// How many cells can the player be away from the monster he is trying to catch?
-// The client automatically walks into this range when you try to catch a monster.
+// How many cells away can the player be from the monster they are trying to catch?
+// The client automatically walks the player into range when trying to catch a monster.
 // Default: 5
 // Set to 0 to disable the check
 pet_distance_check: 5
--- a/conf/battle/pet.conf
+++ b/conf/battle/pet.conf
@@ -9,8 +9,24 @@
 //         assume unit types (1: Pc, 2: Mob, 4: Pet, 8: Homun, 16: Mercenary, 128: NPC, 512: Elemental)
 //--------------------------------------------------------------
 
+// Should the legacy formula for pet catch rate calculation be used? (Note 1) 
+// Legacy Athena formula:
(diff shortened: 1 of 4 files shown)


conf/battle/pet.conf
```diff
@@ -9,9 +9,25 @@
...
+//		( Base rate + ( player level - monster level ) * 30 + player luk * 20 ) * ( 200 - current monster hp percentage ) / 100
+// Official formula: (Default)
+//		Base rate + ( 100 - current monster hp percentage ) * base rate / 100
+pet_legacy_formula: no
+
 // Rate for catching pets (Note 2)
 pet_catch_rate: 100
 
+// How many cells can the player be away from the monster he is trying to catch?
+// The client automatically walks into this range when you try to catch a monster.
```
aleos89, line 23:
```suggestion
// How many cells away can the player be from the monster they are trying to catch?
// The client automatically walks the player into range when trying to catch a monster.
```

src/map/pet.cpp
```diff
@@ -1271,7 +1271,27 @@ int pet_catch_process2(struct map_session_data* sd, int target_id)
...
+	if( battle_config.pet_distance_check && distance_bl( &sd->bl, &md->bl ) > battle_config.pet_distance_check ){
+		clif_pet_roulette( sd, 0 );
+		sd->catch_target_class = PET_CATCH_FAIL;
+
+		return 1;
+	}
+
+	struct status_change* tsc = status_get_sc( &md->bl );
+
+	if( battle_config.pet_hide_check && tsc && ( tsc->data[SC_HIDING] || tsc->data[SC_CLOAKING] || tsc->data[SC_CAMOUFLAGE] || tsc->data[SC_NEWMOON] || tsc->data[SC_CLOAKINGEXCEED] ) ){
```
aleos89, line 1283: Just a thought, but I don't think mobs have the Camouflage, New Moon, or Cloaking Exceed skills.

aleos89 (approved):

src/map/pet.cpp
Lemongrass3110: I just copied the check from somewhere else and thought why leave one out.
If someone gives the skill to mobs they might have it.

PR #6488, opened 2022-01-09
Converted player groups to YAML

Lemongrass3110:

* **Server Mode**: Both

Converts the player groups to YAML.
After this only channels are remaining to be able to drop libconfig dependency completely.


conf/groups.yml
```diff
@@ -0,0 +1,245 @@
...
+#   Level               GM Level to compare which group is ranked higher
+#   LogCommands         Whether atcommands should be logged or not (Default: false)
+#   Commands:           List of atcommands that can be used by this group (Default: none)
+#     <atcommand name>  Whether an atcommand can be used by this group or not (true/false)
+#   CharCommands:       List of charcommands that can be used by this group (Default: none)
+#     <atcommand name>  Whether an atcommand can be used by this group or not (true/false)
+#   Permissions:        List of permissions the this group has (Default: none)
+#     <permission name> Whether the group has this permission or not (true/false)
+#   Inherit:            List of groups that will be inherited (Default: none)
+#     <group name>      Whether this group will be inherited or not (true/false)
```
aleos89, line 36:
```suggestion
# - Id                    Group ID.
#   Name                  Group name.
#   Level                 GM Level used for ranking groups. (Default: 0)
#   LogCommands           Whether atcommands should be logged or not. (Default: false)
#   Commands:             List of atcommands that can be used by this group. (Default: none)
#     <atcommand name>    Whether the specified atcommand can be used by this group or not.
#   CharCommands:         List of charcommands that can be used by this group. (Default: none)
#     <charcommand name>  Whether the specified charcommand can be used by this group or not.
#   Permissions:          List of permissions the group has. (Default: none)
#     <permission name>   Whether the group has this permission or not.
#   Inherit:              List of groups that will be inherited. (Default: none)
#     <group name>        Whether this group will be inherited or not.
```

conf/import-tmpl/groups.yml
```diff
@@ -0,0 +1,41 @@
...
+#   Level               GM Level to compare which group is ranked higher
+#   LogCommands         Whether atcommands should be logged or not (Default: false)
+#   Commands:           List of atcommands that can be used by this group (Default: none)
+#     <atcommand name>  Whether an atcommand can be used by this group or not (true/false)
+#   CharCommands:       List of charcommands that can be used by this group (Default: none)
+#     <atcommand name>  Whether an atcommand can be used by this group or not (true/false)
+#   Permissions:        List of permissions the this group has (Default: none)
+#     <permission name> Whether the group has this permission or not (true/false)
+#   Inherit:            List of groups that will be inherited (Default: none)
+#     <group name>      Whether this group will be inherited or not (true/false)
```
aleos89, line 36:
```suggestion
# - Id                    Group ID.
#   Name                  Group name.
#   Level                 GM Level used for ranking groups. (Default: 0)
#   LogCommands           Whether atcommands should be logged or not. (Default: false)
#   Commands:             List of atcommands that can be used by this group. (Default: none)
#     <atcommand name>    Whether the specified atcommand can be used by this group or not.
#   CharCommands:         List of charcommands that can be used by this group. (Default: none)
#     <charcommand name>  Whether the specified charcommand can be used by this group or not.
#   Permissions:          List of permissions the group has. (Default: none)
#     <permission name>   Whether the group has this permission or not.
#   Inherit:              List of groups that will be inherited. (Default: none)
#     <group name>        Whether this group will be inherited or not.
```

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
+	}
+
+	if( this->nodeExists( node, "Name" ) ){
+		std::string name;
+
+		if( !this->asString( node, "Name", name ) ){
+			return 0;
+		}
+
+		group->name = name;
```
aleos89, line 47: Do we care if a group shares a name? I know we separate on ID, not sure if we should restrict this or not. Just a thought!

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
+	}
+
+	if( this->nodeExists( node, "Level" ) ){
+		uint32 level;
```
aleos89, line 51:
```suggestion
		uint16 level;
```
We limit GM level to 99, guess there's no reason to go this high.

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
+
+		if( !this->asUInt32( node, "Level", level ) ){
```
aleos89, line 53:
```suggestion
		if( !this->asUInt16( node, "Level", level ) ){
```

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
+			return 0;
+		}
+
+		group->level = level;
```
aleos89, line 57:
```suggestion
		if (level > 99) {
			this->invalidWarning(node["Level"], "Group level %d exceeds 99, capping.\n", level);
			level = 99;
		}

		group->level = level;
```

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
+	}
```
aleos89, line 58:
```suggestion
	} else {
		if (!exists)
			group->level = 0;
	}
```

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
+		for( const auto& it : permissions ){
+			std::string permission = it.first.as<std::string>();
+			bool allowed = it.second.as<bool>();
+			const char* str = permission.c_str();
+			bool found = false;
+
+			for( const auto& permission_name : pc_g_permission_name ){
+				if( strcmpi( permission_name.name, str ) == 0 ){
+					if( allowed ){
+						group->permissions |= permission_name.permission;
```
aleos89, line 122: Maybe change the permissions to bitset so we won't be over indulged by large bitmasks again.

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
+
+	return 1;
+}
 
-			if (++loop > group_count) {
-				ShowWarning("pc_groups:read_config: Could not process inheritance rules, check your config '%s' for cycles...\n",
-				            config_filename);
-				break;
+void PlayerGroupDatabase::loadingFinished(){
+	static const int MAX_CYCLES = 10;
```
aleos89, line 173: Should we note in the docs that we only inherit up to 10 groups?

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,356 @@
...
-
-	if (pc_group_has_permission(group_id, PC_PERM_USE_ALL_COMMANDS))
+bool s_player_group::can_use_command( const std::string& command, AtCommandType type ){
+	if( this->has_permission( PC_PERM_USE_ALL_COMMANDS ) ){
 		return true;
+	}
 
-	if ((group = id2group(group_id)) == NULL)
-		return false;
+	std::unordered_map<std::string, bool>* map;
```
aleos89, line 277: We could possibly speed this up with `std::vector<std::string>`.

src/map/pc_groups.hpp
```diff
@@ -91,4 +88,36 @@ static const struct s_pcg_permission_name {
 	{ "all_permission", PC_PERM_ALLPERMISSION },
 };
 
+struct s_player_group{
+	uint32 id;
+	std::string name;
+	uint32 level;
```
aleos89, line 94:
```suggestion
	uint16 level;
```

src/map/pc_groups.hpp
```diff
@@ -91,4 +88,36 @@ static const struct s_pcg_permission_name {
...
+	bool log_commands;
+	std::unordered_map<std::string, bool> commands;
+	std::unordered_map<std::string, bool> char_commands;
+	uint32 permissions;
```
aleos89, line 98:
```suggestion
	std::bitset<uint8> permissions;
```
Setting to a bitset makes this enum much smaller!

src/map/pc_groups.hpp
```diff
@@ -91,4 +88,36 @@ static const struct s_pcg_permission_name {
 	{ "all_permission", PC_PERM_ALLPERMISSION },
 };
 
+struct s_player_group{
+	uint32 id;
+	std::string name;
+	uint32 level;
+	bool log_commands;
+	std::unordered_map<std::string, bool> commands;
+	std::unordered_map<std::string, bool> char_commands;
```
aleos89, line 97: Setting these as string vectors could speed up lookups!

aleos89 (commented):
Just keep in mind once ryml is merged this will break!

src/map/pc_groups.hpp
Lemongrass3110: These are only used once during loading. Afterwards they are cached in atcommand.cpp

src/map/pc_groups.cpp
Lemongrass3110: Since it is only printed in some GM commands I would not have bothered if someone is stupid enough to do that.
But if you want we can validate that.

src/map/pc_groups.cpp
Lemongrass3110: Its only if those groups inherit in circles. Unless that happens it is not limited intentionally.

src/map/pc_groups.cpp
aleos89: It's not of any super importance. I figured all we used it for was simply display a name, but should be no harm if a server wants groups to have similar names.

src/map/pc_groups.cpp
Lemongrass3110: I will leave it as uint32 for the time being, as this is how it was before.
We can do another cleanup later.

src/map/pc_groups.cpp
Lemongrass3110: I did not add this, as level is a mandatory node in the !exists case, see above.

src/map/pc_groups.hpp
Lemongrass3110: Done!

src/map/pc_groups.cpp
Lemongrass3110: This is only used during load time.
Do you still think this is necessary?

Lemongrass3110: Done!

src/map/pc_groups.cpp
```diff
@@ -3,462 +3,394 @@
...
+		if( !atcommand_exists( command.c_str() ) ){
+			this->invalidWarning( it, "Unknown atcommand: %s\n", command.c_str() );
+			return false;
+		}
+
+		util::tolower( command );
+
+		if( allowed ){
+			if( util::vector_exists( commands, command ) ){
+				this->invalidWarning( it, "Group does already have command \"%s\". Please check your data.\n", command.c_str() );
```
aleos89, line 41:
```suggestion
				this->invalidWarning( it, "Group already has command \"%s\". Please check your data.\n", command.c_str() );
```

aleos89 (approved):
Few minor things, everything else looks good from my tests!

Issue #6497, opened 2022-01-11
Level 2 Shields are not recognized by skills as shield


Relliksuriv:
* **rAthena Hash**: dcd01a0

* **Client Date**: 20211103

* **Server Mode**: Renewal

* **Description of Issue**: Level 2 Shields are not seen as shield by skills that require shield e.g reflect shield, auto guard and all other skills
  * Result: 
  * Expected Result: Wearing Level 2 shields should enable you to cast this skills 
  * How to Reproduce: Wear any Level 2 shields e.g Penitentia Aegis, Mad Bunny LT
  * Official Information: 
  

* **Modifications that may affect results**:

laziem:
looks the item missing their View field. Try add `View: 1`

PR #6498, opened 2022-01-11
Added some missing shield view ids

Lemongrass3110:
* **Addressed Issue(s)**: #6497

* **Server Mode**: Renewal

Shield state check for skills is implemented to check if a view id of a shield is set on the character.
Therefore it is required that the shields have their view id set.

Thanks to @Relliksuriv and @laziem


aleos89 (approved):

Atemo:
How about adding a warning during the parsing for `Left_Hand` since the view is required?

Atemo (approved):

Issue #6492, opened 2022-01-10
Add AP in percentheal


RagnaWay:
**Describe the solution you'd like**

With the official command you can restore HP and SP, so it would be very interesting to add AP in the same role.

Example: `percentheal HP,SP,AP;` =  `percentheal 100,100,100;`

thank you **Maple** for the idea.

aleos89:
While this will eventually need to be added, the only problem with utilizing the current commands (IE: percentheal, heal, etc) is that the final optional argument on these is the character ID. This means we'd most likely end up having to do percentheal2, heal2, etc. which is not a very elegant solution.

Atemo:
How about breaking the rule saying the char id should be the last argument and update the command to make it uglier?
`percentheal <hp>,<sp>{,<char_id>,<ap>};`

aleos89:
The -100~100 checks for percent heal work good for checking the character ID but the other heal functions that heal a specific value won't work well in this case for anything over that starting character value.

Lemongrass3110:
Oh well true. I was only thinking about percentheal now.
Since Rytech added a new atcommand for apheal as well and it is not included with the normal "heal", should we create additional script commands too?

aleos89:
Yeah, I was thinking then the only other solution aside from “2” would be just “apheal” or “healap”.

PR #6514, opened 2022-01-13
Script command healap

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/6492

* **Server Mode**: renewal

Added `healap` script command


aleos89 (approved):

Lemongrass3110 (approved):
Shouldnt we add a percent version too?

aleos89:
I guess I'm 50/50 on the percent version since a player can only have 200AP max. Unless we think there will be items and such that will heal AP by %, then we can go for it!

Lemongrass3110:
> I guess I'm 50/50 on the percent version since a player can only have 200AP max. Unless we think there will be items and such that will heal AP by %, then we can go for it!

True that!

PR #6520, opened 2022-01-14
Github Action for NPC and DB Validation

Akkarinage:
* **Addressed Issue(s)**: Lack of automated testing

* **Server Mode**: Both

* **Description of Pull Request**: 

Enables Github Actions to validate NPC and database changes.
This Action will only run if changes have been made to *either* npc/ or db/

skip-ci

PR #6530, opened 2022-01-18
Add more Github Actions

Lemongrass3110:

* **Server Mode**: Both

Renames "build_servers.yml" to "build_servers_gcc.yml"
Removed Pre-Renewal and Renewal from GCC compilation
Added an action for Renewal and one Pre-Renewal
Added an action for VIP
Added an action for different packet versions
Adds a MSVS build


Akkarinage:
There is a whole lot of duplication in there now.

We don't need separate renewal and pre-renewal Action files if you're already running build test against both of those for every packetver listed in the packet version file.

.github/workflows/build_servers_msbuild.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1.1
+
+      - name: Build solution in Debug
+        if: ${{ matrix.mode == 'PRE' }}
+        run: msbuild rAthena.sln -t:rebuild -property:Configuration=Debug /p:DefineConstants="BUILDBOT%3BPRERE"
+
+      - name: Build solution in Debug
+        if: ${{ matrix.mode == 'RE' }}
+        run: msbuild rAthena.sln -t:rebuild -property:Configuration=Debug /p:DefineConstants="BUILDBOT"
```
aleos89, line 40: Just missing a new line for the end of the file.

.github/workflows/build_servers_packetversions.yml
```diff
@@ -0,0 +1,67 @@
+name: Build servers in Renewal
```
aleos89, line 1:
```suggestion
name: Build servers with different packet versions
```

.github/workflows/build_servers_prerenewal.yml
```diff
@@ -0,0 +1,70 @@
+name: Build servers in Pre-Renewal
```
aleos89, line 1: Guess this one isn't needed so much since #6520 will handle the running of the servers? Plus `build_servers_gcc` and `build_servers_msbuild` will handle the compiling process. Unless this will be overtaken with SQL.

.github/workflows/build_servers_renewal.yml
```diff
@@ -0,0 +1,70 @@
+name: Build servers in Renewal
```
aleos89, line 1: Same comment as `build_servers_prerenewal`.

.github/workflows/build_servers_vip.yml
```diff
@@ -1,10 +1,13 @@
-name: Build Servers from a Pull Request
-# build_servers.yml
+name: Build servers in VIP mode
```
aleos89, line 1: If we want to build VIP, just include it in `build_servers_gcc`/`build_servers_msbuild`?

.github/workflows/npc_db_validation.yml
```diff
@@ -6,6 +6,9 @@ name: Validate NPC Scripts and DB Changes
 
 on:
   push:
+    branches:
+      - master
+  pull:
     paths:
       # Always trigger all Github Actions if an action or something CI related was changed
       - '.github/workflows/**'
```
Akkarinage, line 14:
```suggestion
on: [push, pull_request]
  branches:
    - master
  paths:
      # Always trigger all Github Actions if an action or something CI related was changed
    - '.github/workflows/**'
```

Akkarinage: `pull:` isn't a valid trigger - silly inline suggestions wouldn't let me change the indent of the other path lines o.O

PR #6553, opened 2022-01-25
getitemdata/setitemdata

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

This is a suggestion of item data manipulation.
Idea: `getitemdata` copies the characteristics of an existing item in an array. The user changes the data in the array as he wants. He can now create the new item using `setitemdata`.

Why?
--------------------------------------------------------------------------------------------------
Because of the enchant grade `getitem4` will (probably) come out. These commands will give less trouble to the user (hopefully) than writing 14 parameters.


src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+		break;
+	case SID_CARD2:
+		itm->card[1] = script_getnum(st, 4);
+		break;
+	case SID_CARD3:
+		itm->card[2] = script_getnum(st, 4);
+		break;
+	case SID_CARD4:
+		itm->card[3] = script_getnum(st, 4);
+		break;
```
Lemongrass3110, line 25869: If card 0 is not a special card these should have item db checks too.

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+	case SID_REFINE:
+		itm->refine = cap_value( script_getnum(st, 4), 0, MAX_REFINE );
+		break;
+	case SID_ATTRIBUTE:
+		itm->attribute = script_getnum(st, 4) > 0;
+		break;
+	case SID_CARD1:
+		itm->card[0] = script_getnum(st, 4);
+		break;
```
Lemongrass3110, line 25860: If the value to be set is not a special value it should have an item db check too.

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+			id = item_db.searchname( script_getstr(st, 4) );
+		else
+			id = item_db.find( script_getnum(st, 4) );
+
+		if (id == nullptr) {
+			ShowError("buildin_setitemdata: Nonexistant item '%s'.\n", script_getstr(st, 4));
+			st->state = END;
+			return SCRIPT_CMD_FAILURE;
+		}
+		if (id->type != IT_ARMOR && id->type != IT_WEAPON && id->type != IT_SHADOWGEAR) {
```
Lemongrass3110, line 25841: if !itemdb_isequip2?

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (sd->inventory_data[idx] == nullptr) {
+		ShowWarning("buildin_setitemdata: No item can be deleted from index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+	if (sd->inventory_data[idx]->type != IT_ARMOR && sd->inventory_data[idx]->type != IT_WEAPON && sd->inventory_data[idx]->type != IT_SHADOWGEAR) {
```
Lemongrass3110, line 25819: if !itemdb_isequip2?

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+
+	int idx = script_getnum(st, 2);
+	if (idx < 0 || idx >= MAX_INVENTORY) {
+		ShowWarning("buildin_setitemdata: Index %d is out of the range 0-%d.\n", idx, MAX_INVENTORY - 1);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (sd->inventory_data[idx] == nullptr) {
+		ShowWarning("buildin_setitemdata: No item can be deleted from index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
```
Lemongrass3110, line 25815:
```suggestion
		ShowWarning("buildin_setitemdata: No item can be updated at index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
```

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (sd->inventory_data[idx] == nullptr) {
+		ShowWarning("buildin_getitemdata: No item can be deleted from index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+	if (sd->inventory_data[idx]->type != IT_ARMOR && sd->inventory_data[idx]->type != IT_WEAPON && sd->inventory_data[idx]->type != IT_SHADOWGEAR) {
```
Lemongrass3110, line 25754: if !itemdb_isequip2?

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+
+	int idx = script_getnum(st, 2);
+	if (idx < 0 || idx >= MAX_INVENTORY) {
+		ShowWarning("buildin_getitemdata: Index %d is out of the range 0-%d.\n", idx, MAX_INVENTORY - 1);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	if (sd->inventory_data[idx] == nullptr) {
+		ShowWarning("buildin_getitemdata: No item can be deleted from index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
```
Lemongrass3110, line 25750:
```suggestion
		ShowWarning("buildin_getitemdata: No item can be found at index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
```

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+	struct map_session_data* sd;
+
+	if (!script_rid2sd(sd)) {
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	struct script_data *data = script_getdata(st, 3);
+
+	if (!data_isreference(data)) {
```
Lemongrass3110, line 25736: Does the script engine not do that already?
You defined it as "r".

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+			return SCRIPT_CMD_FAILURE;
+		}
+		itm->nameid = id->nameid;
+		break;
+	}
+	case SID_IDENTIFY:
+		itm->identify = script_getnum(st, 4) > 0;
+		break;
+	case SID_REFINE:
+		itm->refine = cap_value( script_getnum(st, 4), 0, MAX_REFINE );
```
Lemongrass3110, line 25853: Maybe report to the user and not cap silently?

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+		itm->card[1] = script_getnum(st, 4);
+		break;
+	case SID_CARD3:
+		itm->card[2] = script_getnum(st, 4);
+		break;
+	case SID_CARD4:
+		itm->card[3] = script_getnum(st, 4);
+		break;
+	case SID_EXPIRE:
+		itm->expire_time = (unsigned int)(time(NULL) + script_getnum(st, 4));
```
Lemongrass3110, line 25871: This should be commented in script doc or use absolute value here.
If absolute value then there should be a check if the time is in the past.

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+		break;
+	case SID_BOUND: {
+		int bound = script_getnum(st, 4);
+		if (bound < BOUND_NONE || bound >= BOUND_MAX)
+			itm->bound = BOUND_NONE;
```
Lemongrass3110, line 25876: Report to console. Removing bound could be seen critical?!

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+		else
+			itm->bound = static_cast<bound_type>( bound );
+		break;
+	}
+	case SID_GRADE:
+		itm->enchantgrade = cap_value( script_getnum(st, 4), 0, 3 );
```
Lemongrass3110, line 25882: Please report to the user again.
Currently 4 grades exist.
I will add a define for it later on.
Or you can if you want?
It should be in mmo.hpp where MAX_ARMOR_LEVEL is defined.

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+		break;
+	case SID_OPTIONID3:
+		itm->option[2].id = script_getnum(st, 4);
+		break;
+	case SID_OPTIONID4:
+		itm->option[3].id = script_getnum(st, 4);
+		break;
+	case SID_OPTIONID5:
+		itm->option[4].id = script_getnum(st, 4);
+		break;
```
Lemongrass3110, line 25898: Should be checked if the option exists in the database.

src/map/script.cpp
```diff
@@ -25686,6 +25722,221 @@ BUILDIN_FUNC( openstylist ){
...
+	case SID_OPTIONPARAM4:
+		itm->option[3].param = script_getnum(st, 4);
+		break;
+	case SID_OPTIONPARAM5:
+		itm->option[4].param = script_getnum(st, 4);
+		break;
+	}
+
+	if (itm->equip > 0)
```
Lemongrass3110, line 25931: Maybe store equip flag here and equip the item again afterwards?

doc/script_commands.txt
```diff
@@ -5558,6 +5566,60 @@ These commands toggle the ability to use atcommand while interacting with an NPC
 
 The default setting, 'atcommand_disable_npc', is defined in 'conf/battle/gm.conf'.
 
+---------------------------------------
+
+*getitemdata(<inventory index>,<arrayname>);
+*setitemdata(<arrayname>);
```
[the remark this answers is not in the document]
Atemo: This has been updated in the last rebase
`setitemdata(<inventory index>,<parameter>,<new value>; `

src/map/script.cpp
Atemo: I left it like that because it's more versatile. User may (or not) want to auto-reequip the item. `equip` script command can do the job
```
*equip <item id>{,<char_id>};
*autoequip <item id>,<option>;

These commands are to equip a equipment on the attached character.
The equip function will equip the item ID given when the player has
this item in his/her inventory, while the autoequip function will
equip the given item ID when this is looted. The option parameter of
the autoequip is 1 or 0, 1 to turn it on, and 0 to turn it off.
```

src/map/script.cpp
Atemo: Maybe I could just remove that check for `getitemdata`

src/map/script.cpp
```diff
@@ -25686,6 +25722,309 @@ BUILDIN_FUNC( openstylist ){
...
+
+		if (!random_option_db.exists(id)) {
+			ShowError("buildin_setitemdata: SID_OPTIONID: Random option ID %d does not exists.\n", id);
+			st->state = END;
+			return SCRIPT_CMD_FAILURE;
+		}
+		itm->option[index].id = id;
+		break;
+	}
+	case SID_OPTIONVALUE1:
```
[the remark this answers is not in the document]
Atemo: If you ask for it, both are fine

aleos89 (commented):
Small review.

laziem:
How we can reset the value? The script doesn't accept `setitemdata( .@index, SID_CARD4, 0 );`

src/map/script.cpp
```diff
@@ -25686,6 +25722,281 @@ BUILDIN_FUNC( openstylist ){
...
+
+	return SCRIPT_CMD_SUCCESS;
+}
+
+BUILDIN_FUNC(setitemdata)
+{
+	map_session_data* sd;
+
+	if (!script_rid2sd(sd)) {
+		st->state = END;
```
Lemongrass3110, line 25785:
```suggestion
```
This is already done inside script_rid2sd

src/map/script.cpp
```diff
@@ -25686,6 +25722,281 @@ BUILDIN_FUNC( openstylist ){
 #endif
 }
 
+BUILDIN_FUNC(getitemdata)
+{
+	struct map_session_data* sd;
+
+	if (!script_rid2sd(sd)) {
+		st->state = END;
```
Lemongrass3110, line 25730:
```suggestion
```
This is already done inside script_rid2sd

src/map/script.cpp
```diff
@@ -25686,6 +25722,281 @@ BUILDIN_FUNC( openstylist ){
...
+				return SCRIPT_CMD_FAILURE;
+			}
+		}
+
+		itm->card[index] = card;
+		break;
+	}
+	case SID_EXPIRE: {
+		int t = script_getnum(st, 4);
+		if (t <= 0) {
```
Lemongrass3110, line 25880: Should we support removing "rental" status here?

src/map/script.cpp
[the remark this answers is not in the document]
Lemongrass3110: Yes I do agree. Both are fine. I will leave it to your choice what you like better.

src/map/script.cpp
```diff
@@ -25686,6 +25722,281 @@ BUILDIN_FUNC( openstylist ){
...
+	case SID_OPTIONID1:
+	case SID_OPTIONID2:
+	case SID_OPTIONID3:
+	case SID_OPTIONID4:
+	case SID_OPTIONID5: {
+		int index = type - SID_OPTIONID1;
+
+		t_itemid id = script_getnum(st, 4);
+
+		if (!random_option_db.exists(id)) {
```
Lemongrass3110, line 25917: Maybe here we should also add a "special case" for id = 0 (which would also need to remove optionvalue and optionparam of the same index)?

Lemongrass3110 (approved):

src/map/script.cpp
```diff
@@ -25686,6 +25722,281 @@ BUILDIN_FUNC( openstylist ){
...
+		ShowError("buildin_getitemdata: No item found on index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+	if (!itemdb_isequip2(sd->inventory_data[idx]) || sd->inventory_data[idx]->type == IT_AMMO) {
+		ShowError("buildin_getitemdata: Invalid type for item %u. The item must be equippable.\n", sd->inventory_data[idx]->nameid);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
```
Lemongrass3110, line 25751: Just a thought for the "set" function:
Should we block calls if the target item is equipped or in equip switch?
And/or should we run status calculations after changing data?

src/map/script.cpp
```diff
@@ -25686,6 +25722,281 @@ BUILDIN_FUNC( openstylist ){
...
+		ShowError("buildin_setitemdata: No item found on index %d of player %s (AID: %u, CID: %u).\n", idx, sd->status.name, sd->status.account_id, sd->status.char_id);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+	if (!itemdb_isequip2(sd->inventory_data[idx]) || sd->inventory_data[idx]->type == IT_AMMO) {
+		ShowError("buildin_setitemdata: Invalid type for item %u. The item must be equippable.\n", sd->inventory_data[idx]->nameid);
+		st->state = END;
+		return SCRIPT_CMD_FAILURE;
+	}
+
```
Lemongrass3110, line 25806: Just a thought for the "set" function:
Should we block calls if the target item is equipped or in equip switch?
(If we allow during equip: should we run status calculations or whatever after changing data?)

Atemo:
I wonder if the format
`setitemdata(<inventory index>,<parameter>,<new value>,<parameter>,<new value>...; `
could be better, to avoid spamming the command (and avoid the spam of the message to the player each times) when the user want to change several data (for example changing the item random option requires several call of the command)

Lemongrass3110:
> I wonder if the format `setitemdata(<inventory index>,<parameter>,<new value>,<parameter>,<new value>...; ` could be better, to avoid spamming the command (and avoid the spam of the message to the player each times) when the user want to change several data (for example changing the item random option requires several call of the command)..?

Maybe support both?
In getitemdata you get an array right now.
So you could use the same (modified) array to update the data.
All you need to do is check if the <parameter> is an array in source.
Or maybe simply if parameter count is not <parameter>, <new value>.
Know what I mean?

kikebarriga (commented):
test

laziem:
Please update to current rAthena

laziem:
Please update to current rAthena

PR #6560, opened 2022-01-26
Resolves cardfix calculation going below zero

aleos89:
* **Addressed Issue(s)**: N/A

  * Fixes an issue where cardfix bonuses cause the final result to go below zero ultimately giving unwanted effects.
Thanks to @Tokeiburu!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -675,7 +675,7 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 	s_defele = (tsd) ? (enum e_element)status_get_element(src) : ELE_NONE;
 
 //Official servers apply the cardfix value on a base of 1000 and round down the reduction/increase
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -675,7 +675,7 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 	s_defele = (tsd) ? (enum e_element)status_get_element(src) : ELE_NONE;
 
 //Official servers apply the cardfix value on a base of 1000 and round down the reduction/increase
-#define APPLY_CARDFIX(damage, fix) { (damage) = (damage) - (int64)(((damage) * (1000 - (fix))) / 1000); }
+#define APPLY_CARDFIX(damage, fix) { (damage) = (damage) - (int64)(((damage) * (1000 - max(0, fix))) / 1000); }
```
alycecil, line 678: `min(max(0,fix),1000)` since were here?

Lemongrass3110: Sounds like a good suggestion.

Lemongrass3110 (approved):

PR #6562, opened 2022-01-26
Updates Golden Thief Bug card behavior

### LINKED ISSUE #5918: GTB card should not block status changes from self skills on self even if they are magic
* **rAthena Hash**: 60b1b51

* **Client Date**: -

* **Server Mode**: Pre-Renewal (but probably applies to both)

* **Description of Issue**: 

I saw reports that skills like Ruwach and Sight are blocked by GTB since our last change but they shouldn't. Originally I thought we might have incorrectly defined these as "magic", but when testing Magnificat and Gloria on Aegis I noticed that these skills apply status change on self, but are still blocked by GTB on all other targets (so if all in the party wear GTB only caster gets the buff).

Which made me think that maybe for self skills there's a special rule that makes GTB not block status changes. So we actually went and tested all the self-skills that are defined as "magic" in rAthena and all of them applied their status changes to the caster.

Target skills should get blocked even when cast on self.

Self skills should get blocked on all targets except self.

Here's a list of skills that work on self even with GTB on (giving the status change to self but not any other targets with GTB):
Ruwach
Sight
Angelus
Magnificat
Gloria
Energy Coat
Cast Cancel
Magic Rod
Hindsight
Hocus Pocus
Basilica
Mystical Amplification
Indulge
Foresight
Star Gladiator Feel / Comfort Skills
Double Casting
Mirror Image

Note that Grand Cross still doesn't work on self. The whole skill actually fails to cast if you wear GTB.

aleos89:
* **Addressed Issue(s)**: Fixes #5918

  * Target magic skills should get blocked even when cast on self.
  * Self magic skills should get blocked on all targets except self.
  * Adds an IgnoreGtb skill flag to explicitly allow a skill to bypass these checks.
Thanks to @Playtester!

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -34211,6 +34211,8 @@ Body:
     MaxLevel: 5
     Type: Magic
     TargetType: Attack
+    Flags:
+      IgnoreGtb: true
     Range: 9
     Hit: Single
     HitCount: 1
--- a/doc/skill_db.txt
+++ b/doc/skill_db.txt
@@ -3,7 +3,7 @@
 //===== By: ==================================================
 //= rAthena Dev Team
 //===== Last Updated: ========================================
-//= 20200324
+//= 20220126
 //===== Description: =========================================
 //= Explanation of the skill_db.yml file and structure.
 //============================================================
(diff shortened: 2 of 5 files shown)


Daraen1:
`ServerMode: Pre-renewal and Renewal` 

missing modification in `db/pre-re/skill_db.yml`

is it an oversight?

aleos89:
No, the only skill which has that flag in the skill_db is a 4th class skill. The pre-renewal part is in the source, same as renewal.

Daraen1:
Okay thanks @aleos89

Issue #6566, opened 2022-01-27
PCBLOCK_NPC should be used by default


Balferian:
* **rAthena Hash**: 0fc05ce
* **Client Date**: 
* **Server Mode**: Pre/Re
* **Description of Issue**: 
Now players can use items, skills and change equip when they use npc if pcblock not used.
  * Result: 
  * Expected Result: Players should not use items, skills and change equip when they use npc.
  * How to Reproduce: Talk with any npc and try to use Hide / potions / change equip.
  * Official Information: I tested a few NPCs, I can do more tests if necessary.
  Check video from kRo.
  

https://user-images.githubusercontent.com/16213511/151297938-a70357bc-651e-40ff-862c-746687211b63.mp4

https://user-images.githubusercontent.com/16213511/151297955-73a523f4-75b5-472f-980b-5bd561c9e55f.mp4

  
* **Modifications that may affect results**:

Atemo:
Do you know what message ID the client uses when the player tries to change equipment?

PR #6574, opened 2022-01-29
Changes the default value of the conf item_enabled_npc to no

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/6566

* **Server Mode**: both

Changes the default value of the conf `item_enabled_npc` to `no` toprevent players to swap equipment/use consumables while interacting with npc.
Note: Was the previous setting from rAthena legacy or PRE-renewal behaviour? Should we add a setting for PRE-renewal?

### DIFF
--- a/conf/battle/items.conf
+++ b/conf/battle/items.conf
@@ -90,8 +90,8 @@ allow_consume_restricted_item: no
 allow_equip_restricted_item: yes
 
 // Allow changing of equipment while interacting with NPCs? (Note 1)
-// Default: yes
-item_enabled_npc: yes
+// Default: no
+item_enabled_npc: no
 
 // Allow map_flooritem to check if item is droppable? (Note 1)
 // If yes, undroppable items will be destroyed instead of appearing on the map when a player's inventory is full.


mazvi:
> Note: Was the previous setting from rAthena legacy or PRE-renewal behaviour? Should we add a setting for PRE-renewal?

I think no need. maybe only note. So pre renewal who need this setting to yes, can use import folder, or can change it self

Atemo:
Thanks you. Does that means you know the behaviour in PRE-renewal?

mazvi:
Sorry idk. :)

Lemongrass3110:
Why not go the way that @aleos89 invented?
Comment it in the .conf file and use the RENEWAL specific setting in src?

Atemo:
Do you have a line as example?
How about splitting re/pre conf like db ?

conf/battle/items.conf
```diff
@@ -89,12 +89,9 @@ allow_consume_restricted_item: no
...
 
-// Allow changing of equipment while interacting with NPCs? (Note 3)
-// 0: Disabled. (Default on renewal)
-// 1: Enabled on pre-renewal. (Default on pre-renewal)
-// 2: Enabled on renewal.
-// 3: 1+2
-item_enabled_npc: 1
+// Allow changing of equipment while interacting with NPCs? (Note 1)
+// Default on official servers: yes for Pre-renewal, no for Renewal
+//item_enabled_npc: yes
```
mazvi, line 94: why? 

//item_enabled_npc: yes

shouldn't be, 

item_enabled_npc: yes

without // ?

Atemo: the conf is using the default value when it doesn't exist in conf folder

Lemongrass3110 (approved):

aleos89 (approved):

PR #6577, opened 2022-02-01
Changes the default value of itemDB's "Override" to 100

Everade:
* **Addressed Issue(s)**:  -

* **Server Mode**: 
Pre-Renewal
Renewal

Changed the default of NoUse's and Trade's **Override** setting from **0** to **100**.
Removed all the Override definitions that are with that no longer required.

### DIFF
--- a/db/import-tmpl/item_db.yml
+++ b/db/import-tmpl/item_db.yml
@@ -65,10 +65,10 @@
 #     Storage               If the stack is applied to the player's storage. (Default: false)
 #     GuildStorage          If the stack is applied to the player's guild storage. (Default: false)
 #   NoUse:                  Conditions when the item is unusable. (Default: null)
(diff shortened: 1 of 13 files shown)


aleos89 (approved):
Looks good! There's just one more location that needs the header update: `doc/yaml/db/item_db.yml`

mazvi:
That's right because GM doesn't need to trade either :D

Everade:
> That's right because GM doesn't need to trade either :D
This isn't any different to what we have right now. The items that did not have any Override settings till now, **have no trading restrictions set**, and **items with restrictions already had it set to 100**. That's why this PR removes 6,335 lines.

Although you might want to take a closer look at your custom items in case you **haven't set an override while using trading or use restrictions**.

PR #6586, opened 2022-02-03
Added support for Cards SubType (yaml2sql)

Balferian:
* **Addressed Issue(s)**:  -
* **Server Mode**: -
  Added Cards SubType support for yaml2sql converter.


Balferian:
> Card SubType (normal/enchant) has been added in item_db yaml. The SubType didn't exist in txt version (at least not supported) so nothing to convert. Or I'm missing something?

Oh, i make mistake in the title.
SubType don`t work when you convert yaml db to sql.

Issue #6580, opened 2022-02-02
Mob walk randomly on spawn


Atemo:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/65af393097582669c5e27a5196bf165b4b0720c1

* **Client Date**: -

* **Server Mode**: Both

* **Description of Issue**: 
  * __Result__: On rAthena, monsters, after being spawn, walk randomly on the map without any conditions.
 
  * __Expected Result__: On the official server, after spawning, monsters are inactive on the map until the player finds them. The difference is noticeable in memorial dungeons where monsters spawn at given coordinates.

  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

PR #6589, opened 2022-02-03
Non spotted mobs shouldn't walk

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/6580

* **Server Mode**: Both

On rAthena, monsters, after being spawn, walk randomly on the map without any conditions.
On the official server, after spawning, monsters are inactive on the map until the player finds them. The difference is noticeable in memorial dungeons where monsters spawn at given coordinates.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -9988,6 +9988,7 @@ static const struct _battle_data {
 	{ "save_body_style",                    &battle_config.save_body_style,                 1,      0,      1,              },
 	{ "monster_eye_range_bonus",            &battle_config.mob_eye_range_bonus,             0,      0,      10,             },
 	{ "monster_stuck_warning",              &battle_config.mob_stuck_warning,               0,      0,      1,              },
+	{ "mob_walk_onspawn",             		 &battle_config.mob_walk_onspawn,               0,      0,      1,              },
 	{ "skill_eightpath_algorithm",          &battle_config.skill_eightpath_algorithm,       1,      0,      1,              },
 	{ "skill_eightpath_same_cell",          &battle_config.skill_eightpath_same_cell,       1,      0,      1,              },
 	{ "death_penalty_maxlv",                &battle_config.death_penalty_maxlv,             0,      0,      3,              },
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1157,7 +1157,7 @@ int mob_spawn (struct mob_data *md)
 	if (battle_config.mob_walk_onspawn)
 		md->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME;
 	else
-		md->next_walktime = 0;
+		md->next_walktime = -1;
 	md->last_linktime = tick;
 	md->dmgtick = tick - 5000;
 	md->last_pcneartime = 0;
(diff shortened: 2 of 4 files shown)


aleos89 (approved):
Looks good! My only comment which doesn’t have to do with this PR is that we don’t really seem to justify -1 or 0 as “invalid” like we normally use for players (-1). Also saw that we missed next_walktime for NPCs to be converted to t_tick. These can be covered in another cleanup!

src/map/mob.cpp
```diff
@@ -1154,7 +1154,10 @@ int mob_spawn (struct mob_data *md)
 
 	md->state.aggressive = status_has_mode(&md->status,MD_ANGRY)?1:0;
 	md->state.skillstate = MSS_IDLE;
-	md->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME;
+	if (battle_config.mob_walk_onspawn)
+		md->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME;
+	else
+		md->next_walktime = 0;
```
aleos89, line 1160:
```suggestion
		md->next_walktime = -1;
```
Actually you might want to make this to -1, because there is one diff check that looks for <0.

aleos89 (approved):

Atemo:
From my understanding `next_walktime` contains a timer, the system checks the difference between now and that timer (`md->next_walktime` is never used alone). For me that means all value below "now" when the mob spawn prevents the mob from walking.
When spawning the mob is in lazy mode (no player around), the time diff is < 0 and nothing happens because `mob_is_spotted` returns false :
https://github.com/rathena/rathena/blob/3da8426fe743ab4bab644453bf62585d2986a48e/src/map/mob.cpp#L2062

src/map/battle.cpp
```diff
@@ -9988,6 +9988,7 @@ static const struct _battle_data {
 	{ "save_body_style",                    &battle_config.save_body_style,                 1,      0,      1,              },
 	{ "monster_eye_range_bonus",            &battle_config.mob_eye_range_bonus,             0,      0,      10,             },
 	{ "monster_stuck_warning",              &battle_config.mob_stuck_warning,               0,      0,      1,              },
+	{ "mob_walk_onspawn",             		 &battle_config.mob_walk_onspawn,               0,      0,      1,              },
```
Lemongrass3110, line 9991: Whitespaces look wrong, but might be github again.

src/map/mob.cpp
```diff
@@ -1154,7 +1154,10 @@ int mob_spawn (struct mob_data *md)
 
 	md->state.aggressive = status_has_mode(&md->status,MD_ANGRY)?1:0;
 	md->state.skillstate = MSS_IDLE;
-	md->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME;
+	if (battle_config.mob_walk_onspawn)
+		md->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME;
+	else
+		md->next_walktime = -1;
```
Lemongrass3110, line 1160: From how I understand it anything < current tick will trigger walking on the next iteration.
So you might want to set it to INT64_MAX instead, since you want to prevent that it triggers walking?
Maybe I have read something wrong though I just looked over it quickly.

PR #6599, opened 2022-02-07
Implements Synthesis and Upgrade UI items

Everade:
* **Addressed Issue(s)**:

* **Server Mode**:
Renewal

Implements Synthesis and Upgrade UI items for an upcoming release.

### DIFF
--- a/db/re/item_db_usable.yml
+++ b/db/re/item_db_usable.yml
@@ -49602,6 +49602,22 @@ Body:
     Weight: 10
     Script: |
       /* upgradeui 100745; */
+  - Id: 100783
+    AegisName: MajorAutoSpell_Mix
+    Name: Auto Spell Shadow Pick-Up Box    # !todo check english name
+    Type: DelayConsume
+    Weight: 10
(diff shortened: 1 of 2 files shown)


db/re/item_db_usable.yml
```diff
@@ -49602,6 +49602,22 @@ Body:
     Weight: 10
     Script: |
       /* upgradeui 100745; */
+  - Id: 100783
+    AegisName: MajorAutoSpell_Mix
+    Name: Auto Spell Shadow Pick-Up Box    # !todo check english name
```
Atemo, line 49607: How about using the name from https://github.com/llchrisll/ROenglishRE/tree/master/Renewal/System

Everade: It took a while to translate them, don't feel like throwing that work into the bin in exchange for mass google translations.

Atemo: imo you can remove the `# !todo check english name` if you're confident in the name

Issue #6596, opened 2022-02-06
Error HP on Expanded Job after Base Level 175+


mazvi:
* **rAthena Hash**: e7f58b1

* **Client Date**: 20211103

* **Server Mode**: RENEWAL

![gambar](https://user-images.githubusercontent.com/8316103/152697288-7af7c7a0-f276-4d03-ade6-f171f87061b1.png)

![gambar](https://user-images.githubusercontent.com/8316103/152697473-570460f3-1a9c-4994-aa6b-162d6d87f833.png)

![gambar](https://user-images.githubusercontent.com/8316103/152697530-f69e2aaa-078b-45c7-a2ad-8ef73cc59af4.png)

* **Description of Issue**: 
  * Result:  HP ERROR After Base Level 175+
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

ryukazuna:
Missing HP Tables. Summoner,Soul Reaper and Star Emperor still have unknown base points for 176 and higher.

mazvi:
> Missing HP Tables. Summoner,Soul Reaper and Star Emperor still have unknown base points for 176 and higher.

So why are there levels above 175 enabled for this job? At least there should be a dummy for this.

Pokye:
> Missing HP Tables. Summoner,Soul Reaper and Star Emperor still have unknown base points for 176 and higher.

I found this, but I don't know if it's official from kRO, but for now I think rA should follow.
https://github.com/PandasWS/Pandas/commit/9ec9a9aa6eb033efb6a9fa416acf0407c5ae9b61

mazvi:
> > Missing HP Tables. Summoner,Soul Reaper and Star Emperor still have unknown base points for 176 and higher.
> 
> I found this, but I don't know if it's official from kRO, but for now I think rA should follow. [PandasWS/Pandas@9ec9a9a](https://github.com/PandasWS/Pandas/commit/9ec9a9aa6eb033efb6a9fa416acf0407c5ae9b61)

so true. because the level can rise above 175. and has exp/stat above 175 there should be a temporary dummy level until it is known that the actual base hp sp from kro.

ryomahostingph:
https://raw.githubusercontent.com/PandasWS/Pandas/9ec9a9aa6eb033efb6a9fa416acf0407c5ae9b61/db/re/job_basepoints.yml
Tested this and i think its good now 
[...]

PR #6600, opened 2022-02-07
Dummy Summoner Star Emperor Soul Reaper Level 175+

mazvi:
Dummy data for Summoner, Star Emperor, Soul Reaper Level 176-200
Will be removed when get actual data from kRO 

Get from PandasWS/Pandas@9ec9a9a

* **Addressed Issue(s)**: #6596

* **Server Mode**: 

### DIFF
--- a/db/re/job_basepoints.yml
+++ b/db/re/job_basepoints.yml
@@ -9893,6 +9893,67 @@ Body:
       - Level: 175
         Hp: 23220
       # Levels 176-200 are unknown
+      # -----------------------------------------------------------------
+      # Provided by SSboyz
+      # data source: https://landgris.github.io/ROCalculator/
+      #
+      # Levels 176-200 HP Data Begin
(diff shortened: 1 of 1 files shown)


Pokye:
Why created again
```
   - Jobs:
       Summoner: true
       Baby_Summoner: true
     BaseHp:
```
Wouldn't it have been better to have put it below the "# Levels 176-200 are unknown " of the existing one?

mazvi:
I put it at the bottom so it's easy to remove later.

Lemongrass3110:
I would be in favor of moving it below the current levels, just like @Pokye suggested and as originally implemented by @CairoLee on PandaWS.

Maybe we can get the input from @aleos89 and @Atemo on this. But basically it looks good.

mazvi:
Edited please check @Lemongrass3110 :)

PR #6615, opened 2022-02-15
Fix Satan Morocc on Warper

mazvi:
Fix Satan Morocc on Warper

* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/npc/custom/warper.txt
+++ b/npc/custom/warper.txt
@@ -14,6 +14,7 @@
 //= 1.4b Added Wolfchev's Laboratory warp.
 //= 1.5 Added Lasagna ,Para Market ,WOE TE ,Instances and settings [sader1992].
 //= 1.5a Fix Bifrost Tower
+//= 1.5b Fix Satan Morocc [Mazvi]
 //============================================================
 
 -	script	Warper	-1,{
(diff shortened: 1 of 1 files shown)


npc/custom/warper.txt
```diff
@@ -257,7 +258,7 @@ F23: setarray @c[0],208,227,190,206,240,206,190,143,307,252,239,213,185,188,193,
 F24: Restrict("Pre-RE",2,7,9,10,11,13);
 	 setarray @c[2],192,162,235,166,202,206,202,208,225,202,202,214,263,196,217,201,87,121,277,181,221,185,175,200,174,197;
 	 Disp("Rachel Field",1,13); Pick("ra_fild");
-F25: if(.Satan_Morocc){
+F25: if(.Satan_Morocc=1){
```
Daegaladh, line 261: That makes no sense, it will set .Satan_Morocc to 1/true. Even with == will do exactly the same as before.

Daegaladh (changes requested):

npc/custom/warper.txt
Lemongrass3110, line 261: This is an assignment and no compare now.
Therefore this does not fix something, rather break something.

Lemongrass3110: Oh... Sorry github did not refresh it seems.
Totally agree with @Daegaladh here.

mazvi:
I have tried it. if without =1 then the function doesn't work, try it.

mazvi:
Sorry I was wrong, my function didn't work because I was using callfunc. :)

PR #6617, opened 2022-02-15
Added mob_names_type battle config

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

Adds a config to choose which field to use by default between `Name` and `JapaneseName` in `mob_db.yml` when a name is not provided.


conf/battle/monster.conf
```diff
@@ -99,9 +99,14 @@ monster_active_enable: yes
 // Should the mob_db names override the mob names specified in the spawn files?
 // 0: No
 // 1: always use the mob_db Name column (english mob name)
-// 2: always use the mob_db JName column (original Kro mob name)
+// 2: always use the mob_db JapaneseName column (original Kro mob name)
 override_mob_names: 0
 
+// Which field should be used by the function by default?
```
Lemongrass3110, line 105: What is "the function"?

Lemongrass3110 (commented):
I could not care less which name we use, but if possible please provide a function that has this logic centralized and not add it to each call.

Pokye:
In this case, would this also work for those who use mobs in sql? would it read 1 - name_english 2 - name_japanese ?

Lemongrass3110:
> In this case, would this also work for those who use mobs in sql? would it read 1 - name_english 2 - name_japanese ?

SQL vs YAML is dealt with during load time. This is about run time.
Both SQL and YAML store the loaded data in the same fields of the structure. So it should work equally for both versions.

src/map/atcommand.cpp
```diff
@@ -6990,8 +6991,7 @@ ACMD_FUNC(mobsearch)
 		return -1;
 	}
 
-	strcpy(mob_name, mob->jname.c_str());	// --ja--
-//	strcpy(mob_name, mob->name.c_str());	// --en--
+	strcpy(mob_name, get_mob_names_type());
```
Lemongrass3110, line 6994: Here output will be wrong now.
See comment below.

src/map/atcommand.cpp
```diff
@@ -7799,9 +7799,12 @@ ACMD_FUNC(showmobs)
...
 	}
 
-	if(mob_id == strtol(mob_name, nullptr, 10) && !mob->jname.empty())
-		strcpy(mob_name, mob->jname.c_str());    // --ja--
-		//strcpy(mob_name, mob->name.c_str());    // --en--
+	if (mob_id == strtol(mob_name, nullptr, 10)) {
+		if (battle_config.mob_names_type == 0 && !mob->name.empty())
+			strcpy(mob_name, mob->name.c_str());    // --en--
+		else if (battle_config.mob_names_type == 1 && !mob->jname.empty())
+			strcpy(mob_name, mob->jname.c_str());    // --ja--
```
Lemongrass3110, line 7806: Here you should probably just obey the config.
If they did not set the respective name it is a user fault.
Maybe we could validate in loadingFinished of mob db based on the config?

Additionally this logic could be hidden inside a function of s_mob_db.
```cpp
std::string& s_mob_db::get_name(){
	if( battle_config.mob_names_type == 0 ){
		return this->name;
	}else{
		return this->jname;
	}
}
```

This function could be used anywhere, where you already have s_mob_db looked up. :)

PR #6618, opened 2022-02-15
Implements MF_NOPETCAPTURE

### LINKED ISSUE #6556: Mapflag NoCapture
**NoCapture**
https://github.com/PandasWS/Pandas/commit/639713916ea87fdb2de0ffc2585d798585621619

This mapflag would be very useful and important to be used mainly in instances to avoid the use of capturing monsters to not bug instances, events and etc.

aleos89:
* **Addressed Issue(s)**: Fixes #6556

  * Implements the nopetcapture mapflag which is used to stop pet-type monsters from being caught.
Thanks to @RagnaWay's suggestion!


src/map/pet.cpp
```diff
@@ -1248,6 +1253,11 @@ int pet_catch_process2(struct map_session_data* sd, int target_id)
 		return 1;
 	}
 
+	if (map_getmapflag(sd->bl.m, MF_NOCAPTURE)) {
+		clif_displaymessage(sd->fd, msg_txt(sd, 669)); // You can't catch any pet on this map.
```
Lemongrass3110, line 1257: Maybe remove the pet related variables in this case too?

src/map/map.hpp
```diff
@@ -647,6 +647,7 @@ enum e_mapflag : int16 {
 	MF_NORODEX,
 	MF_NORENEWALEXPPENALTY,
 	MF_NORENEWALDROPPENALTY,
+	MF_NOCAPTURE,
```
Lemongrass3110, line 650: Maybe call it NOPETCAPTURE?

Lemongrass3110 (approved):

PR #6621, opened 2022-02-16
Add getitempos script

attackjom:
item script that return equip position.

* **Addressed Issue(s)**: -

* **Server Mode**: Renewal

Add script to return equip position that alway uses in JRO item.

ex.  itemID 27345

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -25736,6 +25736,20 @@ BUILDIN_FUNC( openstylist ){
 #endif
 }
 
+BUILDIN_FUNC(getitempos) {
+	TBL_PC *sd;
+	if (script_rid2sd(sd)){
+		if( current_equip_item_index == -1 ){
+			script_pushint(st, 0);
+			return SCRIPT_CMD_FAILURE;
+		}
+
+		script_pushint(st, sd->inventory.u.items_inventory[current_equip_item_index].equip);
+	}else
+		script_pushint(st,0);
+	return SCRIPT_CMD_SUCCESS;
+}
+
 #include "../custom/script.inc"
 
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -25736,6 +25736,20 @@ BUILDIN_FUNC( openstylist ){
 #endif
 }
 
+BUILDIN_FUNC(getitempos) {
+	TBL_PC *sd;
```
Lemongrass3110, line 25740: Please dont use TBL_PC, but map_session_data.

src/map/script.cpp
```diff
@@ -25736,6 +25736,20 @@ BUILDIN_FUNC( openstylist ){
...
+	if (script_rid2sd(sd)){
```
Lemongrass3110, line 25741: Please invert the logic and return SCRIPT_CMD_FAILURE here.
If script_rid2sd fails it reports this to the script engine and actually cancels the script execution.
So nothing will be pushed to the script engine in this case.

src/map/script.cpp
```diff
@@ -25736,6 +25736,20 @@ BUILDIN_FUNC( openstylist ){
...
+		if( current_equip_item_index == -1 ){
```
Lemongrass3110, line 25742: Please add an error message here.
For example:
ShowError( "buildin_getitempos: Invalid usage detected. This command should only be used inside item scripts.\n" );

Lemongrass3110 (approved):
Basically good suggestion. Just some little improvements!

Lemongrass3110 (approved):

aleos89 (approved):

PR #252, opened 2015-01-26
* Various server config cleanups

cydh:
* Change some config variables that use `char` to `StringBuf`. So, don't need to change the char length anymore.
  * Renamed some inter table config variables postfix "_db" to "_table".
  * Added table checks for login-server (Account, IP Ban, & Login Log), map-server.
  * Regroup map-server table variables to `struct MapServer_Schema mapserv_schema_config`.
  * Regroup map-server global config to `struct Map_Config map_config`.
  * Split log file or table names in `log_athena.conf`. File path to `log_path: log/` and extension to `log_extension: .log`. So, if logging is enabled by using file or table, don't need to rename all log names.
  * Removed unnecessary configs in map_athena.conf: `help_txt`, `help2_txt`, and `charhelp_txt`.
  * Removed unused file 'conf/charhelp.txt'.

* Added new StringBuf function
  * `StringBuf* StringBuf_MallocInitial`
  * `StringBuf* StringBuf_FromStr`
  * `void StringBuf_InitialInit`
  * `int StringBuf_PrintfClear`

* Thanks to @aleos89 and @Mendonn at #91 for that StringBuf

Signed-off-by: Cydh Ramdh <house.bad@gmail.com>

PR #6622, opened 2022-02-16
[RFC] Replace StringBuf with std::string

merryhime:
This is a PR to seek if there is any interest in receiving PRs for simple changes that modernize the code base.

One such change is replacing of StringBuf with std::string, an example of which on a single file below.

Would be happy to submit such PRs if they are welcome.

### DIFF
--- a/src/char/char.cpp
+++ b/src/char/char.cpp
@@ -4,11 +4,12 @@
 #pragma warning(disable:4800)
 #include "char.hpp"
 
-#include <stdarg.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-#include <time.h>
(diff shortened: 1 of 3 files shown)


Lemongrass3110:
I guess it might be better to do this in two phases.
1) hide std::string behind typedef StringBuf and implement all related functions
2) replace all previously defined typedefs/macros.

As for the header cleanup, I would do this per header maybe, if possible. But i do agree that we should switch to the C++ variants where possible.

aleos89:
This was actually part of #252 but it became too large to maintain. Might be best to keep it small with the initial StringBuf -> std::string conversions and then adjust anything else down the line.

PR #6654, opened 2022-03-10
Added some more checks to Laphine UIs

Lemongrass3110:

* **Server Mode**: Both

Adds some additional checks and safeguards.
Fixes a parameter being ignored
Fixed logging type to be more clear why the delayed consumable was deleted from inventory

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -23190,6 +23190,16 @@ void clif_parse_laphine_synthesis( int fd, struct map_session_data* sd ){
 			clif_laphine_synthesis_result( sd, LAPHINE_SYNTHESIS_ITEM );
 			return;
 		}
+
+		if( item->equip != 0 ){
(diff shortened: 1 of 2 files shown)


src/map/clif.cpp
```diff
@@ -23190,6 +23190,16 @@ void clif_parse_laphine_synthesis( int fd, struct map_session_data* sd ){
 			clif_laphine_synthesis_result( sd, LAPHINE_SYNTHESIS_ITEM );
 			return;
 		}
+
+		if( item->equip != 0 ){
+			clif_laphine_synthesis_result( sd, LAPHINE_SYNTHESIS_ITEM );
+			return;
+		}
+
+		if( item->equipSwitch != 0 ){
```
aleos89, line 23199:
```suggestion
		if( item->equip != 0 || item->equipSwitch != 0 ){
```
Since it calls the same clif function, just merge the check?

aleos89 (approved):

PR #6661, opened 2022-03-11
Converts the SCB flags to bitset

aleos89:
* **Addressed Issue(s)**: N/A

  * Cleans up the status change battle flags to utilize bitset.
  * Fixes an issue with status_change_clear_buffs comparing checks against the wrong status flags.


src/map/status.hpp
```diff
@@ -2693,62 +2693,64 @@ enum e_scs_flag : uint8 {
...
-	// Special flags for updating all stat/sub-stat stuff on request.
-	SCB_BATTLE	= 0x07FFFFFFFFFE, // All except BASE and extra flags.
-	SCB_ALL		= 0x07FFFFFFFFFF, // All except extra flags.
-	SCB_MAX		= 0x0FFFFFFFFFFF,
+	SCB_MAX,
+};
+
+// Special flags for updating all stat/sub-stat stuff on request.
+enum e_scb_flag_special : uint64 {
+	SCB_BATTLE = 0x7FFFFFFFFFE, // All except BASE and extra flags.
```
Lemongrass3110, line 2752: SCB_BATTLE is only used by setunitdata.
Better set up the calc flags properly there?
Or have a "global" (readonly) bitset to use and hand over as calc flag when this is needed?

src/map/status.hpp
```diff
@@ -2693,62 +2693,64 @@ enum e_scs_flag : uint8 {
...
+	SCB_ALL = 0x7FFFFFFFFFF, // All except extra flags.
```
Lemongrass3110, line 2753: SCB_ALL could be moved directly after SCB_MAX so that it does not affect the bitset size and given proper handling when being parsed during loadtime?

src/map/status.cpp
```diff
@@ -558,7 +558,7 @@ efst_type StatusDatabase::getIcon(sc_type type) {
  * @param type: SC type
  * @return cal_flag: Calc value 
  **/
-uint64 StatusDatabase::getCalcFlag(sc_type type) {
+std::bitset<SCB_MAX> StatusDatabase::getCalcFlag(sc_type type) {
 	std::shared_ptr<s_status_change_db> status = status_db.find(type);
 
 	return status ? status->calc_flag : SCB_NONE;
```
Lemongrass3110, line 564: Returning SCB_NONE here looks strange?
I think this should be an empty bitset now?

Lemongrass3110 (commented):
All in all looks good.
Maybe we should check where we could use references to prevent a lot of allocations?

src/map/status.cpp
```diff
@@ -5457,10 +5457,10 @@ void status_calc_bl_main(struct block_list *bl, uint64 flag)
 	}
 #endif
 
-	if(flag&(SCB_VIT|SCB_MAXHP|SCB_INT|SCB_MAXSP) && bl->type&BL_REGEN)
+	if(flag[SCB_VIT|SCB_MAXHP|SCB_INT|SCB_MAXSP] && bl->type&BL_REGEN)
```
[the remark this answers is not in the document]
aleos89: It didn't complain but I read that bitset supports all masking features, but for what it's worth I'll fix this up. ;)

Lemongrass3110 (approved):

PR #6662, opened 2022-03-11
open_quest_ui script command

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

Script command to force open the quest UI for the attached player.
Thanks to @Lemongrass3110 !

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -9714,7 +9714,7 @@ QMARK_PURPLE - Purple Marker
 
 *open_quest_ui {<quest ID>,{<char ID>}};
 
-Opens the quest UI for the attached player or the given character id.
+Opens the quest UI for the attached player or the given character ID.
 If the quest ID is not 0, the UI is opened at the given quest with the condtions that the quest must exist in quest_db.yml and in the character quest log.
 
 ---------------------------------------
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -9712,9 +9712,9 @@ QMARK_PURPLE - Purple Marker
 
 ---------------------------------------
 
-*open_quest_ui({<quest ID>});
+*open_quest_ui {<quest ID>,{<char ID>}};
 
-Opens the quest UI for the attached player.
+Opens the quest UI for the attached player or the given character id.
 The UI is opened at the given quest if the quest exists in quest_db.yml and in the character quest log.
 
 ---------------------------------------
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -9715,7 +9715,7 @@ QMARK_PURPLE - Purple Marker
 *open_quest_ui {<quest ID>,{<char ID>}};
 
 Opens the quest UI for the attached player or the given character id.
-The UI is opened at the given quest if the quest exists in quest_db.yml and in the character quest log.
+If the quest ID is not 0, the UI is opened at the given quest with the condtions that the quest must exist in quest_db.yml and in the character quest log.
 
 ---------------------------------------
 
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -9715,7 +9715,7 @@ QMARK_PURPLE - Purple Marker
 *open_quest_ui {<quest ID>,{<char ID>}};
 
 Opens the quest UI for the attached player or the given character ID.
-If the quest ID is not 0, the UI is opened at the given quest with the condtions that the quest must exist in quest_db.yml and in the character quest log.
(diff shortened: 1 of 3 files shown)


src/map/script.cpp
```diff
@@ -25812,6 +25812,17 @@ BUILDIN_FUNC(randomoptgroup)
 	return SCRIPT_CMD_SUCCESS;
 }
 
+BUILDIN_FUNC( open_quest_ui ){
+	struct map_session_data* sd;
+
+	if (!script_rid2sd(sd))
+		return SCRIPT_CMD_FAILURE;
+
```
Lemongrass3110, line 25820: Maybe add a check here if the quest exists?

doc/script_commands.txt
```diff
@@ -9712,6 +9712,13 @@ QMARK_PURPLE - Purple Marker
 
 ---------------------------------------
 
+*open_quest_ui({<quest ID>});
```
Lemongrass3110, line 9715: Since we almost do it everywhere - add optional char id?

src/map/script.cpp
```diff
@@ -26524,6 +26535,7 @@ struct script_function buildin_func[] = {
 	BUILDIN_DEF(laphine_synthesis, ""),
 	BUILDIN_DEF(laphine_upgrade, ""),
 	BUILDIN_DEF(randomoptgroup,"i"),
+	BUILDIN_DEF(open_quest_ui, "?"),
```
Lemongrass3110, line 26538: Here and in script documentation you made quest id optional, but in reality it is mandatory.
Maybe make it optional for real - if the client supports sending 0 as quest id.

Lemongrass3110 (approved):

src/map/script.cpp
Atemo: When the quest doesn't exist (example `open_quest_ui(50)`) the client display

Capture d’écran (651)
it doesn't seem necessary to check if the quest exists. Do you want to check anyway?

src/map/script.cpp
Atemo: The client just open the UI when the quest id is 0

src/map/script.cpp
aleos89: I guess if it's checked with the LUB the client won't crash, which shouldn't matter since this is only to visualize a quest to a player, not activate one.

I say we should be ok without including a quest_db check.

Lemongrass3110: Basically it is ok, but could be helpful to get a server side message.
You never know if players report it or not.

src/map/script.cpp
Lemongrass3110: Nice! :)

doc/script_commands.txt
```diff
@@ -9712,6 +9712,13 @@ QMARK_PURPLE - Purple Marker
 
 ---------------------------------------
 
+*open_quest_ui {<quest ID>,{<char ID>}};
+
+Opens the quest UI for the attached player or the given character ID.
+If the quest ID is not 0, the UI is opened at the given quest with the condtions that the quest must exist in quest_db.yml and in the character quest log.
```
aleos89, line 9718:
```suggestion
Use 0 as the quest ID to open the main quest UI. If the quest ID is not 0 then the quest UI is opened to the given quest. If the quest data is not populated in the client LUB then a message will be displayed saying the quest doesn't exist.
```

aleos89 (approved):

src/map/script.cpp
```diff
@@ -25812,6 +25812,30 @@ BUILDIN_FUNC(randomoptgroup)
...
+
+	if (!script_hasdata(st, 2))
+		quest_id = 0;
+	else {
+		int i;
+		quest_id = script_getnum(st, 2);
+
+		ARR_FIND(0, sd->avail_quests, i, sd->quest_log[i].quest_id == quest_id);
+		if (i == sd->avail_quests)
+			ShowWarning("buildin_open_quest_ui: Character %d doesn't have quest %d.\n", sd->status.char_id, quest_id);
```
Lemongrass3110, line 25831:
```suggestion
		quest_id = script_getnum(st, 2);

		if( quest_id != 0 ){
			int i;
			ARR_FIND(0, sd->avail_quests, i, sd->quest_log[i].quest_id == quest_id);
			if (i == sd->avail_quests)
				ShowWarning("buildin_open_quest_ui: Character %d doesn't have quest %d.\n", sd->status.char_id, quest_id);
		}
```

Lemongrass3110 (approved):

src/map/script.cpp
aleos89: If we do this, now we need to change the doc again ;)

PR #6711, opened 2022-03-17
Fixes Reading Spellbook behavior

### LINKED ISSUE #6705: Reading Spell Book and Comet not working properly
* **rAthena Hash**: 

* **Client Date**:  20210107

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result:  Reading Spell Book sometimes cannot be casted and or cannot use any book for that matter especially above 3 spells. 
  Additional: Comet For some reason doesn't use the Magic Poison Status effect even if it is indicated on Status DB and was instead using burn.
  * Expected Result:  You can do the things above
  * How to Reproduce:  Use reading spell book Issue started after the Status Change DB update
  * Official Information: 
  

aleos89:
* **Addressed Issue(s)**: Fixes #6705

* **Server Mode**: Renewal

  * Confirmed on kRO that Reading Spellbook persists through relog but not death.
  * Comet should now give the target Magic Poison instead of Burning.
Thanks to @Relliksuriv!

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -1785,7 +1785,7 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
 		sc_start(src,bl, SC_ADORAMUS, skill_lv * 4 + (sd ? sd->status.job_level : 50) / 2, skill_lv, skill_get_time2(skill_id, skill_lv));
 		break;
 	case WL_COMET:
-		sc_start(src, bl, SC_BURNING, 100, skill_lv, 20000);
+		sc_start(src, bl, SC_MAGIC_POISON, 100, skill_lv, 20000);
 		break;
 	case NPC_COMET:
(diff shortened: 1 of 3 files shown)


Lemongrass3110 (approved):

src/map/skill.cpp
```diff
@@ -21941,19 +21941,19 @@ int skill_magicdecoy(struct map_session_data *sd, t_itemid nameid) {
  * @param sd: Player data
  * @param nameid: Spellbook item used
  */
-void skill_spellbook(struct map_session_data *sd, t_itemid nameid) {
-	nullpo_retv(sd);
+void skill_spellbook(map_session_data &sd, t_itemid nameid) {
+	nullpo_retv(&sd);
```
Lemongrass3110, line 21945:
```suggestion
```
NO!

PR #6744, opened 2022-03-24
Implemented CountSlave in quest_db.yml

Atemo:
* **Addressed Issue(s)**: ?

* **Server Mode**: Both

Suggestion to replace the current slave/temporary monster check by a new setting in quest_db.yml
By default now all monsters increase the counter. When `CountSlave` is set to false temporary monsters are ignored.


aleos89:
I know we talked about this a bit but should we check for GM spawned mobs to avoid abuse? Just a thought.

Atemo:
Well I think we should only care if a normal player can abuse of the system. btw the setting only gives the user the possibility to decide what to do with temporary monsters. Maybe the temporary monsters should be ignored by default like before?

aleos89:
The only time it becomes troublesome is if someone uses NPC script commands to spawn monsters because they want to utilize get/setunitdata. The only notable workaround is to use the getmapunits script commands to grab all the mobs and do manual checks, but that is also not reliable if multiple mobs are spawned with the same ID.

src/map/quest.cpp
```diff
@@ -709,32 +719,32 @@ void quest_update_objective(struct map_session_data *sd, struct mob_data* md)
...
+						continue;
+					if (qi->objectives[j]->mapid >= 0) {
+						map_data *mapdata = map_getmapdata(sd->bl.m);
+
+						if (mapdata->instance_id < 1 && qi->objectives[j]->mapid != sd->bl.m)
+							continue;
+						if (mapdata->instance_id > 0 && mapdata->instance_src_map != qi->objectives[j]->mapid)
+							continue;
+					}
+					if (!qi->objectives[j]->count_slave && md->spawn == nullptr)
```
Lemongrass3110, line 744:
```yaml
#       CountSlave  (Only when the monster to kill is global) Whether the temporary monsters increase the counter. (Default: true)
```
In the documentation you say only applicable to global ones. I think then this should be "else if"

db/import-tmpl/quest_db.yml
```diff
@@ -46,6 +46,7 @@
 #                   On normal maps only kills of permanently spawned monsters increase the counter.
 #                   On instance maps all monster kills increase the counter.
 #       MapName     Displayed map name in quest UI. (Default: empty string)
+#       CountSlave  (Only when the monster to kill is global) Whether the temporary monsters increase the counter. (Default: true)
```
Lemongrass3110, line 49: For me the name "CountSlave" is a little misleading.

src/map/quest.cpp
```diff
@@ -709,32 +719,32 @@ void quest_update_objective(struct map_session_data *sd, struct mob_data* md)
...
-				if (qi->objectives[j]->element == ELE_ALL || qi->objectives[j]->element == md->status.def_ele)
-					objective_check++;
-				if (qi->objectives[j]->mapid < 0 || (qi->objectives[j]->mapid == sd->bl.m && md->spawn != nullptr))
-					objective_check++;
-				else if (qi->objectives[j]->mapid >= 0) {
-					struct map_data *mapdata = map_getmapdata(sd->bl.m);
-
-					if (mapdata->instance_id && mapdata->instance_src_map == qi->objectives[j]->mapid)
-						objective_check++;
+			if (sd->quest_log[i].count[j] < qi->objectives[j]->count)  {
```
Lemongrass3110, line 722: Please also invert this like you did below.

src/map/quest.cpp
```diff
@@ -709,32 +719,32 @@ void quest_update_objective(struct map_session_data *sd, struct mob_data* md)
...
+				if (qi->objectives[j]->mob_id != 0 && qi->objectives[j]->mob_id != md->mob_id)
+					continue;
+				if (qi->objectives[j]->mob_id == 0) {
```
Lemongrass3110, line 725:
```suggestion
				if (qi->objectives[j]->mob_id != 0) {
					if (qi->objectives[j]->mob_id != md->mob_id)
						continue;
				} else {
```

src/map/quest.cpp
```diff
@@ -709,32 +719,32 @@ void quest_update_objective(struct map_session_data *sd, struct mob_data* md)
...
+					if (qi->objectives[j]->min_level != 0 && qi->objectives[j]->min_level > md->level)
```
Lemongrass3110, line 726:
```suggestion
					if (qi->objectives[j]->min_level != 0 && md->level < qi->objectives[j]->min_level)
```

src/map/quest.cpp
```diff
@@ -709,32 +719,32 @@ void quest_update_objective(struct map_session_data *sd, struct mob_data* md)
...
+						continue;
+					if (qi->objectives[j]->max_level != 0 && qi->objectives[j]->max_level < md->level)
```
Lemongrass3110, line 728:
```suggestion
					if (qi->objectives[j]->max_level != 0 && md->level > qi->objectives[j]->max_level)
```

src/map/quest.cpp
```diff
@@ -709,32 +719,32 @@ void quest_update_objective(struct map_session_data *sd, struct mob_data* md)
...
+					if (qi->objectives[j]->size != SZ_ALL && qi->objectives[j]->size != md->status.size)
+						continue;
+					if (qi->objectives[j]->element != ELE_ALL && qi->objectives[j]->element != md->status.def_ele)
+						continue;
+					if (qi->objectives[j]->mapid >= 0) {
+						map_data *mapdata = map_getmapdata(sd->bl.m);
+
+						if (mapdata->instance_id < 1 && qi->objectives[j]->mapid != sd->bl.m)
+							continue;
+						if (mapdata->instance_id > 0 && mapdata->instance_src_map != qi->objectives[j]->mapid)
```
Lemongrass3110, line 741:
```suggestion
						if (qi->objectives[j]->mapid != sd->bl.m && mapdata->instance_src_map != qi->objectives[j]->mapid)
```
Would it not work like this too?

db/import-tmpl/quest_db.yml
```diff
@@ -46,6 +46,7 @@
 #                   On normal maps only kills of permanently spawned monsters increase the counter.
 #                   On instance maps all monster kills increase the counter.
```
Lemongrass3110, line 47: Doesnt this need to be adjusted too?
I dont see any checks like this anymore.

src/map/quest.cpp
Lemongrass3110: Doesnt checking md->spawn still give the problem with the "npc/script spawned" mobs not being counted?

src/map/quest.cpp
```diff
@@ -179,6 +179,7 @@ uint64 QuestDatabase::parseBodyNode(const YAML::Node &node) {
 				target->element = ELE_ALL;
 				target->mapid = -1;
 				target->map_name = "";
+				target->count_slave = true;
```
Lemongrass3110, line 182: I am not sure if we should make it true by default, like you also said in your comment on this PR.

Lemongrass3110 (commented):
I think it might still be a better idea to set a flag/state that a monster was GM spawned and maybe spawned from an item (branches, etc), because this is what we actually intend to prevent or not?

PR #6643, opened 2022-03-03
Episode 17.1 - Illusion

Everade:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/6463

* **Server Mode**: 
Renewal

* **Description of Pull Request**: 
Implementation of Episode 17.1 - Illusion (Walkthrough Conversion)

**ToDo:**
- [x] Monster
- [x] Monster Skills (only custom)
- [x] Add getexpratio script command for official exp reward mechanics
- [x] Add viewpointmap script command to trigger a map wide viewpoint
- [x] Fix Quest ID missmatch with quest IDs 17000+ that have been used in custom pre-re scripts
- [x] Match Illusion Enchanter mechanics to kRO (Some modules must be item restricted)
- [ ] Independent Tester Feedback

**Enchanter Info**
https://www.divine-pride.net/forum/index.php?/topic/3324-kro-episode-171-enchant-info-illusion-equipment-and-os-weapons/

Credits to @crazy-arashi for the initial script/content release

PR #6486, opened 2022-01-08
17.1 Dungeon - Odin Past.

taig40:
* **Addressed Issue(s)**:
Partially #6469 

* **Server Mode**: 
Renewal

* **Description of Pull Request**:
Add warps for Odin Past, at Odin Temple F3, add merchants for enchanting, add mobs, mob skills, spawns and items relative to that update.

Todos:
- [x] Check official enchantment NPC script
- [x] Monsters (custom)
- [x] Monster skills (custom)
- [ ] https://github.com/rathena/rathena/pull/6753

PR #6753, opened 2022-03-28
New boss skills

eppc0330:
* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

Updated new boss skills from https://ro.gnjoy.com/news/devnote/View.asp?category=1&seq=3846495

Todos:

1. NPC_IMMUNE_PROPERTY skill has special mechanism.
   - The skill has 5 max level, though it has 10 immune stasus. So skill level affects only skill duration(DP max skill level is wrong). The element of immune status will follow an element which has done the highest damage before the buff turned on.

   - How to implement above though?
2. Below is a screenshot from skill patch note of KRo.
NPC_DAMAGE_HEAL

   - As I see in screenshot, NPC_DAMAGE_HEAL should have different effect between magic heal and weapon heal.
 I found effect number of blue magic heal effect, which is in right side screenshot, but how to replace the effect from original AL_HEAL effect?

   

~~2. NPC_ALL_STAT_DOWN has no effect & sound with my code.~~ Solved
~~SCs about NPC_IMMUNE_PROPERTY doesn't display effects.~~ Solved thanks to @[attackjom](https://github.com/attackjom)


attackjom:
SC_IMMUNE_PROPERTY_XXX  try to add ->
```
Flags:
      BlEffect: true
      DisplayPc: true
```

if you don't want korean word above skill user  you need to set skill_lv in clif_skill_nodamage to -1.

eppc0330:
> SC_IMMUNE_PROPERTY_XXX try to add ->
> 
> ```
> Flags:
>       BlEffect: true
>       DisplayPc: true
> ```
> 
> if you don't want korean word above skill user you need to set skill_lv in clif_skill_nodamage to -1.

Ooooh...! Thanks I will try.

And Im Korean. :)

Haydrich:
NPC_DAMAGE_HEAL monster activates it, but it's ineffective
NPC_ALL_STAT_DOWN and NPC_MOVE_COORDINATE works fine but missing effect
image

Haydrich:
NPC_DAMAGE_HEAL monster activates it, but it's ineffective
NPC_ALL_STAT_DOWN and NPC_MOVE_COORDINATE works fine but missing effect
all stats down effect:
[image
](https://browiki.org/images/a/af/Perda_Total.gif)

eppc0330:
> NPC_DAMAGE_HEAL monster activates it, but it's ineffective NPC_ALL_STAT_DOWN and NPC_MOVE_COORDINATE works fine but missing effect all stats down effect: [image ](https://browiki.org/images/a/af/Perda_Total.gif)

NPC_ALL_STAT_DOWN : Sorry I still have no idea how to implement this skill's effect.
I have tried ```clif_specialeffect(bl,EF_ENERVATION7, AREA)``` but it still does not have sound.

NPC_MOVE_COORDINATE : I need information about whether the effect is shown under both target and caster, or just caster.
Temporarily I would change effect to giving both target and caster.

And NPC_DAMAGE_HEAL : I have tested enough, but couldn't find problem. Can you explain detail about 'ineffective'?

Haydrich:
> 

how it works with you? when the monster cast the skill, should be my damage is miss and monster heal herself for the duration of the skill, but actually when the monster cast it i can be damage the monster and monster not healing

eppc0330:
> > 
> 
> how it works with you? when the monster cast the skill, should be my damage is miss and monster heal herself for the duration of the skill, but actually when the monster cast it i can be damage the monster and monster not healing

Did you tested with skill level? On level 1 caster ignore weapon damage, on 2 ignore magic damage, and on 3 ignore misc damage.

And on my server it works fine.

Haydrich:
> > > 
> > 
> > 
> > how it works with you? when the monster cast the skill, should be my damage is miss and monster heal herself for the duration of the skill, but actually when the monster cast it i can be damage the monster and monster not healing
> 
> Did you tested with skill level? On level 1 caster heals weapon damage, on 2 magic damage, and on 3 misc damage.
> 
> And my server it works fine.

it's my fault, sry, you are right

limitro:
Is this EF_MOVE_COORDINATE declared in rAthena? :O Can't seem to find it.

eppc0330:
> Is this EF_MOVE_COORDINATE declared in rAthena? :O Can't seem to find it.

Yes, it has been added on #6643

src/map/skill.cpp
```diff
@@ -7483,6 +7483,53 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			case 1  : type = SC_IMMUNE_PROPERTY_NOTHING;	break;
+			case 2  : type = SC_IMMUNE_PROPERTY_WATER;		break;
+			case 3  : type = SC_IMMUNE_PROPERTY_GROUND;		break;
+			case 4  : type = SC_IMMUNE_PROPERTY_FIRE;		break;
+			case 5  : type = SC_IMMUNE_PROPERTY_WIND;		break;
+			case 6  : type = SC_IMMUNE_PROPERTY_DARKNESS;	break;
+			case 7  : type = SC_IMMUNE_PROPERTY_SAINT;		break;
+			case 8  : type = SC_IMMUNE_PROPERTY_POISON;		break;
+			case 9  : type = SC_IMMUNE_PROPERTY_TELEKINESIS;break;
+			case 10 : type = SC_IMMUNE_PROPERTY_UNDEAD;		break;
```
aleos89, line 7497:
```suggestion
			case 1:
				type = SC_IMMUNE_PROPERTY_NOTHING;
				break;
			case 2:
				type = SC_IMMUNE_PROPERTY_WATER;
				break;
			case 3:
				type = SC_IMMUNE_PROPERTY_GROUND;
				break;
			case 4:
				type = SC_IMMUNE_PROPERTY_FIRE;
				break;
			case 5:
				type = SC_IMMUNE_PROPERTY_WIND;
				break;
			case 6:
				type = SC_IMMUNE_PROPERTY_DARKNESS;
				break;
			case 7:
				type = SC_IMMUNE_PROPERTY_SAINT;
				break;
			case 8:
				type = SC_IMMUNE_PROPERTY_POISON;
				break;
			case 9:
				type = SC_IMMUNE_PROPERTY_TELEKINESIS;
				break;
			case 10:
				type = SC_IMMUNE_PROPERTY_UNDEAD;
				break;
```
White space is free ;)

src/map/skill.cpp
```diff
@@ -7483,6 +7483,53 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+				else if (dir > 4) x = -1;
+				else x = 0;
+				if (dir > 2 && dir < 6) y = 1;
+				else if (dir == 7 || dir < 2) y = -1;
+				else y = 0;
+
+				if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
+					skill_blown(src, bl, 1, map_calc_dir_xy(src->x + x, src->y + y, bl->x, bl->y, unit_getdir(src)), BLOWN_NONE);
+				}
+			}
```
aleos89, line 7518:
```suggestion
			int16 x, y;
			uint8 dir = map_calc_dir(bl, src->x, src->y);

			if (dir > 0 && dir < 4)
				x = 1;
			else if (dir > 4)
				x = -1;
			else
				x = 0;
			if (dir > 2 && dir < 6)
				y = 1;
			else if (dir == 7 || dir < 2)
				y = -1;
			else
				y = 0;

			if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
				skill_blown(src, bl, 1, map_calc_dir_xy(src->x + x, src->y + y, bl->x, bl->y, unit_getdir(src)), BLOWN_NONE);
			}
```

src/map/skill.cpp
```diff
@@ -7483,6 +7483,53 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+		} else {
+			int x = src->x;
+			int y = src->y;
```
aleos89, line 7521:
```suggestion
			int16 x = src->x, y = src->y;
```

src/map/skill.cpp
```diff
@@ -7483,6 +7483,53 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			if (unit_movepos(src, bl->x, bl->y, 1, 1)) {
+				clif_blown(src);
+			}
+			if (unit_movepos(bl, x, y, 1, 1)) {
+				clif_blown(bl);
+			}
+		}
+		clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+		clif_specialeffect(bl, EF_MOVE_COORDINATE, AREA);
```
aleos89, line 7530: Are you sure the effect isn't showing up because of an outdated client? The client effect should be handled by the `clif_skill_nodamage` call right above.

src/map/skill.cpp
```diff
@@ -7649,9 +7696,16 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 	case EM_SPELL_ENCHANTING:
+	case NPC_GRADUAL_GRAVITY:
+	case NPC_DAMAGE_HEAL:
 		clif_skill_nodamage(src,bl,skill_id,skill_lv,
 			sc_start(src,bl,type,100,skill_lv,skill_get_time(skill_id,skill_lv)));
 		break;
+	case NPC_ALL_STAT_DOWN:
+		clif_skill_nodamage(src, bl, skill_id, skill_lv,
+			sc_start(src, bl, type, 100, skill_lv, skill_get_time(skill_id, skill_lv)));
+		clif_specialeffect(bl, EF_ENERVATION7, AREA); //Add effect, but still need to add sound.
```
aleos89, line 7707: Guess the same question again here. We shouldn't have to forcefully send out the client effects as it's handled via the `clif_skill_*` packet.

aleos89 (commented):
Looks good so far, just a few things I saw.

aleos89:
> If KRo one is correct,
> 
>     * How to produce 10 immune statuses with 5 skill levels?
> 
>     * Why does AttackRange needed? I thought this is self casting skill.
> 
> 
>     2. NPC_ALL_STAT_DOWN has no effect & sound with my code.
>        ~SCs about NPC_IMMUNE_PROPERTY doesn't display effects.~ Solved thanks to @[attackjom](https://github.com/attackjom)

* A skill's info is capped to MAX_SKILL_LEVEL macro which is 13. So a skill can have a MaxLevel of 5 but it will have values filled depending on how it's written in the skill_db. If we need to define the Element label for each level, that works as well.

* Not sure why Gravity gave the skill a range, but we don't need to define it in our code if it's self.

* Skill effects and sounds are handled via `clif_skill_*` packets and don't need to be explicitly defined in source. It could be the client version you have doesn't support the new skills.

src/map/skill.cpp
eppc0330: Im not sure 🤔 These skills are updated in 2018, which is done so long time ago to doubt about client version, and my client version is 20211117. 

Does this can be due to inconsistency between client version and lua version?

src/map/skill.cpp
eppc0330: Same as above 🙂 I will remove both ```clif_specialeffect``` lines if the problem is doing with client version.

eppc0330:
> > If KRo one is correct,
> > ```
> > * How to produce 10 immune statuses with 5 skill levels?
> > 
> > * Why does AttackRange needed? I thought this is self casting skill.
> > 
> > 
> > 2. NPC_ALL_STAT_DOWN has no effect & sound with my code.
> >    ~SCs about NPC_IMMUNE_PROPERTY doesn't display effects.~ Solved thanks to @[attackjom](https://github.com/attackjom)
> > ```
> 
> * A skill's info is capped to MAX_SKILL_LEVEL macro which is 13. So a skill can have a MaxLevel of 5 but it will have values filled depending on how it's written in the skill_db. If we need to define the Element label for each level, that works as well.
> * Not sure why Gravity gave the skill a range, but we don't need to define it in our code if it's self.
> * Skill effects and sounds are handled via `clif_skill_*` packets and don't need to be explicitly defined in source. It could be the client version you have doesn't support the new skills.

Thanks for review.

1. Then is it okay to implement the skill different way with official? I have some other ideas to produce immune statuses with 5 skill levels, but these are all not verified becuase i cant find information about it. 

2. Maybe

eppc0330:
> Looks good so far, just a few things I saw.

Thank you. I will do all changes soon.

SeyrenDev:
@eppc0330   when I did a git pull and compiled, I got an error that there were two equal tags, case NPC_MOVE_COORDINATE.  
Screenshot_3

Screenshot_2

eppc0330:
> @eppc0330 when I did a git pull and compiled, I got an error that there were two equal tags, case NPC_MOVE_COORDINATE. Screenshot_3
> 
> Screenshot_2

Sorry can you explain detail? I cant understand equal tags because this skill is not implemented on Rathena yet, and I have not worked two codes. Also I have no equal tags error in my server. Or am I misunderstanding error?

Everade:
@SeyrenDev Please test with a clean environment, you must have merged the Odin Temple pull request. Your problem isn't related to this pull request.

SeyrenDev:
I found why I had duplicated the NPC_MOVE_COORDINATE. I had previously added https://github.com/rathena/rathena/pull/6486/files.

eppc0330:
> I found why I had duplicated the NPC_MOVE_COORDINATE. I had previously added https://github.com/rathena/rathena/pull/6486/files. my mistake

Oooh nice 😀

zdurexz:
> Oooh nice 😀 I didnt know this skill is already worked on other pull.

Don't care Other PR because not merge Master

eppc0330:
@aleos89 Some news. I got detail info of NPC_IMMUNE_PROPERTY 
and issue of skilleffect.
There's more other effectless implemented skills. I will check compare with skilleffectinfolist.lub and separtely open new issue with this.

src/map/skill.cpp
eppc0330: > Are you sure the effect isn't showing up because of an outdated client? The client effect should be handled by the `clif_skill_nodamage` call right above.

 Solved and erased ```clif_specialeffect```

src/map/skill.cpp
eppc0330: Solved and erased ```clif_specialeffect```

db/re/skill_db.yml
```diff
@@ -17911,6 +17911,72 @@ Body:
...
+    DamageFlags:
+      NoDamage: true
+    Flags:
+      IsNpc: true
+    Duration1: 30000
+    Status: DAMAGE_HEAL
+  - Id: 754
+    Name: NPC_IMMUNE_PROPERTY
+    Description: Property Immune
+    MaxLevel: 10 #Check: Max level is 5?
```
Pokye, line 17952: according to kRO's official current lub it is level 5.
```c
	[SKID.NPC_IMMUNE_PROPERTY] = {
		"NPC_IMMUNE_PROPERTY",
		SkillName = "이뮨 엘리멘탈",
		MaxLv = 5,
		SpAmount = { 0, 0, 0, 0, 0 },
		bSeperateLv = false,
		AttackRange = { 5, 10, 15, 20, 25 }
	},

Comment (eppc0330) [reply db/re/skill_db.yml]: I fixed max level, but actually to apply this, skill mechanism must be fixed first. 
I have no idea how to implement right now...

Without completion of the skill it would be better leave max level to 10.

File: src/map/skill.cpp [hunk 7]
```diff
@@ -7476,6 +7476,79 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+				break;
+			case 10:
+				type = SC_IMMUNE_PROPERTY_UNDEAD;
+				break;
+		}
+		clif_skill_nodamage(src,bl, skill_id,skill_lv,sc_start(src,bl,type,100,skill_lv,skill_get_time(skill_id,skill_lv)));
+		sc_start(src,bl, SC_IMMUNE_PROPERTY,100,skill_lv,skill_get_time(skill_id,skill_lv));
+		break;
+
+	case NPC_MOVE_COORDINATE:
```
Pokye, line 7516: I found the skill in that form, isn't that correct too?
```c
	case NPC_MOVE_COORDINATE:
		if (src != bl) {
			int x = src->x, y = src->y;
			if (unit_movepos(src,bl->x,bl->y,0,0)) {
				clif_skill_nodamage(src,src,skill_id,skill_lv,1);
				clif_blown(src);
				if (unit_movepos(bl,x,y,0,0)) {
					clif_skill_nodamage(bl,bl,skill_id,skill_lv,1);
					clif_blown(bl);
				}
				map_foreachinallrange(unit_changetarget,src,AREA_SIZE,BL_MOB,bl,src);
			}
		}
		break;

Comment (eppc0330) [reply src/map/skill.cpp:7516]: Skill needs two kinds of form. I see this has only a single form.
Skill needs ```clif_skill_damage``` to show effect on target.
Also, ```clif_blown``` will not remove specific SCs must be removed on moving, for example SC_ROLLINGCUTTER.

**Comment by OptimusM** [comment]
i have been testing this today, and i think it is pretty neccesary to have some additional conditions to trigger skills, like those ones you mentioned on the PR.

these 2 are really good for this i think, and would probably help whit this
https://github.com/rathena/rathena/pull/7138

Also, i think it could be a good idea to add a condition specially for "matk" skills, like "magicattacked" to trigger NPC_DAMAGEHEAL lvl 2 or just to be used to enhance the functionallity of triggers.

something like:
`
	else if (flag&BF_MAGIC)
		res = mobskill_use(md, tick, MSC_MAGICATTACKED);
`
		
(that´s obviously no the complete code, but that´s the idea, personally i think it could be usefull for mob skill db features)

File: src/map/skill.cpp [hunk 8]
```diff
@@ -7647,9 +7720,16 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case TR_MYSTIC_SYMPHONY:
 	case TR_KVASIR_SONATA:
 	case EM_SPELL_ENCHANTING:
+	case NPC_GRADUAL_GRAVITY:
+	case NPC_DAMAGE_HEAL:
```
Atemo, line 7724: From my test using your PR, the monster using NPC_IMMUNE PROPERTY + NPC_DAMAGE_HEAL is still healed by a positive number. No idea how it works on KRO, so for me the heal should be 0

Atemo (changes requested):
Small review!

src/map/status.cpp
```diff
@@ -10609,6 +10621,16 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				val4 = 1;
 			tick_time = val2 * 1000; // [GodLesZ] tick time
 			break;
+
+		case SC_GRADUAL_GRAVITY:
+			val2 = 10 * val1;
+			val4 = tick / 1000;
+			tick_time = 1000;
```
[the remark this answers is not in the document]
eppc0330: Sorry I guess I missed this part during solving conflicts

src/map/skill.cpp
```diff
@@ -7647,9 +7720,16 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 		clif_skill_nodamage(src,bl,skill_id,skill_lv,
 			sc_start(src,bl,type,100,skill_lv,skill_get_time(skill_id,skill_lv)));
```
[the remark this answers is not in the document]
eppc0330: Thanks for review. I added it.

src/map/skill.cpp
eppc0330: My bad. Fixed.

src/map/skill.cpp
```diff
@@ -7647,9 +7720,16 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 		break;
+	case NPC_ALL_STAT_DOWN:
```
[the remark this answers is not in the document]
eppc0330: I changed stats' cap value. Thanks.

src/map/skill.cpp
```diff
@@ -7567,6 +7567,79 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+				y = -1;
+			else
+				y = 0;
+			if (status_get_class_(src) == CLASS_BOSS) {//if caster is a boss, the skill will pull target ahead of the boss, instead of switch coordinates
+				if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
+					clif_blown(src);
+				}
+			}
+			else {
+				int16 px = bl->x, py = bl->y;
```
aleos89, line 7630:
```suggestion
```

src/map/skill.cpp
```diff
@@ -7567,6 +7567,79 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+				if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
+					clif_blown(bl);
+				}
+				if (unit_movepos(src, px, py, 1, 1)) {
```
aleos89, line 7634:
```suggestion
				if (unit_movepos(src, bl->x, bl->y, 1, 1)) {
```

src/map/status.cpp
```diff
@@ -6323,8 +6323,10 @@ static unsigned short status_calc_str(struct block_list *bl, struct status_chang
 		str += sc->data[SC_ALMIGHTY]->val1;
 	if (sc->data[SC_ULTIMATECOOK])
 		str += sc->data[SC_ULTIMATECOOK]->val1;
+	if (sc->data[SC_ALL_STAT_DOWN])
+		str -= sc->data[SC_ALL_STAT_DOWN]->val2;
 
-	return (unsigned short)cap_value(str,0,USHRT_MAX);
+	return (unsigned short)cap_value(str,SHRT_MIN,USHRT_MAX);
```
aleos89, line 6329: These stats can go below 0 now?

src/map/status.cpp
```diff
@@ -13859,6 +13879,18 @@ TIMER_FUNC(status_change_timer){
 		}
 		break;
 
+	case SC_GRADUAL_GRAVITY:
+		if (--(sce->val4) >= 0) {
+			int hp = 0;
+			hp = status->max_hp * sce->val2 / 100;
```
aleos89, line 13885:
```suggestion
			int hp = status->max_hp * sce->val2 / 100;

```

aleos89 (commented):
Just a few things!

src/map/status.cpp
eppc0330: Thanks for review. Atemo said yes it is.
This is KRo screenshot from Atemo.
195630729-54a72ea9-ba57-482a-8845-23ab49d92e9f

aleos89: Thanks, yeah I had talked to Atemo too who confirmed it as well. So the issue here is that the `status_get_<stat>` functions are `unsigned short` which means they won't return negative values, even with your cap adjustment. This will require a bit more of refactoring to get these changes in line as we will need to confirm all the other functions and variables are adjusted properly as well.

eppc0330: Ohhh...! You are right. Then ```singed short``` would be okay...? 
Actually I can't expect what detail problems or variables will be arise.
Do you have any thoughts?

aleos89: Don't worry. I wouldn't expect to implement that behavior in this PR as you would lose the scope of this one. I would just revert these pieces that you changed here for the `cap_value()` and we can then merge this and in a separate PR we can clean up the functions to support negative values!

eppc0330: Alright! Thanks.

src/map/battle.cpp
```diff
@@ -1450,6 +1450,44 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
...
+		}
+	}
+
+	if (sc && sc->data[SC_DAMAGE_HEAL]) {
+		int dmg_heal_lv = sc->data[SC_DAMAGE_HEAL]->val1;
+		if (damage > 0 && ((flag & BF_WEAPON && dmg_heal_lv == 1) || (flag & BF_MAGIC && dmg_heal_lv == 2) || (flag & BF_MISC && dmg_heal_lv == 3))) {//Absorb MISC damage or WEAPON & MAGIC damage on level 3?
+			clif_skill_nodamage(NULL, bl, AL_HEAL, (int)damage, 1);
+			status_heal(bl, damage, 0, 0);
+			damage = 0;
+		}
```
aleos89, line 1488:
```suggestion

		if (sc->data[SC_DAMAGE_HEAL]) {
			int dmg_heal_lv = sc->data[SC_DAMAGE_HEAL]->val1;
			if (damage > 0 && ((flag & BF_WEAPON && dmg_heal_lv == 1) || (flag & BF_MAGIC && dmg_heal_lv == 2) || (flag & BF_MISC && dmg_heal_lv == 3))) {//Absorb MISC damage or WEAPON & MAGIC damage on level 3?
				clif_skill_nodamage(NULL, bl, AL_HEAL, (int)damage, 1);
				status_heal(bl, damage, 0, 0);
				damage = 0;
			}
		}
```

src/map/skill.cpp
```diff
@@ -7574,6 +7574,78 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+				}
+			}
+			else {
+				if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
+					clif_blown(bl);
+				}
+				if (unit_movepos(src, bl->x, bl->y, 1, 1)) {
+					clif_blown(src);
+				}
+			}
```
aleos89, line 7643:
```suggestion
			if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
				clif_blown(bl);
			}
			if (status_get_class_(src) != CLASS_BOSS && unit_movepos(src, bl->x, bl->y, 1, 1)) { // If caster is not a boss, the skill will pull target ahead of the caster, instead of switching coordinates
				clif_blown(src);
			}
```

aleos89 (approved):
Just some minor code cleanup and removal of duplicate code!

src/map/skill.cpp
eppc0330: Sorry my mistake. This part should not be same but I missed. I fixed it with another small fix.

src/map/skill.cpp
```diff
@@ -7759,6 +7832,15 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		clif_skill_nodamage(src,bl,skill_id,skill_lv,
 			sc_start(src,bl,type,100,skill_lv,skill_get_time(skill_id,skill_lv)));
 		break;
+	case NPC_GRADUAL_GRAVITY:
+		clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+		status_change_start(src, bl, type, 10000, skill_lv, 0, 0, 0, skill_get_time(skill_id, skill_lv), SCSTART_NOAVOID|SCSTART_NOTICKDEF|SCSTART_NORATEDEF);
+		break;
+	case NPC_ALL_STAT_DOWN:
```
Atemo, line 7839:
```suggestion
	case NPC_ALL_STAT_DOWN:
```

src/map/skill.cpp
```diff
@@ -7759,6 +7832,15 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		clif_skill_nodamage(src,bl,skill_id,skill_lv,
```
Atemo, line 7832: NPC_DAMAGE_HEAL is missing in skill_castend_nodamage_id
```suggestion
	case NPC_DAMAGE_HEAL:
		clif_skill_nodamage(src,bl,skill_id,skill_lv,
```

Atemo (changes requested):

src/map/skill.cpp
eppc0330: Thanks

Atemo (approved):

src/map/battle.cpp
```diff
@@ -1450,6 +1450,44 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
...
+			|| (sc->data[SC_IMMUNE_PROPERTY_WATER] && element == ELE_WATER)
+			|| (sc->data[SC_IMMUNE_PROPERTY_GROUND] && element == ELE_EARTH)
+			|| (sc->data[SC_IMMUNE_PROPERTY_FIRE] && element == ELE_FIRE)
+			|| (sc->data[SC_IMMUNE_PROPERTY_WIND] && element == ELE_WIND)
+			|| (sc->data[SC_IMMUNE_PROPERTY_DARKNESS] && element == ELE_DARK)
+			|| (sc->data[SC_IMMUNE_PROPERTY_SAINT] && element == ELE_HOLY)
+			|| (sc->data[SC_IMMUNE_PROPERTY_POISON] && element == ELE_POISON)
+			|| (sc->data[SC_IMMUNE_PROPERTY_TELEKINESIS] && element == ELE_GHOST)
+			|| (sc->data[SC_IMMUNE_PROPERTY_UNDEAD] && element == ELE_UNDEAD)) {
+				damage = 0;
```
Lemongrass3110, line 1477: Shouldnt it return right after?

src/map/skill.cpp
```diff
@@ -7574,6 +7574,79 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+				int16 px = bl->x, py = bl->y; //save target's coordinates before target move
+				if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
+					clif_blown(bl);
+				}
+				if (unit_movepos(src, px, py, 1, 1)) {
+					clif_blown(src);
+				}
+			}
+			clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+			clif_skill_damage(src, bl, tick, status_get_amotion(src), 0, -30000, 1, skill_id, skill_lv, DMG_SINGLE);
```
Lemongrass3110, line 7646: Why do we tell the client -30000 damage?

src/map/skill.cpp
```diff
@@ -7574,6 +7574,79 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			else
+				y = 0;
+			if (status_get_class_(src) == CLASS_BOSS) {//if caster is a boss, the skill will pull target ahead of the boss, instead of switch coordinates
+				if (skill_check_unit_movepos(1, bl, src->x+x, src->y+y, 1, 1)) {
+					clif_blown(bl);
+				}
+			}
+			else {
+				int16 px = bl->x, py = bl->y; //save target's coordinates before target move
+				if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
```
Lemongrass3110, line 7638: Why does this care about battleground? (the first 1 in the call)

src/map/skill.cpp
```diff
@@ -7574,6 +7574,79 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			else
+				x = 0;
+			if (dir > 2 && dir < 6)
+				y = 1;
+			else if (dir == 7 || dir < 2)
+				y = -1;
+			else
+				y = 0;
+			if (status_get_class_(src) == CLASS_BOSS) {//if caster is a boss, the skill will pull target ahead of the boss, instead of switch coordinates
+				if (skill_check_unit_movepos(1, bl, src->x+x, src->y+y, 1, 1)) {
```
Lemongrass3110, line 7632: Why does this care about battleground? (the first 1 in the call)

src/map/skill.cpp
```diff
@@ -7574,6 +7574,79 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+					clif_blown(bl);
+				}
+			}
+			else {
+				int16 px = bl->x, py = bl->y; //save target's coordinates before target move
+				if (skill_check_unit_movepos(1, bl, src->x, src->y, 1, 1)) {
+					clif_blown(bl);
+				}
+				if (unit_movepos(src, px, py, 1, 1)) {
```
Lemongrass3110, line 7641: Should the "caster" not be checked with skill_check_unit_movepos?
Or to be more explicit: If a monster is under SC_SV_ROOTTWIST should it really be able to hop out of it by casting that skill?

src/map/skill.cpp
```diff
@@ -7756,9 +7829,19 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 			sc_start(src,bl,type,100,skill_lv,skill_get_time(skill_id,skill_lv)));
 		break;
+	case NPC_GRADUAL_GRAVITY:
+		clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+		status_change_start(src, bl, type, 10000, skill_lv, 0, 0, 0, skill_get_time(skill_id, skill_lv), SCSTART_NOAVOID|SCSTART_NOTICKDEF|SCSTART_NORATEDEF);
+		break;
+	case NPC_ALL_STAT_DOWN:
+		clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+		status_change_start(src, bl, type, 10000, skill_lv, 0, 0, 0, skill_get_time(skill_id, skill_lv), SCSTART_NOAVOID|SCSTART_NOTICKDEF|SCSTART_NORATEDEF);
+		clif_skill_damage(src,bl,tick, status_get_amotion(src), 0, -30000, 1, skill_id, skill_lv, DMG_SINGLE);
```
Lemongrass3110, line 7843: Why do we tell the client -30000 damage?

src/map/skill.cpp
```diff
@@ -7756,9 +7829,19 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case TR_MYSTIC_SYMPHONY:
 	case TR_KVASIR_SONATA:
 	case EM_SPELL_ENCHANTING:
+	case NPC_DAMAGE_HEAL:
 		clif_skill_nodamage(src,bl,skill_id,skill_lv,
 			sc_start(src,bl,type,100,skill_lv,skill_get_time(skill_id,skill_lv)));
 		break;
+	case NPC_GRADUAL_GRAVITY:
+		clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+		status_change_start(src, bl, type, 10000, skill_lv, 0, 0, 0, skill_get_time(skill_id, skill_lv), SCSTART_NOAVOID|SCSTART_NOTICKDEF|SCSTART_NORATEDEF);
```
Lemongrass3110, line 7838: status_change_start result should be given to clif_skill_nodamage in last parameter. So if the SC cannot be started it does not display skill success.

src/map/skill.cpp
```diff
@@ -7756,9 +7829,19 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+		break;
+	case NPC_ALL_STAT_DOWN:
+		clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+		status_change_start(src, bl, type, 10000, skill_lv, 0, 0, 0, skill_get_time(skill_id, skill_lv), SCSTART_NOAVOID|SCSTART_NOTICKDEF|SCSTART_NORATEDEF);
```
Lemongrass3110, line 7842: status_change_start result should be given to clif_skill_nodamage in last parameter. So if the SC cannot be started it does not display skill success.

src/map/status.cpp
```diff
@@ -10792,6 +10804,14 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				val4 = 1;
 			tick_time = val2 * 1000; // [GodLesZ] tick time
 			break;
+		case SC_GRADUAL_GRAVITY:
+			val2 = 10 * val1;
+			val4 = tick / 1000;
+			tick_time = 1000;
```
Lemongrass3110, line 10810: It is better to use status_get_sc_interval function.

src/map/status.cpp
```diff
@@ -13859,6 +13879,18 @@ TIMER_FUNC(status_change_timer){
...
 
+	case SC_GRADUAL_GRAVITY:
+		if (--(sce->val4) >= 0) {
+			int hp = status->max_hp * sce->val2 / 100;
+
+			if (!status_charge(bl, hp, 0))
+				status_zap(bl, hp, 0);
+			if (sc->data[type])
+				sc_timer_next(1000 + tick);
+			return 0;
```
Lemongrass3110, line 13890: Should be rewritten to use status_get_sc_interval. Check SC_SHIELDSPELL_HP for reference.

src/map/status.hpp
```diff
@@ -1257,6 +1257,21 @@ enum sc_type : int16 {
...
+	SC_IMMUNE_PROPERTY_NOTHING,
+	SC_IMMUNE_PROPERTY_WATER,
+	SC_IMMUNE_PROPERTY_GROUND,
+	SC_IMMUNE_PROPERTY_FIRE,
+	SC_IMMUNE_PROPERTY_WIND,
+	SC_IMMUNE_PROPERTY_POISON,
+	SC_IMMUNE_PROPERTY_SAINT,
+	SC_IMMUNE_PROPERTY_DARKNESS,
+	SC_IMMUNE_PROPERTY_TELEKINESIS,
+	SC_IMMUNE_PROPERTY_UNDEAD,
```
Lemongrass3110, line 1273: I am a little 50:50 on if we should define 1 SC per element, as they cannot stack anyway.

Lemongrass3110 (commented):
Good work so far!

The feedback is mostly for the team so that they can also comment on my thoughts.
You do not have to fix all of those yourself, we will help you out on it.

src/map/status.hpp
eppc0330: Yes, as I explained on top of this pull as todos, this skill is not completed. This skill's max level is 5, and level only affects skill's duration. You can check this in ```skillinfolist.lub``` of KRo.

src/map/status.cpp
eppc0330: Yes but I guess we need to add `if (sc->data[type])` on using `status_get_sc_interval` part because without it server will print error when player die during duration of `SC_GRADUAL_GRAVITY`.

eppc0330:
Sorry I deleted wrong repository.... I can't edit this pull anymore :(

eppc0330:
Sorry I have deleted wrong repository...

db/re/status.yml
```diff
@@ -8438,3 +8438,186 @@ Body:
...
+      Agi: true
+      Vit: true
+      Int: true
+      Dex: true
+      Luk: true
+  - Status: GRADUAL_GRAVITY
+    Icon: EFST_GRADUAL_GRAVITY
+    DurationLookup: NPC_GRADUAL_GRAVITY
+    Flags:
+      BlEffect: true
```
attackjom, line 8455: NoDispell + NoClearance

src/map/skill.cpp
attackjom: image
official

eppc0330:
I will close this pull and open new one so we can fix any issue comes out. I can't fix this pull anymore because I've accidentally deleted repository before sorry.

Lemongrass3110:
@eppc0330 can you make 1 PR per skill?
Then it should be easier for us too.

eppc0330:
@Lemongrass3110 Okay, actually I was working on this several days currently and I guess almost finished.
The behaviors are quite accurate as KRo. I will open pulls soon as separate PR as you request.

PR #6764, opened 2022-03-30
Updates the status resistances for renewal

### LINKED ISSUE #4694: SC_STONE (petrified), LUK suposed to reduce duration
https://irowiki.org/wiki/Status_Effects#Stone

> Luk reduces the duration.

```
		case SC_STONE:
			sc_def = status->mdef*100;
			sc_def2 = status->luk*10 + status_get_lv(bl)*10 - status_get_lv(src)*10;
			tick_def = 0; // No duration reduction
			break;
```

aleos89:
* **Addressed Issue(s)**: Fixes #4694

* **Server Mode**: Renewal

  * Updates the status resistance formulas for several of the common statuses.
  * Renewal only protects against a single base stat.
  * When the caster is a higher base level, some statuses now take this difference into effect.
Thanks to @Playtester!

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -18657,7 +18657,7 @@ Body:
       - Level: 10
         Time: 2000
     AfterCastActDelay: 2000
-    Duration1: 8000
+    Duration1: 18000
     FixedCastTime: 500
     Requires:
       SpCost:
@@ -18704,7 +18704,7 @@ Body:
         Area: 6
       - Level: 5
         Area: 7
-    Duration1: 18000
+    Duration1: 15000
     Cooldown: 10000
     Requires:
       SpCost: 30
@@ -19421,7 +19421,6 @@ Body:
     CastCancel: true
     CastTime: 3000
     AfterCastActDelay: 500
-    Duration1: 13000
     Cooldown: 3000
     FixedCastTime: 1000
     Requires:
@@ -19436,7 +19435,6 @@ Body:
           Amount: 82
         - Level: 5
           Amount: 88
-    Status: Burning
   - Id: 2213
     Name: WL_COMET
     Description: Comet
@@ -23715,11 +23713,11 @@ Body:
         Time: 15000
     Duration2:
       - Level: 1
-        Time: 2000
+        Time: 4000
       - Level: 2
-        Time: 6000
+        Time: 8000
       - Level: 3
-        Time: 10000
+        Time: 12000
     Requires:
       SpCost:
         - Level: 1
@@ -27264,7 +27262,7 @@ Body:
       - Level: 5
         Time: 6000
     AfterCastActDelay: 1000
-    Duration: 18000
+    Duration1: 18000
     Cooldown: 2000
     FixedCastTime:
       - Level: 1
@@ -42854,15 +42852,15 @@ Body:
         Time: 30000
     Duration2:
       - Level: 1
-        Time: 3500
+        Time: 3000
       - Level: 2
-        Time: 5500
+        Time: 6000
       - Level: 3
-        Time: 8500
+        Time: 9000
       - Level: 4
-        Time: 11500
+        Time: 12000
       - Level: 5
-        Time: 14500
+        Time: 15000
     Requires:
       SpCost: 10
     Unit:
@@ -43507,25 +43505,25 @@ Body:
     AfterCastActDelay: 3000
     Duration2:
       - Level: 1
-        Time: 28000
+        Time: 30000
       - Level: 2
-        Time: 33000
+        Time: 35000
       - Level: 3
-        Time: 38000
+        Time: 40000
       - Level: 4
-        Time: 43000
+        Time: 45000
       - Level: 5
-        Time: 48000
+        Time: 50000
       - Level: 6
-        Time: 58000
+        Time: 60000
       - Level: 7
-        Time: 58000
+        Time: 60000
       - Level: 8
-        Time: 58000
+        Time: 60000
       - Level: 9
-        Time: 58000
+        Time: 60000
       - Level: 10
-        Time: 58000
+        Time: 60000
     Requires:
       SpCost:
         - Level: 1
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -741,7 +741,7 @@ Body:
       - Level: 10
         Time: 30000
       - Level: 11
-        Time: 33000
+        Time: 30000
     FixedCastTime: 160
     Requires:
       SpCost:
@@ -5947,7 +5947,7 @@ Body:
     Hit: Single
     HitCount: 1
     Element: Dark
-    Duration2: 18000
+    Duration2: 28000
     Status: Curse
   - Id: 182
     Name: NPC_SLEEPATTACK
@@ -15596,7 +15596,7 @@ Body:
       - Level: 1
         Time: 5000
       - Level: 2
-        Time: 1000
+        Time: 10000
       - Level: 3
         Time: 15000
       - Level: 4
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -175,6 +175,7 @@ Body:
       Skill:
         Plagiarism: true
         Reproduce: true
+    Duration2: 4500
     Requires:
       SpCost:
         - Level: 1
@@ -720,25 +721,25 @@ Body:
     AfterCastActDelay: 500
     Duration2:
       - Level: 1
-        Time: 5500
+        Time: 2500
       - Level: 2
-        Time: 8500
+        Time: 5500
       - Level: 3
-        Time: 11500
+        Time: 8500
       - Level: 4
-        Time: 14500
+        Time: 11500
       - Level: 5
-        Time: 17500
+        Time: 14500
       - Level: 6
-        Time: 20500
+        Time: 17500
       - Level: 7
-        Time: 23500
+        Time: 20500
       - Level: 8
-        Time: 26500
+        Time: 23500
       - Level: 9
-        Time: 29500
+        Time: 26500
       - Level: 10
-        Time: 32500
+        Time: 29500
       - Level: 11
         Time: 32500
     FixedCastTime: 160
@@ -780,7 +781,7 @@ Body:
     CastCancel: true
     CastTime: 800
     Duration1: 5000
-    Duration2: 20000
+    Duration2: 17000
     FixedCastTime: 200
     Requires:
       SpCost:
@@ -2022,7 +2023,7 @@ Body:
       Skill:
         Plagiarism: true
         Reproduce: true
-    Duration2: 20000
+    Duration2: 18000
     Requires:
       SpCost: 12
     Status: Poison
@@ -2617,7 +2618,7 @@ Body:
     HitCount: 1
     CastCancel: true
     AfterCastActDelay: 2000
-    Duration2: 20000
+    Duration2: 18000
     Requires:
       SpCost: 5
     Status: Blind
@@ -2732,25 +2733,25 @@ Body:
     AfterCastActDelay: 3000
     Duration2:
       - Level: 1
-        Time: 30000
+        Time: 28000
       - Level: 2
-        Time: 35000
+        Time: 33000
       - Level: 3
-        Time: 40000
+        Time: 38000
       - Level: 4
-        Time: 45000
+        Time: 43000
       - Level: 5
-        Time: 50000
+        Time: 48000
       - Level: 6
-        Time: 60000
+        Time: 58000
       - Level: 7
-        Time: 60000
+        Time: 58000
       - Level: 8
-        Time: 60000
+        Time: 58000
       - Level: 9
-        Time: 60000
+        Time: 58000
       - Level: 10
-        Time: 60000
+        Time: 58000
     Requires:
       SpCost:
         - Level: 1
@@ -3217,7 +3218,7 @@ Body:
         Time: 6000
       - Level: 10
         Time: 7000
-    Duration2: 5000
+    Duration2: 4500
     Cooldown: 7000
     FixedCastTime:
       - Level: 1
@@ -3432,7 +3433,7 @@ Body:
         Time: 4500
     AfterCastActDelay: 1000
     Duration1: 1000
-    Duration2: 20000
+    Duration2: 18000
     Cooldown: 5000
     FixedCastTime:
       - Level: 1
@@ -3664,25 +3665,25 @@ Body:
     AfterCastActDelay: 200
     Duration2:
       - Level: 1
-        Time: 4000
+        Time: 1000
       - Level: 2
-        Time: 5500
+        Time: 2500
       - Level: 3
-        Time: 7000
+        Time: 4000
       - Level: 4
-        Time: 8500
+        Time: 5500
       - Level: 5
-        Time: 10000
+        Time: 7000
       - Level: 6
-        Time: 11500
+        Time: 8500
       - Level: 7
-        Time: 13000
+        Time: 10000
       - Level: 8
-        Time: 14500
+        Time: 11500
       - Level: 9
-        Time: 16000
+        Time: 13000
       - Level: 10
-        Time: 17500
+        Time: 14500
     FixedCastTime:
       - Level: 1
         Time: 160
@@ -3768,7 +3769,7 @@ Body:
         Time: 6300
     AfterCastActDelay: 1000
     Duration1: 4500
-    Duration2: 30000
+    Duration2: 27000
     Cooldown: 6000
     FixedCastTime:
       - Level: 1
@@ -4161,7 +4162,7 @@ Body:
         Area: 2
       - Level: 6
         Area: 12
-    Duration2: 5000
+    Duration2: 4500
     Requires:
       SpCost: 10
       Weapon:
@@ -4428,7 +4429,7 @@ Body:
         Time: 80000
       - Level: 5
         Time: 40000
-    Duration2: 5000
+    Duration2: 4500
     FixedCastTime: 300
     Requires:
       SpCost: 10
@@ -4571,7 +4572,7 @@ Body:
         Time: 60000
       - Level: 5
         Time: 30000
-    Duration2: 20000
+    Duration2: 18000
     Requires:
       SpCost: 12
       ItemCost:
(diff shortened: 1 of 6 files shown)


src/map/status.cpp
```diff
@@ -8838,17 +8878,20 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
...
-			if( bl->type == BL_PC )
-				tick_def2 = status_get_lv(bl)*20 + status->vit*25 + status->agi*10;
-			else
-				tick_def2 = (status->vit + status->luk)*50;
+			sc_def = status->str * 10 + status_get_lv(bl) * 10 + status->luk * 10;
+			tick_def2 = -2000;
+			break;
+		case SC_FEAR:
+			sc_def = status->int_ * 10 + status_get_lv(bl) * 10 + status->luk * 10;
+			tick_def2 = -2000;
```
Playtester, line 8888: Funnily it should be -4000 because it gets added twice on official servers. (There is an ankle snare effect for two seconds, I think they add 2000ms because of this.)

src/map/status.cpp
```diff
@@ -8838,17 +8878,20 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
...
 			break;
 		case SC_BURNING:
-			tick_def2 = 75*status->luk + 125*status->agi;
+			sc_def = status->agi * 10 + status_get_lv(bl) * 10 + status->luk * 10;
+			tick_def2 = -2000;
 			break;
-		case SC_FREEZING:
+		case SC_FREEZING: // TODO: Confirm
```
Playtester, line 8894: Please also check the minimum defined below (very end of this function), it still says it needs to be tested.

Playtester (changes requested):
Reviewed all changes, some changes required and some optional suggestions. Please see comments.

src/map/status.cpp
```diff
@@ -8981,8 +9021,6 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
 
 	// Cap minimum rate
 	rate = max(rate, scdb->min_rate);
-	// Cap minimum duration
-	tick = i64max(tick, scdb->min_duration);
```
Playtester, line 8985: Keep in mind that if you remove this here, then status changes with flag&SCSTART_NOTICKDEF cannot have a minimum duration applied.

src/map/status.cpp
```diff
@@ -8742,71 +8746,104 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
...
+#endif
 			break;
 		case SC_FREEZE:
+#ifndef RENEWAL
 			sc_def = status->mdef*100;
 			sc_def2 = status->luk*10 + status_get_lv(bl)*10 - status_get_lv(src)*10;
 			tick_def2 = status_src->luk*-10; // Caster can increase final duration with luk
+#else
+			sc_def = status->mdef * 100 - levelAdv;
+			tick_def2 = -3000;
```
Playtester, line 8811: Just as reminder (already talked with you about it on Discord): When I implementated the official status change durations in skill_cast_db, I included those times in the variable casting time as I set those times based on my tests on official servers. Now that we properly implemented it as fixed duration here, we need to remove those added times from the skill_db.yml files.

80-90% of all skills just use the default duration, but those default durations are not implemented yet.

The remaining skills are usually those that have a different durations per level, such a Frost Diver, which has a variable duration of 3000ms*level. I currently put as 2500ms+3000ms*level as duration in the file (the times I measured during my tests were off by 500ms, probably due to latency). But now we add those 3000ms here, so the duration in the skill_db.yml should not include it.

So either we comment the fixed duration out in this commit for now, or we update the skill_db.yml at the same time.

Playtester (commented):
All my previous comments were fixed.

There's a big TODO left in regards to the variable and default durations and one hint (optional). See my comments.

db/re/skill_db.yml
```diff
@@ -720,25 +721,25 @@ Body:
...
+        Time: 20500
       - Level: 8
-        Time: 26500
+        Time: 23500
       - Level: 9
-        Time: 29500
+        Time: 26500
       - Level: 10
-        Time: 32500
+        Time: 29500
```
Playtester, line 742: Pretty sure the duration here should be 3s*level just like pre-renewal and this was just a measuring error on my part due to latency.

db/re/skill_db.yml
```diff
@@ -2732,25 +2733,25 @@ Body:
...
+        Time: 58000
       - Level: 8
-        Time: 60000
+        Time: 58000
       - Level: 9
-        Time: 60000
+        Time: 58000
       - Level: 10
-        Time: 60000
+        Time: 58000
```
Playtester, line 2754: This probably should be tested because I doubt they changed the durations that were already custom in pre-renewal. I assume it should not be changed here.

db/re/skill_db.yml
```diff
@@ -3664,25 +3665,25 @@ Body:
...
+        Time: 10000
       - Level: 8
-        Time: 14500
+        Time: 11500
       - Level: 9
-        Time: 16000
+        Time: 13000
       - Level: 10
-        Time: 17500
+        Time: 14500
```
Playtester, line 3686: Pretty sure these should be the same as the pre-renewal duration (1500s*level).

db/re/skill_db.yml
```diff
@@ -4669,15 +4670,15 @@ Body:
...
+        Time: 5500
       - Level: 3
-        Time: 11500
+        Time: 8500
       - Level: 4
-        Time: 14500
+        Time: 11500
       - Level: 5
-        Time: 17500
+        Time: 14500
```
Playtester, line 4681: Probably should be 3s*level same as pre-re.

db/re/skill_db.yml
```diff
@@ -10511,15 +10510,15 @@ Body:
...
+        Time: 28000
       - Level: 3
-        Time: 40000
+        Time: 46000
       - Level: 4
-        Time: 50000
+        Time: 48000
       - Level: 5
-        Time: 60000
+        Time: 58000
```
Playtester, line 10521: Don't think this would be changed in renewal, might need a test.

db/re/skill_db.yml
```diff
@@ -15595,15 +15594,15 @@ Body:
...
+        Time: 9500
       - Level: 3
-        Time: 17500
+        Time: 14500
       - Level: 4
-        Time: 22500
+        Time: 19500
       - Level: 5
-        Time: 27500
+        Time: 24500
```
Playtester, line 15605: Should be 5s*level.

db/re/skill_db.yml
```diff
@@ -16338,7 +16336,7 @@ Body:
         Area: 11
       - Level: 5
         Area: 14
-    Duration2: 30000
+    Duration2: 28000
```
Playtester, line 16339: Don't have any data on confusion but the default duration is 18000.

db/re/skill_db.yml
```diff
@@ -17251,15 +17250,15 @@ Body:
...
+        Time: 14500
       - Level: 3
-        Time: 22500
+        Time: 19500
       - Level: 4
-        Time: 27500
+        Time: 24500
       - Level: 5
-        Time: 32500
+        Time: 29500
```
Playtester, line 17261: Probably should be 500ms higher.

Playtester (commented):
Browsed a bit through the duration changes and commented suspicious values.

db/re/skill_db.yml
```diff
@@ -15594,15 +15594,15 @@ Body:
     Duration1: 100
     Duration2:
       - Level: 1
-        Time: 4500
+        Time: 5000
       - Level: 2
-        Time: 9500
+        Time: 1000
```
Playtester, line 15599: Should be 10000.

db/re/skill_db.yml
```diff
@@ -721,27 +721,27 @@ Body:
...
+        Time: 24000
       - Level: 9
-        Time: 26500
+        Time: 27000
       - Level: 10
-        Time: 29500
+        Time: 30000
       - Level: 11
-        Time: 32500
+        Time: 33000
```
Playtester, line 744: Should be 30000. (This is special boss skill marker for us but it should have the same time as level 10.)

Playtester (changes requested):
Two incorrect durations sneaked in. Rest looks good now.

db/re/skill_db.yml
```diff
@@ -5948,7 +5947,7 @@ Body:
     Hit: Single
     HitCount: 1
     Element: Dark
-    Duration2: 30000
+    Duration2: 18000
```
Playtester, line 5950: Can you test NPC_CURSEATTACK please? My last tests showed the duration was 30s unlike all other curse skills which were 20s.

Based on that test the value here should be 28000 (or maybe 30000?).

Also, since it then can't be used as default duration, you need to change status.yml as well. Use NPC_WIDECURSE for the default duration instead. We already did the same for freeze which references NPC_WIDEFREEZE, so it would be consistent.

src/map/status.cpp
```diff
@@ -8742,71 +8746,104 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
...
+#endif
 			break;
 		case SC_CURSE:
+#ifndef RENEWAL
 			// Special property: immunity when luk is zero
 			if (status->luk == 0)
 				return 0;
```
Playtester, line 8818: Immunity at 0 Luk is actually also in the renewal code, so should apply to both.

src/map/status.cpp
```diff
@@ -8838,18 +8875,21 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
...
 			break;
 		case SC_WHITEIMPRISON:
-			if( tick == 5000 ) // 100% on caster
+			if( src == bl ) // 100% on caster
 				break;
-			if( bl->type == BL_PC )
-				tick_def2 = status_get_lv(bl)*20 + status->vit*25 + status->agi*10;
-			else
-				tick_def2 = (status->vit + status->luk)*50;
+			sc_def = status->str * 10 + status_get_lv(bl) * 10 + status->luk * 10;
```
Playtester, line 8880: iRO wiki confirm that STR indeed resists this status change for players, but claims that for monsters it's still Vit and Luk. Should we check that? Could be that iROwiki is wrong, though.

Playtester (commented):
Looks pretty good now. I have one minor doubt about Imprison after checking Irowiki.
Also curse is missing its Luk=0 resist.
Maybe other resists are missing too?

src/map/status.cpp
```diff
@@ -8838,18 +8875,24 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
...
-				tick_def2 = status_get_lv(bl)*20 + status->vit*25 + status->agi*10;
+			if (bl->type == BL_PC)
+				sc_def = status->str * 10 + status_get_lv(bl) * 10 + status->luk * 10;
 			else
-				tick_def2 = (status->vit + status->luk)*50;
+				tick_def2 = (status->vit + status->luk) * 50;
+			tick_def2 = -2000;
+			break;
+		case SC_FEAR:
+			sc_def = status->int_ * 10 + status_get_lv(bl) * 10 + status->luk * 10;
```
Playtester, line 8887: Should be: `[INT*20] + [BASELEVEL*20] + [LUK*10]`

src/map/status.cpp
```diff
@@ -8838,18 +8875,24 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
...
+			tick_def2 = -4000; // 2 seconds is applied twice on Aegis
 			break;
 		case SC_BURNING:
-			tick_def2 = 75*status->luk + 125*status->agi;
+			sc_def = status->agi * 10 + status_get_lv(bl) * 10 + status->luk * 10;
```
Playtester, line 8891: Should be: `[AGI*20] + [BASELEVEL*20] + [LUK*10]`

Playtester (changes requested):
Sorry made a small mistake in the code interpretation before.

src/map/status.cpp
```diff
@@ -8838,18 +8875,24 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
 			tick_def2 = (status->vit + status->dex) * 50;
 			break;
 		case SC_WHITEIMPRISON:
-			if( tick == 5000 ) // 100% on caster
+			if( src == bl ) // 100% on caster
 				break;
-			if( bl->type == BL_PC )
-				tick_def2 = status_get_lv(bl)*20 + status->vit*25 + status->agi*10;
+			if (bl->type == BL_PC)
+				sc_def = status->str * 10 + status_get_lv(bl) * 10 + status->luk * 10;
```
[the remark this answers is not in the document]
Playtester: Based on our tests just now, this formula also applies to monsters!

db/re/skill_db.yml
```diff
@@ -18706,7 +18704,7 @@ Body:
         Area: 6
       - Level: 5
         Area: 7
-    Duration1: 15000
+    Duration1: 18000
```
Playtester, line 18707: Note: For all other skills you assumed the current value on rA is already correc,t but for this one you decided to change it to its default value. This is just a hint as I don't know much about renewal skills or where their durations come from.

(You also did it for Whiteimprison but our tests clearly showed that the base duration is 18+2, so that's correct.)

db/re/skill_db.yml
```diff
@@ -27266,6 +27264,7 @@ Body:
       - Level: 5
         Time: 6000
     AfterCastActDelay: 1000
+    Duration: 18000
```
Playtester, line 27267: Github glitches out for me at the end of the file so I can't see the context, but is this really "Duration" and not "Duration1" or "Duration2"?

db/re/skill_db.yml
```diff
@@ -42856,15 +42854,15 @@ Body:
...
+        Time: 5500
       - Level: 3
-        Time: 11500
+        Time: 8500
       - Level: 4
-        Time: 14500
+        Time: 11500
       - Level: 5
-        Time: 17500
+        Time: 14500
```
Playtester, line 42865: This looks like it should be 3s*level, but I'm just guessing (also you substraced 2000ms for level 1, but 3000ms for all other levels.

db/re/skill_db.yml
```diff
@@ -23717,11 +23715,11 @@ Body:
...
     Duration2:
       - Level: 1
-        Time: 4000
+        Time: 2000
       - Level: 2
-        Time: 8000
+        Time: 6000
       - Level: 3
-        Time: 12000
+        Time: 10000
```
Playtester, line 23722: I wouldn't be surprised if this was actually 4s*level + fixed duration. But again just guessing.

Playtester (commented):
I cannot really review much of the renewal skills in skill.yml as I don't know much about them and Github glitches out for me after line 24740 so I'm unable to see the context (like status change or skill name), but I marked some things I found at least strange.

I assume many of these durations have not actually been tested yet and were just copied from iROwiki. At least it looks like it.

By the way, everything else looks good for me now, so if you think you're done I'll approve it.

db/re/skill_db.yml
```diff
@@ -43509,25 +43507,25 @@ Body:
...
+        Time: 58000
       - Level: 8
-        Time: 60000
+        Time: 58000
       - Level: 9
-        Time: 60000
+        Time: 58000
       - Level: 10
-        Time: 60000
+        Time: 58000
```
Playtester, line 43528: Probably should remain unchanged (it's not using the default duration), same as Lex Divina.

db/re/skill_db.yml
```diff
@@ -42803,15 +42801,15 @@ Body:
...
+        Time: 12000
       - Level: 3
-        Time: 16000
+        Time: 14000
       - Level: 4
-        Time: 18000
+        Time: 16000
       - Level: 5
-        Time: 20000
+        Time: 18000
```
Playtester, line 42812: It's interesting that this has a different duration per level while the original Sandman trap does not.

db/re/skill_db.yml
```diff
@@ -21913,7 +21911,7 @@ Body:
       Skill:
         Reproduce: true
     Duration1: 15000
-    Duration2: 15000
+    Duration2: 13000
```
Playtester, line 21914: Maybe suspicious.

db/re/skill_db.yml
```diff
@@ -19423,7 +19421,7 @@ Body:
     CastCancel: true
     CastTime: 3000
     AfterCastActDelay: 500
-    Duration1: 15000
+    Duration1: 13000
```
Playtester, line 19424: Maybe suspicious.

db/re/skill_db.yml
```diff
@@ -18659,7 +18657,7 @@ Body:
       - Level: 10
         Time: 2000
     AfterCastActDelay: 2000
-    Duration1: 10000
+    Duration1: 8000
```
Playtester, line 18660: Maybe suspicious.

db/re/skill_db.yml
aleos89: Confirmed that it is indeed 15 seconds.

db/re/skill_db.yml
aleos89: Confirmed.

db/re/skill_db.yml
aleos89: I confirmed it to be 13 seconds on Aegis.

db/re/skill_db.yml
aleos89: Cleaned up as Hell Inferno no longer inflicts Burning after the Warlock Improvement patch.

db/re/skill_db.yml
aleos89: I don't have the 100 loyalty points to test this (You need it in order to get the level 8 Bowman Mercenary which has the Sandman skill). We've had this status duration for years so I doubt it's changed if that's what was leaked for it before.

Playtester: Probably! Because I couldn't test Mercenaries either, I didn't touch their skills. But  maybe we should just make them the same as the original skills.

Playtester (approved):
I recommend still adjusting Mercenary Sandman trap to be the same as HT_SANDMAN, but since we can't test it, I'll approve anyway.
I'll rely on you on the renewal durations (I assume you consider the additional fixed duration in your tests).

PR #6717, opened 2022-03-18
Fixes reflected damage from overflowing

aleos89:
* **Addressed Issue(s)**: Fixes #6693

* **Server Mode**: Pre-renewal and Renewal

* **Description of Pull Request**: 
  * Reflected damage should never be negative.
  * Reflect Damage Reduction should only reduce in normal cases (outside of WoE/BG).
  * Adjusts the variable names to better match their use.
  * General cleanups to speed up certain scenarios.
Thanks to @kaninhot004!

TODO:
- [x] Confirm if Reflect Damage Reduction reduces reflected damage to 0 or 1.

PR #6767, opened 2022-03-31
Fixed some LGTM warnings

Lemongrass3110:

* **Server Mode**: Both

Fixed some LGTM warnings


src/map/battle.cpp
```diff
@@ -8348,7 +8348,7 @@ int64 battle_calc_return_damage(struct block_list* bl, struct block_list *src, i
 			rdamage = damage * sc->data[SC_MAXPAIN]->val1 * 10 / 100;
 	}
 
-	return cap_value(min(rdamage,max_damage),INT_MIN,INT_MAX);
+	return cap_value(i64min(rdamage,max_damage),INT_MIN,INT_MAX);
```
CairoLee, line 8351: Do we need to adjust INT_MIN and INT_MAX to INT64_MIN and INT64_MAX here?

Pokye:
Need add (int) too in skill.cpp
https://github.com/rathena/rathena/pull/6769/files#diff-35e92cf6786714f700005941b58d63e942fb78afd9330d2027cf30e7c5c3dedaL18460

src/map/battle.cpp
aleos89: So I guess the difference here is that in renewal the reflected damage is capped to the target's max HP (which is INT_MAX) while in pre-renewal that isn't taken into account. It's also something I need to fix in #6717.

Issue #6681, opened 2022-03-14
issue main.sql table skill type column lv


zdurexz:
* **rAthena Hash**: 

* **Client Date**: 20211117

* **Server Mode**: Both

* **Description of Issue**: 
  * Result:  
![image](https://user-images.githubusercontent.com/59825922/158214609-a7050c17-0ba4-4a6c-846e-410990917a2f.png)

  * Expected Result: type tinyint(4) to int(11)
  * How to Reproduce: type tinyint(4) to int(11)
  * Official Information: 
  

* **Modifications that may affect results**:

zdurexz:
Rogue's use skill snatch

Lemongrass3110:
Yeah but then increasing from tinyint to int does not help.
The problem is the signedness.
But I always thought that those rogue skills were stored in player variables.

Lemongrass3110:
@zdurexz did you find the problem?

zdurexz:
@Lemongrass3110 sql error skill lost

PR #6771, opened 2022-03-31
Fixed PLAGIARIZED skill would cause skill data loss

CairoLee:
* **Addressed Issue(s)**: #6681

* **Server Mode**: Both

  - This problem is caused by shadowcaster's PLAGIARIZED skill.
  - The flag of the copied skill is SKILL_FLAG_PLAGIARIZED with a value of 2, but before it is written to the database rAthena calculates the skill level by taking the flag - SKILL_FLAG_REPLACED_LV_0 (with a value of 10)
This makes the copied skill level -8, and the field in the database can't hold negative numbers, so it thinks it's out of range and reports an error.

* **中文说明 (Description of Chinese)**: 

  - 这应该是一个非常久远的问题
  - 流氓系列职业使用抄袭技能，抄袭到一个技能后的处理存在问题（技能等级变成 -8）
  - 数据库中的字段定义是无符号的，因此技能等级不能是也不应该是负数
  - 在这种情况下，角色的技能数据写入将会因为 -8 而导致 SQL 报错，最终导致全部技能数据丢失

### DIFF
--- a/src/char/char.cpp
+++ b/src/char/char.cpp
@@ -433,7 +433,7 @@ int char_mmo_char_tosql(uint32 char_id, struct mmo_charstatus* p){
 
 				if( p->skill[i].lv == 0 && ( p->skill[i].flag == SKILL_FLAG_PERM_GRANTED || p->skill[i].flag == SKILL_FLAG_PERMANENT ) )
 					continue;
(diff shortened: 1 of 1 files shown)


alycecil (approved):
Nice catch, thank you.

LGTM

PR #6793, opened 2022-04-04
Fixes statuses with SCB_ALL not recalculating

### LINKED ISSUE #6784: Status_db.yml : SCB_ALL for SC does not work
* **rAthena Hash**:  master

* **Client Date**: 20211117

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: SCB_ALL for SC does not work. Check below:
    Frostweapon
    For example I got SC_WATERWEAPON of SA_FROSTWEAPON, which increases 5% of Water element magic damage, and has no 
    effect.
    Frostweapon2
    The SC finally works after relogin.

  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
  

aleos89:
* **Addressed Issue(s)**: Fixes #6784

  * Adds a missing call for statuses that give the "All" CalcFlags.
Thanks to @eppc0330!

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -12099,8 +12099,21 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 	if (sc_isnew && scdb->state.any() && type != SC_STONE)
 		status_calc_state(bl, sc, scdb->state, true);
 
-	if (sd && sd->pd)
-		pet_sc_check(sd, type); // Skotlex: Pet Status Effect Healing
+	if (sd) {
+		if (sd->pd)
+			pet_sc_check(sd, type); // Skotlex: Pet Status Effect Healing
+		switch (type) {
+			case SC_BERSERK:
+			case SC_MERC_HPUP:
+			case SC_MERC_SPUP:
+				status_calc_pc(sd, SCO_FORCE);
+				break;
+			default:
+				if (!sd->state.connect_new)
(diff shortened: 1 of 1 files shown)


CairoLee:
The status recalculate from this commit breaks autospell's ability to prevent duplicate executions.

autospell relies on the `lock` field of the `s_autospell` structure to prevent duplicate executions, but this commit trigger the status recalculate will force clear `sd->autospell\sd->autospell2\sd->autospell3`, that will make the `lock` field reset to false.

Here is a reproduce step to make the server entering the infinite loop, eventually crashes due to stack overflow.

## Reproduce

- modify the `skill_db.yml` to make sure `SM_MAGNUM`'s cooldown is 0
- modify the `item_db_equip.yml` and give the equip 1201 a script like: `bonus5 bAutoSpellOnSkill,7,7,5,1000,0;`
- create a swordman charactor and use the `Magnum Break` skill
- the map-server will crash immediately

CairoLee:
Another example is that using the `sc_start` script command in a `bonus script` with a 100% chance of triggering the `autobonus` script command will also cause an infinite loop and eventually crash

aleos89:
If you could create a separate issue for this I would appreciate it since this PR is already merged to master. Thanks!

CairoLee:
> If you could create a separate issue for this I would appreciate it since this PR is already merged to master. Thanks!

Okay, here a new issue:
https://github.com/rathena/rathena/issues/6930

Issue #6573, opened 2022-01-28
Skill Item cost perlevel doesn't work properly


dimasshotta:
rAthena Hash: 3fb8268
Client Date: 2021-11-17
Server Mode: Re
Description of Issue: The skill that needs item perlevel is not working. It doesn't read the itemcost per level. But the itemcost working if there is no level requirement in the itemcost.

Result:
Expected Result: Supposedly the skilldb reads the itemcost per skill level
![image](https://user-images.githubusercontent.com/74591268/151537317-fd2570ba-2d5b-4ba3-a4f2-5f74000ddc44.png)
here i dont have gemstone as the requirement to cast lv 10 fire pillar.
but its still working without the blue gemstn.
![image](https://user-images.githubusercontent.com/74591268/151537614-d31a7096-e65d-41d9-bdba-deb5459e2545.png)
Please help

mazvi:
Please provide rAthena Hash

Sinnerseiei:
comfirm

Daraen1:
I can confirm the behavior.
Is there a fix planned in the coming days?
If not, I'll take a look at it when I have a bit more time.

CairoLee:
Tomorrow I will submit a PR to fix it

Daraen1:
Allright! Waiting for your PR ;)

PR #6796, opened 2022-04-06
Fixed ItemCost is invalid when specified level

CairoLee:
* **Addressed Issue(s)**: #6573 

* **Server Mode**: Both

  - Thanks for @dimasshotta to report this

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -18224,6 +18224,13 @@ struct s_skill_condition skill_get_requirement(struct map_session_data* sd, uint
 						break;
 					}
 				}
+				else {
+					// Process level_dependent requirement
+					if (level_dependent && skill_lv <= MAX_SKILL_ITEM_REQUIRE) {
+						req.itemid[0] = skill->require.itemid[skill_lv - 1];
+						req.amount[0] = skill->require.amount[skill_lv - 1];
+					}
+				}
 
 				// Check requirement for gemstone.
 				if (itemdb_group.item_exists(IG_GEMSTONE, req.itemid[i])) {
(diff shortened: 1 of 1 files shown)


Daraen1 (approved):

Daraen1:
I just tested the PR and now fire pillar is working fine, thanks @CairoLee

src/map/skill.cpp
```diff
@@ -18243,6 +18243,12 @@ struct s_skill_condition skill_get_requirement(struct map_session_data* sd, uint
 				if (req.itemid[i] == ITEMID_MAGIC_GEAR_FUEL && sd->special_state.no_mado_fuel)
 					req.itemid[i] = req.amount[i] = 0;
 			}
+
+			// Process level_dependent requirement
+			if (level_dependent && skill_lv <= MAX_SKILL_ITEM_REQUIRE) {
+				req.itemid[0] = skill->require.itemid[skill_lv - 1];
+				req.amount[0] = skill->require.amount[skill_lv - 1];
+			}
```
aleos89, line 18251: I agree with the fix, but there's just one issue with it's location. GitHub won't let me suggest the location of the fix, but it needs to be moved up because where you have it now, you're overwriting a few checks that remove item requirements through skill or item bonuses.

Suggested fix:
image

aleos89 (changes requested):

src/map/skill.cpp
CairoLee: okay, thanks for you review.
I will push a new commit to adjust it after few mins.

aleos89 (approved):

Lemongrass3110 (approved):

PR #6807, opened 2022-04-08
Updates stacking for common statuses

### LINKED ISSUE #6798: Information: Status change fail/end lists
* **rAthena Hash**: a7ee51d

Creating this issue to give us an overview on the fail/end conditions that can be drawn from the sources we currently have. This list may not be 100% complete as "end" or "fail" condition can also be part of a specific skill's code, but if anything shows in the list below, it should definitely be in the fail/end.

Every player can only have one "Opt1" status change, so one will overwrite another if it isn't blocked. 
**Opt1:** Whiteimprison, Stone, Stonewait, Freeze, Stun, Sleep, Burning, Undead

But if we define the status fail/end conditions as below we don't actually need any special handling in the code (note: Undead Opt1 is not implemented in rAthena).

**Opt1 (Bodystate);**

**Stone/Stonewait:**
Fail: Whiteimprison, Freeze, Stun, Sleep, Burning, Undead
End: Stone, Stonewait
EndReturn: true

**Freeze:**
Fail: Whiteimprison, Stone, Stonewait, Freeze, Stun, Sleep, Burning, Undead, EFST_WARMER
End: -

**Stun:**
Fail: Stun
End: Whiteimprison, Stone, Stonewait, Freeze, Sleep, Burning, Undead

**Sleep:**
Fail: Sleep
End: Whiteimprison, Stone, Stonewait, Freeze, Stun, Burning, Undead

**Burning:**
Fail: Whiteimprison
End: Stone, Stonewait, Freeze, Stun, Sleep, Undead

**Whiteimprison:**
Fail: -
End: Stone, Stonewait, Freeze, Stun, Sleep, Burning, Undead

**Opt2 (Healthstate):**

**Poison:**
Fail: Poison, Dpoison
End: -

**Curse:**
Fail: Curse
End: -

**Silence:**
Fail: Silence
End: -

**Confusion:**
Fail: -
End: Confusion
EndReturn: true

**Blind:**
Fail: Blind, Fear
End: -

**Dpoison:**
Fail: -
End: -

**Bleeding:**
Fail: -
End: -

**Fear:**
Fail: -
End: Blind

aleos89:
* **Addressed Issue(s)**: Fixes #6798

  * Updates the Fail, End, and EndReturn lists for OPT1 and OPT2 statuses.
  * Removes the hardcoded OPT1 overwrite prevention check.
Thanks to @Playtester!

### DIFF
--- a/db/pre-re/status.yml
+++ b/db/pre-re/status.yml
@@ -214,7 +214,6 @@ Body:
     Flags:
       BossResist: true
       StopWalking: true
-      RemoveOnDamaged: true
       OverlapFail: true
     Fail:
       Refresh: true
--- a/db/pre-re/status.yml
+++ b/db/pre-re/status.yml
@@ -70,7 +70,6 @@ Body:
       Stun: true
       Sleep: true
       Burning: true
-      #Undead: true
     End:
       Aeterna: true
     EndReturn:
@@ -89,7 +88,6 @@ Body:
       Stun: true
       Sleep: true
       Burning: true
-      #Undead: true
     EndReturn:
       StoneWait: true
       Stone: true
@@ -121,9 +119,7 @@ Body:
       Stun: true
       Sleep: true
       Burning: true
-      #Undead: true
     End:
-      Dancing: true
       Aeterna: true
   - Status: Stun
     DurationLookup: NPC_STUNATTACK
@@ -142,14 +138,13 @@ Body:
       Refresh: true
       Inspiration: true
       Gvg_Stun: true
+      Stun: true
     End:
-      Dancing: true
       Stone: true
       StoneWait: true
       Freeze: true
       Sleep: true
       Burning: true
-      #Undead: true
   - Status: Sleep
     DurationLookup: NPC_SLEEPATTACK
     States:
@@ -168,6 +163,7 @@ Body:
       Refresh: true
       Inspiration: true
       Gvg_Sleep: true
+      Sleep: true
     End:
       Dancing: true
       Stone: true
@@ -175,7 +171,6 @@ Body:
       Freeze: true
       Stun: true
       Burning: true
-      #Undead: true
   - Status: Poison
     DurationLookup: NPC_POISON
     CalcFlags:
@@ -2896,7 +2891,6 @@ Body:
       Freeze: true
       Stun: true
       Sleep: true
-      #Undead: true
   - Status: Freezing
     Icon: EFST_FROSTMISTY
     DurationLookup: WL_FROSTMISTY
@@ -3112,11 +3106,12 @@ Body:
       StopAttacking: true
       OverlapFail: true
     End:
-      Dancing: true
       Burning: true
       Freezing: true
       Freeze: true
       Stone: true
+      StoneWait: true
+      Sleep: true
   - Status: Marshofabyss
     Icon: EFST_MARSHOFABYSS
     DurationLookup: WL_MARSHOFABYSS
--- a/db/pre-re/status.yml
+++ b/db/pre-re/status.yml
@@ -165,7 +165,6 @@ Body:
       Gvg_Sleep: true
       Sleep: true
     End:
-      Dancing: true
       Stone: true
       StoneWait: true
       Freeze: true
--- a/db/pre-re/status.yml
+++ b/db/pre-re/status.yml
@@ -115,6 +115,13 @@ Body:
       Inspiration: true
       Warmer: true
       Gvg_Freez: true
+      Stone: true
+      StoneWait: true
+      Freeze: true
+      Stun: true
+      Sleep: true
+      Burning: true
+      #Undead: true
     End:
       Dancing: true
       Aeterna: true
@@ -137,6 +144,12 @@ Body:
       Gvg_Stun: true
     End:
       Dancing: true
+      Stone: true
+      StoneWait: true
+      Freeze: true
+      Sleep: true
+      Burning: true
+      #Undead: true
   - Status: Sleep
     DurationLookup: NPC_SLEEPATTACK
     States:
@@ -157,6 +170,12 @@ Body:
       Gvg_Sleep: true
     End:
       Dancing: true
+      Stone: true
+      StoneWait: true
+      Freeze: true
+      Stun: true
+      Burning: true
+      #Undead: true
   - Status: Poison
     DurationLookup: NPC_POISON
     CalcFlags:
@@ -172,6 +191,8 @@ Body:
     Fail:
       Refresh: true
       Inspiration: true
+      Poison: true
+      Dpoison: true
   - Status: Curse
     DurationLookup: NPC_CURSEATTACK
     CalcFlags:
@@ -189,6 +210,7 @@ Body:
       Refresh: true
       Inspiration: true
       Gvg_Curse: true
+      Curse: true
   - Status: Silence
     DurationLookup: NPC_SILENCEATTACK
     States:
@@ -204,6 +226,7 @@ Body:
       Refresh: true
       Inspiration: true
       Gvg_Silence: true
+      Silence: true
   - Status: Confusion
     DurationLookup: NPC_WIDECONFUSE
     Flags:
@@ -214,6 +237,7 @@ Body:
     Fail:
       Refresh: true
       Inspiration: true
+    EndReturn:
       Confusion: true
   - Status: Blind
     DurationLookup: NPC_BLINDATTACK
@@ -231,6 +255,7 @@ Body:
       Inspiration: true
       Fear: true
       Gvg_Blind: true
+      Blind: true
   - Status: Bleeding
     Icon: EFST_BLOODING
     DurationLookup: NPC_BLEEDING
@@ -415,6 +440,8 @@ Body:
     Flags:
       BossResist: true
       TaekwonAngel: true
+    EndReturn:
+      Curse: true
   - Status: Signumcrucis
     Icon: EFST_CRUCIS
     DurationLookup: AL_CRUCIS
@@ -2849,6 +2876,8 @@ Body:
   - Status: Burning
     Icon: EFST_BURNT
     DurationLookup: RK_DRAGONBREATH
+    CalcFlags:
+      Mdef: true
     Opt1: Burning
     Flags:
       SendOption: true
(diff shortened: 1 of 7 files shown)


db/pre-re/status.yml
```diff
@@ -115,6 +115,13 @@ Body:
...
       Gvg_Freez: true
+      Stone: true
+      StoneWait: true
+      Freeze: true
+      Stun: true
+      Sleep: true
+      Burning: true
+      #Undead: true
     End:
       Dancing: true
```
Playtester, line 126: Freeze and Stun make you stop only when you are singing/dancing, similar to Stone. Do we need to apply the special handling for Stone also for these?

And I never tested Whiteimprison in that regard.

db/pre-re/status.yml
```diff
@@ -137,6 +144,12 @@ Body:
       Gvg_Stun: true
     End:
       Dancing: true
+      Stone: true
```
Playtester, line 147: Any OPT1 status change should end any other OPT1 status change (unless it makes it fail). Do we still want to define "End" here?

db/pre-re/status.yml
```diff
@@ -115,6 +115,13 @@ Body:
       Inspiration: true
       Warmer: true
       Gvg_Freez: true
+      Stone: true
```
Playtester, line 118: Need to consider what Lemongrass said, that we could do something like "FailAllOpt1Except:" instead. (Or we do that later.)

db/pre-re/status.yml
```diff
@@ -137,6 +144,12 @@ Body:
       Gvg_Stun: true
```
Playtester, line 144: Stun should fail on Stun.

db/pre-re/status.yml
```diff
@@ -157,6 +170,12 @@ Body:
       Gvg_Sleep: true
```
Playtester, line 170: Sleep should fail on Sleep.

db/pre-re/status.yml
```diff
@@ -415,6 +440,8 @@ Body:
     Flags:
       BossResist: true
       TaekwonAngel: true
+    EndReturn:
+      Curse: true
```
Playtester, line 444: There must be some hardcoded code that removes Curse by Blessing, you can remove that code now. (There is probably more code that can be removed because of this update.)

db/re/status.yml
```diff
@@ -159,6 +175,13 @@ Body:
       Gvg_Sleep: true
```
Playtester, line 175: Sleep fails on Sleep.

db/re/status.yml
```diff
@@ -2976,6 +3007,13 @@ Body:
       Refresh: true
       Inspiration: true
       Whiteimprison: true
```
Playtester, line 3009: I don't see that you updated Whiteimprison, was that already correct?

Playtester (changes requested):
Commented all things I noticed.
Probably need to decide on how we want to do OPT1 ending and failing.

db/pre-re/status.yml
aleos89: For White Imprison, I confirmed that it doesn't stop my current walk, only the click-to-walk after White Imprison is active.

db/pre-re/status.yml
aleos89: Will resolve this in #6834.

db/pre-re/status.yml
```diff
@@ -3112,11 +3106,12 @@ Body:
       StopAttacking: true
       OverlapFail: true
     End:
-      Dancing: true
       Burning: true
       Freezing: true
       Freeze: true
       Stone: true
+      StoneWait: true
+      Sleep: true
```
Playtester, line 3114: I didn't know Whiteimprison is on pre-re, but since it is, shouldn't we also add it to the fail/end of other OPT1 skills? (This is mainly to make it easier for people who want to use this in pre-re.)
It's also still missing "Stun" on fail.

db/re/status.yml
```diff
@@ -3232,6 +3225,8 @@ Body:
       Freezing: true
       Freeze: true
       Stone: true
+      StoneWait: true
+      Sleep: true
```
Playtester, line 3229: Missing "Stun".

db/pre-re/status.yml
Playtester: Shouldn't you undo your change here then?

db/re/status.yml
```diff
@@ -157,8 +167,15 @@ Body:
       Refresh: true
       Inspiration: true
       Gvg_Sleep: true
+      Sleep: true
     End:
       Dancing: true
```
Playtester, line 172: Sleep should not end Dancing as far as I know.

db/pre-re/status.yml
```diff
@@ -155,8 +163,14 @@ Body:
       Refresh: true
       Inspiration: true
       Gvg_Sleep: true
+      Sleep: true
     End:
       Dancing: true
```
Playtester, line 168: Sleep should not end Dancing as far as I know.

db/re/status.yml
```diff
@@ -3227,6 +3224,8 @@ Body:
       Stone: true
       StoneWait: true
       Sleep: true
+      Stun: true
+      Sleep: true
```
Playtester, line 3228: Now sleep appears twice here.

src/map/status.cpp
```diff
@@ -9258,6 +9265,16 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 
 	// Check failing SCs from list
 	if (!scdb->fail.empty()) {
+		// Don't let OPT1 that have RemoveOnDamaged start a new effect
+		if (sc->opt1 > OPT1_NONE && scdb->flag[SCF_REMOVEONDAMAGED]) {
+			unit_data *ud = unit_bl2ud(bl);
+
+			if (ud != nullptr && ud->lastEffect == type) {
+				ud->lastEffect = SC_NONE;
+				return 0;
```
Playtester, line 9274: This code seems weird to me. You don't actually compare lastEffect against the fail list here. Also what does "RemoveOnDamaged" of the new SC have to do with the skill failing?
And "lastEffect" should also be reset if it does not lead to failing.

For example if you attack a frozen target with 100% sleep attack. Then sleep should still happen because freeze does not make sleep fail. But attacking a frozen target with a 100% petrify attack should just remove freeze but not petrify the target. Whiteimprison also should not block itself from being reapplied.

Basically you don't want to check for OPT1 or RemoveOnDamaged here at all, instead you want to put this logic inside the scdb->fail iteration and just check for "it == ud->lastEffect" (not sure about syntax) as fail condition.

db/pre-re/status.yml
```diff
@@ -214,6 +219,7 @@ Body:
     Fail:
       Refresh: true
       Inspiration: true
+    EndReturn:
```
Playtester, line 222: I just tested Confusion and it should not be removed on damage. Only causing Confusion again removes it.

src/map/status.cpp
```diff
@@ -1000,8 +1000,15 @@ int status_damage(struct block_list *src,struct block_list *target,int64 dhp, in
 			for (const auto &it : status_db) {
 				sc_type type = static_cast<sc_type>(it.first);
 
-				if (sc->data[type] && it.second->flag[SCF_REMOVEONDAMAGED])
+				if (sc->data[type] && it.second->flag[SCF_REMOVEONDAMAGED]) {
+					if (sc->opt1 > OPT1_NONE) {
+						unit_data *ud = unit_bl2ud(target);
```
Playtester, line 1005: Considering you have so much work with storing this in ud and then also loading it back (because in both places you can't access ud directly), wouldn't it be better to store it inside sc?

src/map/status.cpp
```diff
@@ -1000,8 +1000,15 @@ int status_damage(struct block_list *src,struct block_list *target,int64 dhp, in
...
+
+						if (ud != nullptr)
+							ud->lastEffect = type;
```
Playtester, line 1008: The name "lastEffect" is not very clear in what it does. We should at least put a comment like Skotlex did.
e.g.
// A status change that gets broken by damage should still be considered when calculating if a status change can be applied or not (for the same attack)
// This is a temporary solution until we refactor the code so that the calculation of an SC is done at the start of an attack rather than after the damage was applied

Playtester (commented):
Some additional fixes / improvements.

src/map/status.cpp
```diff
@@ -9265,18 +9263,14 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
...
-		// Don't let OPT1 that have RemoveOnDamaged start a new effect
-		if (sc->opt1 > OPT1_NONE && scdb->flag[SCF_REMOVEONDAMAGED]) {
-			unit_data *ud = unit_bl2ud(bl);
-
-			if (ud != nullptr && ud->lastEffect == type) {
-				ud->lastEffect = SC_NONE;
+		for (const auto &it : scdb->fail) {
+			// Don't let OPT1 that have RemoveOnDamaged start a new effect in the same attack.
+			if (sc->lastEffect == it) {
+				sc->lastEffect = SC_NONE;
```
Playtester, line 9269: Setting sc->lastEffect to SC_NONE should also happen if the status does not fail.

In fact it would be better if this is reset after all the status changes were applied (just in case there is a skill that applies two status changes, there's also certain cards that can apply multiple status changes).

If we move it elsewhere you can simplify the code here to:
```if (sc->data[it] || sc->lastEffect == it)
  return 0;```

File: src/map/status.cpp [hunk 19]
```diff
@@ -9266,10 +9268,10 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 		for (const auto &it : scdb->fail) {
 			// Don't let OPT1 that have RemoveOnDamaged start a new effect in the same attack.
 			if (sc->lastEffect == it) {
-				sc->lastEffect = SC_NONE;
+				if (sc->lastEffectTimer == INVALID_TIMER)
+					sc->lastEffectTimer = add_timer(gettick() + 10, status_clear_lastEffect_timer, bl->id, 0);
```
Playtester, line 9272: Wrong place to start the timer, you want to start the timer the moment you set lastEffect.

db/pre-re/status.yml
```diff
@@ -231,6 +236,7 @@ Body:
       Inspiration: true
       Fear: true
       Gvg_Blind: true
+      Blind: true
   - Status: Bleeding
     Icon: EFST_BLOODING
     DurationLookup: NPC_BLEEDING
```
Playtester, line 242: Currently Bleeding is blocking itself because of the "OverlapFail" flag. It should not. If get inflicted by Bleeding while you still have Bleeding, it will reset the timer.

Please also check Whiteimprison, Confusion, DPoison and Fear, those can also overlap!

Playtester (approved):
Fully tested in pre-renewal and working fine there now.

PR #6829, opened 2022-04-12
Minor Adjustment to Expanded Class skills

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Renewal

  * Kihop now applies physical damage bonus directly no longer depending on the number of party members. Physical damage bonus: (10 + (15 x skill level))%.
  * Catnip Meteor:
    - Reduces fixed casting time from 3 seconds to 1.5 seconds.
    - Reduces variable casting time from 4 seconds to 2 seconds.
    - From base level 100, deals additional damage depending on the user's base level and INT.
  * Lunatic Carrot Beat now deals additional damage (from base level 100) depending on user's base level and STR.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3715,15 +3715,19 @@ static void battle_calc_skill_base_damage(struct Damage* wd, struct block_list *
 				if(sd->bonus.crit_atk_rate && is_attack_critical(wd, src, target, skill_id, skill_lv, false)) { // add +crit damage bonuses here in pre-renewal mode [helvetica]
 					ATK_ADDRATE(wd->damage, wd->damage2, sd->bonus.crit_atk_rate);
 				}
-#endif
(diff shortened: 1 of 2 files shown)


src/map/battle.cpp
```diff
@@ -5003,7 +5007,8 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 			break;
 		case SU_LUNATICCARROTBEAT:
 		case SU_LUNATICCARROTBEAT2:
-			skillratio += 100 + 100 * skill_lv;
+			skillratio += 100 + 100 * skill_lv + sstatus->str / 6; // !TODO: What's the STR bonus?
```
attackjom, line 5010: Status x 5 is a magic number that they alway use. Maybe it better to use this formula?

Lemongrass3110 (approved):

PR #6845, opened 2022-04-17
Rename according to date upgrade_20220422.sql

mazvi:
Rename according to date upgrade_20220422.sql

* **Addressed Issue(s)**: 

* **Server Mode**: 


Everade:
The file was committed on the date it is currently set, it has just been merged later to master, which is to be expected.
Additionaly, the merge date was at the 13th of april, not 22nd.

There's really no need to change the file name each time we merge another upgrade.

Issue #6861, opened 2022-04-19
Incorrect probabilities in Random Option Groups


Toshiro90:
* **rAthena Hash**: 2042262

* **Client Date**: Not relevant

* **Server Mode**: Both

* **Description of Issue**: 
Since #5494 was merged, the probablities of applied random options are incorrect.
Using slot 2 of `MAGIC_CAT_HAND_SCROLL` random option group as an example for now (data source: https://ro.gnjoy.com/news/probability/View.asp?seq=3964879 ), but applies to other option groups as well.

In this example all options should have equal probability (`12,5% = 1/8`) to be applied. 

Option | %
-- | --:
힐량 5 ~ 15% 증가 | 12,50
받는 힐량 5 ~ 15% 증가 | 12,50
ATK + 1 ~ 7% | 12,50
MATK + 1 ~ 7% | 12,50
원거리 물리 데미지 1 ~ 7 % 증가 | 12,50
변동 캐스팅 시간 5 ~ 10% 감소 | 12,50
크리티컬 데미지 5 ~ 10% 증가 | 12,50
ASPD +1 | 12,50

However with the current implementation (since #5494) the probabilities any option other than the first one tried is incorrect.
https://github.com/rathena/rathena/blob/20422622aab62e34da2c19b03a6d8513f69abeab/src/map/itemdb.cpp#L3228-L3271
For simplicity assume `vector_random` yields option entries in linear order.

\# | Option | Expected % | Actual % | Fail % | -
-- | -- | --: | --: | --: | :-- 
1 | 힐량 5 ~ 15% 증가 | 12,50 | 12,50 | 87,50 | `P(O1) = p1`
2 | 받는 힐량 5 ~ 15% 증가 | 12,50 | 10,94 | 89,06 | `P(O2) = p2 * (1-p1)` 
3 | ATK + 1 ~ 7% | 12,50 | 9,57 | 90,43 | `P(O3) = p3 * (1-p1) * (1-p2)` 
4 | MATK + 1 ~ 7% | 12,50 | 9,37 | 91,63 | `P(O4) = p4 * (1-p1) * (1-p2) * (1-p3)` 
5 | 원거리 물리 데미지 1 ~ 7 % 증가 | 12,50 | 7,33 | 92,67 | `P(O5) = p5 * (1-p1) * (1-p2) * (1-p3) * (1-p4)`
[...]

PR #6864, opened 2022-04-21
Incorrect probabilities in Random Option Groups & Random Item Groups

Toshiro90:
* **Addressed Issue(s)**: #6861

* **Server Mode**: Both

WIP

### DIFF
--- a/src/map/itemdb.cpp
+++ b/src/map/itemdb.cpp
@@ -3247,22 +3247,26 @@ void s_random_opt_group::apply( struct item& item ){
 
 	// Apply Must options
 	for( size_t i = 0; i < this->slots.size(); i++ ){
-		// Try to apply an entry
-		for( size_t j = 0, max = this->slots[static_cast<uint16>(i)].size() * 3; j < max; j++ ){
-			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->slots[static_cast<uint16>(i)] );
 
-			if( rnd() % 10000 < option->chance ){
-				apply_sub( item.option[i], option );
-				break;
-			}
+		if (this->slots[static_cast<uint16>(i)]->data.size() == 0)
+		{
+			continue;
 		}
 
-		// If no entry was applied, assign one
-		if( item.option[i].id == 0 ){
-			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->slots[static_cast<uint16>(i)] );
+		if (this->slots[static_cast<uint16>(i)]->total_rate == 0)
+		{
+			continue;
+		}
 
-			// Apply an entry without checking the chance
-			apply_sub( item.option[i], option );
+		int rndVal = rnd() % this->slots[static_cast<uint16>(i)]->total_rate;
+
+		for (const auto& option : this->slots[static_cast<uint16>(i)]->data)
+		{
+			if (rndVal < option->chance) {
+				apply_sub(item.option[i], option);
+				break;
+			}
+			rndVal -= option->chance;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/map/itemdb.cpp
```diff
@@ -3234,38 +3247,30 @@ void s_random_opt_group::apply( struct item& item ){
...
-	if( this->max_random > 0 ){
-		for( size_t i = 0; i < min( this->max_random, MAX_ITEM_RDM_OPT ); i++ ){
-			// If item already has an option in this slot, skip it
-			if( item.option[i].id > 0 ){
-				continue;
-			}
+		if (this->slots[static_cast<uint16>(i)]->total_rate == 0)
+		{
+			continue;
+		}
```
Daegaladh, line 3263: Maybe combine the ifs and remove the unnecessary brackets

src/tool/csv2yaml.cpp
```diff
@@ -3107,20 +3112,21 @@ static bool itemdb_read_randomopt_group(char* str[], int columns, int current) {
...
 		entry->max_value = 0;
 		entry->param = (int8)strtoul(str[k + 2], nullptr, 10);
-		entry->chance = 0;
+		entry->rate = 1;
 		entries.push_back(entry);
-		if (group == nullptr)
-			group_entry.slots[j] = entries;
+		if (group == nullptr) {
+			group_entry.slots[j]->data = entries;
+		}
```
Daegaladh, line 3127: Why did you add the brackets?

src/map/itemdb.cpp
```diff
@@ -3320,58 +3325,62 @@ uint64 RandomOptionGroupDatabase::parseBodyNode(const ryml::NodeRef& node) {
...
+				random->data = inheritRandom->data;
 			}
 
-			randopt->slots[slot - 1] = entries;
-		}
-	}
-
-	if (this->nodeExists(node, "MaxRandom")) {
-		uint16 max;
+			if (!this->nodeExists(slotNode, "Options") && !inheritSlot) {
```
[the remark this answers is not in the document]
Toshiro90: This is intentional to allow not only using inherited options for the specified slot, but also adding more options.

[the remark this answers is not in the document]
Daegaladh:
```suggestion
			else if (!this->nodeExists(slotNode, "Options")) {
```
Then this should do the same

src/tool/csv2yaml.cpp
```diff
@@ -3082,12 +3082,25 @@ static bool itemdb_read_randomopt_group(char* str[], int columns, int current) {
 		return false;
 	}
 
-	uint16 id = static_cast<uint16>(rand_opt_group.size() + 1);
-	s_random_opt_group_csv *group = util::umap_find(rand_opt_group, id);
-	s_random_opt_group_csv group_entry;
+	s_random_opt_group group_entry;
```
aleos89, line 3085: Just to note, this is the wrong way to go about this fix here. The csv2yaml should not be touched, but instead we use the yamlupgrade when drastic changes are needed to be made to the YAML structure. This way the content does not have to be converted from TXT to YAML again, simply just running yamlupgrade will convert the current YAML to the new YAML structure.

Toshiro90: Will look into yamlupgrade.
But shouldn't csv2yaml provide yaml files with the current format, instead of having to additionally run yamlupgrade (when not already having the yaml version)?

aleos89: Sorry, thought I replied before.

So let me take back some of what I stated before. To your question, yes we still want to maintain csv2yaml with the "latest" format but since some databases are conversions from csv and then we add features that weren't part of it, updating csv2yaml is moot since those pieces weren't in the old format.

In your case here you're adjusting prior content so it's ok!

eppc0330:
Is this pull completed?

PR #6865, opened 2022-04-21
Removes SCF_OVERLAPFAIL in favor of Fail list

aleos89:
* **Addressed Issue(s)**: N/A

  * Removes the SCF_OVERLAPFAIL status flag in favor of adding the status to the Fail list.
  * Burning is now able to overlap itself (per #6798).


Playtester (commented):
Frostmisty (Freezing) should probably be able to overwrite itself but currently doesn't. Wasn't it so that the whole skill fails due to the target being water element? Frostmisty also lists "Warmer" as fail condition but I only see that it makes "Freeze" fail, not Frostmisty. Maybe it's rather that if "Freeze" fails to apply then Frostmisty should not even be attempted. You could try vs. high MDEF for example.

Behavior of burning was also changed as part of this commit, can you add at least the information to the PR and link to the orignal issue #6798 ?

Can't see any other issues, but did not test ingame.

PR #6867, opened 2022-04-21
Updates SC_CHANGEUNDEAD behavior

### LINKED ISSUE #6834: Blessing (and Agi Up) behaviors vs SC_CHANGEUNDEAD
* **rAthena Hash**: [f083caf](https://github.com/rathena/rathena/commit/f083caf28d5184f4ca4f5dbad968fb92d6a2f0f0)

* **Client Date**: 2021-11-03

* **Server Mode**: Both

* **Description of Issue**: @aleos89 and me tested around with the skills of official servers, especially Blessing, here are the full results:

Blessing works completely different vs. players and vs. monsters (no shared code):

**vs. Player:**

1. If the player has the the status change SC_CHANGEUNDEAD, then for both Agi and Blessing you still see the skill animation of Agi and Blessing (currently not working on rAthena) and additionally you see the SM_BASH animation with 1 HP damage. You don't get any buffs. As it uses SM_BASH / level 1 as skill_id / skill_lv, you cannot plagiarize Agi and Blessing with it (probably not working?). Also, if you use Agi or Blessing skill while you have 1 HP, it will simply do nothing. You cannot die from it (not working on rAthena).

2. If the player has the status change "Curse" or "Stone" (second phase), then Blessing will remove one of these status changes (it will remove Curse first, Stone is only removed if the target is not cursed - not working on rAthena) and do nothing else (probably not working?). Only when the target is neither cursed nor petrified, the buff effect is applied (probably not working?).

3. Shadow or Undead armor have no impact on Blessing or Agi at all.

**vs. Monster:**

1. If not "Undead element" and not "Demon race" -> Apply normal buff.

2. If buff not applied and not "Debuff immune" (boss?) -> Halves Str, Int and Dex (the decrease is rounded down, so 31 INT becomes 16 INT).

3. No checks for any status changes. No removal either.

aleos89:
* **Addressed Issue(s)**: Fixes #6834

  * Versus Players
    - Animation will be properly displayed for Blessing/Increase Agility when the target has Change Undead active (buffs are not applied even though animation is displayed).
    - Target can no longer be killed through the single damage applied by Blessing/Increase Agility and Change Undead.
    - If the target has Curse and Stone active, only Curse is removed by Blessing first (buffs are not applied).
    - Shadow or Undead armor have no impact on Blessing or Increase Agility at all.
  * Versus Monsters
    - Blessing is applied normally to the target as long as it's not an Undead element or Demon race.
    - Blessing does not cancel out Curse or Stone.

Thanks to @Playtester!


src/map/status.cpp
```diff
@@ -1064,6 +1064,12 @@ int status_damage(struct block_list *src,struct block_list *target,int64 dhp, in
 		unit_stop_walking( target, 1 );
 	}
 
+	if (status->hp == 0 &&
+		((skill_id == AL_INCAGI || skill_id == AL_BLESSING ||
+		  skill_id == CASH_BLESSING || skill_id == CASH_INCAGI ||
+		  skill_id == MER_INCAGI || skill_id == MER_BLESSING) && sc && sc->data[SC_CHANGEUNDEAD]))
+		status->hp = 1; // Can't be killed
```
Playtester, line 1071: On official the whole skill does nothing when you are at 1 HP. In this version it doesn't kill you but still shows the skill animation and the damage.

So basically instead of calling skill_attack you should just do "break;" in that case rather than adding this code.

src/map/status.cpp
```diff
@@ -10501,7 +10507,7 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			val3 = 5*val1; // SP cost
 			break;
 		case SC_BLESSING:
-			if ((!undead_flag && status->race!=RC_DEMON) || bl->type == BL_PC)
+			if (bl->type == BL_PC || (!undead_flag && status->race != RC_DEMON))
 				val2 = val1;
 			else
 				val2 = 0; // 0 -> Half stat.
```
Playtester, line 10513: Note: Currently the stats are halved and then rounded down, but actually the decrease should be rounded down.

Basically:
stat -= stat/2;

15 dex should be become 8 dex.

Playtester (changes requested):
Some minor improvements requested.

src/map/skill.cpp
```diff
@@ -3870,6 +3865,21 @@ int64 skill_attack (int attack_type, struct block_list* src, struct block_list *
...
+		case MER_BLESSING:
+		case MER_INCAGI:
+			if (tsc && tsc->data[SC_CHANGEUNDEAD]) {
+				damage = 1;
+				dmg.dmotion = clif_skill_damage(dsrc, bl, tick, dmg.amotion, dmg.dmotion, damage, dmg.div_, skill_id, skill_lv, dmg_type);
+
+				if (tstatus->hp == 1)
+					return 0; // No actual damage is given if the target has 1 HP.
+			} else
+				dmg.dmotion = clif_skill_damage(dsrc, bl, tick, dmg.amotion, dmg.dmotion, damage, dmg.div_, skill_id, skill_lv, dmg_type);
```
Playtester, line 3881: Do we need this if / else? I thought we only call skill_attack when the target has SC_CHANGEUNDEAD in the first place.

Playtester (approved):
I approb.

PR #6869, opened 2022-04-22
Adjusts swapping equipment behavior

### LINKED ISSUE #6564: Swapping equipment/cards with combo leads to incorrect HP/SP
* **rAthena Hash**: b13bc7402

* **Client Date**: 2018-04-18b

* **Description of Issue**: 
  - Result: HP/SP are not properly set when swapping equipment/cards involving combo
image  
image

  - Expected Result: HP/SP should remain the same value as MaxHP/MaxSP
  * How to Reproduce: Issue can be reproduced with the Fricco's Shoes[0]-2417, Odin's Blessing[1]-2353 and Falcon Muffler[0]-2516 for equipment and the Dark Illusion-4169+ Dark Lord-4168 combo for cards. By simply swapping the same equipment, MaxHP and MaxSP will remain the same, but HP and SP will not and be set to the value without the combo. 
  * Official Information: N/A
  

  * This is mainly due to the fact that during unequip, if a combo is removed it will force update the status even in the context of swapping equipment.
  * https://github.com/rathena/rathena/blob/master/src/map/pc.cpp#L11597 for equipment
  * https://github.com/rathena/rathena/blob/master/src/map/pc.cpp#L11611 for cards
  * Most probably above statements should take into consideration the passed flag (_&& flag & 1_)

aleos89:
* **Addressed Issue(s)**: Fixes #6564

  * When swapping equipment, even if it's the same equipment, a player's status should always be recalculated.
Thanks to @kutsuru!

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -11446,7 +11446,7 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos,bool equipswit
 		for(i=0;i<EQI_MAX;i++) {
 			if(pos & equip_bitmask[i]) {
 				if(sd->equip_index[i] >= 0) //Slot taken, remove item from there.
-					pc_unequipitem(sd,sd->equip_index[i],2 | 4);
+					pc_unequipitem(sd,sd->equip_index[i], 1 | 2 | 4);
 
 				sd->equip_index[i] = n;
 			}


Playtester (commented):
This comment makes it work exactly as official servers, I tested it.

However, I want to at least putting making this configurable (for both equip and combo) to discussion.

Playtester (approved):
Discussed, no config needed. Waiting for Suggestion PR to consider it.

PR #6876, opened 2022-04-23
Fix equip order

eppc0330:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

*  Weapon and accessory should always be equipped for Right side of character, left side of equip window,  first.

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -11406,7 +11406,7 @@ bool pc_equipitem(struct map_session_data *sd,short n,int req_pos,bool equipswit
 	} else if(pos == EQP_ARMS && id->equip == EQP_HAND_R) { //Dual wield capable weapon.
 		pos = (req_pos&EQP_ARMS);
 		if (pos == EQP_ARMS) //User specified both slots, pick one for them.
-			pos = equip_index[EQI_HAND_R] >= 0 ? EQP_HAND_L : EQP_HAND_R;
+			pos = (equip_index[EQI_HAND_L] < 0) ? EQP_HAND_L : EQP_HAND_R;
 	} else if(pos == EQP_SHADOW_ACC) { // Shadow System
 		pos = req_pos&EQP_SHADOW_ACC;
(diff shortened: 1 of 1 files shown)


Playtester:
From my tests on official servers, they are equipped on the left side first.

eppc0330:
> From my tests on official servers, they are equipped on the left side first.

How about accessories? I tested on KRo today and they are equipped on the right first.

eppc0330:
> From my tests on official servers, they are equipped on the left side first.
Equiporderweapon
Equiporderacc
These are all from KRo, and both are equipped right side first.

aleos89:
Confirmed the behavior as well, the items are applied to the player's right-hand, which is on the left side of the equip window.

aleos89 (approved):

Playtester:
Yes, sorry I misunderstood. Correct, it's the left side of the equipment window, so it's the right side of the character.

eppc0330:
> Yes, sorry I misunderstood. Correct, it's the left side of the equipment window, so it's the right side of the character.

Sorry my description was kind of confusing. I edited it.

Playtester:
@eppc0330 
Hi, I tested this on pre-renewal now and I think there's a difference. On pre-renewal, if you have both hands equipped and you equip an item, it will always go to your shield hand (right side).
However, if both your right hand is free, then it will go to your right hand (left side).

For accessoires it works perfectly fine! There it indeed always prioritizes the left side.

We might need a switch for renewal here.

eppc0330:
> @eppc0330 Hi, I tested this on pre-renewal now and I think there's a difference. On pre-renewal, if you have both hands equipped and you equip an item, it will always go to your shield hand (right side). However, if both your right hand is free, then it will go to your right hand (left side).
> 
> For accessoires it works perfectly fine! There it indeed always prioritizes the left side.
> 
> We might need a switch for renewal here.

Thanks for info. I will add RENEWAL switching.

Playtester (approved):
Tested in pre-re and approved.

PR #6892, opened 2022-04-25
Minor SC_STONEWAIT fixes

aleos89:
* **Addressed Issue(s)**: N/A

  * SC_STONEWAIT status should not be applicable to the Undead element.
  * Adds the status resistance case for SC_STONEWAIT.
Thanks to @Playtester!

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -10128,7 +10128,7 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 
 		case SC_STONEWAIT:
 			val4 = val3;
-			val3 = tick - val3; // Petrify time - Incubation time
+			val3 = max(1, tick - val3); // Petrify time - Incubation time
 			tick = val4; // Original incubation time
 			break;
 
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -9298,6 +9298,7 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				return 0;
 			break;
 		case SC_STONE:
+		case SC_STONEWAIT:
 		case SC_FREEZE:
 			// Undead are immune to Freeze/Stone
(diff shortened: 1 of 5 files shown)


Playtester (approved):
Edit: Please see below.

src/map/status.cpp
```diff
@@ -9298,6 +9298,7 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				return 0;
 			break;
 		case SC_STONE:
+		case SC_STONEWAIT:
```
Playtester, line 9301: We also need to add SC_STONEWAIT in status_get_sc_def or else the resistance there won't work either.

Playtester (changes requested):

src/map/status.cpp
```diff
@@ -8804,6 +8804,7 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
 #endif
 			break;
 		case SC_STONE:
```
Playtester, line 8806: I think we need to remove SC_STONE here because otherwise you can resist the transition from SC_STONEWAIT to SC_STONE.

Alternatively you could also make the transition itself unavoidable.

Playtester (changes requested):
Found one more issue when testing.

src/map/status.cpp
```diff
@@ -10126,7 +10127,9 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			break;
 
 		case SC_STONEWAIT:
-			val3 -= tick; // Petrify time - Incubation time
+			val4 = val3;
+			val3 = tick - val3; // Petrify time - Incubation time
```
aleos89, line 10131:
```suggestion
			val3 = max(1, tick - val3); // Petrify time - Incubation time
```
Prevent any possibility of a negative value.

Playtester (approved):
Works great now as far as I could test!

Issue #6894, opened 2022-04-26
Suggestion: Defaulting drop effect to DROPEFFECT_CLIENT.


eppc0330:
**Is your request related to a missing feature?** No.

**Describe the solution you'd like**

Because recent client, supporting dropeffect, already defines all items' dropeffect, I think it is inefficient to work DB again for dropeffect. 
So isn't it better to default dropeffect to DROPEFFECT_CLIENT, instead of DROPEFFECT_NONE?
Then we will just need to touch dropeffect only when we find something special item needed special dropeffect.

**Describe alternatives you've considered**

**Additional context**

Balferian:
If set droplight effect = 'CLIENT' for all items by default then users can highlight any items by modify iteminfo.
Players can control the highlight by themselves and gain an advantage for example in events.
I think it's better to make this feature optional or make a parser for itemdb that set dropeffect_client based on received data from iteminfo ("EffectID").

PR #6895, opened 2022-04-26
Defaulting dropeffect client

eppc0330:
* **Addressed Issue(s)**: #6894 

* **Server Mode**: Renewal

 This pull will default dropeffect to DROPEFFECT_CLIENT instead of DROPEFFECT_NONE.
 I just work this pull for suggestion but not sure about stability.
 If there is other reason not defauting dropeffect to CLIENT or better way to implement suggestion, I will close this pull.


eppc0330:
I close this because without fixed item option, user can freely custom dropeffect by iteminfo.lub.

Sorry I was careless.

PR #6926, opened 2022-05-06
Removes the 100% rate cap for SC defense

aleos89:
* **Addressed Issue(s)**: N/A

  * Status changes are simply not capped to 100%.
Thanks to @Playtester!

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -2762,8 +2762,8 @@ static void pc_bonus_addeff(std::vector<s_addeffect> &effect, enum sc_type sc, i
 
 	for (auto &it : effect) {
 		if (it.sc == sc && it.flag == flag) {
-			it.rate = util::safe_addition_cap(it.rate, rate, SHRT_MAX);
-			it.arrow_rate += arrow_rate;
+			it.rate = util::safe_addition_cap(it.rate, rate, INT_MAX);
+			it.arrow_rate = util::safe_addition_cap(it.arrow_rate, arrow_rate, (short)SHRT_MAX);
 			it.duration = umax(it.duration, duration);
 			return;
 		}
@@ -2771,11 +2771,8 @@ static void pc_bonus_addeff(std::vector<s_addeffect> &effect, enum sc_type sc, i
 
 	struct s_addeffect entry = {};
 
-	if (rate < SHRT_MIN || rate > SHRT_MAX)
-		ShowWarning("pc_bonus_addeff: Item bonus rate %d exceeds -%d~%d range, capping.\n", rate, SHRT_MIN, SHRT_MAX);
-
 	entry.sc = sc;
-	entry.rate = cap_value(rate, SHRT_MIN, SHRT_MAX);
+	entry.rate = rate;
 	entry.arrow_rate = arrow_rate;
 	entry.flag = flag;
 	entry.duration = duration;
@@ -2804,7 +2801,7 @@ static void pc_bonus_addeff_onskill(std::vector<s_addeffectonskill> &effect, enu
 
 	for (auto &it : effect) {
 		if (it.sc == sc && it.skill_id == skill_id && it.target == target) {
-			it.rate = util::safe_addition_cap(it.rate, rate, SHRT_MAX);
+			it.rate = util::safe_addition_cap(it.rate, rate, INT_MAX);
 			it.duration = umax(it.duration, duration);
 			return;
 		}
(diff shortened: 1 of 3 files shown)


src/map/status.cpp
```diff
@@ -8733,7 +8733,7 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
 			return 0;
 	}
 
-	rate = cap_value(rate, 0, 10000);
+	rate = max(0, rate);
```
Playtester, line 8736: This doesn't fully fix the issue. If you equip stuff that has an added chance of over 100%, it will still go with 100% into this method.

I also noticed another issue we added here: https://github.com/rathena/rathena/commit/2993c0baa64ec41181d0cf465635f988e12f3893

We should not give sc_def2 a lower cap as that breaks how pre-renewal chances work (you can increase chance if you're higher level than target).
Alternatively you can add a lower cap only for renewal and leave it as it was before for pre-renewal.

Playtester (changes requested):

src/map/pc.cpp
```diff
@@ -3333,7 +3333,7 @@ void pc_bonus(struct map_session_data *sd,int type,int val)
 				bonus = status->def + val;
 #ifdef RENEWAL
 				status->def = cap_value(bonus, SHRT_MIN, SHRT_MAX);
-#else
+#elsef
```
Playtester: Typo?

src/map/status.cpp
```diff
@@ -8733,7 +8733,7 @@ t_tick status_get_sc_def(struct block_list *src, struct block_list *bl, enum sc_
 			return 0;
 	}
 
-	rate = max(0, rate);
+	rate = cap_value(rate, SHRT_MIN, SHRT_MAX);
```
Playtester, line 8736: Wrong location to cap it. As discussed before you need to remove this line.

src/map/pc.cpp
```diff
@@ -2762,7 +2762,7 @@ static void pc_bonus_addeff(std::vector<s_addeffect> &effect, enum sc_type sc, s
 
 	for (auto &it : effect) {
 		if (it.sc == sc && it.flag == flag) {
-			it.rate = cap_value(it.rate + rate, -10000, 10000);
```
Playtester, line 2765: I think I will need to test this later. Looks like we are capping the sum of the bonus rather than the individual bonus. But even beforehand when I had a combined bonus of >10000 I didn't get a warning in the console, so maybe I'm wrong.

Playtester (changes requested):

src/map/pc.cpp
Playtester: Ah I see you already fixed this.

src/map/pc.cpp
aleos89: The initial bonus is capped just a few lines below when the item bonus is initially created. This line is only specific to stacking the same bonus.

aleos89: Would this matter though? As long as we are capping the final result is all that we are concerned about. The other method would just be capping in two places: at the parse and here at the summation.

src/map/pc.cpp
```diff
@@ -2804,19 +2804,19 @@ static void pc_bonus_addeff_onskill(std::vector<s_addeffectonskill> &effect, enu
...
 			return;
 		}
 	}
 
 	struct s_addeffectonskill entry = {};
 
-	if (rate < -10000 || rate > 10000)
-		ShowWarning("pc_bonus_addeff_onskill: Item bonus rate %d exceeds -10000~10000 range, capping.\n", rate);
+	if (rate < SHRT_MIN || rate > SHRT_MAX)
+		ShowWarning("pc_bonus_addeff_onskill: Item bonus rate %d exceeds -%d~%d range, capping.\n", rate, SHRT_MIN, SHRT_MAX);
```
Playtester, line 2816: Duplicate minus sign here.

src/map/pc.cpp
```diff
@@ -2771,11 +2771,11 @@ static void pc_bonus_addeff(std::vector<s_addeffect> &effect, enum sc_type sc, s
 
 	struct s_addeffect entry = {};
 
-	if (rate < -10000 || rate > 10000)
-		ShowWarning("pc_bonus_addeff: Item bonus rate %d exceeds -10000~10000 range, capping.\n", rate);
+	if (rate < SHRT_MIN || rate > SHRT_MAX)
+		ShowWarning("pc_bonus_addeff: Item bonus rate %d exceeds -%d~%d range, capping.\n", rate, SHRT_MIN, SHRT_MAX);
```
Playtester, line 2775: Duplicate minus sign here.

src/map/pc.cpp
```diff
@@ -2804,19 +2804,19 @@ static void pc_bonus_addeff_onskill(std::vector<s_addeffectonskill> &effect, enu
...
 	struct s_addeffectonskill entry = {};
 
-	if (rate < -10000 || rate > 10000)
-		ShowWarning("pc_bonus_addeff_onskill: Item bonus rate %d exceeds -10000~10000 range, capping.\n", rate);
+	if (rate < SHRT_MIN || rate > SHRT_MAX)
+		ShowWarning("pc_bonus_addeff_onskill: Item bonus rate %d exceeds -%d~%d range, capping.\n", rate, SHRT_MIN, SHRT_MAX);
 
 	entry.sc = sc;
-	entry.rate = cap_value(rate, -10000, 10000);
+	entry.rate = cap_value(rate, SHRT_MIN, SHRT_MAX);
```
Playtester, line 2819: The cap seems to be only working for the first item.
For example I had two cards that gave a 40000 (400%) chance to freeze.
The resulting chance was 32767+40000=72767.

Playtester (approved):
Status changes from items and spells now seem to be working fine!

Issue #6647, opened 2022-03-06
"NPC_EARTHQUAKE" and "CR_GRANDCROSS" bug


eppc0330:
* **rAthena Hash**: master

* **Client Date**: 20211117

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: 
      1. When a character uses skill CR_Grandcross, it displays miss on character.    
![Grandcross](https://user-images.githubusercontent.com/47050704/156920876-f19a6128-6417-4fa1-aecd-17b8edaa70a8.gif)

      2. SKill NPC_EARTHQUAKE is disabled.
      
![Earthquake](https://user-images.githubusercontent.com/47050704/156920992-5f54d5b0-a4f3-47c4-a2b7-40a0af5e6474.png)

  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

Lemongrass3110:
What do you mean with "SKill NPC_EARTHQUAKE is disabled."?

eppc0330:
> What do you mean with "SKill NPC_EARTHQUAKE is disabled."?

Player can't use skill. When character uses Earthquake, the skill does not casted and that error message appears.

Lemongrass3110:
This is a NPC skill. It is not intended for player use. O_o

eppc0330:
> This is a NPC skill. It is not intended for player use. O_o

Course I know. I mean when character uses skill by aubospell.

eppc0330:
> This is a NPC skill. It is not intended for player use. O_o

Oh sorry. Just my bad. EARTHQUAKE is good.

dimasshotta:
I thought that's the way Grand cross does in renewal server?

it's this line that makes it appear as Miss..

#ifdef RENEWAL
					if (src == target)
						ad.damage = 0;
#else

eppc0330:
> 

Oh yeah it's right... I should figure out from there. Thanks.

aleos89:
Does Grand Cross display miss after the 1st/2nd skill balance page?

eppc0330:
> Does Grand Cross display miss after the 1st/2nd skill balance page?

Sorry I Im not sure.

eppc0330:
[...]

PR #6927, opened 2022-05-07
Fixes Grand Cross and Grand Darkness

Playtester:
* **Addressed Issue(s)**: #6647 (b)

* **Server Mode**: Both

- Follow-up to ce4aed2
- Fixed Grand Cross doing 4 hits instead of 3
- Fixed Grand Darkness doing 30 hits instead of 3

### DIFF
--- a/db/pre-re/skill_db.yml
+++ b/db/pre-re/skill_db.yml
@@ -6972,7 +6972,7 @@ Body:
     Unit:
       Id: Dummyskill
       Layout: -1
-      Interval: 300
+      Interval: 320
       Target: Enemy
       Flag:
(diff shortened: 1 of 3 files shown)


aleos89 (approved):

Playtester:
@eppc0330
I reworked this a bit. The interval was actually correct, the skill can hit up to 4 times if it's a moving target. Each cell itself does expire after it hit 3 times, though, so a single standing monster should always take exactly 3 hits.
If you have any concerns, please tell me.

eppc0330:
@Playtester I didn't know at all about this. Amazing detail. 👍

PR #6934, opened 2022-05-09
cloakonnpcself/cloakoffnpcself

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

Added 2 cheat codes derived from cloakonnpc/cloakoffnpc, shorter. Only for lazy people.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -11892,7 +11892,6 @@ BUILDIN_FUNC(enablenpc)
 {
 	npc_data* nd;
 	e_npcv_status flag = NPCVIEW_DISABLE;
-	int char_id = script_hasdata(st, 3) ? script_getnum(st, 3) : 0;
 	const char* command = script_getfuncname(st);
 
 	if (script_hasdata(st, 2))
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -11925,6 +11924,19 @@ BUILDIN_FUNC(enablenpc)
...
 
+	int char_id = 0;
+
+	if (script_hasdata(st, 3))
+		char_id = script_getnum(st, 3);
+	else if (!strcmp(command,"cloakoffnpcself") || !strcmp(command,"cloakonnpcself")) {
+		map_session_data* sd;
+
+		if (!script_rid2sd(sd))
+			return SCRIPT_CMD_SUCCESS;
```
aleos89, line 11935:
```suggestion
			return SCRIPT_CMD_FAILURE;
```
Shouldn't we technically return a failure here?

aleos89 (commented):
Looks good overall!

src/map/script.cpp
Atemo: `script_rid2sd` already reports the error displayed by `SCRIPT_CMD_FAILURE`, so `SCRIPT_CMD_SUCCESS` is the right value to return
https://github.com/rathena/rathena/blob/cfa9776a281a5e9431b3e4880a1300c0a7ac0100/src/map/script.cpp#L2672
https://github.com/rathena/rathena/blob/cfa9776a281a5e9431b3e4880a1300c0a7ac0100/src/map/script.cpp#L4134

PR #6945, opened 2022-05-15
Updated pre-renewal mob_db for Brasilis monsters

Playtester:
* **Addressed Issue(s)**: -

* **Server Mode**: Pre-Re

* **Description of Pull Request**: Current out pre-renewal mob_db.yml file only contains a mix of guessed an renewal values for Brasilis. Brasilis never official existed on kRO but it existed on bRO and jRO. The bRO version is a bit exploitable as monsters have really low ATK, that's why jRO increased the attack significantly. I decided to go with jRO stats as they are less exploitable, but I will provide a DB release on forums to revert to bRO ATK values if this PR goes through.

The reason why I want to do this change now is because the YAML format allows to overwrite stats of a mob only partially, so it's better if the base file has as many official values as possible (instead of e.g. even having wrong values for aMotion). Also jRO officially published their ATK stats and I saw it fits with what Doddler had originally reported, so no more estimating needed.

### DIFF
--- a/db/pre-re/mob_db.yml
+++ b/db/pre-re/mob_db.yml
@@ -40631,10 +40631,10 @@ Body:
     BaseExp: 74288
     JobExp: 77950
     MvpExp: 37144
-    Attack: 1060
-    Attack2: 2022
-    Defense: 7
-    MagicDefense: 36
+    Attack: 3304
+    Attack2: 4266
+    Defense: 32
+    MagicDefense: 66
     Str: 140
     Agi: 99
     Vit: 30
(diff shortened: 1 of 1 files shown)


db/pre-re/mob_db.yml
```diff
@@ -40645,42 +40645,49 @@ Body:
...
     Class: Boss
     Modes:
       Mvp: true
+    MvpDrops:
+      - Item: Old_Violet_Box
+        Rate: 5500
+      - Item: Old_Violet_Box
+        Rate: 5000
+      - Item: Old_Card_Album
+        Rate: 2000
```
Daegaladh, line 40665: This is the only difference I've found compared with the data I got from bRO back in the day, mine says the rate for OCA is 3000, but can't remember where that came from, probably from bRO forums.

Daegaladh (commented):
Also got different data for E_HYDRA using Sense, they probably used normal Hydra data as a base, and jRO or kRO changed it afterwards. But yeah on bRO these monsters hit a lot and were very annoying.

```
  - Id: 2081
    #AegisName: E_HYDRA
    Level: 14
    Hp: 660
    Attack: 22
    Attack2: 28
    Agi: 14
    Vit: 14
    Dex: 40
    Luk: 2
```

Daegaladh (approved):

PR #6968, opened 2022-05-20
Fixed Safety Wall not blocking reflected physical damage

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**:  Fixes Safety Wall not blocking reflected physical damage in certain cases, like items that that reflect damage.


eppc0330:
This is becuase in rA reflected damage is not weapon damage, while in KRo it is.

Basically reflecting damage type might be fixed.

Daegaladh:
> The root cause of this issue is about reflected damage type.
> 
> In rA reflected damage is not weapon damage, while in KRo it is.
> 
> So in Ra user can't trigger weapon-triggered autospell by reflected damage, while in KRo also it does.
> 
> Basically reflected damage type might be fixed, then safetywall issue will be solved without any touch.
> 
> Corresponding reflecting skills are below:
> 
>     * CR_REFLECTSHIELD (including bonus bShortWeaponDamageReturn).
> 
>     * NPC_MAXPAIN
> 
>     * Specially NPC_MAGICMIRROR, too (not bonus bMagicDamageReturn, only magicmirror).

You're right, I'll try to redo this to fix CR_REFLECTSHIELD and bShortWeaponDamageReturn, since you already started PRs on the other 2

eppc0330:
@Daegaladh Thanks

Daegaladh:
> @Daegaladh Thanks

I'm not sure how to proceed with this, do you have any clue or information on how it works on official servers?

eppc0330:
Hi, some news. Ive tried fix Reflectshield damage type and I guess I made it.
However about safetywall, I think additional implement needed as this pull does.
Sorry please forget what I said before because Safetywall still does not block reflected damage after the damage type fixed...

So I have a few questions to develop thiw PR.

1. For now this is just my guess but as PR_KYRIE blocks NPC_MAXPAIN damage, if Kyrie also blocks Reflectshield damage too,
this issue would not only doing with Safetywall.

2. This PR makes Safetywall block Reflectshield damage, but safetywall shield will also not be damaged either.
As I tested safetywall lasts untill its duration is over, no matter how much damage has been blocked.

eppc0330:
Hello. Im now actually have other query. Does safety wall blocks reflected damage in official KRo?
Because as I tested today in KRo Kyrie cannot block it.

eppc0330:
Hello. Now I have new query. Today I suddenly tested in KRo about this, with PR_KYRIE.
Kyrie cannot block reflected damage in KRo. So it would re reasonable to doubt does safetywall blocks reflected damage in KRo.
I couldn't test it because I had to reset my skill to test with safetywall.

secretdataz:
I just ran a couple of tests on kRO.

Safety wall on attacker unit does not block reflected physical damage (attacker receives reflected damage).
This is the case for Kyrie too.

Safety wall on the attacked unit prevents the attacked unit from receiving the damage but the attacker still receives the reflected damage.
Although, Kyrie on the attacked unit prevents the damage on both the attacker and attacked unit.

Forgot to record a video so "trust me bro" on this one 🤣

eppc0330:
@secretdataz Thanks for testing!

eppc0330:
I think this pull is no longer necessary because @secretdataz confirmed that Safety wall cannot block reflected damage in KRo.

PR #6969, opened 2022-05-20
Add QuestInfo display for Lost Knife quest in Kunlun

KrokusPokus:
OngoingQuestInfo (both in official kro version and in the english translation) contains data for the Lost Knife quest in Kunlun at ID 10034 - 10035. This PR adds the missing setquest/changequest/completequest calls to the rAthena NPC script.

### DIFF
--- a/npc/quests/quests_gonryun.txt
+++ b/npc/quests/quests_gonryun.txt
@@ -3805,6 +3805,7 @@ gon_fild01,245,257,0	script	#gonknife	111,3,3,{
 OnTouch_:
 	if (nakha == 1) {
 		set nakha,2;
+		changequest 10034,10035;
 		mes "^3355FFHm? What's this?";
 		mes " ";
 		mes "Something was hidden beneath the leaves...^000000";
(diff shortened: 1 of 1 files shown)


npc/quests/quests_gonryun.txt
```diff
@@ -3810,6 +3812,7 @@ OnTouch_:
 		mes "- You have found an old knife -";
 		close2;
 		getitem 1201,1; //Knife
+		changequest 10034,10035;
```
Atemo, line 3815: Pleae move `changequest` near `set nakha,2;` (few lines above) to avoid a potential warning

KrokusPokus: I moved it up, but I'm kind of wondering if it wouldn't have made more sense for me to move `set nakha,2;` down instead. It seems like the more natural flow of things, to first read the window that tells you that you find something, then close that window, get the item and the new questlog entry.

Atemo:
Thanks

PR #6984, opened 2022-05-25
Clean up to how map cache is loaded by mode

aleos89:
* **Addressed Issue(s)**: N/A

  * The db/(pre-)re/map_cache.dat now contain maps that are only different between the modes.
  * Moves all general maps into db/map_cache.dat for loading across both modes.
  * Adds support for the new RSW water height.
  * Adds an error message when loading GRFs if the file size is over 2GB.


Playtester (changes requested):
Pre-renewal map cache doesn't seem to be correct, can't walk around Izlude properly for example with the maps from the pre-re translation project:
https://github.com/llchrisll/ROenglishRE/tree/master/Pre-Renewal/data

Playtester (approved):
Pre-renewal maps are now working.
Also tested water level on the 2021-11-03 client and it still works fine.

PR #6997, opened 2022-06-01
Renewal Physical damage cleanup

attackjom:
* **Addressed Issue(s)**: correction order and formula of calculation physical damage in renewal.

* **Server Mode**: Renewal

- add new itembonus ATK%  ( It's called VAR_ATKPERCENT in aegis )
- correction order of calculation of physical attack
- fix skill EDP that give 25% poison pseudo elemental bonus to user. also Magnum break that give 20% fire pseudo elemental damage.
- correction advance katar mastery bonus.
- fix critical damage to always do max-variance attack same as Maximize power buff.

- [x] Correction item-script of many items/combo from bAddClass,Class_All  to bAtkRate according to aegis.
- [ ] Test & Review.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3494,10 +3494,7 @@ static void battle_calc_damage_parts(struct Damage* wd, struct block_list *src,s
 		critical = true;
 
 	wd->weaponAtk += battle_calc_base_weapon_attack(src, tstatus, &sstatus->rhw, sd, critical);
-	wd->weaponAtk = battle_attr_fix(src, target, wd->weaponAtk, right_element, tstatus->def_ele, tstatus->ele_lv);
-
 	wd->weaponAtk2 += battle_calc_base_weapon_attack(src, tstatus, &sstatus->lhw, sd, critical);
-	wd->weaponAtk2 = battle_attr_fix(src, target, wd->weaponAtk2, left_element, tstatus->def_ele, tstatus->ele_lv);
 
 	// Weapon ATK gain bonus from SC_SUB_WEAPONPROPERTY here ( +x% pseudo element damage)
 	if (sd && sd->sc.data[SC_SUB_WEAPONPROPERTY]) {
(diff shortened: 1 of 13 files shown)


Daegaladh (changes requested):

src/map/battle.cpp
```diff
@@ -6407,15 +6440,18 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 				ATK_ADD(wd.weaponAtk, wd.weaponAtk2, sstatus->matk_min);
 		}
 #endif
+
+int i = 0;
+
+#ifndef RENEWAL
 		// add any miscellaneous player ATK bonuses
 		if( sd && skill_id && (i = pc_skillatk_bonus(sd, skill_id))) {
```
Daegaladh, line 6448: Please, remove the unnecessary brackets in this and the other if below

Daegaladh:
Are you sure the calculation order in _battle_calc_weapon_attack_ is different in pre-renewal and renewal?
I mean, you splitted it and misc ATK bonuses are calculated later in renewal, but what about pre-re? did we got wrong that too? or is really calculated earlier?

src/map/battle.cpp
attackjom: done

attackjom:
> battle_calc_weapon_attack

I'm not sure about pre-renewal, so this pull request will focus only renewal damage that I already test many times in KRO and rathena.

attackjom:
> Are you sure the calculation order in _battle_calc_weapon_attack_ is different in pre-renewal and renewal? I mean, you splitted it and misc ATK bonuses are calculated later in renewal, but what about pre-re? did we got wrong that too? or is really calculated earlier?

I'm not sure about pre-renewal, so this pull request will focus only renewal damage that I already test many times in KRO and rathena.

src/map/battle.cpp
```diff
@@ -6410,6 +6410,7 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 	}
 
 	std::bitset<NK_MAX> nk = battle_skill_get_damage_properties(skill_id, wd.miscflag);
+	int i = 0;
```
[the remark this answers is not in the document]
attackjom: It was compiled error earlier ,  fixed now.

attackjom:
> Why did you close the PR?

I'm sorry, just a misclick.

aleos89 (approved):
Looks good to me!

attackjom:
@aleos89  thank you for review, fixed.

Daegaladh (approved):
Everything looks good, but I'd like to know @Playtester 's opinion on the ATK bonuses order just to be sure it's different in pre-re

src/map/battle.cpp
```diff
@@ -2270,7 +2272,11 @@ static int battle_calc_base_weapon_attack(struct block_list *src, struct status_
 		atkmin = max(0, (int)(atkmin - variance + base_stat_bonus));
 		atkmax = min(UINT16_MAX, (int)(atkmax + variance + base_stat_bonus));
 
+#ifdef RENEWAL
+		if ((sc && sc->data[SC_MAXIMIZEPOWER]) || critical == true)
+#else
 		if (sc && sc->data[SC_MAXIMIZEPOWER])
+#endif
```
secretdataz, line 2279:
```suggestion
		if ((sc && sc->data[SC_MAXIMIZEPOWER]) || critical == true)
```
Pre-re also uses maxAtk for critical hits

secretdataz (changes requested):
This small change for pre-renewal would be nice, albeit being out of scope of this PR.

PR #7007, opened 2022-06-03
Fixes script commands unitwalkto/unitwalk

### LINKED ISSUE #5340: unitwalkto/unitwalk sometimes stops path too early with mobs
* **rAthena Hash**: ab8ce69629e88095d72625cc96f4900ed53e16bf

* **Client Date**: 20180621

* **Server Mode**: pre-re

* **Description of Issue**: When using unitwalkto/unitwalk (in this case my source GID was a mob and my target was an NPC for unitwalkto) the source object will sometimes stop earlier than the intended destination. Despite this if an event label is passed the logic will still trigger anyways even if the mob hasn't actually "touched" that intended location.
  * **Result:** In my case, if the mob has a slow enough mob speed and/or the object is far enough away seemingly (in this case I used a poring mob with 50, 150, 250, and 500 speed(s) for testing purposes, and the target NPC destination was roughly 25 cells away), instead of continuing the movement until it reaches the NPC, it will instead stop somewhere in between the path rather than reaching the intended target and proceed with its random walk idle behavior. Meanwhile if I increase the poring's speed to be fast enough, it will reach the target 100% of the time.
  * **Expected Result:** The source object should reach the intended target (whether it's an X/Y or a unitwalkto destination) regardless of the GID Object's speed. 
  * **How to Reproduce:** Here is a sample script I created illustrating the buggy behavior at hand. As you can see here, speeds 50 and 150 seem to work fine. But once you try speeds 250 and 500 where it starts to get slower, the problem clearly starts to emerge: 

```
prt_fild08,127,264,4	script	start#testwalk	100,{
	switch (select("50 Speed:150 Speed:250 Speed:500 Speed"))
	{
	case 1: .@speed = 50;	break;
	case 2: .@speed = 150;	break;
	case 3: .@speed = 250;	break;
	case 4: .@speed = 500;	break;
	default:	close;
	}
	monster "prt_fild08",127,264,"Poring",1002,1;
	.@gid = $@mobid[0];
	setunitdata .@gid,UMOB_SPEED,.@speed;
	unitwalkto .@gid,getnpcid(0,"end#testwalk"),"::OnDone";
	end;
}

prt_fild08,153,264,4	script	end#testwalk	100,{
	end;
OnDone:
	announce "I've reached my destination.",0;
	end;
}
```

Upon a quick investigation on my end, it appears that this bug might be related to the fact that mobs have their random walk behavior on idle which triggers alongside the walkto timer without any kind of check. As a result, if the mob isn't fast enough before reaching its walkto destination, the walkto path gets overwritten with the idle random walk timer, stopping the mob's walkto path. However, the walkto timer still thinks the mob is on that path, and as a result the label will trigger eventually anyways.

Additionally, in the case of Poring, random loot laying on the floor in the middle of their walkto path also appears to interrupt it due to looting behavior, and like before, trigger the label anyways as if nothing happened and they kept going their path. So it would appear that the unitwalkto timer can get interrupted several different ways beyond my initial find.

aleos89:
* **Addressed Issue(s)**: Fixes #5340

  * Resolves an issue where monsters may tend to lose their target if they are too slow and the destination is too far when using script commands unitwalkto/unitwalk.
Thanks to @NeutralDev!

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -19321,8 +19321,10 @@ BUILDIN_FUNC(unitwalk)
 		int x = script_getnum(st,3);
 		int y = script_getnum(st,4);
 
-		if (script_pushint(st, unit_can_reach_pos(bl,x,y,0)))
+		if (script_pushint(st, unit_can_reach_pos(bl,x,y,0))) {
+			ud->state.force_walk = true;
 			add_timer(gettick()+50, unit_delay_walktoxy_timer, bl->id, (x<<16)|(y&0xFFFF)); // Need timer to avoid mismatches
+		}
 	} else {
 		struct block_list* tbl = map_id2bl(script_getnum(st,3));
 
@@ -19330,8 +19332,10 @@ BUILDIN_FUNC(unitwalk)
 			ShowError("buildin_unitwalk: Bad target destination.\n");
 			script_pushint(st, 0);
 			return SCRIPT_CMD_FAILURE;
-		} else if (script_pushint(st, unit_can_reach_bl(bl, tbl, distance_bl(bl, tbl)+1, 0, NULL, NULL)))
+		} else if (script_pushint(st, unit_can_reach_bl(bl, tbl, distance_bl(bl, tbl)+1, 0, NULL, NULL))) {
+			ud->state.force_walk = true;
 			add_timer(gettick()+50, unit_delay_walktobl_timer, bl->id, tbl->id); // Need timer to avoid mismatches
+		}
 		off = 4;
 	}
 
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1687,6 +1687,10 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 	if(md->bl.prev == nullptr || md->status.hp == 0)
 		return false;
 
+	// Monsters force-walked by script commands should not be searching for targets.
+	if (md->ud.state.force_walk)
+		return false;
+
 	if (DIFF_TICK(tick, md->last_thinktime) < MIN_MOBTHINKTIME)
 		return false;
 
@@ -2004,6 +2008,10 @@ static int mob_ai_sub_lazy(struct mob_data *md, va_list args)
 
 	if(md->bl.prev == NULL)
 		return 0;
+
+	// Monsters force-walked by script commands should not be searching for targets.
+	if (md->ud.state.force_walk)
+		return false;
 
 	t_tick tick = va_arg(args,t_tick);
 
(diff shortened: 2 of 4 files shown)


Lemongrass3110 (approved):
Looks good!
But what I think now that we have state tracking: Should we not report or respect if the monster is still "unitwalking" from the last script call?

Daraen1:
error
does it fix these problems (appear sometimes after **_charming wink_** Skill but not all the time)

Atemo:
I haven't checked (maybe you have), will an aggressive monster be able to attack the player if the monster is ordered to move through a chain of unitwalk (from a position to another), like it's the case for wave mode instance (even through the instance is currently disabled)

src/map/script.cpp
```diff
@@ -19310,6 +19310,13 @@ BUILDIN_FUNC(unitwalk)
 
 	ud = unit_bl2ud(bl);
 
+	// Unit was already forced to walk.
+	if (ud != nullptr && ud->state.force_walk) {
+		script_pushint(st, 0);
+		ShowWarning("buildin_%s: Unit has already been foced to walk and not reached its destination yet.\n", cmd);
```
Lemongrass3110, line 19316:
```suggestion
		ShowWarning("buildin_%s: Unit has already been forced to walk and not reached it's destination yet.\n", cmd);
```

src/map/script.cpp
```diff
@@ -19477,8 +19490,13 @@ BUILDIN_FUNC(unitstopwalk)
...
-	if(script_rid2bl(2,bl))
+	if(script_rid2bl(2,bl)) {
+		unit_data *ud = unit_bl2ud(bl);
+
+		if (ud != nullptr)
+			ud->state.force_walk = false;
 		unit_stop_walking(bl, flag);
+	}
 
 	return SCRIPT_CMD_SUCCESS;
```
Lemongrass3110, line 19501:
```suggestion
		return SCRIPT_CMD_SUCCESS;
	}else{
		return SCRIPT_CMD_FAILURE;
	}
```

src/map/script.cpp
```diff
@@ -19477,8 +19490,13 @@ BUILDIN_FUNC(unitstopwalk)
 	if (script_hasdata(st, 3))
 		flag = script_getnum(st, 3);
 
-	if(script_rid2bl(2,bl))
+	if(script_rid2bl(2,bl)) {
+		unit_data *ud = unit_bl2ud(bl);
+
+		if (ud != nullptr)
```
Lemongrass3110, line 19496: Should we also add a warning/failure here, if flag != USW_FORCE_STOP and the unit is not walking?

Lemongrass3110 (approved):

PR #7008, opened 2022-06-03
Fixes an item comparison check when loading chars

### LINKED ISSUE #4933: Wrong item equality validation in char_memitemdata_to_sql?
* **rAthena Hash**: 

8c26298

* **Client Date**: 

2018-06-21

* **Server Mode**: 

Renewall

* **Description of Issue**: 
 
image

In function char_memitemdata_to_sql, is missing the  card[1]

aleos89:
* **Addressed Issue(s)**: Fixes #4933

  * Adds a missing card slot check when validating items during character loading.
  * Also adds the random options to the check.
Thanks to @lukasrmattos!

### DIFF
--- a/src/char/char.cpp
+++ b/src/char/char.cpp
@@ -629,9 +629,8 @@ int char_memitemdata_to_sql(const struct item items[], int max, int id, enum sto
 				continue;
 
 			if( items[i].nameid == item.nameid
-			&&  items[i].card[0] == item.card[0]
-			&&  items[i].card[2] == item.card[2]
-			&&  items[i].card[3] == item.card[3]
+			&& (memcmp(items[i].card, item.card, sizeof(int) * MAX_SLOTS) == 0)
+			&& (memcmp(items[i].option, item.option, 5 * MAX_ITEM_RDM_OPT) == 0)
 			&&  items[i].unique_id == item.unique_id
 			) {	//They are the same item.
 				int k;


src/char/char.cpp
```diff
@@ -629,9 +629,8 @@ int char_memitemdata_to_sql(const struct item items[], int max, int id, enum sto
 				continue;
 
 			if( items[i].nameid == item.nameid
-			&&  items[i].card[0] == item.card[0]
-			&&  items[i].card[2] == item.card[2]
-			&&  items[i].card[3] == item.card[3]
+			&& (memcmp(items[i].card, item.card, sizeof(int) * MAX_SLOTS) == 0)
+			&& (memcmp(items[i].option, item.option, 5 * MAX_ITEM_RDM_OPT) == 0)
```
Lemongrass3110, line 633: Arent both checks done below?
I think from what I remember this was why I never touched it.
Because those were only the initial checks and there are more below. But maybe they should be merged or something. Cannot check it on phone right now.

Lemongrass3110 (commented):
I am still not 100% sure if this is better as it is right now.
The old "assumption" was like outer if is "most likely" the same. And then check the details (could still be wrong though), but now we assume that even earlier (without checking the cards) and might continue with our wrong assumption.

Basically I am curious if this really saves anything "load wise", it just moves the load from the database server to rAthena's character server. It would be interesting if this approach would still be truly useful nowadays or if modern database servers with proper database design would swallow the load without any hickups.

Also what I do not really understand is that we first check a lot of different attributes and then we update or delete based on `item.id`, which seems to be unique already - so why not use that one?

So my suggestion would be to first compare `items[i].id == item.id` (outer if) and then assume it is the same item and then check for changes (inner if), if there are any update or if there are none do nothing. We could even consider using memcmp on the whole struct at this point. If the item with `item.id` x was not found anymore it should be deleted from the database. The remaining items will be inserted at the bottom.

PR #7014, opened 2022-06-07
updated useskill_fail_cause

datawulf:
* **Server Mode**: Both

Updated useskill_fail_cause with new fail causes, which shall be used by new 4th/expanded job skills


datawulf:
will be part of another PR

Issue #7034, opened 2022-06-16
Spirit Handler not able to use Summoner Items


admkakaroto:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/2992879ee3e6b43a150dcecbd2c3ad384d380023
* **Client Date**: Irrelevant
* **Server Mode**: Renewal
* **Description of Issue**: Items that can be equipped by the Summoner job are not being equipped by the later Spirit Handler class.
  * Result: 
  * How to Reproduce: Try to equip any summoner item in Spirit Handler Class.

Atemo:
`MAPID_SPIRIT_HANDLER = JOBL_FOURTH|JOBL_THIRD|JOBL_UPPER|JOBL_2_1|MAPID_SUMMONER,`
Shouldn't the item be equippable by SPIRIT_HANDLER, `Classes` from item_db being `All` by default ?

Atemo:
Ha I see, spirit handler is 2-1 so it should have it own "field"

PR #7035, opened 2022-06-16
Add support for Spirit Handler equip Summoner items

admkakaroto:
Added definition for Spirit Handler class to be able to equip the same items as the Summoner class.
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/7034
* **Server Mode**: Renewal
* **Description of Pull Request**: This PR will add a basic definition so that the Spirit Handler class can equip the same items as the Summoner class.

### DIFF
--- a/db/re/item_db_equip.yml
+++ b/db/re/item_db_equip.yml
@@ -14814,6 +14814,7 @@ Body:
     Range: 1
     Jobs:
       Summoner: true
+      Spirit_Handler: true
     Locations:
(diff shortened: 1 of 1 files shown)


aleos89 (approved):

Lemongrass3110 (approved):

Lemongrass3110:
> Shouldn't we add a Spirit_handle field in sql table ?

If it is not there, yes. But unrelated to this PR.

PR #7058, opened 2022-06-23
Fixes mob and item atcommand sorting issues

### LINKED ISSUE #7055: Atcommands @ii, @mi, @whodrops, @monster are lacking sorting
* **rAthena Hash**: [563a701](https://github.com/rathena/rathena/commit/563a7012ff01163d3bc499f03b75c44f4c8b765c)

* **Client Date**: 2021-11-03

* **Server Mode**: Both

* **Description of Issue**: The result when using `@ii, @mi, @whodrops and @monster` among other commands does not turn the results as expected.

Examples:

1. `@ii Carnium` does not actually return the original Carnium at all, but instead different item that contain Carnium in there name such as HD Carnium. Same issue with `@whodrops Carnium` or `@whodrops Bradium`, can't even find out who drops them as it just lists of bunch of items that don't drop at all!

2. `@mi Poring` does not return monsters in their database order. A lot of event mobs come before the base one.

3. `@monster Savage` spawns an event version of the Savage monster rather than the original Savage.

aleos89:
* **Addressed Issue(s)**: Fixes #7055

  * Fixes several atcommands used for mob and item information were not properly accounting for order.
  * Resolves an issue where YAML databases were not properly utilizing the cache feature.
Thanks to @Playtester!

### DIFF
--- a/src/map/itemdb.hpp
+++ b/src/map/itemdb.hpp
@@ -22,7 +22,9 @@ const t_itemid UNKNOWN_ITEM_ID = 512;
 /// The maximum number of item delays
 #define MAX_ITEMDELAYS	10
 ///Designed for search functions, species max number of matches to display.
+#ifndef MAX_SEARCH
 #define MAX_SEARCH	10
+#endif
 
 #define MAX_ROULETTE_LEVEL 7 /** client-defined value **/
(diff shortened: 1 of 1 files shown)


Playtester (approved):
I fixed an issue that could make the map server crash, everything else seems fine.

Issue #3020, opened 2018-04-01
Baby Dragon Hat [1]


MrAntares:
The item has wrong script and the double attack part is not working at all:

Current script:
bonus bDex,2; skill "TF_DOUBLE",max(getskilllv("TF_DOUBLE"),5);

Correct script:
bonus bDex,2; skill "TF_DOUBLE",5; bonus bDoubleRate,25;

Daegaladh:
Right, besides that max(getskilllv("TF_DOUBLE") doesn't make sense, because if you have this skill at a higher level, will already be triggered at that level.
Btw this item is commented out from the pre-re db for no reason.

BrOgBr:
I think that according to the DP _(kRO)_, the correct script is:

`bonus bDex,2; skill "TF_DOUBLE",max(5,getskilllv("TF_DOUBLE"));`

> Enable Double Attack level 5, if a higher level of this skill is learned then that level will be enable.

But how Daegaladh said, it doesn't make sense.

Mystou:
@BrOgBr 
What's the diffrence between 
max(getskilllv("TF_DOUBLE"),5)
and
max(5,getskilllv("TF_DOUBLE"))
?

For me, the command is the same. It will always get the highest value between the two.

*max(<number or array>{,<number or array>,...})
Returns the smallest (or biggest) from the set of given parameters.
.@maximum = max( 0, 5, 10, 4 ); // .@maximum will be equal to 10

So the current script is correct, no?

BrOgBr:
I really don't know the difference, but I did so and it's working perfectly here.
I always do it that way.

Daegaladh:
skill "TF_DOUBLE" gives you the skill in your skill tree, but does nothing, bDoubleRate is what really does the effect (yeah we should fix that, has been like this for years D:)
Also, as I said, there's no point on using max, since if you already have the skill at a higher level, it will do the higher effect.

Mystou:
Dumb question : so if I have the skill at level 7, the "skill "TF_DOUBLE",5" in the item will change the skill to level 5 or keep it to level 7 (in my skill tree) ?

Daegaladh:
I've already answered that question in my previous comment: "if you already have the skill at a higher level, it will do the higher effect" so yeah, it will keep the level 7 one.
But now I've realized i don't know how it would work, 7 for daggers only and 5 for the rest? 7 for all weapons? Need to test this on official servers.

BrOgBr:
[...]

PR #7070, opened 2022-06-27
Cleans up Thief's Double Attack item bonuses

aleos89:
* **Addressed Issue(s)**: N/A

  * Items that grant TF_DOUBLE now no longer require bDoubleRate.
  * Adds pc_checkskill_flag() to return a skill's flag value.
Thanks to @eppc0330 and @secretdataz!

### DIFF
--- a/db/pre-re/item_db_equip.yml
+++ b/db/pre-re/item_db_equip.yml
@@ -844,7 +844,6 @@ Body:
     Refineable: true
     Script: |
       skill "TF_DOUBLE",5;
-      bonus bDoubleRate,25;
       bonus2 bAddRace,RC_DemiHuman,5;
       bonus2 bAddRace,RC_Player_Human,5;
   - Id: 1131
@@ -24702,7 +24701,6 @@ Body:
     Script: |
       bonus bMaxSP,100;
       skill "TF_DOUBLE",3;
-      bonus bDoubleRate,15;
       skill "TF_STEAL",1;
       skill "TF_HIDING",1;
       skill "TF_POISON",1;
@@ -29767,7 +29765,6 @@ Body:
       bonus bMaxHP,50;
       bonus bMaxSP,50;
       skill "TF_DOUBLE",2;
-      bonus bDoubleRate,10;
       bonus2 bSubRace,RC_DemiHuman,3;
       bonus2 bSubRace,RC_Player_Human,3;
   - Id: 5284
@@ -31728,7 +31725,6 @@ Body:
       NoGuildStorage: true
     Script: |
       skill "TF_DOUBLE",5;
-      bonus bDoubleRate,25;
   - Id: 5389
     AegisName: Angel_Spirit
     Name: Angel Spirit
@@ -34171,9 +34167,7 @@ Body:
     View: 525
     Script: |
       bonus bDex,2;
-      .@bonus = max(getskilllv("TF_DOUBLE"), 5);
-      skill "TF_DOUBLE",.@bonus;
-      bonus bDoubleRate,.@bonus * 5;
+      skill "TF_DOUBLE",5;
   - Id: 5532
     AegisName: Pirate_Dagger_J
     Name: Pirate Dagger
@@ -39963,7 +39957,6 @@ Body:
     WeaponLevel: 4
     Script: |
       skill "TF_DOUBLE",5;
-      bonus bDoubleRate,25;
       bonus2 bAddRace,RC_DemiHuman,40;
       bonus2 bAddRace,RC_Player_Human,40;
   - Id: 13408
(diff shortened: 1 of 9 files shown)


eppc0330:
I think this PR will make all doubleattack items check character's TF_DOUBLE skill level right? But some items do not check it. I think defaulting this is kind of dangerous.

eppc0330:
For example if assassin character, using katar, learning 10lv TF_DOUBLE, equips Snakehead hat (ID 5388), character should have 35% double rate, not 70%, because the item doesnt check character's TF_DOUBLE skill level.

aleos89:
We don't have to check for the level since `pc_skill()` which is called from the `buildin_skill` function automatically detects the level learned and whether to replace it or not.

aleos89:
> For example if assassin character, using katar, learning 10lv TF_DOUBLE, equips Snakehead hat (ID 5388: Grants 5lv TF_DOUBLE without skill level check), character should have 35% double rate, not 70%, because the item doesnt check character's TF_DOUBLE skill level.

Do you know of other items that grant skills that behave like this, or is it only select to TF_DOUBLE?

eppc0330:
@aleos89 Quite a lot items behave like that but I will mention only ones in case of thief can equip, because in normal case other jobs else theif do not have TF_DOUBLE.  Nagan(ID 1130, not ID 13407) and snakehead hat(ID 5388), two items behave as I told you. In case of items not from KRo, sorry I have no idea because my information is based on KRo lub file.

secretdataz:
@eppc0330 From my test on kRO, equipping a stalker (double attack level 10, 70% proc chance) with a chick hat (double attack level 2, 21% proc chance) procs double attack like a lot it's hard to believe it's just 21% chance.

eppc0330:
> @eppc0330 From my test on kRO, equipping a stalker (double attack level 10, 70% proc chance) with a chick hat (double attack level 2, 14% proc chance) procs double attack like a lot it's hard to believe it's just 14% chance.
> 
> https://media.discordapp.net/attachments/120596466661195778/991052756154474496/Ragexe_RL9JvBM7S0.gif

WTF really? Its nonsense that iteminfo already says double rate will follow character's TF_DOUBLE level.
If you are true than it must be a bug.

eppc0330:
@secretdataz 
But your screenshot...
A character hits doubleattack 8 times during 13 attacks.
It is 61%. No matter how small the number of trials, your screenshot shows 14% is literally nonsense. 
Are you really sure...? 

I have experience of using chick hat when I was playin assassin before and I remember it works well. Also Idk if you read Korean but KRo chick hat description says double rate will follow character's TF_DOUBLE level. 
Sorry for having doubts about your test but I need more explain because your clue(the screenshot) does not back up your opinion.

secretdataz:
> @secretdataz But your screenshot... A character hits doubleattack 7times during 12attacks. It is 58%. No matter how small the number of trials, your screenshot shows 14% is literally nonsense. Are you really sure...?
> 
> I have experience of using chick hat when I was playing assassin before and I remember it works well. Also Idk if you read Korean but KRo chick hat description says double rate will follow character's TF_DOUBLE level. Sorry for having doubts about your test but I need more explain because your clue(the screenshot) does not back up your opinion.

I attacked the egg over 100 times. It definitely procced more than 50 times so I'm sure it's more than 14%.  
Also, chick hat's Korean description seems to be more elaborate than other double attack item for some reason.

Chick Hat's description states that.
```더블어택 2레벨 사용 가능. 더 높은 레벨을 배우고 있을 시, 해당 레벨로 발동.```

While Poison Forged Spear's description only states.
```더블어택 5Lv 사용 가능.```

But both items are scripted in the same way on official servers, only granting the skill via `EnableSkill`. This applies to every item that grants double attack. They are all scripted this way. So it's reasonable to expect them all to behave the same.

Example official scripts with irrelevant statements removed:
```
item  Poison_Forged_Spear								
	event OnStartEquip:										
		EnableSkill TF_DOUBLE 5		
```

```
item Nagan
	event OnStartEquip:
		EnableSkill TF_DOUBLE 5
```

```
item Side_Winder_Card
	event OnStartEquip:
		EnableSkill TF_DOUBLE 1
```

```
item Chick_Hat
	event OnStartEquip:
		EnableSkill TF_DOUBLE 2
```
```
item Snake_Head
	event OnStartEquip:
		EnableSkill TF_DOUBLE 5
```

Also a longer test, for larger sample size is here, https://www.youtube.com/watch?v=JzY3Ahs4EbI.

In conclusion, I believe, in all cases, unless stated otherwise, double attack with higher level learned by the player should takes a higher priority.

aleos89:
Will also mention #3020, as it has also been confirmed on previously to take into account the player's learned level versus the granted level.

eppc0330:
@secretdataz Oh I think I missinderstood you. I mean chick hat should have 70% chance if char has 10lv TF_DOUBLE  so you are definitely right. 

I thought you are saying chick hat grants only 14% chance sorry my English was bad.

PR #7093, opened 2022-07-05
Fix displayed damage when it is dodged

eppc0330:
Fix miss when a unit dodges attack from player

Thanks to @Haydrich!

* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

    As I tested in KRo, when player attack is dodged, miss is shown just once though dodged attack is muti hit skill.


Haydrich:
Great work 🪄

PR #7094, opened 2022-07-05
Minor change of TF_BACKSLIDING behavior

eppc0330:
Backsliding02

Fix Backsliding blow only when a character is moved by the skill.

If wall is placed behind the character, skill will now show casting animation.

Thanks to @Haydrich!

* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -9248,17 +9248,20 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		break;
 
 	case TF_BACKSLIDING: //This is the correct implementation as per packet logging information. [Skotlex]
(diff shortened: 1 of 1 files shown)


eppc0330:
@Daegaladh 
Backsliding01

This is my video from KRo yesterday. I have tested it with my character.

eppc0330:
@Daegaladh Yeah actually I didn't know it either until yesterday.

XanKriegor1:
I'd like to have this as a configurable option.

Atemo:
A config just for that seems overkill

Playtester:
I agree that a config doesn't seem to be needed, it's just a visual hint and it only applies when backsliding against a wall anyway. Just make it work like on kRO.

Issue #7114, opened 2022-07-17
Script commands unitwalkto/unitwalk no longer works in loop


Badarosk0:
* **rAthena Hash**: https://github.com/rathena/rathena/commit/aad3a70bbcb9261cb472f03b94eebf8b92be2200

* **Client Date**: 2021-11

* **Server Mode**: RE

* **Description of Issue**: 
  * Result: The unitwalk command is malfunctioning if used in a loop for longer positions. It always returns the message "[Warning]: buildin_unitwalk: Unit has already been forced to walk and not reached it's destination yet.
". This is for a few moments forcing and immobilizing the monster and it crashes and stops attacking the target.
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

Atemo:
After https://github.com/rathena/rathena/commit/aad3a70bbcb9261cb472f03b94eebf8b92be2200, the steps should be
```
unitstopwalk .boss_gid;
unitwalk .boss_gid,.x-1,150;
```
(the tracking need to be removed by unitstopwalk)

Badarosk0:
> After [aad3a70](https://github.com/rathena/rathena/commit/aad3a70bbcb9261cb472f03b94eebf8b92be2200), the steps should be
> 
> ```
> unitstopwalk .boss_gid;
> unitwalk .boss_gid,.x-1,150;
> ```
> 
> (the tracking need to be removed by unitstopwalk)

I've done this modification before and it doesn't work either. Same problem, but the warnings still happen on the console.
"[Warning]: buildin_unitstopwalk: Unable to find unit or unit is not walking."
```
prontera,150,150,3	script	move_test	67,{
	monster strnpcinfo(4),150,150,"--ja--",20784,1,"move_test::OnBossDeath";
	.boss_gid = $@mobid[0];
	getunitdata .boss_gid,.@boss_data;
	while(unitexists(.boss_gid)) 
	{
		getunitdata .boss_gid,.@boss_data;
		.x = .@boss_data[UMOB_X];
		.y = .@boss_data[UMOB_Y];
		announce "x: "+.x+"",0;
		if(.x > 155)
		{
			unitstopwalk .boss_gid;
			unitwalk .boss_gid,.x-1,150;
		}
		sleep(3000);
		
	}
end;

OnWalkEnd:
	announce "finish",0;
end;

OnBossDeath:
	announce "killed",0;	
	end;
}
```

[...]

PR #7116, opened 2022-07-17
unitwalk corrections

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/7114

* **Server Mode**: Both

Removed the warning in unitstopwalk (there is no way to check is the unit is curently walking with npc script command).
Corrected mistakes in unitwalk.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -19325,11 +19325,15 @@ BUILDIN_FUNC(unitwalk)
 			status_calc_npc(((TBL_NPC*)bl), SCO_NONE);
 	}
 
+	bool can_reach = false;
+
 	if (!strcmp(cmd,"unitwalk")) {
 		int x = script_getnum(st,3);
 		int y = script_getnum(st,4);
+		can_reach = unit_can_reach_pos(bl,x,y,0);
+		script_pushint(st, can_reach);
 
-		if (script_pushint(st, unit_can_reach_pos(bl,x,y,0))) {
+		if (can_reach) {
 			if (ud != nullptr)
 				ud->state.force_walk = true;
 			add_timer(gettick()+50, unit_delay_walktoxy_timer, bl->id, (x<<16)|(y&0xFFFF)); // Need timer to avoid mismatches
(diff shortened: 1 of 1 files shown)


Badarosk0:
It works fine now, thanks!

aleos89:
The immobile-ness is because the commit Atemo linked I added a new state to mobs to force them to ignore all outside influences so that the mob will walk to their destination no matter what before resuming their original AI.

I tested your script and increased the sleep to 5 seconds. The monster attacked me and then after a few seconds walked again but then started attacking me again.

Badarosk0:
> The immobile-ness is because the commit Atemo linked I added a new state to mobs to force them to ignore all outside influences so that the mob will walk to their destination no matter what before resuming their original AI.
> 
> I tested your script and increased the sleep to 5 seconds. The monster attacked me and then after a few seconds walked again but then started attacking me again.

I increased the delay time and it wasn't static anymore. Now it's hard to get him out of the accessible area, which is great. Thanks.

Atemo:
The monster is still stuck if, for example, it moves at x, y+31.

Badarosk0:
> The monster is still stuck if, for example, it moves at x, y+31.

Maybe create a check for inaccessible cell. Then it would abort the walk.

src/map/script.cpp
```diff
@@ -19497,11 +19505,6 @@ BUILDIN_FUNC(unitstopwalk)
 		if (ud != nullptr)
 			ud->state.force_walk = false;
 
-		if (unit_stop_walking(bl, flag) == 0 && flag != USW_FORCE_STOP) {
-			ShowWarning("buildin_unitstopwalk: Unable to find unit or unit is not walking.\n");
-			return SCRIPT_CMD_FAILURE;
-		}
-
```
Lemongrass3110, line 19504: Not really sure why you remove this?
The rest looks good!

Atemo: yeah `unit_stop_walking` should not be removed

PR #7124, opened 2022-07-20
Adjusts Enchanting Shadow behavior

### LINKED ISSUE #7104: SHC_ENCHANTING_SHADOW official behavior
* **rAthena Hash**: master

* **Client Date**: 20211117

* **Server Mode**: Renewal

* **Description of Issue**:  SHC_ENCHANTING_SHADOW from rA has different behavior with one from KRo.
  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
    First difference, we need to know that Enchanting Shadow is not only triggered on normal attack.
    It also can be triggered by the skill. 

    Second, amount of increased damage per stack and limit of stack info is lacking. So I have tested it.
    status
    Tested with character above, with critical normal attack.
    Starts with 9973 damage. Below is a damage log.
    Damage log

    9973
    10274
    10575
    10876
    11177
    11480
    11781
    12082
    12383
    and so no....

    Each stack adds about 3% damage more.
    Something interesting is that I couldn't find a limit of the stack. 
    The damage is continuously increased as the stack added.
    However it's not easy to stack more than 8 because triggering stack does not reset duration of the stack.
    New stack will not affect previous stacks' duration. Check below:
    Stack ends
    We can see that as durations of some stacks are ended, the damage bonus is not reset as 0%.
    Instead, 3% damage has been decreased as each stack ended.

    Finally, as we see in screenshot when triggering stack, skill shows scratching effect on target's head. 
    The number of this scratching effect is 677.

  

aleos89:
* **Addressed Issue(s)**: Fixes #7104

* **Server Mode**: Renewal

  * Enchanting Shadow can now be triggered by skills as well as melee.
  * If Shadow Wounds is active the hit count is increased by 1 without resetting the status' duration.
  * Removes the hit count limit.
  * Adjusts the damage rate of Shadow Wounds to increase by 3%.
Thanks to @eppc0330!


src/map/skill.cpp
```diff
@@ -1303,6 +1303,18 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
 			}
 		}
 
+		// Enchanting Shadow gives a chance to inflict Shadow Wounds to the enemy.
+		if (sc != nullptr) {
+			status_change_entry *sce;
+
+			if ((sce = sc->data[SC_SHADOW_WEAPON]) && rnd() % 100 < sce->val2) {
```
Lemongrass3110, line 1310:
```suggestion
			if ((sce = sc->data[SC_SHADOW_WEAPON]) && util::rnd_chance(sce->val2, 100)) {
```

src/map/skill.cpp
```diff
@@ -1303,6 +1303,18 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
...
+				if (tsc && tsc->data[SC_SHADOW_SCAR])
+					sce->val1 += 1; // Directly adjust the rate so the duration is not reset each time.
```
Lemongrass3110, line 1312:
```suggestion
					sce->val1 += 1; // Directly adjust the damage rate so the duration is not reset each time.
```

Lemongrass3110 (approved):
Good from my side.

@eppc0330 can you test if it is working like you reported now?

eppc0330:
@Lemongrass3110 Thanks for alarm me. I've tested right before and actually It is still different with three things.

First the duration of SC is 15sec, not 30sec.
Sorry I forgot report this before in my issue.

Second specialeffect 677 should appear on target when the skill triggered.

Finally, damage bonus mechanism. I'm not sure how to explain this properly.
The total number of triggered shadow scar SCs are affecting damage bonus
but all durations of these SCs are going separately.

For example, If the first trigger has been done at 00:00, second at 00:05, and third at 00:10,
then damage bonus will be 9% until 00:15, but 6% at 00:16, 3% at 00:21, and finally none at 00:26.

eppc0330:
I can't explain in words clearly so I will try with code.
I'm not sure how to simplify the code but the mechanism is something like below:
I'll pretend the stack limit is 10.

```
if ((sce = sc->data[SC_SHADOW_WEAPON]) && rnd_chance(sce->val2, 100)) {
	sc_start(src, bl, SC_SHADOW_SCAR, 100, 1, skill_get_time2(SHC_ENCHANTING_SHADOW, sce->val1));
	if (tsc && tsc->data[SC_SHADOW_SCAR]) {
		sc_start(src, bl, SC_SHADOW_SCAR1, 100, 1, skill_get_time2(SHC_ENCHANTING_SHADOW, sce->val1));
	} else if (tsc && tsc->data[SC_SHADOW_SCAR1]) {
		sc_start(src, bl, SC_SHADOW_SCAR2, 100, 1, skill_get_time2(SHC_ENCHANTING_SHADOW, sce->val1));
	} else if (tsc && tsc->data[SC_SHADOW_SCAR2]) {
		sc_start(src, bl, SC_SHADOW_SCAR3, 100, 1, skill_get_time2(SHC_ENCHANTING_SHADOW, sce->val1));
	.
	.
	.
	.
	} else if (tsc && tsc->data[SC_SHADOW_SCAR9]) {
		sc_start(src, bl, SC_SHADOW_SCAR10, 100, 1, skill_get_time2(SHC_ENCHANTING_SHADOW, sce->val1));
	}
	clif_specialeffect(bl, 677, AREA);
}
```
I feel like individual SCs are applied on target because each shadow scar SC's duration is processing separately.
This is very messy code but I am just trying to explain how the SCs are applied.

aleos89:
Thanks! Sorry I was busy this week and didn't come back to this. I was trying to think of a solution as well. The only things I could think of would be to attach a timer to the player like Spirit Spheres or like you suggested by creating several SC. The second option just seems overkill so I might end up creating a timer on the player.

eppc0330:
> Thanks! Sorry I was busy this week and didn't come back to this. I was trying to think of a solution as well. The only things I could think of would be to attach a timer to the player like Spirit Spheres or like you suggested by creating several SC. The second option just seems overkill so I might end up creating a timer on the player.

No problem. I just hope my intentions are well communicated. Thanks for solutions all the time.

aleos89:
@eppc0330 I've implemented the timer behavior. Let me know if this works better as to how you described the skill behavior.

eppc0330:
Wow I guess now it works. Thank you so much.

src/map/unit.hpp
```diff
@@ -19,6 +19,8 @@ enum clr_type : uint8;
 extern const short dirx[DIR_MAX]; ///lookup to know where will move to x according dir
 extern const short diry[DIR_MAX]; ///lookup to know where will move to y according dir
 
+#define MAX_SHADOW_SCAR 100 /// Max Shadow Scars
```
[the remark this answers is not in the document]
Lemongrass3110: If it is only used in unit.cpp it could also be moved to the top of the .cpp file.

src/map/unit.cpp
```diff
@@ -3622,6 +3625,103 @@ int unit_free(struct block_list *bl, clr_type clrtype)
...
+		util::erase_at(ud.shadow_scar_timer, 0);
+	}
+
+	ud.shadow_scar++;
+	ud.shadow_scar_timer.push_back(add_timer(gettick() + interval, unit_shadowscar_timer, ud.bl->id, 0));
+
+	status_change *sc = status_get_sc(ud.bl);
+
+	if (sc != nullptr && sc->data[SC_SHADOW_SCAR] != nullptr)
+		sc->data[SC_SHADOW_SCAR]->val1 = ud.shadow_scar;
```
Lemongrass3110, line 3685: No clif call needed in this case?

src/map/skill.cpp
```diff
@@ -1303,6 +1303,16 @@ int skill_additional_effect(struct block_list* src, struct block_list *bl, uint1
 			}
 		}
 
+		// Enchanting Shadow gives a chance to inflict Shadow Wounds to the enemy.
+		if (sc != nullptr) {
+			status_change_entry *sce = sc->data[SC_SHADOW_WEAPON];
+			unit_data *ud = unit_bl2ud(bl);
+
+			if (sce != nullptr && ud != nullptr && rnd_chance(sce->val1, 100)) {
```
Lemongrass3110, line 1311: val1 is used for chance and skill lv now?

src/map/unit.cpp
```diff
@@ -3622,6 +3629,59 @@ int unit_free(struct block_list *bl, clr_type clrtype)
...
+ * Adds a Shadow Scar to unit for 'interval' ms.
+ * @param ud: Unit data
+ * @param interval: Duration
+ */
+void unit_addshadowscar(unit_data &ud, int interval) {
+	if (ud.shadow_scar_timer.size() > 0 && ud.shadow_scar_timer.size() >= MAX_SHADOW_SCAR) {
+		if (ud.shadow_scar_timer[0] != INVALID_TIMER)
+			delete_timer(ud.shadow_scar_timer[0], unit_shadowscar_timer);
+		util::erase_at(ud.shadow_scar_timer, 0);
+	}
```
Lemongrass3110, line 3670:
```suggestion
	if (ud.shadow_scar_timer.size() >= MAX_SHADOW_SCAR) {
		// <add warning if you want>
		return;
	}
```

src/map/unit.cpp
```diff
@@ -3622,6 +3629,59 @@ int unit_free(struct block_list *bl, clr_type clrtype)
...
+
+	status_change *sc = status_get_sc(ud.bl);
+
+	if (sc != nullptr && sc->data[SC_SHADOW_SCAR] == nullptr)
+		sc_start(ud.bl, ud.bl, SC_SHADOW_SCAR, 100, 1, INFINITE_TICK);
+
```
Lemongrass3110, line 3676:
```suggestion
```

src/map/unit.cpp
```diff
@@ -3622,6 +3629,59 @@ int unit_free(struct block_list *bl, clr_type clrtype)
...
+	ud.shadow_scar_timer.push_back(add_timer(gettick() + interval, unit_shadowscar_timer, ud.bl->id, 0));
+
+	if (sc != nullptr && sc->data[SC_SHADOW_SCAR] != nullptr) {
+		sc->data[SC_SHADOW_SCAR]->val1 = static_cast<int>(ud.shadow_scar_timer.size());
+		clif_enchantingshadow_spirit(ud);
+	}
```
Lemongrass3110, line 3682:
```suggestion
	status_change *sc = status_get_sc(ud.bl);
	
	if (sc != nullptr) {
		if (sc->data[SC_SHADOW_SCAR] != nullptr) {
			sc->data[SC_SHADOW_SCAR]->val1 = static_cast<int>(ud.shadow_scar_timer.size());
		} else {
			sc_start(ud.bl, ud.bl, SC_SHADOW_SCAR, 100, static_cast<int>(ud.shadow_scar_timer.size()), INFINITE_TICK);
		}
		clif_enchantingshadow_spirit(ud);
	}
```

Lemongrass3110 (approved):
Small style thingies now, looks good!

src/map/skill.cpp
aleos89: Shadow Weapon (Enchanting Shadow) is the base skill that needs to be active to send out Shadow Scars to targets. So here sce->val1 is SC_SHADOW_WEAPON's chance which is skill_lv%.

PR #7138, opened 2022-07-26
adds missing mob skill conditions MSC_GROUNDATTACKED and MSC_DAMAGEDGT

datawulf:
* **Server Mode**:  Renewal (both)

* adds missing mob skill condition MSC_GROUNDATTACKED for IF_GROUNDATTACKCHECK. This triggers when the monster receives damage from any ground targeted skill.
* adds missing mob skill conditions MSC_DAMAGEDGT for IF_HEAVYDAMAGE. This triggers when a monster receives a specified amount of dmg in a single attack.
* adds mob-skills for existing monsters / skills with new conditions

Any input if the naming or solution could be improved are appreciated.

Source: divine-pride.net
Verified on aegis.


src/map/mob.cpp
```diff
@@ -3713,6 +3715,10 @@ int mobskill_use(struct mob_data *md, t_tick tick, int event)
 			flag = 1; //Trigger skill.
 		else if (ms[i]->cond1 == MSC_SKILLUSED)
 			flag = ((event & 0xffff) == MSC_SKILLUSED && ((event >> 16) == c2 || c2 == 0));
+		else if (ms[i]->cond1 == MSC_GROUNDATTACKED && damage)
+			flag = (skill_get_inf((event >> 16))&INF_GROUND_SKILL);
+		else if (ms[i]->cond1 == MSC_DAMAGEDGT && damage)
```
aleos89, line 3720:
```suggestion
		else if (ms[i]->cond1 == MSC_GROUNDATTACKED && damage > 0)
			flag = (event > 0 && skill_get_inf((event >> 16))&INF_GROUND_SKILL);
		else if (ms[i]->cond1 == MSC_DAMAGEDGT && damage > 0)
```

aleos89 (approved):
Minor cleanup. Rest looks good to me!

Daegaladh (approved):

src/map/mob.cpp
```diff
@@ -2415,6 +2415,7 @@ void mob_damage(struct mob_data *md, struct block_list *src, int damage)
 	}
 
 	if (src && damage > 0) { //Store total damage...
+		mobskill_use(md, gettick(), (skill_id<<16), damage);
```
Atemo, line 2418: Maybe the call could be instead from https://github.com/rathena/rathena/blob/97b5f3b4ac63bb920a673043af02518784c386f5/src/map/battle.cpp#L284

src/map/mob.cpp
```diff
@@ -3667,7 +3668,7 @@ bool mob_chat_display_message(mob_data &md, uint16 msg_id) {
 /*==========================================
  * Skill use judging
  *------------------------------------------*/
-int mobskill_use(struct mob_data *md, t_tick tick, int event)
+int mobskill_use(struct mob_data *md, t_tick tick, int event, int64 damage)
```
Atemo, line 3671: Why int64 ? `damage` from `mob_damage` is int

datawulf: I guess that was left from trying to use battle_damage :)

src/map/mob.cpp
datawulf: Ground/Unit skill damage doesn't use battle damage, that's why part isn't correct either like reported here:
https://github.com/rathena/rathena/issues/7141

Lemongrass3110 (approved):

PR #7150, opened 2022-08-01
Fixes Cart Cannon defense reduction

### LINKED ISSUE #6734: Dragon Breath does no work the way it's intended
* **rAthena Hash**: 

* **Client Date**: 2021

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result:  
   Dragon Breath Skills both act as if Hard DEF is Hard DEF.

  * Expected Result:  
  Dragon Breath Skills both should calculate damage just like Cart Cannon and Arms Cannon, treating Hard DEF as Soft DEF (or kind of "ignoring DEF").

  * How to Reproduce: 
  Testing damage, after you use some kind of ignore DEF equipment, the damage is hugely different compared to the same when testing with Cart Cannon that works the same way.

  * Official Information: 
  Many sources from kRO and even Sigma(from Divine-Pride) states himself that Dragon Breath treats Hard DEF just like Cart Cannon: as Soft DEF. You can see it here: https://www.divine-pride.net/forum/index.php?/topic/3723-kro-jobs-improvement-project/page/13/#comment-11057

  

### LINKED ISSUE #7143: Cart Cannon Defence Ignore Bugged
* **rAthena Hash**:  Latest

* **Client Date**:  20200904

* **Server Mode**: Renewal

* **Description of Issue**:  Cart Cannon is supposed to treat Hard defence as soft, however it does not, info i found was from 202203 stating that Cart cannon STILL DOES THIS, on rAthena it does not.
  * Result: 
  * Expected Result:  Using Cart cannon on a target with Hard Physical Def ( left ) should treat it as Right, if it has what would be 350 Left DEF (Hard) which is like 40-50% or so, it should treat that as Soft Def (VIT) and just remove 350 damage instead.
  * How to Reproduce:  Use Cart Cannon with latest rAthena hash and equip Defence penetration gear and see how much it improves your damage, which it should not as you are already or should already be ignoring that def.
  * Official Information: 
  

* **Modifications that may affect results**:  None

aleos89:
* **Addressed Issue(s)**: Fixes #6734 and #7143

* **Server Mode**: Renewal

  * Resolves an issue where Cart Cannon wasn't properly reducing damage based on the target's defense.
Thanks to @LadyNanuia!

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -5756,23 +5756,24 @@ static void battle_calc_defense_reduction(struct Damage* wd, struct block_list *
 	}
 
 #ifdef RENEWAL
-	/**
-	 * RE DEF Reduction
-	 * Damage = Attack * (4000+eDEF)/(4000+eDEF*10) - sDEF
-	 * Pierce defence gains 1 atk per def/2
-	 */
 	switch(skill_id) {
 		case GN_CARTCANNON:
-			// Defense reduction by flat value.
 			if (attack_ignores_def(wd, src, target, skill_id, skill_lv, EQI_HAND_R) || attack_ignores_def(wd, src, target, skill_id, skill_lv, EQI_HAND_L))
 				return;
 			if (is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_R) || is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_L))
 				return;
 
+			// Defense reduction by flat value.
+			// This completely bypasses the normal RE DEF Reduction formula.
 			wd->damage -= (def1 + vit_def);
 			if (is_attack_left_handed(src, skill_id))
 				wd->damage2 -= (def1 + vit_def);
 			break;
+		/**
+		 * RE DEF Reduction
+		 * Damage = Attack * (4000+eDEF)/(4000+eDEF*10) - sDEF
+		 * Pierce defence gains 1 atk per def/2
+		 */
 		default:
 			if( def1 == -400 ) /* -400 creates a division by 0 and subsequently crashes */
 				def1 = -399;
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -5759,6 +5759,7 @@ static void battle_calc_defense_reduction(struct Damage* wd, struct block_list *
 	switch(skill_id) {
 		case RK_DRAGONBREATH:
 		case RK_DRAGONBREATH_WATER:
+		case NC_ARMSCANNON:
 		case GN_CARTCANNON:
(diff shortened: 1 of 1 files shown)


attackjom:
Can this applied to Dragonbreath and Dragonbreath water?  It work on defense in the same way.

src/map/battle.cpp
```diff
@@ -5761,17 +5761,31 @@ static void battle_calc_defense_reduction(struct Damage* wd, struct block_list *
...
-		is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_R) ? (def1*battle_calc_attack_skill_ratio(wd, src, target, skill_id, skill_lv))/200 : 0,
-		is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_L) ? (def1*battle_calc_attack_skill_ratio(wd, src, target, skill_id, skill_lv))/200 : 0
-	);
-	if( !attack_ignores_def(wd, src, target, skill_id, skill_lv, EQI_HAND_R) && !is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_R) )
-		wd->damage = wd->damage * (4000+def1) / (4000+10*def1) - vit_def;
-	if( is_attack_left_handed(src, skill_id) && !attack_ignores_def(wd, src, target, skill_id, skill_lv, EQI_HAND_L) && !is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_L) )
-		wd->damage2 = wd->damage2 * (4000+def1) / (4000+10*def1) - vit_def;
+	switch(skill_id) {
+		case GN_CARTCANNON:
+			// Defense reduction by flat value.
```
Lemongrass3110, line 5766: Comment should be moved down and maybe add something like that eDef and sDef from the normal RE formula are not taken into account.

src/map/battle.cpp
```diff
@@ -5761,17 +5761,31 @@ static void battle_calc_defense_reduction(struct Damage* wd, struct block_list *
...
+			if (attack_ignores_def(wd, src, target, skill_id, skill_lv, EQI_HAND_R) || attack_ignores_def(wd, src, target, skill_id, skill_lv, EQI_HAND_L))
+				return;
+			if (is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_R) || is_attack_piercing(wd, src, target, skill_id, skill_lv, EQI_HAND_L))
+				return;
 
+			wd->damage -= (def1 + vit_def);
```
Lemongrass3110, line 5772: What I miss above is the battle_calc_attack_skill_ratio call.
GN_CARTCANNON has it's own case there so from my understanding it should call it, but I could be wrong with my lacking knowledge about battle.cpp as well.

aleos89: The base damage calculation should be triggered in `battle_calc_weapon_attack`.

Lemongrass3110 (approved):

PR #7157, opened 2022-08-02
Updates the behavior of Overheat

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Renewal

  * Overheat is now triggered via damage received, not damage dealt.
  * Corrects an issue where the HP damage would not trigger because the Overheat status would end too early.
  * Damage will no longer cause a flinch.
  * General cleanups.


Lemongrass3110 (commented):
Rest looks okay

vstumpf (approved):
LGTM!

PR #7170, opened 2022-08-10
fix item job all fourth

null6385:
* **Addressed Issue(s)**: 

* **Server Mode**: 


src/map/itemdb.hpp
```diff
@@ -234,8 +234,8 @@ enum e_item_job : uint16
 
 	ITEMJ_ALL_UPPER = ITEMJ_UPPER | ITEMJ_THIRD_UPPER | ITEMJ_FOURTH,
 	ITEMJ_ALL_BABY = ITEMJ_BABY | ITEMJ_THIRD_BABY,
-	ITEMJ_ALL_THIRD = ITEMJ_THIRD | ITEMJ_THIRD_UPPER | ITEMJ_THIRD_BABY,
-
+    ITEMJ_ALL_THIRD = ITEMJ_THIRD | ITEMJ_THIRD_UPPER | ITEMJ_THIRD_BABY | ITEMJ_FOURTH,
+    ITEMJ_ALL_FOURTH = ITEMJ_FOURTH,
```
Atemo, line 238: `ITEMJ_FOURTH` shouldn't be added in `ITEMJ_ALL_THIRD` (illogical). 
`ALL_FOURTH` is not required because currently there is only normal fourth jobs. We can stick with `FOURTH`

PR #7175, opened 2022-08-12
Fix variable cast reduction bug.

kaninhot004:
* **Addressed Issue(s)**: -

* **Server Mode**: Renewal

* **Description of Pull Request**: Fix variable cast (Millisecond) reduction bonus that made 'Release', 'Improve Concentration' and other 0 second cast time skill to be cast longer.

How to reproduce

1. Add custom item bonus `bonus bVariableCast,-500;` to any equipment
2. and stack it about 2~3 equipment
3. then equip it.
4. Cast Improve Concentration, cast time will increased.


aleos89 (approved):

Pokye:
```
6>D:\rA\src\map\skill.cpp(19157,24): warning C4244: 'argument': converting 'double' to 'int', possible data loss
6>D:\rA\src\map\skill.cpp(19168,24): warning C4244: 'argument': converting 'double' to 'int', possible data loss
```

src/map/skill.cpp
```diff
@@ -18950,7 +18950,7 @@ int skill_vfcastfix(struct block_list *bl, double time, uint16 skill_id, uint16
 		}
 		for (const auto &it : sd->skillvarcast) {
 			if (it.id == skill_id) { // bonus2 bSkillVariableCast
-				time += it.val;
+				time = max(0, time + it.val);
```
Atemo, line 18953: The range is checked later, do these 2 changes change anything?
```
time = max((int)time, 0) + (1 - (float)min(fixcast_r, 100) / 100) * max(fixed, 0); //Underflow checking/capping
```

kaninhot004: Seem like it's no longer problem in latest commits. I will close this PR.

Thank you for reviewed!

PR #7189, opened 2022-08-17
Fix grade upgrade fail announcement

eppc0330:
* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

    Now grade upgrade fail message will be shown only when upgrading to grade A and B.
    This is official behavior.


secretdataz:
Shouldn't this be done in enchantgrade.yml instead?

eppc0330:
> Shouldn't this be done in enchantgrade.yml instead?

This is not affecting success message. Only fail message.

Atemo:
In that case how about replacing `Announce` field by `AnnounceSuccess` and `AnnounceFail` in the yml file to make things clearer?

eppc0330:
> In that case how about replacing `Announce` field by `AnnounceSuccess` and `AnnounceFail` in the yml file to make things clearer?

Right this seems good

eppc0330:
Sorry miss click

src/map/status.cpp
```diff
@@ -674,17 +674,31 @@ uint64 EnchantgradeDatabase::parseBodyNode( const ryml::NodeRef& node ){
 				}
 			}
 
-			if( this->nodeExists( gradeNode, "Announce" ) ){
-				bool announce;
+			if( this->nodeExists( gradeNode, "AnnounceSuccess" ) ){
+				bool announcesucess;
```
aleos89, line 678:
```suggestion
				bool announcesuccess;
```

src/map/status.cpp
```diff
@@ -674,17 +674,31 @@ uint64 EnchantgradeDatabase::parseBodyNode( const ryml::NodeRef& node ){
...
 
-				if( !this->asBool( gradeNode, "Announce", announce ) ){
+				if( !this->asBool( gradeNode, "AnnounceSuccess", announcesucess ) ){
```
aleos89, line 680:
```suggestion
				if( !this->asBool( gradeNode, "AnnounceSuccess", announcesuccess ) ){
```

aleos89 (approved):

Atemo (approved):

Lemongrass3110:
Needs some small adjustments. I will push them to your branch this weekend!

db/enchantgrade.yml
```diff
@@ -30,7 +30,9 @@
 #           Refine                      Required refine level.
 #           Chance                      Base chance of success out of 0~10000.
 #           Bonus                       Enchantgrade bonus. (Default: 0)
-#           Announce                    Announce if someone tries to increase the enchantgrade. (Default: true)
+#           AnnounceSuccess             Announce on upgrade success. (Default: true)
+#           AnnounceFail                Announce on upgrade failure. (Default: false)
+#           Announce                    Announce on upgrade success and failure.
```
Atemo, line 35: Seems a bit redundant as the 2 above do the job, but why not.
The rest looks good !

Atemo (approved):

db/enchantgrade.yml
Lemongrass3110: Thats just to keep support for V1. :)

eppc0330:
Sorry I deleted wrong repository... I can't edit this pull anymore T_T

PR #7222, opened 2022-08-30
laphine_synthesis script command update

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

Added a parameter for the `laphine_synthesis` script command. The NPC can open the UI on official servers.
Another (better?) alternative might be to update `consumeitem` script command to support the `laphine_synthesis` command.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -23242,16 +23242,21 @@ void clif_parse_laphine_synthesis( int fd, struct map_session_data* sd ){
 		}
 	}
 
-	int16 index = pc_search_inventory( sd, sd->state.laphine_synthesis );
+	int16 index;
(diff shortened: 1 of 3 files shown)


src/map/clif.cpp
```diff
@@ -23242,17 +23242,22 @@ void clif_parse_laphine_synthesis( int fd, struct map_session_data* sd ){
 		}
 	}
 
-	int16 index = pc_search_inventory( sd, sd->state.laphine_synthesis );
+	int16 index;
```
Lemongrass3110, line 23245: Move it into the if and for loop now please! :)

Lemongrass3110 (approved):

Issue #3690, opened 2018-11-14
[RE] Gunslinger job change reward


zackdreaver:
* **rAthena Hash**: 78cda50e3b5cfb3d053c7d08181345418e5a83a1
* **Client Date**: N/A
* **Server Mode**: Renewal
* **Description of Issue**: 
  * Result: Missing renewal reward items from changing Novice to Gunslinger, preventing gunslinger changing Novice Revolver into other guns by talking to Sharp Snake's Fang in academy
  * Expected Result: Renewal reward items need to be updated to follow official
  * How to Reproduce: Check script
  * Official Information: https://irowiki.org/wiki/First_Expanded_Job_Change_Guides#Gunslinger

I don't have Gunslinger in iRO to test, but the reward should be same as wiki.
Reward items: 1 Novice Rifle [3], 2 Cartridge and 1 Silver Cartridge
* **Modifications that may affect results**: N/A

Balferian:
Checked on kRo server.
**Changed req.items for Wise Bull Horn:**
	_[6021] = {
		Title = "Wise Bull Horn's voucher",
		Description = {
			"I can make a voucher that will demonstrate your desire to become a warrior of the earth for you to present to Gunslinger drillmasters."
		},
		Summary = "3 Zargons, 10 Shells, 3 Fluffs, 1 Trunk, 3 Rainbow Shells, 3 Green Herbs"
	},_
=>
	[6021] = {
		Title = "황소 뿔이 말하는 징표",
		Description = { "자네의 바람을 계속 이어가려면 내게 징표의 재료를 가져와야 할 것일세." },
		Summary = "버섯포자3개, 초록허브 3개" 
	},
**Change (3 Zargons, 10 Shells, 3 Fluffs, 1 Trunk, 3 Rainbow Shells, 3 Green Herbs)  to (3 Mushroom Spore ID - 921 & 3 Green Herb ID - 511)**
**Reward items:** 1 Novice Rifle [3] (13180), 2 Cartridge (12149) and 1 Silver Cartridge (22738) - **true**

Replay from kRo: [Gunslinger Quest.zip](https://github.com/rathena/rathena/files/2580426/Gunslinger.Quest.zip)

Akkarinage:
Has anyone tested this on kRO recently? The files I have for the Gunslinger Quest provide a `Six_Shooter` **or** a `Branch` when talking to Master Miller and completing quest 6024 and gives no other items.. which I guess means it's a very early version of the script.

PR #7223, opened 2022-08-30
Updated Gunslinger Job Change Quests's Reward Items

Akkarinage:
Checked against reliable source to confirm items.
Added a check instead of a separate file as only the items have changed between Pre-RE and RE.

* **Addressed Issue(s)**:  #3690 

* **Server Mode**:  Renewal

Adds a `checkre()` function to determine which items the new Gunslinger should be rewarded with. This allows players to exchange their rifle instead of being stuck with the Pre-RE items.

I used the function instead of creating separate (pre-)re files as the rewards are the only lines that change between server modes, and code duplication is not beneficial to anybody.

### DIFF
--- a/npc/jobs/1-1e/gunslinger.txt
+++ b/npc/jobs/1-1e/gunslinger.txt
@@ -183,10 +183,16 @@ que_ng,152,167,3	script	Master Miller	901,{
 			callfunc "Job_Change",Job_Gunslinger;
 			set GUNS_Q,6;
 			completequest 6024;
-			if (rand(1,2) == 1) {
-				getitem 13100,1; // Six_Shooter
+			if(checkre(0) == 1){
+				getitem 13180,1;		// Novice_Rifle
+				getitem 12149,2;		// Bullet_Case
+				getitem 12151,1;		// Bullet_Case_Silver
 			} else {
-				getitem 13150,1; // Branch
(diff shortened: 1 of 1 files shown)


Atemo (approved):

aleos89 (approved):

npc/jobs/1-1e/gunslinger.txt
```diff
@@ -183,10 +183,16 @@ que_ng,152,167,3	script	Master Miller	901,{
 			callfunc "Job_Change",Job_Gunslinger;
 			set GUNS_Q,6;
 			completequest 6024;
-			if (rand(1,2) == 1) {
-				getitem 13100,1; // Six_Shooter
+			if(checkre(0) == 1){
```
Daegaladh, line 186:
```suggestion
			if (checkre(0)) {
```

Daegaladh (changes requested):
Just some cleanup

npc/jobs/1-1e/gunslinger.txt
Akkarinage: I don't know why everyone is so fond of adding extra spaces, but I'll apply the suggestion because we don't really need the comparative.

Daegaladh (approved):

PR #7224, opened 2022-08-30
Initial release of adventure agency

Lemongrass3110:

* **Server Mode**: Both

Adds support for the adventure agency feature.
Some additional cleanups in web-server.


JohnnyPlayy:
What beautiful codes! Another incredible job. 😻

sql-files/web.sql
```diff
@@ -48,3 +48,24 @@ CREATE TABLE IF NOT EXISTS `merchant_configs` (
   `data` longtext NOT NULL,
   PRIMARY KEY (`world_name`, `account_id`, `char_id`)
 ) ENGINE=MyISAM;
+
+--
+-- Table structure for table `party_bookings`
+--
+
+CREATE TABLE IF NOT EXISTS `party_bookings` (
```
vstumpf, line 56: if this is handled by the map server, shouldn't it go in main.sql?

src/web/partybooking_controller.cpp
```diff
@@ -0,0 +1,530 @@
...
+	uint16 maximum_level;
+	std::string comment;
+
+public:
+	std::string to_json( std::string& world_name );
+};
+
+std::string s_party_booking_entry::to_json( std::string& world_name ){
+	return
+		"{ \"AID\": " + std::to_string( this->account_id ) +
```
vstumpf, line 50: soon i'll add nlohmann and we can get rid of this :D

gravity-ro:
Bug report: 
When join the party disconnect from server
Error in Map Server

[Warning]: clif_parse: Received packet 0x0ae6 with expected packet length 12, but only 10 bytes remaining, disconnecting session #3.

vstumpf:
> Error in Map Server
> 
> [Warning]: clif_parse: Received packet 0x0ae6 with expected packet length 12, but only 10 bytes remaining, disconnecting session`#3`.

Can you be more specific, what packetver are you using?

ecdarreola:
Hmmm can't seem to find the button to register in party booking using 2020-03-04RagexeRE. See attached.
image

And I don't see any request going to web-server after using the PR. See below.
image

Am I missing something?

secretdataz:
@ecdarreola You client doesn't even try to call the party booking api. Likely a lua file issue.

ecdarreola:
> @ecdarreola Your client doesn't even try to call the party booking api. Likely a lua file issue.

The lua files in service_korea folder that has the externalsettings has been updated to use the IP of web-server which is localhost. Are there any lua files needs to be updated for party booking?

Lemongrass3110:
> > @ecdarreola Your client doesn't even try to call the party booking api. Likely a lua file issue.
> 
> The lua files in service_korea folder that has the externalsettings has been updated to use the IP of web-server which is localhost. Are there any lua files needs to be updated for party booking?

Make sure that you configure the right property. I think in the old clients the property for the URL was called different and for party booking they added a new property: `AssistAddr = "127.0.0.1:8888"`

Lemongrass3110:
> Bug report: When join the party disconnect from server Error in Map Server
> 
> [Warning]: clif_parse: Received packet 0x0ae6 with expected packet length 12, but only 10 bytes remaining, disconnecting session #3.

@gravity-ro oh. you are back? Since you AGAIN were unable to apply the PR correctly (0xae6 is added in this PR) I am thinking of giving you another thinking time before you can spam comments all over our repo again. :) So maybe think twice before commenting and check if you did everything correctly.

ecdarreola:
> > > @ecdarreola Your client doesn't even try to call the party booking api. Likely a lua file issue.
> > 
> > 
> > The lua files in service_korea folder that has the externalsettings has been updated to use the IP of web-server which is localhost. Are there any lua files needs to be updated for party booking?
> 
> Make sure that you configure the right property. I think in the old clients the property for the URL was called different and for party booking they added a new property: `AssistAddr = "127.0.0.1:8888"` But it is as @secretdataz said, there are no requests to your web-server and that is why you cannot see the button to register a booking.

Yeah I noticed that the new client uses AssistAddr. I will try to find old reference of lua for externalsettings for 2020-03-04 client first. Thanks for the info.

gravity-ro:
> > Bug report: When join the party disconnect from server Error in Map Server
> > [Warning]: clif_parse: Received packet 0x0ae6 with expected packet length 12, but only 10 bytes remaining, disconnecting session #3.
> 
> @gravity-ro oh. you are back? Since you AGAIN were unable to apply the PR correctly (0xae6 is added in this PR) I am thinking of giving you another thinking time before you can spam comments all over our repo again. :) So maybe think twice before commenting and check if you did everything correctly.

==============================================
In fact, every script I've followed every line There might be some mistakes. But why do you think I'm spam?

If I find anything wrong, I'll report it. Have good intentions every time why is there a problem

Example of a person who reported that the button was not found. join the party He was wrong to use the wrong Client.
Each person can fail. but came to report problems There might be some points that you really missed.
And I've already reported the problem and it's really wrong.

This is the same problem.
Sorry translated by google.

Screenshot_sssss

fab10Carvalho:
> Bug report: When join the party disconnect from server Error in Map Server
> 
> [Warning]: clif_parse: Received packet 0x0ae6 with expected packet length 12, but only 10 bytes remaining, disconnecting session #3.

Please Update your emulator!!

gravity-ro:
> > Bug report: When join the party disconnect from server Error in Map Server
> > [Warning]: clif_parse: Received packet 0x0ae6 with expected packet length 12, but only 10 bytes remaining, disconnecting session #3.
> 
> Please Update your emulator!!  

 Thank you.
But i update from here everyday I don't know where I went wrong.

ecdarreola:
Once you refuse and block a request it will always be permanent.

zdurexz:
i will waiting update current.

OptimusM:
Thanks @Lemongrass3110 for doing this amazing PR.

2 observations testing this:

1.-
I tried this feature on 2 clients, on clients from 2021, is everything working just fine, but, on 2022 clients, even march 2022, it has some response problem, i mean, everyone can register their own party, but, clicking on Join, does not generate any request to Party leader, and console says "[Info]: *ip* [POST/ party/info] 404
and the windows that should ask the party leader to accept or deny, does not show up. (to be clear, this is working ok on 2021 clients), so, seems like a packet problem maybe.

2.- 
if the map server has some problem, and somehow crash by any reason, the sql tables will keep the information of the registered party there, so you should manually clean the table to drop the information. (so, even if people delete the party, the information of the party, on the window will still show up.)


Thanks for the hard work!

ecdarreola:
> > Once you refuse and block a request it will always be permanent.
> 
> @ecdarreola works normally for me. Make sure you dont tick that box: grafik

Yeah. I accidentally ticked the box. After that I can't seem to receive any invite from same account with different character. I'll take note of it.

tmdwlse1990:
screenrAthena000
The Korean name of the Adventurer Agency is broken.
Do you have any solution?

ecdarreola:
I experienced memory leak in partybooking_controller.cpp as attached.

image

zdurexz:
Could you update this PR with the current version of rAthena?

Lemongrass3110:
> Could you update this PR with the current version of rAthena?

@zdurexz done

zdurexz:
> > Could you update this PR with the current version of rAthena?
> 
> @zdurexz done

Thanks @Lemongrass3110

tmdwlse1990:
There is a phenomenon that Korean characters are broken when used in the euckr character set. Is there any solution?
Sorry for my weak English.
@Lemongrass3110
screenrAthena000

aleos89 (commented):
image

Issue #7244, opened 2022-09-04
Remove old SQL Upgrade files


Akkarinage:
I'm aware that in reality someone is going to curse me for this, but there's no legitimate reason for not updating and applying upgrade files as and when they are available.

Perhaps removal of old files that exist in the repo prior to 2022 is reasonable?

vstumpf:
Personally, I'm not a fan of this. Sometimes my (test) servers go months without updating rathena, I'm sure I'm not the only one. 

While we can and should encourage people to update often, I don't think this is a friendly way of doing so.

secretdataz:
IMO, we can do this once. However, moving forward, there should be an automated database upgrade system in place that this should not have to be done again.

Akkarinage:
Users aren't needing files to upgrade anything if they're not updating their files. If files are needed they're still available through previous revisions. Having an automated process would be great, but that doesn't remove old files from the repo.

I think I'm the only person to have deleted upgrade files in the last 10 years... that's a very long time 😂  https://github.com/rathena/rathena/pulls?q=is%3Apr+author%3Aakkarinage+is%3Aclosed+sql+upgrade unsure how we handled them back on svn days. I vaguely recall a document somewhere that one of the devs typed up about a one year retention, but that was never adhered to and the files just continued to sit there doing nothing. I have no desire to trawl back through the commit log for 10yr+ data. 

In the past I kept files for longer (2 years) and another occasion only 6 months of files were kept. Personally, I see no reason to keep them longer than 6 months.

secretdataz:
> Having an automated process would be great, but that doesn't remove old files from the repo.

I don't think removing old files will be required anymore if we were to implement such automated db upgrade system.
But for tidiness' sake, we could move them into a separate folder from automatable upgrades and leave them alone.

aleos89:
I agree with the others opinions. Maybe finalizing an automated SQL patcher will resolve this for us.

Akkarinage:
I'll close this off, as the above linked PR is for an automated system for *applying* the upgrades, not *removing* old ones.

PR #4916, opened 2020-05-09
Adds support for automatic SQL updates

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Pre-renewal and Renewal

* **Description of Pull Request**: 
  * SQL updates are checked at individual server start up and applied based on several unique factors.
  * Adds a master sql_update_db.yml database that houses everything in one.
  * Progress is saved into a local flat file for each available schema.
  * Removes usage of individual SQL files.

Further details:
- Individual updates can be further defined based on server mode, SQL schema (if tables are housed separately; ie: `log` database from normal `ragnarok` database), or completely skipped.
- The progress of determining if an update has been applied is done through a flat file depending on the 4 different schemas (login, char, map, and log).
- Useful for future features such as deploying RPM, PKG, Docker, etc.

PR #7245, opened 2022-09-04
Removed old SQL Upgrade files

Akkarinage:
* **Addressed Issue(s)**: #7244 

This PR removes SQL Upgrade files older than 2022.


Lemongrass3110:
We are not Microsoft and forcing people to update daily.
Better keep it for now until #4916 is finished.

PR #7265, opened 2022-09-13
Adds SC_WEAPONBREAKER status

### LINKED ISSUE #6915: NPC_WEAPONBRAKER should be a buff skill
* **rAthena Hash**: b25fda1

* **Client Date**: 2022-11-03

* **Server Mode**: Both

* **Description of Issue**: NPC_WEAPONBRAKER should be a self-skill that grants a buff to self with the property "Adds a 2%*Level chance to break a weapon for 30 seconds." (similar to an item bonus that gives the chance to break weapons)
  * Result: It's a target attack skill with range=9 that breaks the weapon on target.
  * Expected Result: It's a self skill with range=NULL that grants a buff that gives a 2%*Level chance to break a weapon for 30 seconds.
  * How to Reproduce: Fight Bloody Knight
  * Official Information: Range of the skill is undefined on official servers which is a good indication it's supposed to be a self skill. The chance and duration I got from a friend. Might require an actual test to confirm.

aleos89:
* **Addressed Issue(s)**: Fixes #6915

  * Adds the SC_WEAPONBREAKER status which is applied via NPC_WEAPONBRAKER's self cast ability.
  * Adds a chance to break a target's weapon at 2 * skill_lv% for 30 seconds.
Thanks to @Playtester!

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -12473,7 +12473,7 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			}
 			break;
 		case SC_WEAPONBREAKER:
-			val2 = val1 * 2; // Chance to break weapon
+			val2 = val1 * 2 * 100; // Chance to break weapon
 			break;
 
 		default:
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -12472,6 +12472,9 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				}
 			}
 			break;
+		case SC_WEAPONBREAKER:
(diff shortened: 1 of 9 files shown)


Lemongrass3110 (approved):

src/map/status.cpp
```diff
@@ -12472,6 +12472,9 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				}
 			}
 			break;
+		case SC_WEAPONBREAKER:
+			val2 = val1 * 2; // Chance to break weapon
```
Playtester, line 12476: Shouldn't it be 200?

PR #7288, opened 2022-09-27
Fixes an issue with weapon swapping and statuses

aleos89:
* **Addressed Issue(s)**: N/A

  * Resolves statuses ending early when swapping weapons of the same type.
  * Removes a duplicate check.
Thanks to @Aristeus09!

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -11720,8 +11720,6 @@ static void pc_unequipitem_sub(struct map_session_data *sd, int n, int flag) {
 	pc_deleteautobonus( sd->autobonus3, sd->inventory.u.items_inventory[n].equip );
 
 	sd->inventory.u.items_inventory[n].equip = 0;
-	if (!(flag & 4))
-		pc_checkallowskill(sd);
 	iflag = sd->npc_item_flag;
 
 	// Check for combos (MUST be before status_calc_pc)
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -11751,7 +11749,8 @@ static void pc_unequipitem_sub(struct map_session_data *sd, int n, int flag) {
 	}
 
 	if (flag & 1 || status_calc) {
-		pc_checkallowskill(sd);
+		if (!(flag & 4))
+			pc_checkallowskill(sd);
```
Lemongrass3110, line 11753:
```suggestion
		pc_checkallowskill(sd);
```
Shouldnt that still be outside of the if?
Otherwise we lose checks that dont use flag 4 and the combo is broken?

src/map/pc.cpp
```diff
@@ -11751,7 +11749,8 @@ static void pc_unequipitem_sub(struct map_session_data *sd, int n, int flag) {
...
 		status_calc_pc(sd, SCO_FORCE);
 	}
```
Lemongrass3110, line 11755:
```suggestion
	}else if( !( flag & 4 ) ){
		pc_checkallowskill( sd );
	}
```

Lemongrass3110 (approved):

src/map/pc.cpp
aleos89: So `pc_checkallowskill()` only looks for required weapon or shield items. I don't think combos or anything else should affect this check specifically. Unless some combo is auto-unequipped when one of them is off?

PR #7301, opened 2022-10-01
Implemented refine result broadcast packet

secretdataz:
* **Addressed Issue(s)**: N/A
* **Server Mode**: Renewal
Implemented official refine announcement packet. Should work the way kRO works except name censoring which I don't know its logic.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -22408,6 +22408,9 @@ void clif_parse_refineui_refine( int fd, struct map_session_data* sd ){
 		log_pick_pc( sd, LOG_TYPE_OTHER, 1, item );
 		clif_misceffect( &sd->bl, 3 );
 		clif_refine( fd, 0, index, item->refine );
+		if (info->broadcast_success) {
+			clif_broadcast_refine_result(*sd, item->nameid, item->refine, true);
+		}
 		if( id->type == IT_WEAPON ){
 			achievement_update_objective( sd, AG_ENCHANT_SUCCESS, 2, id->weapon_level, item->refine );
 		}
@@ -22415,6 +22418,9 @@ void clif_parse_refineui_refine( int fd, struct map_session_data* sd ){
 	}else{
 		// Failure
 
+		if (info->broadcast_failure) {
+			clif_broadcast_refine_result(*sd, item->nameid, item->refine, false);
+		}
 		// Blacksmith blessings were used to prevent breaking and downgrading
 		if( blacksmith_amount > 0 ){
 			clif_refine( fd, 3, index, item->refine );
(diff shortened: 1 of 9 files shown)


iberryRO:
WoW

src/map/status.cpp
```diff
@@ -230,6 +230,32 @@ uint64 RefineDatabase::parseBodyNode( const ryml::NodeRef& node ){
 						}
 					}
 
+					if (this->nodeExists(refineLevelNode, "BroadcastSuccess")) {
+						bool bcast = false;
```
Lemongrass3110, line 234:
```suggestion
						bool bcast;
```

src/map/status.cpp
```diff
@@ -230,6 +230,32 @@ uint64 RefineDatabase::parseBodyNode( const ryml::NodeRef& node ){
...
+						level_info->broadcast_success = bcast;
+					}
+					else {
+						if (!level_exists) {
+							level_info->broadcast_success = false;
+						}
+					}
+
+					if (this->nodeExists(refineLevelNode, "BroadcastFailure")) {
+						bool bcast = false;
```
Lemongrass3110, line 247:
```suggestion
						bool bcast;
```

src/map/clif.cpp
```diff
@@ -24693,6 +24699,29 @@ void clif_parse_itempackage_select( int fd, struct map_session_data* sd ){
...
+
+	if (battle_config.broadcast_hide_name) {
+		std::string dispname = std::string(sd.status.name);
+		int hide = min(battle_config.broadcast_hide_name, dispname.length() - 1);
+		dispname.replace(dispname.length() - hide, hide, hide, '*');
+		safestrncpy(packet.name, dispname.c_str(), NAME_LENGTH);
+	}
+	else {
+		safestrncpy(packet.name, sd.status.name, NAME_LENGTH);
+	}
```
Lemongrass3110, line 24719: Can we put this in a separate function?
As this is used in clif_broadcast_obtain_special_item and here (and maybe should also be used in grade announce)?

src/map/clif.cpp
```diff
@@ -24693,6 +24699,29 @@ void clif_parse_itempackage_select( int fd, struct map_session_data* sd ){
 #endif
 }
 
+void clif_broadcast_refine_result(map_session_data& sd, t_itemid itemId, int8 level, bool success)
+{
+#if PACKETVER_MAIN_NUM >= 20170906 || PACKETVER_RE_NUM >= 20170830 || defined(PACKETVER_ZERO)
+	PACKET_ZC_BROADCAST_ITEMREFINING_RESULT packet{};
```
Lemongrass3110, line 24705: Maybe rename from "packet" to "p" to match other functions.

Lemongrass3110 (approved):

src/map/clif.cpp
secretdataz: kRO does not hide the player's name. It could be an oversight, but I'm keeping it as is for now.

PR #7303, opened 2022-10-01
Start refactoring the core to C++

Lemongrass3110:

* **Server Mode**: Both

Started to refactor the core to make use of C++.


src/char/char.hpp
```diff
@@ -14,19 +14,30 @@
 
 #include "packets.hpp"
 
+using namespace rathena::server_core;
+
+namespace rathena{
+	namespace server_character{
```
vstumpf, line 20: just make this `namespace rathena::server_character`

also, can we choose not to indent in namespaces? Otherwise we gonna get wiiiiiiiiiiiiiiide

src/char/char.hpp
```diff
@@ -14,19 +14,30 @@
 
 #include "packets.hpp"
 
+using namespace rathena::server_core;
```
vstumpf, line 17: we shouldn't be doing this in a header file, we're polluting the global namespace of every file that includes `char.hpp`

instead, we can specifically import Core, with
`using rathena::server_core::Core`

vstumpf (changes requested):
repeat these for login/map/web

src/char/char.hpp
Lemongrass3110: As discussed on Discord - not yet possible.

aleos89 (approved):

Issue #7194, opened 2022-08-19
show debug message every time login game


idk-whoami:
* **rAthena Hash**: ae74eb840480d358c8250d35bd9d28c097794e28

* **Client Date**: 20211103

* **Server Mode**: re

* **Description of Issue**: 
```
every time login game
show debug message
```

![บันทึก_2022_08_19_22_16_50_504](https://user-images.githubusercontent.com/60462533/185652743-bdc0af56-e45d-41dd-9030-e42516d7cad8.gif)

```
[Debug]: [AccountID: 2000001, CharID: 150001, World: "rAthena"] Not found in table, sending new info.
```
  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

mazvi:
Are you runing web-server or not? related webservice? 

https://github.com/rathena/rathena/commit/0ac52d0bb49e858f7ff7774685bb0d18af6d1cf7

idk-whoami:
how to run web-server 0ac52d0bb49e858f7ff7774685bb0d18af6d1cf7

idk-whoami:
forgot to add description

run server on Ubuntu 21.04 machine

idk-whoami:
guild emblems works normally


![บันทึก_2022_08_20_21_43_38_485](https://user-images.githubusercontent.com/60462533/185752193-f24897fc-6858-440f-ad9c-22e44d15b989.gif)

```
[Debug]: [AccountID: 2000000, CharID: 150000, World: "rAthena"] Not found in table, sending new info.

**Comment by Akkarinage** [comment]
I'm unsure on why it does this, as I get it to.

As a temporary measure, you can use the `console_silent: ` option to hide the debug messages.

*config/import/web_conf.txt*
```
console_silent: 32
```

aleos89:
Maybe @vstumpf can provide some insight since it's his baby.

vstumpf:
I guess we can remove the log message.
For some reason the client isn't sending a save request when the client closes, so nothing ever gets added to the db

PR #7305, opened 2022-10-01
Removed a debug message in web-server

Lemongrass3110:
* **Addressed Issue(s)**: #7194

* **Server Mode**: Both

If a character/account logs in the first time, we silently store the default configuration in the foreseen tables now. Fixes the problem that the client does not call account configuration save in newer versions and therefore always triggers the debug message.

Thanks to @idk-whoami

### DIFF
--- a/src/web/charconfig_controller.cpp
+++ b/src/web/charconfig_controller.cpp
@@ -135,9 +135,26 @@ HANDLER_FUNC(charconfig_load) {
 
 	if (SqlStmt_NumRows(stmt) <= 0) {
 		SqlStmt_Free(stmt);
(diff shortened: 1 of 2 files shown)


ecdarreola:
After applying PR I'm receiving the error below

[Error]: Memory manager: args of aFree 0x0x7feeb2f74efc is overflowed pointer sql.cpp line 689

idk-whoami:
Just like me!

Capture
```
[Error]: Memory manager: args of aFree 0x0x7fbbc9a72ed4 is overflowed pointer sql.cpp line 689
```
and can't shutdown server

Capture

ecdarreola:
Works fine now. No more memory leaks. Thanks!

idk-whoami:
It works.

aleos89 (approved):

Akkarinage (approved):

PR #7307, opened 2022-10-01
Add cap_value script command

Emistry:
* **Addressed Issue(s)**: 
we have min/max but dont have cap_value

* **Server Mode**: 
both

return a value but capped between min and max.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -24130,7 +24130,7 @@ BUILDIN_FUNC(minmax){
 
 BUILDIN_FUNC(cap_value)
 {
-	script_pushint64(st, (int64)cap_value(script_getnum64(st, 2), script_getnum64(st, 3), script_getnum64(st, 4)));
+	script_pushint64(st, cap_value(script_getnum64(st, 2), script_getnum64(st, 3), script_getnum64(st, 4)));
 	return SCRIPT_CMD_SUCCESS;
 }
 
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -24130,12 +24130,7 @@ BUILDIN_FUNC(minmax){
 
 BUILDIN_FUNC(cap_value)
 {
-	int value = script_getnum(st, 2);
-	int min = script_getnum(st, 3);
-	int max = script_getnum(st, 4);
-
-	script_pushint(st, (int)cap_value(value, min, max));
-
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -24128,6 +24128,17 @@ BUILDIN_FUNC(minmax){
 	return SCRIPT_CMD_SUCCESS;
 }
 
+BUILDIN_FUNC(cap_value)
+{
+	int value = script_getnum(st, 2);
+	int min = script_getnum(st, 3);
+	int max = script_getnum(st, 4);
+
+	script_pushint(st, (int)cap_value(value, min, max));
```
aleos89, line 24137:
```suggestion
	script_pushint(st, (int)cap_value(script_getnum(st, 2), script_getnum(st, 3), script_getnum(st, 4)));
```

aleos89 (approved):

src/map/script.cpp
```diff
@@ -24128,6 +24128,12 @@ BUILDIN_FUNC(minmax){
 	return SCRIPT_CMD_SUCCESS;
 }
 
+BUILDIN_FUNC(cap_value)
+{
+	script_pushint(st, (int)cap_value(script_getnum(st, 2), script_getnum(st, 3), script_getnum(st, 4)));
+	return true;
```
vstumpf, line 24134: just a nitpick, can we make this `SCRIPT_CMD_SUCCESS`

vstumpf (approved):

src/map/script.cpp
```diff
@@ -24128,6 +24128,12 @@ BUILDIN_FUNC(minmax){
 	return SCRIPT_CMD_SUCCESS;
 }
 
+BUILDIN_FUNC(cap_value)
+{
+	script_pushint64(st, (int64)cap_value(script_getnum64(st, 2), script_getnum64(st, 3), script_getnum64(st, 4)));
```
Lemongrass3110, line 24133:
```suggestion
	script_pushint64(st, cap_value(script_getnum64(st, 2), script_getnum64(st, 3), script_getnum64(st, 4)));
```
Should not be necessary to cast here. You call the funciton with all variables being int64, so the macro should return an int64 as well. See:
```cpp
#define cap_value(a, min, max) (((a) >= (max)) ? (max) : ((a) <= (min)) ? (min) : (a))
```

Lemongrass3110 (approved):

vstumpf (approved):

PR #7311, opened 2022-10-01
Fixes a possible size mismatch for isspace()

### LINKED ISSUE #5928: isspace() could invoke unpredictable results.
* **rAthena Hash**: the latest version

* **Client Date**: 2021.05.10

* **Server Mode**: -

* **Description of Issue**: isspace() could invoke unpredictable results.
  * Result: isspace() could invoke unpredictable results.
  * Expected Result: isspace() returns predictable result.
  * How to Reproduce:  The parameter of the isspace() should be cast with 'unsigned int'.
  * Official Information: 
  -Affecting code path: /src/common/strlib.cpp (exists in the latest version)
-Function name: cchar *trim2(char *str,char flag)
-Line number: 158, 166
-Description:
isspace() could be invoked unpredictable results.
So, the parameter of the isspace() should be cast with 'unsigned int'. 
We found out that information through these sites. (Microsoft, Stack Overflow)
url: [https://docs.microsoft.com/en-us/cpp/code-quality/c6328?view=msvc-160]
url: [https://stackoverflow.com/questions/122721]

aleos89:
* **Addressed Issue(s)**: Fixes #5928

  * Passing a char type argument to isspace() could result in unpredictable results.
Thanks to @hyunji-Hong!

### DIFF
--- a/src/common/strlib.cpp
+++ b/src/common/strlib.cpp
@@ -155,7 +155,7 @@ char* trim(char* str)
 // value must NOT be deallocated using free() etc.
 char *trim2(char *str,char flag) {
 	if(flag&1) { // Trim leading space
-		while(isspace(*str)) str++;
+		while(isspace(static_cast<unsigned char>(*str))) str++;
 		if(*str == 0)  // All spaces?
 			return str;
 	}
@@ -163,7 +163,7 @@ char *trim2(char *str,char flag) {
 		char *end;
 
 		end = str + strlen(str) - 1;
-		while(end > str && isspace(*end)) end--;
+		while(end > str && isspace(static_cast<unsigned char>(*end))) end--;
 		*(end+1) = 0; // Write new null terminator
 	}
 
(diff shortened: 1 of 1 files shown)


src/common/strlib.cpp
```diff
@@ -155,15 +155,15 @@ char* trim(char* str)
 // value must NOT be deallocated using free() etc.
 char *trim2(char *str,char flag) {
 	if(flag&1) { // Trim leading space
-		while(isspace(*str)) str++;
+		while(isspace(static_cast<unsigned char>(*str))) str++;
```
Lemongrass3110, line 158:
```suggestion
		while(ISSPACE(*str)) str++;
```

src/common/strlib.cpp
```diff
@@ -155,15 +155,15 @@ char* trim(char* str)
...
 		if(*str == 0)  // All spaces?
 			return str;
 	}
 	if(flag&2) { // Trim trailing space
 		char *end;
 
 		end = str + strlen(str) - 1;
-		while(end > str && isspace(*end)) end--;
+		while(end > str && isspace(static_cast<unsigned char>(*end))) end--;
```
Lemongrass3110, line 166:
```suggestion
		while(end > str && ISSPACE(*end)) end--;
```

Lemongrass3110 (approved):
Our predecessors already thought of that and created a macro for this.

PR #7312, opened 2022-10-01
Adds battle config mob_respawn_time

aleos89:
* **Addressed Issue(s)**: N/A

  * Adjusts the minimum respawn time of a monster to be 1 second as it is officially.
  * Adds a battle config to allow for easy adjustment.
Thanks to @mrjnumber1!


src/map/mob.cpp
```diff
@@ -696,7 +696,7 @@ int mob_once_spawn(struct map_session_data* sd, int16 m, int16 x, int16 y, const
 					memcpy(md->guardian_data->guild_name, g->name, NAME_LENGTH);
 				}
 				else if (gc->guild_id) //Guild not yet available, retry in 5.
```
vstumpf, line 698: Change the comment here to `in one second`?

conf/battle/monster.conf
```diff
@@ -297,3 +297,7 @@ achievement_mob_share: no
 // Should slaves teleport back to their master if they get too far during chase? (Note 1)
 // Default (Official): no
 slave_stick_with_master: no
+
+// Absolute minimum respawn time of a monster.
```
vstumpf, line 301: in milliseconds?

vstumpf (approved):
just some nitpicks

conf/battle/monster.conf
Lemongrass3110: Agree that the units should be made clear!

src/map/mob.cpp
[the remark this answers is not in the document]
aleos89: I'm only going off an assumption here that this timer was a copy from `mob_spawn_guardian` which is called via the script which also most likely copied the `mob_once_spawn` delay. That was the reasoning of me adding the config in each of these places.

Atemo:
I agree with this change but I haven't seen a respawn timer below 5000ms in the Aegis file where permanent mobs are defined. How about keeping our default at 5000ms in `npc_parse_mob` when no value is passed so we don't have to modify our permanent monsters (minimum will always be 1000ms but default 5000ms)?

src/map/npc.cpp
```diff
@@ -5181,6 +5181,12 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
 		return strchr(start,'\n');// skip and continue
 	}
 
+	if (delay != 5000 && (delay < 1000 || delay > UINT_MAX)) {
```
Lemongrass3110, line 5184: delay is an int32 so it can never be > UINT32_MAX.
Maybe just make it delay < battle_config.mob_respawn_time.
Also to check if delay was given correctly you need to introduce w4count, like we have w1count.
Otherwise it allows setting smaller delays even if you set the battle_config to 10 seconds for example.

conf/battle/monster.conf
```diff
@@ -297,3 +297,7 @@ achievement_mob_share: no
 // Should slaves teleport back to their master if they get too far during chase? (Note 1)
 // Default (Official): no
 slave_stick_with_master: no
+
+// Absolute minimum respawn time in milliseconds of a monster.
```
Lemongrass3110, line 301: Maybe leave a small comment that it is also used to delay spawning of guardians when the guild is not yet loaded to retry every x seconds?

Lemongrass3110 (approved):

Lemongrass3110:
CI complains about PRE:
```
2022-12-12T21:03:33.1323911Z [Error]: npc_parse_mob: Invalid delay 0 for mob ID 2014 (file 'npc/pre-re/mobs/dungeons/nyd_dun.txt', line '17').
2022-12-12T21:03:33.1330067Z [Error]: npc_parse_mob: Invalid delay 0 for mob ID 1101 (file 'npc/pre-re/mobs/dungeons/prt_maze.txt', line '87').
```

PR #7313, opened 2022-10-01
Adds Cancel node to status database

aleos89:
* **Addressed Issue(s)**: N/A

  * Any status defined under this node will be removed when the status becomes inactive.
  * Removes some hard coded statuses.

### DIFF
--- a/db/import-tmpl/status.yml
+++ b/db/import-tmpl/status.yml
@@ -37,8 +37,9 @@
 #   Fail:                     List of Status Changes that causes the status to fail to activate. (Optional)
 #   End:                      List of Status Changes that will end when the status activates. (Optional)
 #   EndReturn:                List of Status Changes that will end when the status activates and won't give its effect. (Optional)
+#   Cancel:                   List of Status Changes that will be canceled when the status becomes inactive. (Optional)
 ###########################################################################
 
 Header:
   Type: STATUS_DB
-  Version: 2
+  Version: 3
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -13044,11 +13044,6 @@ int status_change_end(struct block_list* bl, enum sc_type type, int tid)
 						((TBL_MER*)d_bl)->devotion_flag = 0;
 					clif_devotion(d_bl, NULL);
 				}
-
-				status_change_end(bl, SC_AUTOGUARD);
-				status_change_end(bl, SC_DEFENDER);
-				status_change_end(bl, SC_REFLECTSHIELD);
-				status_change_end(bl, SC_ENDURE);
 			}
 			break;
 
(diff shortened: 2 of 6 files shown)


Akkarinage (approved):

db/import-tmpl/status.yml
```diff
@@ -37,8 +37,9 @@
 #   Fail:                     List of Status Changes that causes the status to fail to activate. (Optional)
 #   End:                      List of Status Changes that will end when the status activates. (Optional)
 #   EndReturn:                List of Status Changes that will end when the status activates and won't give its effect. (Optional)
+#   Cancel:                   List of Status Changes that will be canceled when the status becomes inactive. (Optional)
```
Lemongrass3110, line 40: Cancel somewhat sounds ambigous.
People that dont read closely will think "End" and "Cancel" is the same.
Dont really have a better suggestion though. :-(

src/map/status.cpp
```diff
@@ -13495,16 +13455,21 @@ int status_change_end(struct block_list* bl, enum sc_type type, int tid)
...
 			if( sd ){
 				pc_delabyssball( *sd, sd->abyssball );
 			}
 			break;
 	}
 
+	// End statuses found in the Cancel list.
+	if (!scdb->cancel.empty()) {
+		for (const auto &it : scdb->cancel) {
+			if (it == type)
```
Lemongrass3110, line 13468: This if looks weird.
scdb is already the database link for this specific SC.
So the list should be all SCs that should be cancelled and those should most likely be other SCs.
So maybe you simply inverted the check, because you wanted to prevent infinity loops?
But even infinity loops cant be a thing, because sce is already set to NULL at this point.

src/map/status.hpp
```diff
@@ -2950,13 +2950,14 @@ struct s_status_change_db {
...
 	std::vector<sc_type> endreturn;	///< List of SC that will be ended when this SC is activated and then immediately return
+	std::vector<sc_type> cancel;	///< List of SC that will be canceled when this SC ends
 	t_tick min_duration;			///< Minimum duration effect (after all status reduction)
 	uint16 min_rate;				///< Minimum rate to be applied (after all status reduction)
 };
 
 class StatusDatabase : public TypesafeCachedYamlDatabase<uint16, s_status_change_db> {
 public:
-	StatusDatabase() : TypesafeCachedYamlDatabase("STATUS_DB", 2) {
+	StatusDatabase() : TypesafeCachedYamlDatabase("STATUS_DB", 3) {
```
Lemongrass3110, line 2960:
```suggestion
	StatusDatabase() : TypesafeCachedYamlDatabase("STATUS_DB", 3, 2) {
```
It is not really a breaking change, right?

Lemongrass3110 (approved):

db/import-tmpl/status.yml
aleos89: Agreed, I thought about the confusion as I was working on this. I'll see if I can come up with a better label name.

src/map/status.hpp
aleos89: It's not breaking for it to be missing but I was 50/50 on this since this could break current skill behavior if it's missing, but maybe it's better to make it like you suggested.

PR #7315, opened 2022-10-01
Adds support for macro detection

### LINKED ISSUE #6297: Adding macro detection features
Macro features coming with clients 2016+

https://github.com/HerculesWS/Hercules/pull/3051/commits/03c54ee9ac1fc6f9fa4e3425a92ccf16df229d33

aleos89:
* **Addressed Issue(s)**: Fixes #6297

* **Server Mode**: Renewal

  * Adds the official client macro detection system.
  * Includes the ability to load imagery at server boot.
Thanks to @Asheraf!

Here is the list of client commands used for this feature:
 * `/macro_register`: Allows authorized accounts (with `macro_register` group permission) to upload captcha images to the server.
 * `/macro_detector`: Allows authorized accounts (with `macro_detect` group permission) to send captcha requests to a single player or as a batch.
 * `/macro_preview <captcha_id>`: Allows for authorized accounts (with `macro_register` group permission) to preview the saved images in the captcha database.


conf/groups.yml
```diff
@@ -216,6 +216,7 @@ Body:
       hack_info: true
       any_warp: true
       view_hpmeter: true
+      macro_detect: true
```
ecdarreola, line 219: Causes crash on windows when running map-server it stops in groups.yml [6/8]

Lemongrass3110: We are working on the feedback already. Wait a little before posting further feedback and testing.

lightnow: Me to.

Badarosk0:
Has anyone tested with 2021-11-17_Ragexe? The captcha image is not loaded.

Pokye:
> Has anyone tested with 2021-11-17_Ragexe? The captcha image is not loaded. I had tested it on version 2020-09-02bRagexeRE and it had worked before.

image
Working perfectly on 2021-11-17 and 2022-03-30~31

You need to check the exact weight of the image, and add the value in the client and in pc.hpp
image
In my case the value would be like this.
```
#define CAPTCHA_BMP_SIZE (4 + 52 + (3 * 220 * 90))

File: src/map/pc.cpp [hunk 2]
```diff
@@ -14989,6 +14994,364 @@ void pc_attendance_claim_reward( struct map_session_data* sd ){
...
+
+		if (answer.length() < 4 || answer.length() > CAPTCHA_ANSWER_SIZE) {
+			this->invalidWarning(node["Answer"], "The captcha answer must be between 4~%d characters, skipping...", CAPTCHA_ANSWER_SIZE);
+			return 0;
+		}
+
+		safestrncpy(cd->captcha_answer, answer.c_str(), sizeof(cd->captcha_answer));
+	}
+
+	safesnprintf(cd->captcha_key, sizeof(cd->captcha_key), "%X", static_cast<int32>(captcha_db.size() + 1));
```
Lemongrass3110, line 15329: Still not sure if we shouldnt remove this.
The whole captcha_key thing is useless.

src/map/pc.cpp
```diff
@@ -14989,6 +14994,364 @@ void pc_attendance_claim_reward( struct map_session_data* sd ){
...
+		}
+
+		if (i == (0xFFF + 1)) {
+			// no free key found...
+			sd.captcha_upload.cd = nullptr;
+			sd.captcha_upload.upload_size = 0;
+			return;
+		}
+
+		safesnprintf(sd.captcha_upload.cd->captcha_key, sizeof(sd.captcha_upload.cd->captcha_key), "%03X", i);
```
Lemongrass3110, line 15063: Still thinking if we should remove all this. It is actually useless.
I mean free id check yes, but limiting us with captcha_key could be removed.

src/map/pc.cpp
```diff
@@ -14989,6 +14994,364 @@ void pc_attendance_claim_reward( struct map_session_data* sd ){
...
+ * @param image_size: Captcha image size
+ * @param captcha_answer: Answer to captcha
+ */
+void pc_macro_captcha_register(map_session_data &sd, int16 image_size, char captcha_answer[CAPTCHA_ANSWER_SIZE]) {
+	nullpo_retv(captcha_answer);
+
+	sd.captcha_upload.cd = nullptr;
+	sd.captcha_upload.upload_size = 0;
+
+	if (strlen(captcha_answer) < 4 || image_size < 0 || image_size > CAPTCHA_BMP_SIZE || captcha_db.size() >= 0xFFF) {
```
Lemongrass3110, line 15009:
```suggestion
	if (strlen(captcha_answer) < 4 || image_size < 0 || image_size > CAPTCHA_BMP_SIZE) {
```
Still thinking if we should remove all this. It is actually useless.
We only limit ourselves with this captcha_key.

src/map/clif.cpp
```diff
@@ -24626,6 +24632,270 @@ void clif_broadcast_refine_result(map_session_data& sd, t_itemid itemId, int8 le
...
+
+	clif_send(&p, sizeof(p), &sd.bl, SELF);
+#endif
+}
+
+void clif_parse_captcha_preview_request(int fd, map_session_data *sd) {
+#if PACKETVER >= 20160323
+	nullpo_retv(sd);
+
+	if (!(pc_has_permission(sd, PC_PERM_MACRO_REGISTER) && pc_has_permission(sd, PC_PERM_MACRO_DETECT))) {
```
Lemongrass3110, line 24703:
```suggestion
	if (!pc_has_permission(sd, PC_PERM_MACRO_REGISTER)) {
```
That was my mistake during our session!

src/map/pc.hpp
```diff
@@ -120,6 +120,56 @@ enum e_additem_result : uint8 {
 	ADDITEM_STACKLIMIT
 };
 
+#define CAPTCHA_ANSWER_SIZE 16
```
Lemongrass3110, line 123:
```suggestion
#ifndef CAPTCHA_ANSWER_SIZE
	#define CAPTCHA_ANSWER_SIZE 16
#endif
```

src/map/pc.hpp
```diff
@@ -120,6 +120,56 @@ enum e_additem_result : uint8 {
...
+#define CAPTCHA_KEY_SIZE 4
```
Lemongrass3110, line 124:
```suggestion
#ifndef CAPTCHA_KEY_SIZE
	#define CAPTCHA_KEY_SIZE 4
#endif
```

src/map/pc.hpp
```diff
@@ -120,6 +120,56 @@ enum e_additem_result : uint8 {
...
+#define CAPTCHA_BMP_SIZE (2 + 52 + (3 * 220 * 90)) // sizeof("BM") + sizeof(BITMAPV2INFOHEADER) + 24bits 220x90 BMP
```
Lemongrass3110, line 125:
```suggestion
#ifndef CAPTCHA_BMP_SIZE
	#define CAPTCHA_BMP_SIZE (2 + 52 + (3 * 220 * 90)) // sizeof("BM") + sizeof(BITMAPV2INFOHEADER) + 24bits 220x90 BMP
#endif
```

src/map/pc.hpp
```diff
@@ -120,6 +120,56 @@ enum e_additem_result : uint8 {
...
+#define MAX_CAPTCHA_CHUNK_SIZE 1024
```
Lemongrass3110, line 126:
```suggestion
#ifndef MAX_CAPTCHA_CHUNK_SIZE
	#define MAX_CAPTCHA_CHUNK_SIZE 1024
#endif
```

src/map/pc.hpp
```diff
@@ -120,6 +120,56 @@ enum e_additem_result : uint8 {
...
+
+class CaptchaDatabase : public TypesafeYamlDatabase<int16, s_captcha_data> {
+public:
+	CaptchaDatabase() : TypesafeYamlDatabase("CAPTCHA_DB", 1) {
+
+	}
+
+	const std::string getDefaultLocation() override;
+	uint64 parseBodyNode(const ryml::NodeRef &node) override;
+	void clear() override {};
```
Lemongrass3110, line 168:
```suggestion
```

doc/permissions.txt
```diff
@@ -212,3 +212,15 @@ Allow to bypass the maximum stat parameter (at conf/player.conf) to
 maximum value 32,767.
 
 ---------------------------------------
+
+*macro_detect
```
Lemongrass3110, line 216: Dont we usually turn `/commands` into `@commands` instead of permissions?

Lemongrass3110 (commented):
Overall looks really good now already!

What I really dont like about this system is that by using this "captcha_key", which is a static value it makes it useless. The bots/players can create a dictionary for each server, so we should really consider not sending this out and only keep it inside server side state tracking instead.
By doing so we would also be able to remove some limitations.

Pokye:
Is there a way to add an option to send to jail instead of banning?

Lemongrass3110:
> Is there a way to add an option to send to jail instead of banning?

We want to add this in V2 of the yaml file. We wanted to release the feature first though.
Or do you think this is a must have for the release?

Badarosk0:
> > Is there a way to add an option to send to jail instead of banning?
> 
> We want to add this later on. We wanted to release the feature first though. Or do you think this is a must have for the release?

Could add to call a script function. So it could be easier for anyone to configure whatever they want, jail, buff, incentives.

PR #7024, opened 2022-06-13
4th / Expanded Job Skill-Updates as of KRO 20220602 incl. 3rd Job fixes / Homunculus skills

datawulf:
* **Server Mode**: Renewal

* **Description of Pull Request**: 

This shall bring all skills to the most recent official state.
4th / Expanded Job Skills, but also some 3rd Jobs skills as well as the 200+ Homunculus skills are included.

This PR's focus is solely on non-NPC skill functionality.

There are many people out there trying to distribute job-skills, but neither are they willing to update / fix them nor try to be accurate with official behaviours. Now there's no guarantee that everything is 100% accurate, but if you have additional official informations I'll gladly update them. Nobody is perfect, right?

As sources I used:
* already available PRs / projects
* replay files from sakray server
* youtube KRO videos
* most recent KRO skill descriptions / patch notes
* thanks @limitro for some fundamental mechanics

Note: The new skills of Dragon Knight (Dragonic Breath), Biolo (Explosive Powder, Mayhemic Thorns), Meister (Mighty Smash, Spark Blaster, Triple Laser) need a client 20220602+.
<details>
  <summary>New skills showcase</summary>

biolo
dragon
meister
</details>

I'm happy to get feedback on the code itself as I'm by no means a rathena/c++-pro and tried to use what was used before or what I thought would be a good solution.

So let's make this great hopefully :)

<details>
  <summary>"Full" changelog</summary>

## General Code Changes
* harmonized order of cardfix bonus applied for physical/magical attacks (according to IRO wiki)
* reworked cardfix for physical attacks to apply proper bonus for melee/ranged attack types, because there are ranged attacks without sd->state.arrow_atk
* updated ST_SHIELD skill fail message
* updated require.weapon skill fail messages
* added new skill fail causes (useskill_fail_cause)
* added flag to va_args in skill_detonator used for UNT_KUNAIWAIKYOKU 
* added new function skill_check_bl_sc to check SC of nearby BLs using va_list
* fixes status_calc_pc_sub to check battle_config.left_cardfix_to_right for left/right hand bonus

### added new global flags for skill controls:
* SKILL_ALTDMG_FLAG: to control alternative skill scalings
* SKILL_NOCONSUME_REQ: make skill ignore requirement consumption
* UNIT_NOCONSUME_AMMO: make skill consume ammo, but not the unit

#### HAMI_DEFENCE:
* fixed AfterCastActDelay/Cooldown 

#### added new Skills:
6048:
* MH_CLASSY_FLUTTER
[...]

PR #7328, opened 2022-10-06
Fixes SA_DRAGONOLOGY MATK behavior

aleos89:
* **Addressed Issue(s)**: N/A

  * The MATK bonus should only be doubled versus Dragons, not quadrupled like the other bonuses.
  * Confirmed the behaviors between pre-renewal and renewal.
Thanks to @mrjnumber1!


Badarosk0:
I think this has already been fixed in this PR:
https://github.com/rathena/rathena/pull/7024

aleos89:
Looks like it is, but it's still a little wrong as the RENEWAL macro is not needed there.

Lemongrass3110 (approved):

PR #7333, opened 2022-10-07
Fixes Marine Sphere behavior

### LINKED ISSUE #7332: Alchemist Marine Sphere Mine Broken
### rAthena Hash

b568f2fb119f247acba459d04f95f7e91c939e69

### Client Date

2022-04-06

### Server Mode

Pre-Renewal, Renewal

### Result

The alchemist skill "Summon Marine Sphere" is not working as intended. It is not moving when attacked and it does not even begin casting the skill to self-destruct.

### Relevant Log Output

_No response_

### Expected Result

Marine Sphere mine should move 7 cells from the attackers position at the moment of the attack the player is facing and the mine should cast it's skill simultaenously.

### How to Reproduce

Plant a marine sphere using alchemist skill.

### Official Information

Can't find any sources but I think what I said is the default behavior of the skill.

none

aleos89:
* **Addressed Issue(s)**: Fixes #7332

  * Resolves an issue where Marine Sphere were not casting their skills after being hit.
Thanks to @violent01 and @Daegaladh!


Daegaladh (approved):

violent01:
@aleos89 the skill is working now as intended but is the marine movement goes always to the west? Is that intentional?

Daegaladh:
> @aleos89 the skill is working now as intended but is the marine movement goes always to the west? Is that intentional?

It should go in the opposite direction you hit it

violent01:
@Daegaladh that's what i remember also. But it always goes to west instead of the opposite direction of the player hit it. It feels like it's like the cart revo skill which forces the mob to go to west.

violent01:
@aleos89 this f694f6142843a53f76c875e274bdab5046a6fcae commit works like a charm now.

PR #7347, opened 2022-10-11
Removed a comment that suggested treating SC_EXTREMITYFIST2 specially

### LINKED ISSUE #7346: SC_EXTREMITYFIST2 and statuses

No

When adding a custom status, you always want to add it before SC_MAX to avoid shifting all the entries that follow. Yet, the comment after SC_EXTREMITYFIST2 tells you explicitly not to do that.
```
#ifdef RENEWAL
	SC_EXTREMITYFIST2, //! NOTE: This SC should be right before SC_MAX, so it doesn't disturb if RENEWAL is disabled
#endif
```
Originally this request was meant towards custom statuses, but it appears rAthena also follows this comment when adding new official statuses. Yet, I believe this should be avoided. Here's an example:

- A player logs out while having the SC_EXTREMITYFIST2 status (let's pretend the value is 1500), and SC_MAX is 1501.
- You do an update, you add statuses, SC_PORK_RIB_STEW is now above SC_EXTREMITYFIST2.
- Their new IDs will now become SC_PORK_RIB_STEW = 1500, and SC_EXTREMITYFIST2 = 1501, and SC_MAX is 1502.
- When the player logs back in, instead of having SC_EXTREMITYFIST2, he will now have SC_PORK_RIB_STEW as his status.

Granted this won't last for long, since SC_EXTREMITYFIST2 usually has a short duration, it can create weird situations or straight up potential bug/crash for the player having a status they shouldn't have.

I would consider simply removing the #ifdef before SC_EXTREMITYFIST2 and remove the comment next to it. As far as I could tell, it is the only status defined in such a way. Meanwhile, there are numerous renewal only statuses defined even for pre-renewal, so it doesn't make much sense to do an exception for this one.

### Additional context

_No response_

secretdataz:
* **Addressed Issue(s)**: Resolves #7346

* **Server Mode**: Both

* **Description of Pull Request**: Removed a comment that suggested treating SC_EXTREMITYFIST2 specially

### DIFF
--- a/src/map/status.hpp
+++ b/src/map/status.hpp
@@ -1254,10 +1254,7 @@ enum sc_type : int16 {
 	SC_SKF_CAST,
 	SC_BEEF_RIB_STEW,
 	SC_PORK_RIB_STEW,
-
-#ifdef RENEWAL
-	SC_EXTREMITYFIST2, //! NOTE: This SC should be right before SC_MAX, so it doesn't disturb if RENEWAL is disabled
-#endif
+	SC_EXTREMITYFIST2,
 	SC_MAX, //Automatically updated max, used in for's to check we are within bounds.
 };
 


Lemongrass3110 (changes requested):
As discussed in discord this only fixes half of the issue.
We should either reserve a range, where to add custom stuff or finally fix saving them (by name instead of ID).

Additionally we should keep in mind that we should refactor this so that pre-renewal does not have too much SCs that are not used, but this does not have to be part of this PR.

aleos89:
Also, the script_constants.hpp has a macro wrap!

Playtester:
Was fixed in 82c37ce already.

PR #7352, opened 2022-10-12
Added support mob name as parameter in mob spawn definition

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

This is a suggestion to support mob name as parameter in mob spawn definition.

Example
----------------------------------------------------------------------------------
Currently permanent monster can be defined with the mob ID only :
`tra_fild,57,159	monster	Dummy (Small)	21064,1`

With this PR (mob ID is still supported)
`tra_fild,57,159	monster	Dummy (Small)	S_DUMMY_100_SMALL,1`

### DIFF
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -5053,7 +5053,7 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
 {
 	int num, mob_id, mob_lv = -1, size = -1, w1count;
 	short m, x = 0, y = 0, xs = -1, ys = -1;
-	char mapname[MAP_NAME_LENGTH_EXT], mobname[NAME_LENGTH];
+	char mapname[MAP_NAME_LENGTH_EXT], mobname[NAME_LENGTH], sprite[NAME_LENGTH];
 	struct spawn_data mob, *data;
 	int ai = AI_NONE; // mob_ai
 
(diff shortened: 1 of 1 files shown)


src/map/npc.cpp
```diff
@@ -5089,6 +5089,17 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
...
 
+	// Check if sprite is the mob name or ID
+	char *pid;
+	mob_id = strtol(sprite, &pid, 0);
+
+	if (pid != nullptr && *pid != '\0') {
+		std::shared_ptr<s_mob_db> mob = mobdb_search_aegisname(sprite);
+
+		if (mob != nullptr)
+			mob_id = mob->id;
```
Lemongrass3110, line 5100: Should be inverted and an error message with %s should be used, the other one is %d.

src/map/npc.cpp
```diff
@@ -5089,6 +5089,17 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
 		return strchr(start,'\n');// skip and continue
 	}
 
+	// Check if sprite is the mob name or ID
+	char *pid;
```
Lemongrass3110, line 5093:
```suggestion
	char *pid;
	sprite[NAME_LENGTH-1] = '\0';
```
sprite is uninitialized and could therefore be not zero terminated.

src/map/npc.cpp
```diff
@@ -5089,6 +5089,17 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
...
+	mob_id = strtol(sprite, &pid, 0);
+
+	if (pid != nullptr && *pid != '\0') {
+		std::shared_ptr<s_mob_db> mob = mobdb_search_aegisname(sprite);
+
+		if (mob != nullptr)
+			mob_id = mob->id;
+	}
```
Lemongrass3110, line 5101: Turn the next if below into an else case of this.

Lemongrass3110 (approved):

PR #7381, opened 2022-10-23
Update of item ID 5054 according to kRO

Pokye:

* **Server Mode**: PRE/RENEWAL

* **Description of Pull Request**: Recently, kRO updated item id 5054 (Assassin Mask) to (Popcorn_Hair), I added the item information, and changed the information I used before item id 5054 (https://www.divine-pride.net/database/item/5054/ ) to 5096 ( https://www.divine-pride.net/database/item/5096/ )


db/pre-re/item_db_equip.yml
```diff
@@ -25687,19 +25687,15 @@ Body:
...
-      Assassin: true
-      Priest: true
+    Weight: 300
     Locations:
-      Head_Low: true
+      Head_Top: true
     ArmorLevel: 1
-    EquipLevelMin: 70
-    View: 180
+    View: 1895
```
Atemo, line 25698: For me, the item shouldn't be changed in pre-re. Whoever wants to use pre-re should also use a pre-re client because there is a lot of change between the items description in pre-re and re

Pokye: but item ID 5054 affects both **custom/itemmail.txt** and **quests/seals/sleipnir_seal.txt** as both are used in PRE/RE

PR #7404, opened 2022-11-04
Removed the strange ACK function concept

Lemongrass3110:

* **Server Mode**: Both

Cleans up packet database a little more.
Turned some more packets into structs.


src/map/status.cpp
```diff
@@ -12815,8 +12815,9 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			break;
 		case SC_C_MARKER:
 			//Send mini-map, don't wait for first timer triggered
-			if (src->type == BL_PC && (sd = map_id2sd(src->id)))
-				clif_crimson_marker(sd, bl, false);
+			if( src->type == BL_PC && ( sd = (struct map_session_data*)( src ) ) ){
+				clif_crimson_marker( *sd, *bl, false );
+			}
```
vstumpf, line 12820: we don't use sd later, so can we just do something like
```suggestion
			if (src->type == BL_PC) {
				clif_crimson_marker(*(struct map_session_data *)(src), *bl, false);
			}
```
imo it's cleaner, but it's up to you

Lemongrass3110: Dont ask me why, but it is used below right now, but i dont think it is correct XD
I like your change better

aleos89 (approved):

PR #7413, opened 2022-11-07
Update warper npc

idk-whoami:
* **Addressed Issue(s)**: 

* **Server Mode**: re

* **Description of Pull Request**: updating the warper npc


Atemo (approved):

vstumpf:
Personally I don't think we should keep the labels sorted, just add to the end. We can keep the menu options sorted, 
Can we change the registered warps at the bottom of the file too?

Lemongrass3110:
> Personally I don't think we should keep the labels sorted, just add to the end. We can keep the menu options sorted, Can we change the registered warps at the bottom of the file too?

Agree with @vstumpf here. There is no use of sorting them.
This is not necessary and there are a lot of changes in this PR that hide the actual changes because of this.

Personally I do not even understand why it is making use of labels, after making everything so dynamic and everything is maintained via functions and arrays and then for the core feature we go back to a hardcoded menu? O_o
But this is not a problem of this PR.

Atemo:
Checked and tested, it's good for me so I took the initiative to merge it.
Thank you for your work !

Issue #7420, opened 2022-11-09
DynamicNPC shouldn't use in instance dungeon


attackjom:
### Is your request related to a missing official feature?

Yes

### Describe the solution you'd like

according to this patch note -> https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7584&curpage=6

`- In all MD-type dungeons, a separate NPC cannot be called due to items, etc. is modified.`

dynamic npc can't use in instance dungeon.

PR #7429, opened 2022-11-10
Block dynamic NPCs in instances

Lemongrass3110:
* **Addressed Issue(s)**: #7420

* **Server Mode**: Both

Adds a mapflag to block the use of dynamic NPCs on certain maps.

Thanks to @attackjom


aleos89 (approved):

idk-whoami:
I think if there is no reply to the client and no mapserver message should be displayed.

image

```
[Debug]: Source (NPC): FAKE_NPC (invisible/not on a map)
[Debug]: Source (NPC): FAKE_NPC is located in: (null)
[Debug]: Source (NPC): FAKE_NPC (invisible/not on a map)
[Debug]: Source (NPC): FAKE_NPC is located in: (null)
[Debug]: Source (NPC): FAKE_NPC (invisible/not on a map)
[Debug]: Source (NPC): FAKE_NPC is located in: (null)
```

Atemo:
This debug is probably because the script command `duplicate_dynamic` return `SCRIPT_CMD_FAILURE` when the player already have a dynamic npc on
https://github.com/rathena/rathena/blob/master/src/map/script.cpp#L25233

idk-whoami:
I fixed /src/map/script.cpp follow up to #7415 it works
Thanks to @Atemo

Issue #758, opened 2015-11-16
Hp displayed when dead


Atemo:
Hash : e27697072317820cab4ef1e551ec50de46fc73a7

On iro/euro the client display 1 hp when you have died (on rathena -> 0 hp).

Image on left, with rathena ; offi on right
![comparaision](https://cloud.githubusercontent.com/assets/5231893/11194170/e65eacea-8caa-11e5-9678-8e678d28873c.png)

Lemongrass3110:
@RagnarokNova could you confirm this?

RagnarokNova:
http://i.imgur.com/RHCp2U6.png
Confirmed on iRO, checking kRO now.

lordganja:
due to this update. is there any work around to check if the player is dead?

Atemo:
This update only affected the hp displayed by the client. Inside server the HP of dead player is still 0.

I will update the docs but I think it will be a bit confusing, specially for those who don't read the docs.. :p We may implement a new command script or use a const to check if the player is dead Just an idea~.

lordganja:
Ohh. Alright. Didn't know that XD. So if we check the hp using a script, it will show 0 hp while we see it as 1?

And yeah a new command script to check if player is dead will really be helpful.

kyeme:
Novice too :D

![image](https://cloud.githubusercontent.com/assets/4656578/11374284/3977a542-9310-11e5-8c9a-8b759aecaa02.png)

Atemo:
Hmm yes :p

icxbb-xx:
@kyeme , you can post screen shot for  high novice and super novice when die
thank you

icxbb-xx:
Fixed in https://github.com/rathena/rathena/commit/54136ddfca421546a6cb84298d212995061829bd
Try check again , Thank you

PR #7434, opened 2022-11-14
Suggestion : pc_isdead script command

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: both

Simple script command to check if a player is dead.
The `readparam(Hp)` script command currently does the job but due to the visual on the client (https://github.com/rathena/rathena/issues/758), the result of this command may be misleading. As such, it seems a good idea to implement a command to clarify if a player is dead or not.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11250,3 +11250,9 @@ Returns fame rank (start from 1 to MAX_FAME_LIST), else 0.
 Note: Only works with classes that use the ranking system.
 
 ---------------------------------------
+
+*pc_isdead({<char id>})
+
+Returns 1 if the player is dead else 0.
+
+---------------------------------------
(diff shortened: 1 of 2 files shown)


aleos89 (approved):

doc/script_commands.txt
```diff
@@ -11250,3 +11250,9 @@ Returns fame rank (start from 1 to MAX_FAME_LIST), else 0.
 Note: Only works with classes that use the ranking system.
 
 ---------------------------------------
+
+*pc_isdead({<char id>})
+
+Returns 1 if the player is dead else 0.
```
Lemongrass3110, line 11256:
```suggestion
Returns true if the player is dead else false.
```

Lemongrass3110 (approved):
Since we do everything for "units" lately, shouldnt we rather introduce a unit_isdead and forward the result of status_isdead?

Atemo:
For non-player cases even though the idea is different the command would work like unitexists since the unit is removed, right? Or are there other units that aren't removed when they die?
https://github.com/rathena/rathena/blob/03f04ed507ef09c84224bbe1e404b5d1c6939ce8/doc/script_commands.txt#L8481

Issue #7447, opened 2022-11-19
Some random option appear invisible on item


Abszoluto:
### rAthena Hash

https://github.com/rathena/rathena/tree/master

### Client Date

2021-11-03 

### Server Mode

Pre-Renewal, Renewal

### Result

When I use random options randomly dropping the amount of options, some options are in random slots.
This bug happens when slots are filled with space between them.
Example:
Item:
- Random option in slot 0
- Random option in slot 2

The random option in slot 2 will be invisible. The item will say it has 2 random options but the random option in slot 2 will be invisible.

Image:

![image](https://user-images.githubusercontent.com/45849310/202864485-0f8d5f44-7be0-4b70-aea4-f7dda2f991e9.png)
![image](https://user-images.githubusercontent.com/45849310/202864494-e6527dbe-b27c-4d94-93b0-e3c507ec96e8.png)

### Expected Result

N/A

### How to Reproduce

At the item_randomopt_group.yml, use the following configuration on a group (to get the number of options random, instead of item always drop with 3 options):
![image](https://user-images.githubusercontent.com/45849310/202864632-e3af74bf-9b1f-44f2-bc7e-aad1b950d1b2.png)

### Official Information

N/A

### Modifications that may affect results

Make random options appear in consecutive slots, for example:
1 random option: On slot 0
2 random options: On slot 0 and slot 1
3 random options: On slot 0, slot 1 and slot 2
... etc

LadyNanuia:
This is also controlled client side, you need to make sure the LUA files matches in the client or it'll be bugged, not sure if this is what youre running into.

Abszoluto:
All my client files and server files are fine. I talked to Lemongrass and Stolao on discord and we found out that the issue is that the options need to be on consecutive slots.
In the image, i changed on sql and put a random option on the slot missing and the 3 options appeared normally

LadyNanuia:
> All my client files and server files are fine. I talked to Lemongrass and Stolao on discord and we found out that the issue is that the options need to be on consecutive slots. In the image, i changed on sql and put a random option on the slot missing and the 3 options appeared normally

oh i see, im sorry then i was just trying to help, i had this issue with my server with client side so i thought it might be similiar but nvm then ^^

Abszoluto:
[...]

PR #7449, opened 2022-11-20
Fixed gaps in random options

Lemongrass3110:
* **Addressed Issue(s)**: #7447

* **Server Mode**: Both

Fixed the gaps in random options by ensuring that after applying a random option group that empty slots will be filled up with options after the gap.
Added a rather complicated update script, as I could not think of an easier way to implement the same "for loop" without a stored procedure. Open for suggestions if someone has an easier idea.

Thanks to @Abszoluto


Tokeiburu:
Not sure if this was considered, but the way we resolved that issue was to skip empty random options when displaying them to the client, on the packet. This was useful since we didn't have to care about gaps anymore. So something like this:

```
static uint8 clif_add_random_options( struct ItemOptions buf[MAX_ITEM_RDM_OPT], struct item* it ){
	nullpo_retr( 0, it );

	uint8 count = 0;
	
	memset(buf, 0, MAX_ITEM_RDM_OPT * 5);
	
	for( int i = 0; i < MAX_ITEM_RDM_OPT; i++ ){
		if( it->option[i].id ){
			buf[count].index = it->option[i].id;	// OptIndex
			buf[count].value = it->option[i].value;	// Value
			buf[count].param = it->option[i].param;	// Param1
			count++;
		}
	}
	
	return count;
}
```

vstumpf:
> Not sure if this was considered, but the way we resolved that issue was to skip empty random options when displaying them to the client, on the packet. This was useful since we didn't have to care about gaps anymore. 

This is what I was thinking too, we kind of squish the empty space out

Lemongrass3110:
> Not sure if this was considered, but the way we resolved that issue was to skip empty random options when displaying them to the client, on the packet. This was useful since we didn't have to care about gaps anymore. So something like this:
> 
> ```
> static uint8 clif_add_random_options( struct ItemOptions buf[MAX_ITEM_RDM_OPT], struct item* it ){
> 	nullpo_retr( 0, it );
> 
> 	uint8 count = 0;
> 	
> 	memset(buf, 0, MAX_ITEM_RDM_OPT * 5);
> 	
> 	for( int i = 0; i < MAX_ITEM_RDM_OPT; i++ ){
> 		if( it->option[i].id ){
> 			buf[count].index = it->option[i].id;	// OptIndex
> 			buf[count].value = it->option[i].value;	// Value
> 			buf[count].param = it->option[i].param;	// Param1
> 			count++;
> 		}
> 	}
> 	
> 	return count;
> }
> ```

I've considered it, but thought that was just a workaround and that it would be better to fix the root cause.
I mean in theory the workaround could be applied additionally.

PR #7462, opened 2022-11-28
Corrected server response when adding an item to a trade session fails.

secretdataz:
* **Addressed Issue(s)**: -
* **Server Mode**: Both
On kRO, when adding an untradable item to a trade, it responds to the client with result=2 in packet ZC_ACK_ADD_EXCHANGE_ITEM.
Also removed a custom response when adding an item to a trade without free slot.


Lemongrass3110 (approved):
If possible would be nice to clean it up while we touch it:

1. sd from pointer to reference
2. index calculation should happen inside the clif function
3. failure reason should be an enum
4. packet should be changed to struct
~~5. packet 0xb42 should be implemented~~ Confused with a different packet, because it has almost the same name... Sorry!

Otherwise approved, if you do not want to do it! :-)

src/map/clif.cpp
```diff
@@ -4671,17 +4671,14 @@ void clif_tradeadditem( struct map_session_data* sd, struct map_session_data* ts
 ///     0 = success
 ///     1 = overweight
 ///     2 = trade canceled
-void clif_tradeitemok(struct map_session_data* sd, int index, int fail)
+void clif_tradeitemok(struct map_session_data& sd, int index, e_exitem_add_result result)
 {
-	int fd;
-	nullpo_retv(sd);
+	PACKET_ZC_ACK_ADD_EXCHANGE_ITEM* p = reinterpret_cast<PACKET_ZC_ACK_ADD_EXCHANGE_ITEM*>(packet_buffer);
```
Lemongrass3110, line 4676:
```suggestion
	PACKET_ZC_ACK_ADD_EXCHANGE_ITEM p = {};
```
The pointer/packet_buffer variant should only be used for dynamic packets.

src/map/clif.cpp
```diff
@@ -4671,17 +4671,14 @@ void clif_tradeadditem( struct map_session_data* sd, struct map_session_data* ts
...
+	p->packetType = HEADER_ZC_ACK_ADD_EXCHANGE_ITEM;
```
Lemongrass3110, line 4677:
```suggestion
	p.packetType = HEADER_ZC_ACK_ADD_EXCHANGE_ITEM;
```

src/map/clif.cpp
```diff
@@ -4671,17 +4671,14 @@ void clif_tradeadditem( struct map_session_data* sd, struct map_session_data* ts
...
+	p->index = client_index(index);
```
Lemongrass3110, line 4678:
```suggestion
	p.index = client_index(index);
```

src/map/clif.cpp
```diff
@@ -4671,17 +4671,14 @@ void clif_tradeadditem( struct map_session_data* sd, struct map_session_data* ts
...
+	p->result = static_cast<uint8>(result);
```
Lemongrass3110, line 4679:
```suggestion
	p.result = static_cast<uint8>(result);
```

src/map/clif.cpp
```diff
@@ -4671,17 +4671,14 @@ void clif_tradeadditem( struct map_session_data* sd, struct map_session_data* ts
...
 
-	fd = sd->fd;
-	WFIFOHEAD(fd,packet_len(0xea));
-	WFIFOW(fd,0) = 0xea;
-	WFIFOW(fd,2) = index;
-	WFIFOB(fd,4) = fail;
-	WFIFOSET(fd,packet_len(0xea));
+	clif_send(p, sizeof(p), &sd.bl, SELF);
```
Lemongrass3110, line 4681:
```suggestion
	clif_send(&p, sizeof(p), &sd.bl, SELF);
```
sizeof(p) was wrong in the pointer variant. sizeof(*p) would have been correct. Therefore the rest can stay.

src/map/trade.cpp
```diff
@@ -362,12 +362,10 @@ void trade_tradeadditem(struct map_session_data *sd, short index, short amount)
 	}
 
 	if( !amount ) { // Why do this.. ~.~ just send an ack, the item won't display on the trade window.
-		clif_tradeitemok(sd, index, 0);
+		clif_tradeitemok(*sd, index, EXITEM_ADD_SUCCEED);
```
Lemongrass3110, line 365: Since we removed the custom response below, we should maybe remove that one (which happens before the index validation) too?

src/map/trade.cpp
```diff
@@ -486,7 +484,7 @@ void trade_tradeok(struct map_session_data *sd)
 	}
 
 	sd->state.deal_locked = 1;
-	clif_tradeitemok(sd, 0, 0);
+	clif_tradeitemok(*sd, 0, 0);
```
Lemongrass3110, line 487:
```suggestion
	clif_tradeitemok(*sd, 0, EXITEM_ADD_SUCCEED);
```
This will fix CI.

src/map/packets.hpp
```diff
@@ -416,6 +422,7 @@ DEFINE_PACKET_HEADER(ZC_NOTIFY_CHAT, 0x8d)
 DEFINE_PACKET_HEADER(ZC_BROADCAST, 0x9a)
 DEFINE_PACKET_HEADER(ZC_ITEM_ENTRY, 0x9d)
 DEFINE_PACKET_HEADER(ZC_PC_PURCHASE_RESULT, 0xca)
+DEFINE_PACKET_HEADER(ZC_ACK_ADD_EXCHANGE_ITEM, 0xea);
```
Lemongrass3110, line 425:
```suggestion
DEFINE_PACKET_HEADER(ZC_ACK_ADD_EXCHANGE_ITEM, 0xea)
```

Lemongrass3110 (approved):

src/map/trade.cpp
```diff
@@ -486,7 +484,7 @@ void trade_tradeok(struct map_session_data *sd)
 	}
 
 	sd->state.deal_locked = 1;
-	clif_tradeitemok(sd, 0, 0);
+	clif_tradeitemok(*sd, 0, EXITEM_ADD_SUCCEED);
```
Lemongrass3110, line 487: Do we need to send -2 here now?

secretdataz: This needs testing. Will test and adjust (or not) accordingly.

secretdataz: This behavior matches official server.

src/map/trade.cpp
secretdataz: Official server sends index 0, and amount 0 in this case.

PR #7470, opened 2022-11-30
Update inter_athena.conf to match the documentation 

emmamarshall451:
* **Server Mode**:  Both

* **Description of Pull Request**:  Just updated "\rathena\conf\inter_athena.conf" to match the documentation which can be seen here 

[https://github.com/rathena/rathena/blob/c9f23e691a1ad26785e6f59f13733fef72e1a83a/sql-files/README.md?plain=1#L16](https://github.com/rathena/rathena/blob/c9f23e691a1ad26785e6f59f13733fef72e1a83a/sql-files/README.md?plain=1#L16)

### DIFF
--- a/conf/inter_athena.conf
+++ b/conf/inter_athena.conf
@@ -76,7 +76,7 @@ log_db_ip: 127.0.0.1
 log_db_port: 3306
 log_db_id: ragnarok
 log_db_pw: ragnarok
-log_db_db: ragnarok
+log_db_db: log
 log_codepage:
 log_login_db: loginlog
 


Lemongrass3110:
I would rather like to improve the README instead of the config file.
The README should always talk about "configured schema for xxx" or something like this.
What do you think @aleos89?

aleos89:
Agreed. The README can be cleaned up some more to be a little more ambiguous based on the user's schema names.

aleos89:
Please see #7484.

PR #7471, opened 2022-12-01
Fixes #7363

### LINKED ISSUE #7363: MAX_CHARS issue
### rAthena Hash

e95eb0ad84999c886d3adb5e11e5fa65c0f80eae

### Client Date

2021-11-03

### Server Mode

Renewal

### Result

My /src/config/core.hpp

```
#ifdef VIP_ENABLE
	#ifndef MIN_STORAGE
		#define MIN_STORAGE 300 // Default number of storage slots.
	#endif
	#ifndef MAX_CHAR_VIP
		#define MAX_CHAR_VIP 10 // This must be less than MAX_CHARS
	#endif
#else
```

VIP ENABLED total char is only 15 but i got this error 

`[Error]: Account '2004000' `character_slots` column is higher than supported MAX_CHARS (25), update MAX_CHARS in mmo.hpp! capping to MAX_CHARS...`

### Relevant Log Output

_No response_

### Expected Result

No error on char server

### How to Reproduce

Just Use as I wrote above.

### Official Information

NA

_No response_

violent01:
* **Addressed Issue(s)**: #7363

* **Server Mode**:  Both

 Removes additional character slots when VIP is enabled which results in an error in the char_server.

### DIFF
--- a/src/login/loginchrif.cpp
+++ b/src/login/loginchrif.cpp
@@ -683,9 +683,6 @@ int logchrif_parse_reqvipdata(int fd) {
 			if( now < vip_time ) { //isvip
 				if(acc.group_id != login_config.vip_sys.group){ //only upd this if we're not vip already
 					acc.old_group = acc.group_id;
-					if( acc.char_slots == 0 ){
-						acc.char_slots = MIN_CHARS;
-					}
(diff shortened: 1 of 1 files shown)


eppc0330:
Thanks

zdurexz:
Thanks

Issue #6709, opened 2022-03-17
Reflecting skills damage type


eppc0330:
* **rAthena Hash**: master

* **Client Date**: 20211117

* **Server Mode**: Renewal

* **Description of Issue**: 
  * Result: Reflecting damage type is not weapon.
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: According to official KRo behavior, damage type of reflected magic skill is special short weapon damage.
  Corresponding skills: NPC_MAGICMIRROR. (```bonus bMagicDamageReturn,n;``` still reflects as magic type damage. Just for NPC_MAGICMIRROR)
So in official KRo I can activate weapon triggered autospell by NPC_MAGICMIRROR. 
  

* **Modifications that may affect results**:

aleos89:
Fixed in 2896ae2.

Daegaladh:
There's still some skills to be fixed

aleos89:
Fixed in [c979ed7](https://github.com/rathena/rathena/commit/c979ed74722d62a64c5883579d0c203105ea4639).

PR #7480, opened 2022-12-11
NPC_MAGICMIRROR damage type

eppc0330:
* **Addressed Issue(s)**: #6709 

* **Server Mode**: Renewal

  - Now type of damage, reflected by NPC_MAGICMIRROR, will be special melee weapon damage.

  - Skill damage reflected by NPC_MAGICMIRROR will be able to trigger autospell.

  - Kyomu will block NPC_MAGICMIRROR reflecting ability.

  - Calculation of reducing reflect damage will be addition, not multiplication.

  Sorry I've deleted repository.. so I closed original pull and opened new one.


CairoLee:
Previously, it was possible to determine whether damage was caused by a reflect skill by checking the skill ID for CR_REFLECTSHIELD. However, this will no longer be easily achievable after the applied changes.

Maybe we can add a flag to indicate that this is a reflect damage?

eppc0330:
@CairoLee Sorry I am not following. More explain please? Why determining needed?

CairoLee:
@eppc0330 Because that reflection damage should not trigger bAutoSpell (the accuracy of this behavior is only confirmed in twRO, I'm not sure about other versions).

I make this change in my own MOD. When this PR is merged, I may need to make bigger adjustments to the code, and it will also increase the probability of conflicts when I merge rathena's code into my own repository in the future. If you could consider this together, it would be better for me.

This is just a suggestion. Thanks again for your PR and attention.

eppc0330:
@CairoLee Hmm.. so you are saying the opposite of mine. Because in KRO reflectied damage should be able to trigger autospell.

eppc0330:
@CairoLee Oh okay I guess I got it. From your point of view, I think I was talking about autospellwhenhit. Current rA's reflect damage does not trigger autospellwhenhit.

eppc0330:
@CairoLee Can you DM me? My discord id eppc0330.

eppc0330:
@CairoLee Confirmed in KRo reflect damage not triggering autospellwhenhit.
Same in case of autobonus2, right?

CairoLee:
@eppc0330 Yes, that's why I am requesting a maker to be provided, which can help me determine if the damage is caused by reflection.

This will allow me to skip the processing logic of autospell2 when necessary.

eppc0330:
@CairoLee Thanks for nice feedback. I've reverted codes about reflectshield.
Now this pull is only doing with `NPC_MAGICMIRROR`.

src/map/battle.cpp
```diff
@@ -8804,7 +8780,29 @@ int64 battle_calc_return_damage(struct block_list* tbl, struct block_list *src,
...
+	if (sc) {
+		if (status_reflect && sc->getSCE(SC_REFLECTDAMAGE)) {
+			reduce += sc->getSCE(SC_REFLECTDAMAGE)->val2;
+		}
+		if (sc->getSCE(SC_VENOMBLEED) && sc->getSCE(SC_VENOMBLEED)->val3 == 0) {
+			reduce += sc->getSCE(SC_VENOMBLEED)->val2;
+		}
+		if (sc->getSCE(SC_REF_T_POTION))
+			reduce += 100;
+	}
```
vstumpf, line 8801: we can probably factor this out, but we'll do this in another pr

vstumpf (approved):
Sorry it took so long! looks good!

src/map/battle.cpp
eppc0330: Oh right it sounds good. Thanks!

aleos89 (approved):

Issue #7481, opened 2022-12-11
Mob_size_influence uses database size instead of spawned size for drop rate


Singe-Horizontal:
### rAthena Hash

1f1b5fe4a

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

When Mob_size_influence is true, the game wrongly uses monster db size instead of spawned size for drop rate, resulting in halfed drop rate for small mobs such as knockers, and x2 drop rate for large mobs.

### Expected Result

Spawned size and not the db one should be taken into consideration.

### How to Reproduce

set Mob_size_influence to true
mobinfo for small and large mobs should yield wrong values

### Official Information

N/A

PR #7482, opened 2022-12-11
Restores correct drop rate for spawned size-modified mobs

Singe-Horizontal:
* **Addressed Issue(s)**: 
#7481

* **Server Mode**: 
Both

Restores correct drop rate for spawned size-modified mobs instead of database size.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2447,22 +2447,7 @@ void mob_damage(struct mob_data *md, struct block_list *src, int damage)
 	}
 #endif
 }
-/**
- * Get modified drop rate
- * @param mob: monster
- * @return Modified drop rate
- */
-int mob_getsizedropmodifier(mob_data* md)
-{
-	int size_rate = 100;
-	if (battle_config.mob_size_influence) {  // Change drops depending on monsters size [Valaris]
-		if (md->special_state.size == SZ_MEDIUM)
-			size_rate /= 2; // SZ_MEDIUM actually is small size modification... this is not a bug!
-		else if (md->special_state.size == SZ_BIG)
-			size_rate *= 2;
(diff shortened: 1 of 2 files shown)


src/map/mob.cpp
```diff
@@ -2459,14 +2474,6 @@ void mob_damage(struct mob_data *md, struct block_list *src, int damage)
 int mob_getdroprate(struct block_list *src, std::shared_ptr<s_mob_db> mob, int base_rate, int drop_modifier)
 {
 	int drop_rate = base_rate;
-
-	if (battle_config.mob_size_influence) { // Change drops depending on monsters size [Valaris]
```
aleos89, line 2463: I guess I'm curious why it was removed from here since any other call to this function will also have to make sure to include the size modifier manually now. Seems more logical to keep this check with the special_state fix in `mob_getdroprate`.

Singe-Horizontal: Yep let's just add a parameter

Lemongrass3110: @aleos89:
It was most likely removed when the atcommand was added that does not have access to any specific spawned monster.

@Singe-Horizontal:
Could you maybe hand over the actual mob_data* to the function? Just in case some other checks will be necessary in the future or if someone wants to hook with a source modification to that function. Also please make sure to revert the empty line removals and style changes wherever possible please.
Thank you!

Ps:
All these assumptions with SZ_MEDIUM actually being SZ_SMALL make me wanna puke btw... XD Just because someone was too lazy to initialize this properly.

Singe-Horizontal: The forgotten curse of the size enum

Lemongrass3110 (approved):

Issue #7473, opened 2022-12-04
dynamic npc doesnt work if another dynamic npc (with different npc id) is being called.


dimasshotta:
### rAthena Hash

a763ad06294f1478343b3e765064e4bcdcb993c0

### Client Date

any

### Server Mode

Pre-Renewal, Renewal

### Result

I believe dynamic NPC can be called once for one npc id, but i guess it could be called twice or more if the source of the npc is different.

### Relevant Log Output

```Shell
struct npc_data* npc_duplicate_npc_for_player( struct npc_data& nd, struct map_session_data& sd ){
	
	if( sd.npc_id_dynamic != 0 ){
		clif_msg_color( &sd, C_DYNAMICNPC_TWICE, color_table[COLOR_LIGHT_YELLOW] );
		return nullptr;
	}
```

### Expected Result

this line above needs to be changed I guess. probablly try to change to npc dynamic name? dynamic npc can't be called if the same npc name is still appear.

dynamic npc works if the source id of the npc is different.

### How to Reproduce

play KRO

### Official Information

here working for the other dynamic npcs
![image](https://user-images.githubusercontent.com/74591268/205474336-dc9875fd-2a01-4d10-9030-73df4dc778d0.png)

eppc0330:
Also dynamic npc should be removed on logout

eppc0330:
@Atemo Ah I misremembered my test.

I got some more infos about dynamic NPC.

First I can confirm this issue, tested on KRo right now.
![dynamic02](https://user-images.githubusercontent.com/47050704/205721518-23b024cb-eddf-4a32-a638-91f0aedab216.gif)

Second dynamic NPC should be removed on map change, too.
![dynamic01](https://user-images.githubusercontent.com/47050704/205721598-1adb97a2-1fc9-49bd-a2d9-a59d03fe66ea.gif)
![dynamic03](https://user-images.githubusercontent.com/47050704/205721621-84b14fc0-7136-482d-ae84-fbf5374264f2.gif)

In this case I think I have seen some case not removing on mapchange and need more info.
However in most of situation as I tested dynamic NPC gets removed on map change.

eppc0330:
Add) Timer issue also exist.
When dynamic NPC goes in idle status, ```dynamicnpc.removal_tid``` should be reset as 60sec, not remaining time before click the NPC.

Atemo:
> Second dynamic NPC should be removed on map change, too.

[...]

PR #7486, opened 2022-12-12
Multiple dynamic npc

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/7473

* **Server Mode**: Both

Vectorize npc_id_dynamic to support multiple dynamic npc with different ids.

Thanks to @dimasshotta !

### DIFF
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -5802,11 +5802,13 @@ TIMER_FUNC(npc_dynamicnpc_removal_timer){
 }
 
 struct npc_data* npc_duplicate_npc_for_player( struct npc_data& nd, struct map_session_data& sd ){
+	// A duplicate of a duplicate is still a duplicate of the same NPC
 	int src_id = nd.src_id > 0 ? nd.src_id : nd.bl.id;
 
 	for (const auto &it : sd.npc_id_dynamic) {
 		struct npc_data* src_nd = map_id2nd( it );
 
+		// Check if the source NPC id of currently active duplicates already exists.
 		if( src_nd != nullptr && src_nd->src_id == src_id ){
 			clif_msg_color( &sd, C_DYNAMICNPC_TWICE, color_table[COLOR_LIGHT_YELLOW] );
 			return nullptr;
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -5790,7 +5790,7 @@ TIMER_FUNC(npc_dynamicnpc_removal_timer){
 			return 0;
 		}
 
-		util::vector_erase_if_exists(sd->npc_id_dynamic, nd->bl.id);
+		// npc id from sd->npc_id_dynamic is removed in npc_unload
 	}
 
 	// Delete the NPC
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -4515,8 +4515,7 @@ const char* npc_parse_duplicate( char* w1, char* w2, char* w3, char* w4, const c
 
 	if( owner != nullptr ){
 		nd->dynamicnpc.owner_char_id = owner->status.char_id;
-		if (!util::vector_exists(owner->npc_id_dynamic, nd->bl.id))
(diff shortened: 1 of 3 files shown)


src/map/npc.cpp
```diff
@@ -4515,7 +4515,8 @@ const char* npc_parse_duplicate( char* w1, char* w2, char* w3, char* w4, const c
 
 	if( owner != nullptr ){
 		nd->dynamicnpc.owner_char_id = owner->status.char_id;
-		owner->npc_id_dynamic = nd->bl.id;
+		if (!util::vector_exists(owner->npc_id_dynamic, nd->bl.id))
+			owner->npc_id_dynamic.push_back(nd->bl.id);
```
Lemongrass3110, line 4519: Its already checked outside either just push back or check earlier above by checking dnd->bl.id and stop it earlier.

src/map/npc.cpp
```diff
@@ -5802,9 +5803,15 @@ TIMER_FUNC(npc_dynamicnpc_removal_timer){
 }
 
 struct npc_data* npc_duplicate_npc_for_player( struct npc_data& nd, struct map_session_data& sd ){
-	if( sd.npc_id_dynamic != 0 ){
-		clif_msg_color( &sd, C_DYNAMICNPC_TWICE, color_table[COLOR_LIGHT_YELLOW] );
-		return nullptr;
+	int src_id = nd.src_id > 0 ? nd.src_id : nd.bl.id;
```
Lemongrass3110, line 5806: What's that new check here? Like if we have a duplication chain?

Atemo: Just in case the user tries to create a dynamic npc from a duplicate npc. It sounds stupid but our system allows it

src/map/npc.cpp
```diff
@@ -5802,9 +5802,15 @@ TIMER_FUNC(npc_dynamicnpc_removal_timer){
...
```
Lemongrass3110, line 5805:
```suggestion
	// A duplicate of a duplicate is still a duplicate of the same NPC
	int src_id = nd.src_id > 0 ? nd.src_id : nd.bl.id;
```

src/map/npc.cpp
```diff
@@ -5802,9 +5802,15 @@ TIMER_FUNC(npc_dynamicnpc_removal_timer){
...
+
+	for (const auto &it : sd.npc_id_dynamic) {
+		struct npc_data* src_nd = map_id2nd( it );
+
+		if( src_nd != nullptr && src_nd->src_id == src_id ){
```
Lemongrass3110, line 5810:
```suggestion
		// Check if the source NPC id of currently active duplicates already exists.
		if( src_nd != nullptr && src_nd->src_id == src_id ){
```

Lemongrass3110 (approved):
Just some small comments to make it easier to understand in the future.

dimasshotta:
I have tested this pull, it works great, but when you call more than 2 dynamic npc the first dynamic npc does not unload when you log out, and then you can call the same dynamic npc from the same source again..

dimasshotta:
Thank you @Atemo now it works good
I'm not really good at src, but i just figure out something, this is just a suggestion what if put this code

			if (sd->state.changemap && !sd->npc_id_dynamic.empty()) {
				for (const auto &it : sd->npc_id_dynamic) {
					struct npc_data* nd = map_id2nd(it);

					if (nd != nullptr) {
						// Erase the owner first to prevent loops from npc_unload
						nd->dynamicnpc.owner_char_id = 0;

						// Delete the NPC
						npc_unload(nd, true);
					}
					// Update NPC event database
					npc_read_event_script();
				}

				sd->npc_id_dynamic.clear();
			}
inside int unit_remove_map_(struct block_list *bl, clr_type clrtype, const char* file, int line, const char* func) in unit.cpp

i have tried this, and it unloads player's dynamic npc when the char making a map change.

Atemo:
As explained in the issue, we are not removing dynamic NPCs when the player changes maps

PR #7490, opened 2022-12-15
Cleaned struct tokens for map_session_data and status_change

Singe-Horizontal:
#7356 updated map_session_data and status_change from struct to class specifiers
This PR replaces elaborated type specifier in declarations from struct to class and removes them when needed ( most were C left-overs )

### DIFF
--- a/src/map/buyingstore.cpp
+++ b/src/map/buyingstore.cpp
@@ -702,7 +702,7 @@ void do_init_buyingstore_autotrade( void ) {
 					at->sit = battle_config.feature_autotrade_sit;
 
 				// initialize player
-				CREATE(at->sd, map_session_data, 1);
+				CREATE(at->sd, map_session_data, 1); // TODO: Dont use Memory Manager allocation anymore and rely on the C++ container
 				pc_setnewpc(at->sd, at->account_id, at->char_id, 0, gettick(), at->sex, 0);
 				at->sd->state.autotrade = 1|4;
 				if (battle_config.autotrade_monsterignore)
--- a/src/map/buyingstore.cpp
+++ b/src/map/buyingstore.cpp
@@ -19,7 +19,7 @@
 #include "clif.hpp"  // clif_buyingstore_*
 #include "log.hpp"  // log_pick_pc, log_zeny
 #include "npc.hpp"
-#include "pc.hpp"  // struct map_session_data
+#include "pc.hpp"  // map_session_data
 
 //Autotrader
 static DBMap *buyingstore_autotrader_db; /// Holds autotrader info: char_id -> struct s_autotrader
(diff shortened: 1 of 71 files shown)


src/map/status.cpp
```diff
@@ -9299,7 +9299,7 @@ struct status_change *status_get_sc(struct block_list *bl)
  */
 void status_change_init(struct block_list *bl)
 {
-	struct status_change *sc = status_get_sc(bl);
+	status_change *sc = status_get_sc(bl);
 	nullpo_retv(sc);
 	memset(sc, 0, sizeof (struct status_change));
```
aleos89, line 9304:
```suggestion
	memset(sc, 0, sizeof (status_change));
```

aleos89 (commented):
Thanks for all the effort you put into this!

There are two lines I can't review which are throwing a couple warnings, both in mob.cpp:
Line 3662:
```cpp
flag=( md->sc.getSCE(cond2) != NULL );
```
Should be:
```cpp
flag=( md->sc.getSCE(static_cast<int32>(cond2)) != NULL );
```

And line 3773:
```cpp
flag = (md->sc.getSCE(ms[i]->cond2)!=NULL);
```
Should be:
```cpp
flag = (md->sc.getSCE(static_cast<int32>(ms[i]->cond2))!=NULL);
```

src/map/status.cpp
Singe-Horizontal: I'm not the person who refactored the statuses, do you want me to do those fixes here?

src/map/buyingstore.cpp
```diff
@@ -702,7 +702,7 @@ void do_init_buyingstore_autotrade( void ) {
 					at->sit = battle_config.feature_autotrade_sit;
 
 				// initialize player
-				CREATE(at->sd, struct map_session_data, 1);
+				CREATE(at->sd, map_session_data, 1);
```
Lemongrass3110, line 705:
```suggestion
				CREATE(at->sd, map_session_data, 1); // TODO: Dont use Memory Manager allocation anymore and rely on the C++ container
```

src/map/vending.cpp
```diff
@@ -616,7 +616,7 @@ void do_init_vending_autotrade(void)
 					at->sit = battle_config.feature_autotrade_sit;
 
 				// initialize player
-				CREATE(at->sd, struct map_session_data, 1);
+				CREATE(at->sd, map_session_data, 1);
```
Lemongrass3110, line 619:
```suggestion
				CREATE(at->sd, map_session_data, 1); // TODO: Dont use Memory Manager allocation anymore and rely on the C++ container
```

src/map/buyingstore.cpp
Lemongrass3110, line 705:
```suggestion
				CREATE(at->sd, map_session_data, 1); // TODO: Dont use Memory Manager allocation anymore and rely on the C++ container
```

Lemongrass3110 (approved):

src/map/status.cpp
aleos89: Yeah, it was just a couple other warnings that are popping up from the change so might as well fix it now!

Issue #7501, opened 2022-12-19
specialeffect doesn't work when passing npc name with SELF target


Singe-Horizontal:
### rAthena Hash

5b8d0ee6a

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

When using the command specialeffect with npc name parameter, it ignores SELF target, meaning you can't use it to display an effect on attached players only.

### Expected Result

The effect should display on named npc for attached player

### How to Reproduce

Example script:
`specialeffect EF_LEVEL99_2,SELF,"Job Master";`

### Official Information

N/A

PR #7502, opened 2022-12-19
Fixed specialeffect with npc names

Singe-Horizontal:
* **Addressed Issue(s)**: 
#7501

* **Server Mode**: 
Both

Restores SELF target to call clif_specialeffect_single in specialeffect when called with npc name argument

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -15370,7 +15370,7 @@ BUILDIN_FUNC(specialeffect)
 	}
 
 	if (target == SELF) {
-		TBL_PC *sd;
+		map_session_data* sd;
 		if (script_rid2sd(sd)){
 			clif_specialeffect_single(bl,type,sd->fd);
 		}
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -15361,17 +15361,18 @@ BUILDIN_FUNC(specialeffect)
 	{
 		TBL_NPC *nd = npc_name2id(script_getstr(st,4));
 		if(nd)
-			clif_specialeffect(&nd->bl, type, target);
+			bl = &nd->bl;
+		else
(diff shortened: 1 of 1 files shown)


src/map/script.cpp
```diff
@@ -15361,17 +15361,18 @@ BUILDIN_FUNC(specialeffect)
...
-	{
-		if (target == SELF) {
-			TBL_PC *sd;
-			if (script_rid2sd(sd))
-				clif_specialeffect_single(bl,type,sd->fd);
-		} else {
-			clif_specialeffect(bl, type, target);
+
+	if (target == SELF) {
+		TBL_PC *sd;
```
Lemongrass3110, line 15370:
```suggestion
		map_session_data* sd;
```

src/map/script.cpp
```diff
@@ -15361,17 +15361,18 @@ BUILDIN_FUNC(specialeffect)
...
+		if (script_rid2sd(sd)){
+			clif_specialeffect_single(bl,type,sd->fd);
```
Lemongrass3110, line 15372:
```suggestion
		if (!script_rid2sd(sd)){
			return SCRIPT_CMD_FAILURE;
```

src/map/script.cpp
```diff
@@ -15361,17 +15361,18 @@ BUILDIN_FUNC(specialeffect)
...
 		}
+	} else {
```
Lemongrass3110, line 15374:
```suggestion
		clif_specialeffect_single(bl,type,sd->fd);
	} else {
```

Lemongrass3110 (changes requested):

src/map/script.cpp
Singe-Horizontal: Thanks Lemon

PR #7523, opened 2023-01-03
Adds char_server config for clearing parties

aleos89:
* **Addressed Issue(s)**: N/A

  * Adds a character server config to clear empty parties at start up.
  * The commented code was hidden in the source and is pretty useless unless you are digging for it.


Lemongrass3110 (approved):
I would vote for default "yes" as it does not hurt at all, but we did not have it forever now, so should be okay.

Issue #7441, opened 2022-11-16
itemlink not working some exe


zdurexz:
### rAthena Hash

-

### Client Date

20200304

### Server Mode

Pre-Renewal, Renewal

### Result

itemlink not working on before 20200724
@mobinfo 1002
![image](https://user-images.githubusercontent.com/59825922/202196144-1c8a3e02-f753-4a16-b9c6-c7a4b7b164dc.png)

@iteminfo 1201
![image](https://user-images.githubusercontent.com/59825922/202196498-04ffac91-6316-4192-8c9a-f734e19f8c1b.png)

### Expected Result

@mobinfo
![image](https://user-images.githubusercontent.com/59825922/202196707-b8c9d628-d162-44e6-839c-759564bde242.png)

@iteminfo
![image](https://user-images.githubusercontent.com/59825922/202196763-35cfee91-8ced-4685-afdb-145787fdccb9.png)

### How to Reproduce

-

### Official Information

-

### Modifications that may affect results

Before 
```
#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200724
	const std::string start_tag = "<ITEML>";
	const std::string closing_tag = "</ITEML>";
#else // PACKETVER >= 20100000
	const std::string start_tag = "<ITEMLINK>";
	const std::string closing_tag = "</ITEMLINK>";
```
to
```
#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200724
	const std::string start_tag = "<ITEML>";
	const std::string closing_tag = "</ITEML>";
#elif PACKETVER >= 20151104
	const std::string start_tag = "<ITEML>";
	const std::string closing_tag = "</ITEML>";
#else // PACKETVER >= 20100000
	const std::string start_tag = "<ITEMLINK>";
	const std::string closing_tag = "</ITEMLINK>";
#endif
```

zdurexz:
> Working for me @mobinfo 1011, client 2022-04-06 ![image](https://user-images.githubusercontent.com/98105181/202208445-d6a690f4-f5ef-479f-9c38-076184d82fee.png)

client less than 20200724 can't working

Aruar:
i managed to solve it, i don't know if it will solve there too:

change:

#if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200724
    const std::string start_tag = "<ITEML>";
    const std::string closing_tag = "</ITEML>";
[...]

PR #7532, opened 2023-01-05
Improved itemlink logic

Leemonn:
* **Addressed Issue(s)**: #7441

* **Server Mode**: Pre-renewal.

1. Modified how `create_item_link()` is called. Now there are 3 overloads:

   - `create_item_link( std::shared_ptr<item_data>& data )` is used when we want to link an item from the DB.
   - `create_item_link(struct item& item)` is used when we want to link an actual item with all its cards, refine, options, etc.
   - `create_item_link(struct item& item, std::shared_ptr<item_data>& data)` is called by the overloads mentioned above.

This allowed me to remove duplicated calls to itemdb->find.

2. If itemlink is disabled it will output the item name with the refine as prefix and the slots as suffix (if equipment). For example: `+6 Cotton Shirt [0]`, `Cotton Shirt [1]`.
3. Fixed tags and separators for various packet versions.
4. Added a `[0]` suffix for items that have 0 slots (when itemlink is enabled). So now it will show `<Cotton Shirt> [0]` instead of `<Cotton Shirt>`. This will make it easier to tell between an equipment with 0 slots and a non-equipment. (otherwise you would get the same format for both, like this: `<Jellopy>`, `<Full Plate>`).
5. `@itemlist` will now correctly link the actual item (instead of looking for the db data) thus showing cards, options, refine, etc.
6. Modified the `map_msg*.conf` files to reflect the changes mentioned above. (more info below)
7. Adapted the calls to `create_item_link` on `atcommand.cpp` to reflect the changes mentioned above. (more info below)

### Important:
- The slot suffix and the map_msg changes were necessary because the commands (`@mi`, `@ii`, `@itemlist`, etc.) appended the slots even if itemlink was enabled resulting in duplicated slots suffix like this: `<Cotton Shirt [1]> [1]`. This change also allows us to centralize the item format (even when itemlink is disabled) instead of manually appending prefixes and suffixes everywhere.
- The `[0]` suffix for items that have 0 slots (when itemlink is enabled) can be removed if needed, let me know if you want me to remove it and I will update the PR.

Only the clients listed in the comment above were tested, it might break itemlink for clients that weren't tested (even if it was working before this commit).

I tested:
- `@mi` for consumables, gear and etc items.
- `@ii` for consumables, gear and etc items.
- Items with refine
- Items with no slots
- Items with slots
- Items with cards
- Items with random options
- Items with no random options


Leemonn:
`20180620` uses:
```
const std::string start_tag = "<ITEML>";
const std::string closing_tag = "</ITEML>";

// & for equipment
// % for refine

const std::string card_sep = "(";
const std::string optid_sep = "*";
const std::string optpar_sep = "+";
const std::string optval_sep = ",";
```

Leemonn:
`20151104` uses:
```
const std::string start_tag = "<ITEM>";
const std::string closing_tag = "</ITEM>";

// % for refine

const std::string card_sep = "'";
const std::string optid_sep = ")";
const std::string optpar_sep = "*";
const std::string optval_sep = "+";
```

Leemonn:
`20220406` uses:
```
const std::string start_tag = "<ITEML>";
const std::string closing_tag = "</ITEML>";

// & for equipment
// % for refine

const std::string card_sep = ")";
const std::string optid_sep = "+";
const std::string optpar_sep = ",";
const std::string optval_sep = "-";
```

Leemonn:
Need help to test what format each client version uses. Please if your client version is not listed above do this:

1. Enable chat logging
2. In-game do `@item2 1734 1 1 5 0 4004 4004 4004 4004`
3. In-game do `shift+click` on the item to link it in the chat and press `enter` to send the message
4. Go to your database and comment here the message logged in the `chatlog` table (in rathena_logs database)

Leemonn:
1. Modified how `create_item_link()` is called. Now there are 3 overloads:

   - `create_item_link( std::shared_ptr<item_data>& data )` is used when we want to link an item from the DB.
   - `create_item_link(struct item& item)` is used when we want to link an actual item with all its cards, refine, options, etc.
   - `create_item_link(struct item& item, std::shared_ptr<item_data>& data)` is called by the overloads mentioned above.

This allowed me to remove duplicated calls to itemdb->find.

2. If itemlink is disabled it will output the item name with the refine as prefix and the slots as suffix (if equipment). For example: `+6 Cotton Shirt [0]`, `Cotton Shirt [1]`.
3. Fixed tags and separators for various packet versions.
4. Added a `[0]` suffix for items that have 0 slots (when itemlink is enabled). So now it will show `<Cotton Shirt> [0]` instead of `<Cotton Shirt>`. This will make it easier to tell between an equipment with 0 slots and a non-equipment. (otherwise you would get the same format for both, like this: `<Jellopy>`, `<Full Plate>`).
5. `@itemlist` will now correctly link the actual item (instead of looking for the db data) thus showing cards, options, refine, etc.
6. Modified the `map_msg*.conf` files to reflect the changes mentioned above. (more info below)
7. Adapted the calls to `create_item_link` on `atcommand.cpp` to reflect the changes mentioned above. (more info below)


### Important:
- The slot suffix and the map_msg changes were necessary because the commands (`@mi`, `@ii`, `@itemlist`, etc.) appended the slots even if itemlink was enabled resulting in duplicated slots suffix like this: `<Cotton Shirt [1]> [1]`. This change also allows us to centralize the item format (even when itemlink is disabled) instead of manually appending prefixes and suffixes everywhere.
- The `[0]` suffix for items that have 0 slots (when itemlink is enabled) can be removed if needed, let me know if you want me to remove it and I will update the PR.

Only the clients listed in the comment above were tested, it might break itemlink for clients that weren't tested (even if it was working before this commit).

Lemongrass3110:
2016-01-13: `<ITEML>0000y1rY%05'12A'12A'12A'12A</ITEML>`
2016-02-03: `<ITEML>0000y1rY%05'12A'12A'12A'12A</ITEML>`
Sadly I have no working clients in between
2016-11-16: `<ITEML>0000y1rY%05&00(12A(12A(12A(12A</ITEML>`

PR #7533, opened 2023-01-05
Removed mapindex from char-server

Lemongrass3110:

* **Server Mode**: Both

Refactored char-server to not use mapindex anymore.
Fixed an ancient bug that did not trigger character saving when only the savepoint was changed.
Removed the default map feature.
Lots of cleanups on the way.

This is required for a few upcoming features and bugfixes!
So make sure to help us with testing so that we can merge it fast.

TODOs:
- [x] Test a lot - especially the affected features
- [x] Test in a multi map-server environment


aleos89 (approved):
🚀

PR #7535, opened 2023-01-08
Added 1@begi mapflag to prevent status abuse

mazvi:
Added 1@begi mapflag

To minimize status abuse sc_start SC_GLASTHEIM_STATE,30000,20; outside the map

* **Addressed Issue(s)**: 

* **Server Mode**: 


Atemo:
Is there a reason you want to add `noreturn` mapflag to the map?

mazvi:
> Is there a reason you want to add `noreturn` mapflag to the map? others instance maps don't have it

its because status abuse sc_start SC_GLASTHEIM_STATE,30000,20; outside the map

player after get this status allstat +20 then player using butterflywing and use the status outside the map

Atemo:
The buff only lasts for 30 seconds and the player can only enter into the instance once per day. Is this change relevant? It doesn't seem abusive or I missed something?

mazvi:
Using status from within the instance to be taken out is an abuse either way.

If there are other map instances that have the potential to do so, they can also be included in noreturn.

Atemo:
What you say is true to some extent, but I don't share your opinion. 30s of a small buff is not exploitable. I'll stick to the official behavior that doesn't remove the buff outside of the instance map. Please revert your change about the `noreturn` mapflag.

Issue #7538, opened 2023-01-09
User logged in already, cannot log in


MaouDaiou:
### rAthena Hash

User logged in already cannot log in

### Client Date

2023-01-09

### Server Mode

Renewal

### Result

User logged in already, cannot log in.
https://media.discordapp.net/attachments/727532828727836743/1061932414475186227/image.png

### Expected Result

I hope you fix it soon.

### How to Reproduce

I have no advice

### Official Information

kRo

### Modifications that may affect results

i'm not modify.

zdurexz:
I confirm that I am unable to log in again.

MaouDaiou:
https://youtu.be/HkHdPzzUvlQ
how to fix T^T

MaouDaiou:
thank you very mush

mazvi:
Another problem is user autotrade cannot login.

PR #7539, opened 2023-01-09
Fixed online_char_db

Lemongrass3110:
* **Addressed Issue(s)**: #7538

* **Server Mode**: Both

Added a missing map entry assignment from an invalid DBMap ensure conversion.

Thanks to @MaouDaiou and @zdurexz


zdurexz:
Thank for support. i try work properly

24msz6eo:
This change will cause the pincode system window chance to fail

Issue #6632, opened 2022-02-20
Save Point Problem


Xypr0:
* **rAthena Hash**: latest git file

* **Client Date**: 
any client version

* **Server Mode**: 
Both Happen in pre-re and renewal

* **Description of Issue**: 
  * Result: when the map is disabled or not existing at all. for example payon town. when you save on that town and you disabled it and restart the server, once you die you will respawn in your current position when you click return to savepoint.
  * Expected Result: its good if the map of your savepoint is not existing there will be default map x and y can teleport your character and show a dispbottom message about "your savepoint is not exist.".
  * How to Reproduce: go save in any map and then disabled it and restart the server and do not forget to move your character first in existing map.
  * Official Information: *i cant test this in official kro

* **Modifications that may affect results**: 
  i didnt change anything. you can use any file even in latest git file.

aleos89:
Is it even a normal scenario that a map is removed from a server? If you're "disabling" maps by not using them anymore, it might be best to just add a NoSave mapflag to it to force players out and possibly having a hidden NPC resetting the save point to a specific map if their save point matches the map you are "disabling".

Gravity doesn't remove maps from their cache, they just disable warps and stuff getting to those maps. The practice can easily be done the same for us especially with the minimal amount of kilobytes a map uses up in rAthena.

Toshiro90:
On official servers (just experienced on the newly launched MY/SG/PH server), when the SavePoint map is not available (e.g. map-server down) it just displays "Unable to move the applied area" and wont resurrect you at all.

DevNoName:
> Is it even a normal scenario that a map is removed from a server? If you're "disabling" maps by not using them anymore, it might be best to just add a NoSave mapflag to it to force players out and possibly having a hidden NPC resetting the save point to a specific map if their save point matches the map you are "disabling".
> 
> Gravity doesn't remove maps from their cache, they just disable warps and stuff getting to those maps. The practice can easily be done the same for us especially with the minimal amount of kilobytes a map uses up in rAthena.

if your server is running episode and you start in lower episode you can experience what i said above.

DevNoName:
> 

this is what im talking about. because its kinda weird if you click return to save point, it will revive you at your current position. for me it's a bug that need to fix.

DevNoName:
[...]

PR #7546, opened 2023-01-11
Added support for map selection

Lemongrass3110:
* **Addressed Issue(s)**: #6632

* **Server Mode**: Both

Added support for the map selection window in case a map is unavailable anymore.
Allows your users to pick a map from a list of defined maps that was previously only used to warp your players automatically to one of those maps, if available. By default you ended up in Prontera.

Thanks to @Xypr0, @Toshiro90, @alisonrag and @aleos89

### DIFF
--- a/src/char/char_clif.cpp
+++ b/src/char/char_clif.cpp
@@ -30,6 +30,10 @@
 
 using namespace rathena;
 
+// Reuseable global packet buffer to prevent too many allocations
+// Take socket.cpp::socket_max_client_packet into consideration
+static int8 packet_buffer[UINT16_MAX];
+
 std::vector<struct s_point_str> accessible_maps{
 	s_point_str{ MAP_PRONTERA, 273, 354 },
 	s_point_str{ MAP_GEFFEN, 120, 100 },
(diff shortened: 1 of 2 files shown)


src/char/char_clif.cpp
```diff
@@ -816,6 +820,165 @@ void chclif_send_map_data( int fd, std::shared_ptr<struct mmo_charstatus> cd, ui
...
+	node->account_id = sd->account_id;
+	node->char_id = cd->char_id;
+	node->login_id1 = sd->login_id1;
+	node->login_id2 = sd->login_id2;
+	node->sex = sd->sex;
+	node->expiration_time = sd->expiration_time;
+	node->group_id = sd->group_id;
+	node->ip = ipl;
+
+	char_get_authdb()[node->account_id] = node;
```
vstumpf, line 943: Lets move this instead of assignment

src/char/char_clif.cpp
```diff
@@ -816,6 +820,165 @@ void chclif_send_map_data( int fd, std::shared_ptr<struct mmo_charstatus> cd, ui
...
+
+	return 1;
+#else
+	return 0;
+#endif
+}
+
+void chclif_accessible_maps( int fd ){
+#if PACKETVER >= 20100714
+	struct PACKET_HC_NOTIFY_ACCESSIBLE_MAPNAME* p = (struct PACKET_HC_NOTIFY_ACCESSIBLE_MAPNAME*)packet_buffer;
```
vstumpf, line 953: If you don't want a global, you can create a static packet buffer here

vstumpf (commented):
Some initial thoughts

src/char/char_clif.cpp
Lemongrass3110: As discussed in Discord this should be dealt with in a broader scope.
But good point!

aleos89 (approved):

Issue #1525, opened 2016-08-28
Implement new mob skills


secretdataz:
There are some new mob skills that are left unimplemented.

Lemongrass3110:
#1893 is one of them

Jeybla:
Can we make up a list of missing skills?
Or else the issue isn't really valuable.

secretdataz:
```
//****
// New NPC Wide Status AoE Skills And Others
//700,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_WIDEHEALTHFEAR,Wide Health Fear
//701,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_WIDEBODYBURNNING,Wide Body Burnning
//702,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_WIDEFROSTMISTY,Wide Freezing
//703,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_WIDECOLD,Wide Crystalize
//704,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_WIDE_DEEP_SLEEP,Wide Deep Sleep
//705,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_WIDESIREN,Wide Siren's Voice
706,0,6,4,5,0,0,10,1,no,0,0x2,0,magic,0,0x1,	NPC_VENOMFOG,Venom Fog
//707,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_MILLENNIUMSHIELD,Millenium Shield 2
//708,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_COMET,Comet 2
//709,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_ICEMINE,Ice Mine
//710,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_ICEEXPLO,Ice Explosion
//711,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_FLAMECROSS,Flame Cross
//712,0,0,0,0,0,0,??,0,no,0,0,0,none,0,0x0,		NPC_PULSESTRIKE2,Pulse Strike 2
[...]

PR #7614, opened 2023-02-16
Implemented the skills NPC_RELIEVE_ON and NPC_RELIEVE_OFF

Atemo:
* **Addressed Issue(s)**: part of https://github.com/rathena/rathena/issues/1525

* **Server Mode**: re

Implemented the skills NPC_RELIEVE_ON and NPC_RELIEVE_OFF as part of the episode 17.2


src/map/battle.cpp
```diff
@@ -1839,6 +1839,13 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 		pc_overheat(*sd, (battle_get_weapon_element(d, src, bl, skill_id, skill_lv, EQI_HAND_R, false) == ELE_FIRE ? 3 : 1));
 	}
 
+	sc = status_get_sc(bl);
```
aleos89, line 1842: I'm only viewing this on the app, but was this not defined prior?

src/map/status.cpp
```diff
@@ -10403,6 +10403,12 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			break;
 		default:
 			break;
+	case SC_RELIEVE_ON:
+		status_change_end(bl, SC_RELIEVE_OFF, INVALID_TIMER);
+		break;
+	case SC_RELIEVE_OFF:
+		status_change_end(bl, SC_RELIEVE_ON, INVALID_TIMER);
+		break;
```
aleos89, line 10411: Maybe move these to EndOnStart?

src/map/battle.cpp
Atemo: It's assigned multiple times but the last assignment of sc is for src
`sc = status_get_sc(src);`
and we want it for bl

aleos89: Variable re-usage, the chef's kiss. Would it be better then for us to just move this check up inside the bl if-statement that ends around line 1735~? Or is it needed to be down here to be after some other check? If it's the latter, maybe add an extra comment stating it needs to be after X, Y, Z check or something so we will know for future reference!

Atemo: I'm not sure but for me putting it at the end of `battle_calc_damage` (like `damagetaken`) is a safe spot

PR #7618, opened 2023-02-20
Made debug ShowStatus calls easier to disable

secretdataz:
* **Addressed Issue(s)**: N/A

* **Server Mode**: N/A

In various places like YAML loading and NPC loading, `ShowStatus` calls are disabled with `DEBUG` ifdefs to increase startup performance. However, in practice, when not dealing with NPCs or YAML databases, these outputs can be disabled safely to reduce startup time when developing rAthena.


vstumpf (approved):
Looks good! Much cleaner that what i did

aleos89 (approved):

Issue #7619, opened 2023-02-22
Skill: Fire Pillar ( ID:80 WZ_FIREPILLAR ) Blue Gemstone will also be consumed in VIP or wearing Mistress Card


Oysica:
### rAthena Hash

198566204

### Client Date

20220602

### Server Mode

Renewal

### Result

Skill: Fire Pillar ( ID:80 WZ_FIREPILLAR ) Blue Gemstone will also be consumed in VIP or wearing Mistress Card
1. Change Job to Wizard.
2. Get blue gemstone
3. Use Fire Pillar above lv6
4. Check whether the blue gemstone is consumed

### Relevant Log Output

```Shell
no
```

### Expected Result

Blue gemstone is not consumed

### How to Reproduce

Skill: Fire Pillar ( ID:80 WZ_FIREPILLAR ) Blue Gemstone will also be consumed in VIP or wearing Mistress Card
1. Change Job to Wizard.
2. Get blue gemstone
3. Use Fire Pillar above lv6
4. Check whether the blue gemstone is consumed

### Official Information

no

aleos89:
Fixed in bb8c4a0.

PR #7623, opened 2023-02-26
Fix Fire Pillar skill to not consume Gemstone if VIP or use Mistress card

aoshfan:
Fix Fire Pillar to not consume Gemstone when VIP or using Mistress card.

* **Addressed Issue(s)**:  #7619 

* **Server Mode**: Both

   * Fix the Fire Pillar skill to not consume Gemstone if VIP or wearing Mistress Card.
   * Thank  you @Oysica for mention this bug.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -18461,7 +18461,7 @@ struct s_skill_condition skill_get_requirement(map_session_data* sd, uint16 skil
 				}
 				else {
 					// Process level_dependent requirement
-					if (level_dependent && skill_lv <= MAX_SKILL_ITEM_REQUIRE) {
+					if (level_dependent && skill_lv <= MAX_SKILL_ITEM_REQUIRE && i == 0) {
 						req.itemid[0] = skill->require.itemid[skill_lv - 1];
 						req.amount[0] = skill->require.amount[skill_lv - 1];
 					}
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -18465,6 +18465,18 @@ struct s_skill_condition skill_get_requirement(map_session_data* sd, uint16 skil
 						req.itemid[0] = skill->require.itemid[skill_lv - 1];
 						req.amount[0] = skill->require.amount[skill_lv - 1];
 					}
+
+					if (sd->special_state.no_gemstone == 2) // Remove all Magic Stone required for all skills for VIP.
+						req.itemid[0] = req.amount[0] = 0;
+					else {
+						if (sd->special_state.no_gemstone || (sc && sc->getSCE(SC_INTOABYSS)))
+						{	// Ignore SA_ABRACADABRA & HW_GANBANTEIN from mistress effect
+							if (skill_id != SA_ABRACADABRA && skill_id != HW_GANBANTEIN)
+								req.itemid[0] = req.amount[0] = 0;
+							else if (--req.amount[0] < 1)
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -18465,6 +18465,18 @@ struct s_skill_condition skill_get_requirement(map_session_data* sd, uint16 skil
 						req.itemid[0] = skill->require.itemid[skill_lv - 1];
 						req.amount[0] = skill->require.amount[skill_lv - 1];
```
vstumpf, line 18466: I believe the issue is we write `req.itemid[0]` and `req.amount[0]` when `i = 1`
so we should only run this code if `i == 0`

vstumpf (changes requested):

src/map/skill.cpp
aoshfan: Hi @vstumpf,
I believe that changes is unrelated or not the fix.

The issue was it did not have a condition to make the `req.itemid[0] = req.amount[0] = 0` which result to no Gemstone will be consume by the skill.

vstumpf: It does set it to 0; the issue is we go through the loop twice, and both times we set `itemid[0]` to the level requirement.

The first loop iteration (i = 0) sets req.itemid[0] correctly, then checks req.itemid[i] for gemstones.
The second iteration (i = 1) sets req.itemid[0] again, then checks req.itemid[i] for gemstones. 

So this is the reason why we should only set req.itemid[0] once on line 18464

aoshfan: I see, then i will change code in line `18464` to below ?
`if (level_dependent && skill_lv <= MAX_SKILL_ITEM_REQUIRE && i == 0)`

vstumpf: Yup! That's perfect

aoshfan: Thank you @vstumpf 😄 ,
done added & commited.

src/map/skill.cpp
```diff
@@ -18461,10 +18461,22 @@ struct s_skill_condition skill_get_requirement(map_session_data* sd, uint16 skil
...
+						req.itemid[0] = req.amount[0] = 0;
+					else {
+						if (sd->special_state.no_gemstone || (sc && sc->getSCE(SC_INTOABYSS)))
+						{	// Ignore SA_ABRACADABRA & HW_GANBANTEIN from mistress effect
+							if (skill_id != SA_ABRACADABRA && skill_id != HW_GANBANTEIN)
+								req.itemid[0] = req.amount[0] = 0;
+							else if (--req.amount[0] < 1)
+								req.amount[0] = 1; // Hocus Pocus always use at least 1 gem
+						}
+					}
```
vstumpf, line 18479: I think with the other fix, we no longer need this check, as it's done below

aoshfan: Yup true, didn't notice that, thank you for your review.
Update changes as requested 😄 .

vstumpf (approved):
Thanks for making all the changes!

I also forgot to say, welcome to rathena!

aoshfan:
Thank you @vstumpf for your warm welcome.

I thought i could not comment here already, lol.

May i know when the merge going to happen ? 😄 .

vstumpf:
Sorry, I just got back from a vacation. Had to wait for the tests to pass!

aoshfan:
Ah, i see,

Noted @vstumpf , and branch deleted.

PR #7635, opened 2023-03-11
Implement NPC_GRADUAL_GRAVITY

eppc0330:
* **Addressed Issue(s)**: #6753

* **Server Mode**: Renewal

 Updated new boss skill NPC_GRADUAL_GRAVITY
from https://ro.gnjoy.com/news/devnote/View.asp?category=1&seq=3846495
Split original pull to put 1 skill per single pull.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -13969,9 +13969,7 @@ TIMER_FUNC(status_change_timer){
 		
 	case SC_GRADUAL_GRAVITY:
 		if (sce->val4 >= 0) {
-			int hp = status->max_hp * sce->val2 / 100;
-
-			status_zap(bl, hp, 0);
+			status_zap(bl, status->max_hp * sce->val2 / 100, 0);
 		}
 		break;
 
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -9352,6 +9352,7 @@ static int status_get_sc_interval(enum sc_type type)
 		case SC_LEECHESEND:
 		case SC_DPOISON:
 		case SC_DEATHHURT:
+		case SC_GRADUAL_GRAVITY:
 			return 1000;
 		case SC_BURNING:
 		case SC_PYREXIA:
@@ -10888,6 +10889,11 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 			if( (val4 = tick/(val2 * 1000)) < 1 )
 				val4 = 1;
 			tick_time = val2 * 1000; // [GodLesZ] tick time
+			break;
+		case SC_GRADUAL_GRAVITY:
+			val2 = 10 * val1;
+			val4 = tick / 1000;
+			tick_time = status_get_sc_interval(type);
 			break;
 		case SC_BOSSMAPINFO:
 			if( sd != NULL ) {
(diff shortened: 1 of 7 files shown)


src/map/status.cpp
```diff
@@ -10889,6 +10890,11 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				val4 = 1;
 			tick_time = val2 * 1000; // [GodLesZ] tick time
 			break;
+		case SC_GRADUAL_GRAVITY:
+			val2 = 10 * val1;
+			val4 = tick / 1000;
+			tick_time = status_get_sc_interval(type);
```
Lemongrass3110, line 10896:
```suggestion
			tick_time = status_get_sc_interval(type);
			val4 = tick - tick_time; // Remaining time
```
For interval based SCs val4 should be the remaining time.

src/map/status.cpp
```diff
@@ -13960,6 +13966,18 @@ TIMER_FUNC(status_change_timer){
 			return 0;
 		}
 		break;
+		
+	case SC_GRADUAL_GRAVITY:
+		if (--(sce->val4) >= 0) {
```
Lemongrass3110, line 13971:
```suggestion
		if (sce->val4 >= 0) {
```
For interval based SCs you only need to check if there was still time remaining.

src/map/status.cpp
```diff
@@ -13960,6 +13966,18 @@ TIMER_FUNC(status_change_timer){
...
+			int hp = status->max_hp * sce->val2 / 100;
+
+			if (!status_charge(bl, hp, 0))
+				status_zap(bl, hp, 0);
+			if (sc->getSCE(type))
+				sc_timer_next(interval + tick);
+			return 0;
```
Lemongrass3110, line 13978:
```suggestion
```
This part will be handled at the bottom of the function.
Looks for `// If status has an interval and there is at least 100ms remaining time, wait for next interval` to find it.

src/map/status.cpp
```diff
@@ -13960,6 +13966,18 @@ TIMER_FUNC(status_change_timer){
 			return 0;
 		}
 		break;
+		
+	case SC_GRADUAL_GRAVITY:
+		if (--(sce->val4) >= 0) {
+			int hp = status->max_hp * sce->val2 / 100;
+
+			if (!status_charge(bl, hp, 0))
+				status_zap(bl, hp, 0);
```
Lemongrass3110, line 13975:
```suggestion
			status_zap(bl, hp, 0);
```
Why first try status_charge (fails if not enough HP) and then force it via status_zap (will reduce it even if it kills) in case of failure?

Lemongrass3110 (commented):
Just some small cleanups. But looks good in general!

src/map/status.cpp
```diff
@@ -13960,6 +13966,14 @@ TIMER_FUNC(status_change_timer){
 			return 0;
 		}
 		break;
+		
+	case SC_GRADUAL_GRAVITY:
+		if (sce->val4 >= 0) {
+			int hp = status->max_hp * sce->val2 / 100;
+
+			status_zap(bl, hp, 0);
```
aleos89, line 13974:
```suggestion
			status_zap(bl, status->max_hp * sce->val2 / 100, 0);
```

aleos89 (approved):

Atemo (approved):
Look good!

PR #7636, opened 2023-03-11
Implement NPC_MOVE_COORDINATE

eppc0330:
* **Addressed Issue(s)**: #6753

* **Server Mode**: Renewal

 Updated new boss skill NPC_MOVE_COORDINATE
from https://ro.gnjoy.com/news/devnote/View.asp?category=1&seq=3846495
Split original pull to put 1 skill per single pull.

### DIFF
--- a/src/common/mmo.hpp
+++ b/src/common/mmo.hpp
@@ -89,7 +89,7 @@ typedef uint32 t_itemid;
 #endif
 #define MAX_FAME 1000000000 ///Max fame points
 #define MAX_CART 100 ///Maximum item in cart
-#define MAX_SKILL 1481 ///Maximum skill can be hold by Player, Homunculus, & Mercenary (skill list) AND skill_db limit
+#define MAX_SKILL 1486 ///Maximum skill can be hold by Player, Homunculus, & Mercenary (skill list) AND skill_db limit
 #define DEFAULT_WALK_SPEED 150 ///Default walk speed
 #define MIN_WALK_SPEED 20 ///Min walk speed
 #define MAX_WALK_SPEED 1000 ///Max walk speed
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -7632,22 +7632,24 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 
 	case NPC_MOVE_COORDINATE:
 		{
-		    clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
-		    clif_skill_damage(src, bl, tick, status_get_amotion(src), 0, -30000, 1, skill_id, skill_lv, DMG_SINGLE);
+			clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+			clif_skill_damage(src, bl, tick, status_get_amotion(src), 0, -30000, 1, skill_id, skill_lv, DMG_SINGLE);
 
-		    int16 px = bl->x, py = bl->y;
-		    if (!skill_check_unit_movepos(0, bl, src->x, src->y, 1, 1)) {
-		        return 0;
-		    }
+			int16 px = bl->x, py = bl->y;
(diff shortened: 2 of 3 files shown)


eppc0330:
Small behavior fixed. Thanks to @Lemongrass3110 for advise!

Lemongrass3110 (approved):

src/common/mmo.hpp
```diff
@@ -89,7 +89,7 @@ typedef uint32 t_itemid;
 #endif
 #define MAX_FAME 1000000000 ///Max fame points
 #define MAX_CART 100 ///Maximum item in cart
-#define MAX_SKILL 1481 ///Maximum skill can be hold by Player, Homunculus, & Mercenary (skill list) AND skill_db limit
+#define MAX_SKILL 1486 ///Maximum skill can be hold by Player, Homunculus, & Mercenary (skill list) AND skill_db limit
```
Pokye, line 92:
```suggestion
#define MAX_SKILL 1482 ///Maximum skill can be hold by Player, Homunculus, & Mercenary (skill list) AND skill_db limit
```

src/map/skill.cpp
```diff
@@ -7630,6 +7630,29 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 
 		break;
 
+	case NPC_MOVE_COORDINATE:
+		{
+			clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+			clif_skill_damage(src, bl, tick, status_get_amotion(src), 0, -30000, 1, skill_id, skill_lv, DMG_SINGLE);
+
+			int16 px = bl->x, py = bl->y;
```
aleos89, line 7638:
```suggestion
```

src/map/skill.cpp
```diff
@@ -7630,6 +7630,29 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			if (!skill_check_unit_movepos(0, bl, src->x, src->y, 1, 1)) {
+				return 0;
+			}
+
+			clif_blown(bl);
+
+			// If caster is not a boss, switch coordinates with the target
+			if (status_get_class_(src) != CLASS_BOSS) {
+				if (!skill_check_unit_movepos(0, src, px, py, 1, 1)) {
```
aleos89, line 7647:
```suggestion
				if (!skill_check_unit_movepos(0, src, bl->x, bl->y, 1, 1)) {
```

aleos89 (approved):
Just to get rid of an extra variable since it's not reused.

src/map/skill.cpp
Lemongrass3110: Thats not possible, because bl was already moved at this point in time!

src/common/mmo.hpp
Lemongrass3110: This is intentional to prevent conflicts between the PRs.
It does not really hurt to increase it more than needed.

src/map/skill.cpp
aleos89: Reviewing from mobile so the "check" in the function name confused me if it actually moved the object or not!

src/map/skill.cpp
```diff
@@ -7630,6 +7630,29 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 
 		break;
 
+	case NPC_MOVE_COORDINATE:
+		{
+			clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+			clif_skill_damage(src, bl, tick, status_get_amotion(src), 0, -30000, 1, skill_id, skill_lv, DMG_SINGLE);
```
Atemo, line 7636: Those should be after `skill_check_unit_movepos`, the skill animation shouldn't be shown when the skill fails

eppc0330: I did like this because the skill animation would not be located proper place.

eppc0330: move_coordinate

Check this from KRo, that the effects created before units moved.

PR #7646, opened 2023-03-14
Added a config for banking state check

secretdataz:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

I've received reports the banking UI no longer work with commit 22abdf8. This seems to be caused by clients not reporting its banking UI state to the server, but instead only querying the player's balance when the UI is opened. I apologize for not testing my previous pull request thoroughly.

I thought of reverting 22abdf8 entirely, but it could still be useful in some scenarios, so I'm adding a battle config to re-enable the check instead.

### DIFF
--- a/conf/battle/feature.conf
+++ b/conf/battle/feature.conf
@@ -33,6 +33,11 @@ feature.warp_suggestions: off
 // Requires: 2013-07-24aRagexe or later
 feature.banking: on
 
+// Should Banking strictly checks the UI state on zeny deposit and withdrawal? (Note 1)
+// Note: Disabled by default because existing official clients do not report the banking UI state to the server.
+// But this config could be useful in case of clients that actually report the UI state or custom clients.
+feature.banking_state_enforce: off
+
 // Autotrade persistency (Note 1)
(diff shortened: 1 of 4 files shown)


conf/battle/feature.conf
```diff
@@ -33,6 +33,11 @@ feature.warp_suggestions: off
 // Requires: 2013-07-24aRagexe or later
 feature.banking: on
 
+// Should Banking strictly checks the UI state on zeny deposit and withdrawal? (Note 1)
+// Note: Disabled by default because existing official clients do not report the banking UI state to the server.
+// But this config could be useful in case of clients that actually report the UI state or custom clients.
+feature.banking_state_enforce: off
```
Lemongrass3110, line 39:
```suggestion
feature.banking_state_enforce: no
```
Just sounds better to me.

Lemongrass3110 (approved):

conf/battle/feature.conf
mazvi: It is best to use the usual conventions used in other confs, which are usually yes and no.

aleos89 (approved):

PR #7651, opened 2023-03-19
Deduplicate strlib code

Lemongrass3110:

* **Server Mode**: Both

Deduplicate code that was added in 4e64623

### DIFF
--- a/src/common/malloc.hpp
+++ b/src/common/malloc.hpp
@@ -4,6 +4,8 @@
 #ifndef MALLOC_HPP
 #define MALLOC_HPP
 
+#include <config/core.hpp>
+
 #include "cbasetypes.hpp"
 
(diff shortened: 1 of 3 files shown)


src/common/malloc.hpp
```diff
@@ -4,6 +4,8 @@
 #ifndef MALLOC_HPP
 #define MALLOC_HPP
 
+#include <config/core.hpp>
```
vstumpf, line 7: just confirming, this for `custom/defines_pre.hpp` right?

vstumpf (approved):
Looks much cleaner, thanks!

src/common/malloc.hpp
Lemongrass3110: Yup! So that you can disable the memory manager there.

Issue #7483, opened 2022-12-11
npcwalkto: npc invisible while walking


adaral:
### rAthena Hash

https://github.com/rathena/rathena/commit/c6c6bc131eff2d4cd1fb22dc3192f057217c7d89

### Client Date

2020-04-01

### Server Mode

Pre-Renewal

### Result

When using `npcwalkto`, while the NPC moves it becomes invisible (can only see the shadow) and unclickable. It's an official sprite that moves just fine on older rAthena versions.
![error](https://user-images.githubusercontent.com/8409570/206925709-eaf1e651-672c-4b92-b2cf-5b4d321deeee.png)

### Expected Result

While moving, the NPC should be visible and clickable.

### How to Reproduce

Make an NPC using `npcwalkto`

### Official Information

-

Lemongrass3110:
Please provide a sample script.

adaral:
@Lemongrass3110 
`prontera,155,180,5	script	Moving	90,{

    npcspeed 600;

    OnTimer2000:
    npcwalkto 149,169;
    end;

    OnTimer21000:
    npcwalkto 162,169;
    end;

    OnTimer35000:
    npcwalkto 155,80;
    end;

    OnInit:
    initnpctimer;
    end;

}
`

Haydrich:
> 

as i remember while working on ep 18.1, it's related to npc sprite? you should choose an npc that have walking position in sprite?

adaral:
> > 
> 
> as i remember while working on ep 18.1, it's related to npc sprite, you should choose an npc that have walking position in sprite

Not really. I have used some sprites in older versions of rAthena that are not working now

Pienne:
For anyone else finding this issue, I've tested this a bit prior and it's because the NPC sprite data doesn't have animation/sprite data for walking.

You'll find that if you find an NPC with valid walking sprites they will walk and be clickable. You can find one using a grf editor, I found some kafra NPCs that have valid walking sprites and they walk around just fine.

PR #7652, opened 2023-03-19
Fixed an issue where normal NPC had no sprite when walking

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/7483

* **Server Mode**: both


vstumpf (approved):
Figuring out how to classify viewids with walking sprites will definitely be an interesting problem!

LGTM

PR #7655, opened 2023-03-19
Removed NoForcedEnd for SC_MONSTER_TRANSFORM

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

On official server the status can be removed by sc_end.

- [ ] Check if we have any reason to keep it on
- [ ] The status remains unchanged on pre-renewal?

### DIFF
--- a/db/re/status.yml
+++ b/db/re/status.yml
@@ -5822,7 +5822,6 @@ Body:
       NoBanishingBuster: true
       NoClearance: true
       SendVal1: true
-      NoForcedEnd: true
   - Status: Angel_Protect
     Icon: EFST_ANGEL_PROTECT
     Flags:


Lemongrass3110:
I dont have any official info, but:
The reason is simple I guess.
Because of our shitty code.

```cpp
		if( sc_db->flag[SCF_NOCLEARBUFF] && sc_db->flag[SCF_NOFORCEDEND] ){
			ShowError( "buildin_sc_end: Status %d cannot be cleared.\n", type );
			return SCRIPT_CMD_FAILURE;
		}

		//This should help status_change_end force disabling the SC in case it has no limit.
		sce->val1 = sce->val2 = sce->val3 = sce->val4 = 0;
		status_change_end(bl, (sc_type)type);
```

Since we remove the vals down there, it probably loses the connection to the second status that is stored in val2.

Atemo:
Removing the flag is not a problem for this sc

PR #7662, opened 2023-03-24
Small correction on the attribute table

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal

Follow-up https://github.com/rathena/rathena/commit/e4e8063ec3ea85e06b0dd367eb0aec0aa9535700

- Defense Wind and Fire values have been reversed for Undead weapon element.
- There should be no negative defense according to the official tables.

### DIFF
--- a/db/re/attr_fix.yml
+++ b/db/re/attr_fix.yml
@@ -120,7 +120,7 @@ Body:
       Holy: 125
       Dark: 0
       Ghost: 100
-      Undead: -25
+      Undead: 0
     Ghost:
       Neutral: 90
       Water: 100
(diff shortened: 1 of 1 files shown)


vstumpf:
Maybe we can have a battle config that will clamp values?

Atemo:
imo it would be a mistake to leave the old values in db/re folder, but we could add them (commented) in import

vstumpf:
You're right, I don't know if we even need it in import. Someone can do it themselves.

vstumpf (approved):

PR #7676, opened 2023-04-03
Updates map_drops.yml 

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: 

Adjusted the rate in map_drops.yml to 1/100000 instead of 1/10000

### DIFF
--- a/src/common/random.cpp
+++ b/src/common/random.cpp
@@ -43,10 +43,3 @@ int32 rnd_value( int32 min, int32 max ){
 	return min + (int32)( rnd_uniform() * ( max - min + 1 ) );
 }
 
-bool rnd_chance( uint16 chance, uint16 base ){
-	return rnd_value( 0, base ) < chance;
-}
-
-bool rnd_chance32( int32 chance, int32 base ){
-	return rnd_value( 0, base ) < chance;
-}
--- a/src/common/random.cpp
+++ b/src/common/random.cpp
@@ -46,3 +46,7 @@ int32 rnd_value( int32 min, int32 max ){
 bool rnd_chance( uint16 chance, uint16 base ){
 	return rnd_value( 0, base ) < chance;
 }
+
+bool rnd_chance32( int32 chance, int32 base ){
+	return rnd_value( 0, base ) < chance;
+}
--- a/src/common/random.hpp
+++ b/src/common/random.hpp
@@ -10,7 +10,9 @@ void rnd_init(void);
 
 int32 rnd(void);// [0, SINT32_MAX]
 int32 rnd_value(int32 min, int32 max);// [min, max]
-bool rnd_chance( uint16 chance, uint16 base );
-bool rnd_chance32( int32 chance, int32 base );
+
+template <typename T> bool rnd_chance( T chance, T base ){
+	return rnd_value( 0, base ) < chance;
+}
 
 #endif /* RANDOM_HPP */
--- a/src/common/random.hpp
+++ b/src/common/random.hpp
@@ -11,5 +11,6 @@ void rnd_init(void);
 int32 rnd(void);// [0, SINT32_MAX]
 int32 rnd_value(int32 min, int32 max);// [min, max]
 bool rnd_chance( uint16 chance, uint16 base );
+bool rnd_chance32( int32 chance, int32 base );
 
 #endif /* RANDOM_HPP */
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2904,7 +2904,7 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
 		if( mapdrops != nullptr ){
 			// Process map wide drops
 			for( const auto& it : mapdrops->globals ){
-				if( rnd_chance32( it.second->rate, 100000u ) ){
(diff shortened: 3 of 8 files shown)


src/map/mob.cpp
```diff
@@ -2904,8 +2900,8 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
 		if( mapdrops != nullptr ){
 			// Process map wide drops
 			for( const auto& it : mapdrops->globals ){
-				if( rnd_chance( it.second->rate, 10000 ) ){
-					mob_item_drop( md, dlist, mob_setdropitem( it.second.get(), 1, md->mob_id ), 0, it.second->rate, homkillonly || merckillonly );
+				if( rnd_value( 0, 100000 ) < it.second->rate ){
+					mob_item_drop( md, dlist, mob_setdropitem( it.second.get(), 1, md->mob_id ), 0, it.second->rate, (homkillonly || merckillonly), true );
```
Lemongrass3110, line 2904: I would prefer multiplying by 10 at these two calls and not touch the other function at all.
The mob_item_drop function is already unreadable enough with all those short ifs... :-(

src/map/mob.cpp
```diff
@@ -2904,8 +2900,8 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
 		if( mapdrops != nullptr ){
 			// Process map wide drops
 			for( const auto& it : mapdrops->globals ){
-				if( rnd_chance( it.second->rate, 10000 ) ){
-					mob_item_drop( md, dlist, mob_setdropitem( it.second.get(), 1, md->mob_id ), 0, it.second->rate, homkillonly || merckillonly );
+				if( rnd_value( 0, 100000 ) < it.second->rate ){
```
Lemongrass3110, line 2903: Can we add a uint32 variant of the rnd_chance function instead?

src/common/random.hpp
```diff
@@ -11,5 +11,6 @@ void rnd_init(void);
 int32 rnd(void);// [0, SINT32_MAX]
 int32 rnd_value(int32 min, int32 max);// [min, max]
 bool rnd_chance( uint16 chance, uint16 base );
+bool rnd_chance32( int32 chance, int32 base );
```
Lemongrass3110, line 14:
```suggestion
bool rnd_chance( uint32 chance, uint32 base );
```
C++ has function overloading.

Lemongrass3110 (approved):
You still need to add parsing logic. It is still capped to uint16rate there!

src/common/random.hpp
Atemo: `rnd_chance` is using `rnd_value` (`int32 rnd_value(int32 min, int32 max);`) leaving it in int32 was less annoying :(

The compiler shows me a warning from `rnd_chance` in `skill.cpp` when using the overloading, so I renamed the function. Not sure how to fix the warning

src/common/random.cpp
```diff
@@ -46,3 +46,7 @@ int32 rnd_value( int32 min, int32 max ){
 bool rnd_chance( uint16 chance, uint16 base ){
 	return rnd_value( 0, base ) < chance;
 }
+
+bool rnd_chance32( int32 chance, int32 base ){
+	return rnd_value( 0, base ) < chance;
+}
```
Lemongrass3110, line 52:
```suggestion
```

src/common/random.hpp
Lemongrass3110, line 14:
```suggestion

template <typename T> bool rnd_chance( T chance, T base ){
	return rnd_value( 0, base ) < chance;
}
```

src/map/mob.cpp
```diff
@@ -2904,8 +2904,10 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
 		if( mapdrops != nullptr ){
 			// Process map wide drops
 			for( const auto& it : mapdrops->globals ){
-				if( rnd_chance( it.second->rate, 10000 ) ){
-					mob_item_drop( md, dlist, mob_setdropitem( it.second.get(), 1, md->mob_id ), 0, it.second->rate, homkillonly || merckillonly );
+				if( rnd_chance32( it.second->rate, 100000 ) ){
```
Lemongrass3110, line 2907:
```suggestion
				if( rnd_chance32( it.second->rate, 100000u ) ){
```

src/map/mob.cpp
```diff
@@ -2914,8 +2916,10 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
 
 			if( specific != mapdrops->specific.end() ){
 				for( const auto& it : specific->second ){
-					if( rnd_chance( it.second->rate, 10000 ) ){
-						mob_item_drop( md, dlist, mob_setdropitem( it.second.get(), 1, md->mob_id ), 0, it.second->rate, homkillonly || merckillonly );
+					if( rnd_chance32( it.second->rate, 100000 ) ){
```
Lemongrass3110, line 2919:
```suggestion
					if( rnd_chance32( it.second->rate, 100000u ) ){
```

Lemongrass3110 (approved):

PR #7685, opened 2023-04-06
Converting scripts_commands.txt to markdown version

SapitoSucio:
* **Addressed Issue(s)**: 
 Cleaning and converting scripts_commands.txt to markdown version

* **Server Mode**: 
 Both

Cleaning and converting scripts_commands.txt to markdown version

### DIFF
--- a/doc/script_commands.md
+++ b/doc/script_commands.md
@@ -235,7 +235,9 @@ ex: if your NPC is named 'Hunter#hunter1', it will be displayed as 'Hunter'
 ## Define a warp point
 
 **\<from mapname\>**,**\<fromX\>**,**\<fromY\>**,**\<facing\>****%TAB%warp\%TAB\%****\<warp name\>** **\%TAB\%** **\<spanx\>**,**\<spany\>**,**\<to mapname\>**,**\<toX\>**,**\<toY\>**\
-**\<from mapname\>**,**\<fromX\>**,**\<fromY\>**,**\<facing\>****%TAB%warp2\%TAB\%****\<warp name\>** **\%TAB\%\<spanx\>**,**\<spany\>**,**\<to mapname\>**,**\<toX\>**,**\<toY\>**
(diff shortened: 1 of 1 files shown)


aleos89:
I guess this is more of a general question, since I know one large file is easier to search locally, but for markdown do we want to separate out commands into their own pages on the doc site?

Akkarinage:
> I guess this is more of a general question, since I know one large file is easier to search locally, but for markdown do we want to separate out commands into their own pages on the doc site?

Maybe not individual pages, but I wouldn't mind seeing them split into the 15 categories.. though some extra work re-positioning some of them would need to be done. Categories 7 and up (from Instance commands onwards) are pretty clear, but the first 6 are kinda.. everywhere.

Lemongrass3110:
I am kind of confusing why we have "user guides" and the normal "doc" now.
It would be nice if we could move all "doc" to the "user guides" and have it in markdown there.

Additionally I am still in favor of splitting it up on a one file per command basis and have a runnable sample for each script command somewhere in /npc/test/ so that it is also run with CI.
Since it was not possible until now: would the new tool support generating an imported codeblock from a file?

Danil0v3s:
Additionally to what have been said, here's my two cents:

Since we're going to use the html generator tool and it can generate a side menu navigation for each of the categories we define, it could be worth checking a way of may be splitting this 12k lines file into smaller files and leveraging that side menu feature.

Also, since we're using markdown now, we should reformulate the sample snippets as well. This:

-**\%TAB\%shop\%TAB\%\<NPC Name\>\%TAB\%\<spriteid\>{,discount},\<itemid\>:\<price\>{,\<itemid\>:\<price\>...}\<mapname\>,\<x\>,\<y\>,\<facing\>\%TAB\%shop\%TAB\%\<NPC Name\>\%TAB\%\<spriteid\>{,discount},\<itemid\>:\<price\>{,\<itemid\>:\<price\>...}**

Doesn't bring as much value at a first glance as this, for example:

- `- <TAB> shop <TAB> <NPC Name> <TAB> <spriteid>,[<itemid>:<price>],[<itemid>:<price>]...`
- `<mapname>,<x>,<y>,<direction> <TAB> shop <TAB> <NPC Name> <TAB> <spriteid>,[<itemid>:<price>],[<itemid>:<price>]...`

Issue #7723, opened 2023-04-19
Status clearing skills such as Gospel and Tarot are flawed


Singe-Horizontal:
### rAthena Hash

ea65c1b9b

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

These skills also including RK_LUXANIMA and RK_REFRESH will fail to dispel buffs and debuffs.

### Expected Result

All indicated buffs/debuffs should be dispelled.

### How to Reproduce

Cast Blessing and Full Chemical Protection on a paladin, then use gospel. Only the chemical protection will get dispelled.

### Official Information

N/A

PR #7724, opened 2023-04-19
Corrects a short-circuit bug in status_change_clear_buffs

Singe-Horizontal:
* **Addressed Issue(s)**: 
#7723

* **Server Mode**: 
Both

Now, previous negative tests will not end the status search loop prematurely.


src/map/status.cpp
```diff
@@ -14926,33 +14926,34 @@ void status_change_clear_buffs(struct block_list* bl, uint8 type)
 	for (const auto &it : status_db) {
 		sc_type status = static_cast<sc_type>(it.first);
 		std::bitset<SCF_MAX> flag = it.second->flag;
```
vstumpf, line 14928: While you're here, can you const ref this? We don't need to make a copy of the bitset.

vstumpf (approved):
Looks good!

Issue #7728, opened 2023-04-20
implode command causes a server freeze when passing an undefined string


Singe-Horizontal:
### rAthena Hash

9051fc644

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

Server freezes when using `implode` command on an undefined string variable.

### Expected Result

The server is supposed to log that an empty array was passed.

### How to Reproduce

Pass an undefined variable .@var$ to implode : `implode(.@var$)`;

### Official Information

N/A

PR #7729, opened 2023-04-20
Prevent an uint underflow and reformat array_size occurences in some commands

Singe-Horizontal:
* **Addressed Issue(s)**: 
#7728

* **Server Mode**: 
Both

This reformats a naming convention related to array_size in some scripts commands that could cause confusion.
By doing so it corrects a server crash/freeze in implode command : 
When an empty variable is passed to `implode`, the line
`size_t array_size = script_array_highest_key( st, sd, name, reference_getref( data ) ) - 1`
causes an unsigned int underflow, not detected in the following `if(array_size == -1)` test.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -6671,15 +6671,15 @@ BUILDIN_FUNC(inarray)
 	if (not_server_variable(*name) && !script_rid2sd(sd))
 		return SCRIPT_CMD_FAILURE;
 
-	array_size = script_array_highest_key(st, sd, name, ref) - 1;
+	array_size = script_array_highest_key(st, sd, name, ref);
 
-	if (array_size < 0)
+	if (array_size <= 0)
 	{
 		script_pushint(st, -1);
 		return SCRIPT_CMD_SUCCESS;
 	}
 
-	if (array_size > SCRIPT_MAX_ARRAYSIZE)
+	if (array_size >= SCRIPT_MAX_ARRAYSIZE)
 	{
 		ShowError("buildin_inarray: The array is too large.\n");
 		script_reportdata(data);
@@ -6692,7 +6692,7 @@ BUILDIN_FUNC(inarray)
 	if( is_string_variable( name ) ){
 		const char* value = script_getstr( st, 3 );
 
-		for( i = 0; i <= array_size; ++i ){
+		for( i = 0; i < array_size; ++i ){
 			const char* temp = get_val2_str( st, reference_uid( id, i ), ref );
 
 			if( !strcmp( temp, value ) ){
@@ -6708,7 +6708,7 @@ BUILDIN_FUNC(inarray)
 	}else{
 		int64 value = script_getnum64( st, 3 );
 
-		for( i = 0; i <= array_size; ++i ){
+		for( i = 0; i < array_size; ++i ){
 			int64 temp = get_val2_num( st, reference_uid( id, i ), ref );
 
 			if( temp == value ){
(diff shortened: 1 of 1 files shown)


vstumpf (commented):
The changes look good, but i just woke up so I'll take a look again in a couple of hours

src/map/script.cpp
```diff
@@ -6671,15 +6671,15 @@ BUILDIN_FUNC(inarray)
 	if (not_server_variable(*name) && !script_rid2sd(sd))
 		return SCRIPT_CMD_FAILURE;
 
-	array_size = script_array_highest_key(st, sd, name, ref) - 1;
+	array_size = script_array_highest_key(st, sd, name, ref);
```
vstumpf, line 6674: Cam we make `array_size` unsigned? if we put it into a signed int, the `array_size >= SCRIPT_MAX_ARRAYSIZE` comparisons don't make sense.

let's make it const as well to prevent anything from changing it after the 0 check.

src/map/script.cpp
```diff
@@ -6671,15 +6671,15 @@ BUILDIN_FUNC(inarray)
...
 
-	if (array_size < 0)
+	if (array_size <= 0)
```
vstumpf, line 6676: If we move to unsigned, then you can just check equality with 0

src/map/script.cpp
```diff
@@ -6774,7 +6774,7 @@ BUILDIN_FUNC(countinarray)
 
 	i = reference_getindex(data1);
 	j = reference_getindex(data2);
-	if (array_size1 < i || array_size2 < j)
+	if (array_size1 - 1 < i || array_size2 - 1 < j)
```
vstumpf, line 6777: Can we change this check to `if (i + 1 > array_size1 || j + 1 > array_size2)`?
This is checking if the reference that we passed in is outside the array 
> for example, `.@arr[10]`
if the highest key is 2, then `.@arr[10]` is outside the size.

src/map/script.cpp
```diff
@@ -17002,16 +17002,16 @@ BUILDIN_FUNC(implode)
 	}
 
 	//count chars
-	size_t array_size = script_array_highest_key( st, sd, name, reference_getref( data ) ) - 1;
+	size_t array_size = script_array_highest_key( st, sd, name, reference_getref( data ) );
 
-	if(array_size == -1) { //empty array check (AmsTaff)
+	if(array_size <= 0) {
```
vstumpf, line 17007: just check if size == 0

src/map/script.cpp
```diff
@@ -17002,16 +17002,16 @@ BUILDIN_FUNC(implode)
...
 		ShowWarning("script:implode: array length = 0\n");
 		script_pushstrcopy( st, "NULL" );
 	} else {
 		const char *glue = nullptr, *temp;
 		size_t len = 0, glue_len = 0, k = 0;
 
-		for( int i = 0; i <= array_size; ++i ){
+		for( int i = 0; i < array_size; ++i ){
```
vstumpf, line 17014: because `array_size` is a `size_t`, let's make `i` the same.

src/map/script.cpp
```diff
@@ -6754,16 +6754,16 @@ BUILDIN_FUNC(countinarray)
 	if (not_server_variable(*name1) && not_server_variable(*name2) && !script_rid2sd(sd))
 		return SCRIPT_CMD_FAILURE;
 
-	array_size1 = script_array_highest_key(st, sd, name1, ref1) - 1;
-	array_size2 = script_array_highest_key(st, sd, name2, ref2) - 1;
+	array_size1 = script_array_highest_key(st, sd, name1, ref1);
+	array_size2 = script_array_highest_key(st, sd, name2, ref2);
```
vstumpf, line 6758: can we make these const and unsigned?

src/map/script.cpp
```diff
@@ -17002,16 +17002,16 @@ BUILDIN_FUNC(implode)
 	}
 
 	//count chars
-	size_t array_size = script_array_highest_key( st, sd, name, reference_getref( data ) ) - 1;
+	size_t array_size = script_array_highest_key( st, sd, name, reference_getref( data ) );
```
vstumpf, line 17005: let's make it const

src/map/script.cpp
```diff
@@ -6754,16 +6754,16 @@ BUILDIN_FUNC(countinarray)
 	if (not_server_variable(*name1) && not_server_variable(*name2) && !script_rid2sd(sd))
 		return SCRIPT_CMD_FAILURE;
 
-	array_size1 = script_array_highest_key(st, sd, name1, ref1) - 1;
-	array_size2 = script_array_highest_key(st, sd, name2, ref2) - 1;
+	array_size1 = script_array_highest_key(st, sd, name1, ref1);
+	array_size2 = script_array_highest_key(st, sd, name2, ref2);
 
-	if (array_size1 < 0 || array_size2 < 0)
+	if (array_size1 <= 0 || array_size2 <= 0)
```
vstumpf, line 6760: just check for == 0

vstumpf (commented):
Overall good, just a few nitpicks

src/map/script.cpp
Singe-Horizontal: Couldn't this overflow?

vstumpf: You're right, it could overflow.

vstumpf (approved):
Thanks for making the changes!

PR #7739, opened 2023-04-28
Remove md->attacked_id at start of mob_ai_sub_hard

Singe-Horizontal:
* **Addressed Issue(s)**: 
N/A

* **Server Mode**: 
Both

This fixes a bug where an unwanted rude cascade.
It is due to attacker id fields not being reset at the start of mob_ai_sub_hard.
That could create situations where mobs after entering idle state would proc rude skills such as teleport once a player would come nearby, because they wouldn't be in range to retaliate to the stored attacked_id.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1702,8 +1702,10 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 
 	md->last_thinktime = tick;
 
-	if (md->ud.skilltimer != INVALID_TIMER)
+	if (md->ud.skilltimer != INVALID_TIMER) {
+		md->attacked_id = md->norm_attacked_id = 0;
 		return false;
+	}
 
 	// Abnormalities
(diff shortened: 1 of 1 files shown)


Singe-Horizontal:
I forgot to add that this happens when a mob is hit while casting and loses its target before it ends its cast.

src/map/mob.cpp
```diff
@@ -1702,8 +1702,10 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 
 	md->last_thinktime = tick;
 
-	if (md->ud.skilltimer != INVALID_TIMER)
+	if (md->ud.skilltimer != INVALID_TIMER) {
+		md->attacked_id = md->norm_attacked_id = 0;
```
vstumpf, line 1706: I'm not sure this is where we want to put this.
This means the mob drops the target while casting? Maybe we should do it at the end of `skill_castend_id`, if the cast fails and it's a mob, set the attacked_id and norm_attacked_id to 0?

Or, can we put this in mob_unlocktarget?

vstumpf (changes requested):

Playtester:
Is this still relevant? The setting of attacked_id should already be working as on Aegis now. The monster AI will be inactive and attacked_id will be set to 0 on teleport, which should already prevent multi-teleporting. So I think this won't be necessary anymore.

Singe-Horizontal:
Not relevant anymore, closing this.

PR #7752, opened 2023-05-10
Clean-up script.cpp

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

Introduced `SCRIPT_CMD_END` which displays an error and set the script state to `END`.

Removed `script_reportfunc(st)`, `script_reportsrc(st)` and `st->state = END;` from `script_rid2sd_`, which are no longer needed.

This also fixes the logic where `script_charid2sd`, `script_accid2sd`, `script_nick2sd`, `script_mapid2sd` and `script_rid2bl` should set the script state to `END` when the command fails.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -173,7 +173,7 @@ static bool script_rid2bl_(struct script_state *st, uint8 loc, struct block_list
 
 	if ( !script_hasdata(st, loc) || ( unit_id = script_getnum(st, loc) ) == 0)
 		unit_id = st->rid;
-		
+
 	*bl =  map_id2bl(unit_id);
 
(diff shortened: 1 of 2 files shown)


Lemongrass3110:
Kind of reminds me of #2063

Do we really need to differ between those two?
In my opinion still any kind of failure should lead to ending the script.

Atemo:
Yeah it's kinda similar to https://github.com/rathena/rathena/pull/2063. This PR was started to fix the logic for `script_charid2sd` etc.. and one thing after another it ends like this. `SCRIPT_CMD_END` was added to follow the old logic, however I don't know any special case where the script should not terminate indeed.

Issue #7743, opened 2023-05-02
sc->cant.attack being checked when sc = NULL is executed on a previous line


Leemonn:
### rAthena Hash

84bf6b87f1946f28055e90bf713cedd4f401c685

### Client Date

2015-11-04

### Server Mode

Pre-Renewal, Renewal

### Result

This method `status_check_skilluse` of `status.cpp` (line 1923) has the following coding:
```c++
bool status_check_skilluse(struct block_list *src, struct block_list *target, uint16 skill_id, int flag) {
	struct status_data *status;
	status_change *sc = NULL, *tsc;
	int hide_flag;

	status = src ? status_get_status_data(src) : &dummy_status;

	if (src && src->type != BL_PC && status_isdead(src))
		return false;

	if (!skill_id) { // Normal attack checks.
		if (sc && sc->cant.attack)
			return false;
```

But as you can see it makes no sense to check for `sc && sc->cant.attack`, because `sc` will always be `NULL`.

### Expected Result

-

### How to Reproduce

-

### Official Information

-

Singe-Horizontal:
Indeed I noticed cases where you can use skills/items while being stunned for instance.

PR #7753, opened 2023-05-11
Fixed a cant.attack bypass

Singe-Horizontal:
* **Addressed Issue(s)**: 
#7743

* **Server Mode**: 
Both

This moves the pointer assignation before the check so cant.attack can be properly tested for normal attacks.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -1922,23 +1922,18 @@ int status_revive(struct block_list *bl, unsigned char per_hp, unsigned char per
  */
 bool status_check_skilluse(struct block_list *src, struct block_list *target, uint16 skill_id, int flag) {
 	struct status_data *status;
-	status_change *sc, *tsc;
 	int hide_flag;
 
 	if (src) {
 		if (src->type != BL_PC && status_isdead(src))
 			return false;
-		sc = status_get_sc(src);
 		status = status_get_status_data(src);
 	}else{
 		status = &dummy_status;
-		sc = nullptr;
 	}
 
-	if(target)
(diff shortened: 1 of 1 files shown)


src/map/status.cpp
```diff
@@ -1925,10 +1925,13 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
 	status_change *sc = NULL, *tsc;
```
Lemongrass3110, line 1925:
```suggestion
	status_change *sc, *tsc;
```

src/map/status.cpp
```diff
@@ -1925,10 +1925,13 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
...
-	status = src ? status_get_status_data(src) : &dummy_status;
-
-	if (src && src->type != BL_PC && status_isdead(src))
-		return false;
+	if (src) {
+		if (src->type != BL_PC && status_isdead(src))
+			return false;
+		sc = status_get_sc(src);
+		status = status_get_status_data(src);
+	}else
```
Lemongrass3110, line 1933:
```suggestion
	}else{
```

src/map/status.cpp
Lemongrass3110: Additionally: Can you move the tsc declaration to the respective line to prevent a similar issue with that variable please?

Lemongrass3110 (approved):

Lemongrass3110 (approved):

aleos89 (approved):

PR #7754, opened 2023-05-12
Removes player item bonus rate cap warnings

aleos89:
* **Addressed Issue(s)**: N/A

  * Removes all of the warnings that were displayed when players had item bonuses that eventually reached the min/max rate.
  * The rate is already capped so the warning was just a redundant message that could spam the console.

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -3087,9 +3087,6 @@ static void pc_bonus_autospell(std::vector<s_autospell> &spell, uint16 id, uint1
 
 	struct s_autospell entry = {};
 
-	if (rate < -1000 || rate > 1000)
-		ShowWarning("pc_bonus_autospell: Item bonus rate %d exceeds -1000~1000 range, capping.\n", rate);
-
 	entry.id = id;
 	entry.lv = lv;
 	entry.rate = cap_value(rate, -1000, 1000);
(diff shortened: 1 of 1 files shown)


Lemongrass3110:
Wasnt that intended to detect overpowered items/item combinations?
But I guess without giving any information about the player and a hint of the items involved or the one triggering the final overflow it will be pretty much worthless?

aleos89:
Correct, the intent was just as you stated, to try to derive possible situations where a player is getting a large rate bonus when they shouldn't be. Since we don't pass any item information this low to these functions this warning is pretty much useless since we can't specify what's invoking it. Another point is that some of these _may_ be valid since many new items have several bonuses that could easily exceed 100%.

Lemongrass3110 (approved):

src/map/pc.cpp
```diff
@@ -3605,13 +3587,8 @@ static void pc_bonus_itembonus(std::vector<s_item_bonus> &bonus, uint16 id, int
...
 	struct s_item_bonus entry = {};
 
-	if (cap_rate && (val < -10000 || val > 10000)) {
-		ShowWarning("pc_bonus_itembonus: Item bonus val %d exceeds -10000~10000 range, capping.\n", val);
-		val = cap_value(val, -10000, 10000);
-	}
-
 	entry.id = id;
-	entry.val = val;
+	entry.val = cap_value(val, -10000, 10000);
```
attackjom, line 3591: It seem cap_rate boolean is deleted , so it make bSkillCooldown can reduce more than 10000 msec.

Issue #7757, opened 2023-05-13
too many cash shop item crash


ChiesaV:
### rAthena Hash

latest

### Client Date

2022-04-06 Kro from Nemo with RSU, Ragnarok English

### Server Mode

Renewal

### Result

[Error]: _mmalloc: 4294950912
[Fatal Error]: Received a crash signal, trying to handle it as good as possible!
[Error]: Server received crash signal! Attempting to save all online characters!
on map server
When too many cash shop item, but it dont happend in the previous versin of rathena (Ep 18, item_cash still using txt).

### Expected Result

Not crash, because old rathena which use item cash txt is capable of it.

### How to Reproduce

i can provide the file.

### Official Information

2022 04 kro one from nemo.

aleos89:
The doubt sets in with you using the latest version of rAthena when you state your item_cash is still in TXT format. I'm not sure what the client limit is but maybe someone else knows. How many items do you have set in your database?

ChiesaV:
No, I mean, it was fine during rathena item_cash txt version.
I have 2 version of rathena and ragnarok.
It upto 16500 item in the txt, and almost the same in the yml version.
But the yml version one is crashed directly after login.

vstumpf:
I don't see a reason why the cashshopdb would allocate that much data.

Are you sure it's not the itemdb that is causing the crash?
Is it _only_ the cashdb changing from txt to yml? Can you run the server in a debugger so we get a full stacktrace?

ChiesaV:
hey vstumpf, 
could you give me link / brief how to do it ?
and for the conclusion i came up with, because I try to reduce it with 3 different items randomly. and it always succeed when the data amount is on 25% of the total I initially try.
and why I submit this with bug tag, because the txt version is capable of doing it. Even though I understand that it for cheating and personal gameplay purpose to have that many of database. 
It would be great if we can have same capability with the old one tho..

ChiesaV:
when I input Infinite_Concentration_Potion into the yml, it said not found and not displayed at cash shop,
but command gm using itemid 12884 it is work, using @item Infinite_Concentration_Potion not work.
Meanwhile, using while using infinite_berserk_potion, both itemid and aegisname work at using command gm @item.

ChiesaV:
seems the root cause is, every tab, can't have more than around 250-260 pages. if more, crash after character selection.

Lemongrass3110:
[...]

PR #7763, opened 2023-05-16
Added support for really big cashshops

Lemongrass3110:
* **Addressed Issue(s)**: #7757

* **Server Mode**: Both

Our predecessors seem to have forgotten to also report the source trying to allocate an invalid size of memory.

Thanks to @ChiesaV


ChiesaV:
[Error]: _mmalloc: Invalid allocation size 4294950912 bytes at F:\RagnarokOffline\Server\rathena-master\src\common\socket.cpp:791
[Fatal Error]: Received a crash signal, trying to handle it as good as possible!
[Error]: Server received crash signal! Attempting to save all online characters!

Issue #7766, opened 2023-05-17
@reloadnpcfile causing scripts to unload, disappear or duplicate


Kiranetix:
### rAthena Hash

5443d566a787da87b3341fe4e07232ee1ee65177

### Client Date

2018-08-08

### Server Mode

Pre-Renewal

### Result

When using @reloadnpcfile, when reloading 2 different script files, the latest script reloaded will unload the previously reloaded script. Further reloads will cause the other reloaded scripts to unload, disappear or duplicate. This also causes dynamic NPC menu, like quest shop, to disappear.

### Expected Result

Scripts will not unload/disappear or duplicate

### How to Reproduce

1. Normally load /npc/custom/quests/quest_shop.txt
2. Normally load /npc/custom/healer.txt
3. use @reloadnpcfile to reload quest_shop.txt (@reloadnpcfile npc/custom/quests/quest_shop.txt)
4. use @reloadnpcfile to reload healer.txt (@reloadnpcfile npc/custom/healer.txt)
5. use @reloadnpcfile to reload quest_shop.txt (@reloadnpcfile npc/custom/quests/quest_shop.txt)

This will cause the healer npc to disappear.

...furthermore use @reloadnpcfile twice on the script that disappeared, console will throw duplicate npc warning

To fix the issue use @reloadscript. But the next time @reloadnpcfile is used on 2 or more different scripts, the same issue will occur.

### Official Information

N/A

### Modifications that may affect results

NONE, I'm using fresh rAthena.

Melk3000:
Confirmed.

Triedge:
Confirm :(

sinju92:
Confirm 1000%

PR #7769, opened 2023-05-18
Fixes npc last loaded path caching

Singe-Horizontal:
* **Addressed Issue(s)**: 
#7766

* **Server Mode**: 
Both

This removes a redundant pointer that kept track of last loaded npc path, which was always pointing to newly loaded paths. This was preventing a correct unloading and causing duplicates.


Singe-Horizontal:
Shall I add something to this PR?

vstumpf (approved):
I think eventually we should probably remove this, do we even need this? Maybe we can use shared_ptrs with strings, but that's for another PR

PR #7774, opened 2023-05-20
Change unit's size dynamically via script

SapitoSucio:
* **Addressed Issue(s)**: 
You can't change a unit's size dynamically once it's already created

* **Server Mode**: 
Both

using the script setunitsize let's change a unit's size dynamically, it let you change Mobs and NPCs

Script for testing
```cpp
prontera,155,185,5	script	UnitSizeTest	4_F_KAFRA1,{
    monster "prontera",159,183,"Poringu",1002,1,strnpcinfo(0)+"::OnPoringDead",0;
	dispbottom "Mob ID: "+$@mobid[0];
	do {
		.@size = select("Make it Small:Make it Big:Cancel");
		setunitsize $@mobid[0], .@size;
	} while (.@menu != 3);
	end;

OnPoringDead:
    mapannounce "prontera","Poringu has been defeated!",bc_map,"0x00FF00";
    end;

OnMy:
	input .@size;
	setunitsize $@mobid[0], .@size;
	end;
OnMy2:
	input .@GID;
	input .@size;
	setunitsize .@GID, .@size;
	end;

OnInit:
	bindatcmd "lastunitsize",strnpcinfo(3)+"::OnMy";
	bindatcmd "unitsize",strnpcinfo(3)+"::OnMy2";
	end;
}
```

If you try the binded unitsize command, it will appear an input window and if you don't have `input_max_value` to `INT_MAX` it will always return `0`

https://github.com/rathena/rathena/assets/7357354/dbae8922-9139-404e-a768-5f622838d618

### DIFF
--- a/src/map/script.hpp
+++ b/src/map/script.hpp
@@ -2176,7 +2176,7 @@ int conv_num(struct script_state *st, struct script_data *data);
 const char* conv_str(struct script_state *st,struct script_data *data);
 void pop_stack(struct script_state* st, int start, int end);
 TIMER_FUNC(run_script_timer);
-TIMER_FUNC(mob_size_timer);
+TIMER_FUNC(unit_size_timer);
 void script_stop_sleeptimers(int id);
 struct linkdb_node *script_erase_sleepdb(struct linkdb_node *n);
 void script_attach_state(struct script_state* st);


aleos89:
We support size adjustment of objects via `setunitdata` (except for NPC, but that could be added). The only difference is currently there's no visual refresh like you have. I guess what I'm getting at is if we should add another sub-script command or expand what we have. The visual portion could be added with a new script command (`unitspecialeffect`) since I think the 2 script commands for special effects only work on NPC and players.

Just my 2 cents!

Lemongrass3110:
> We support size adjustment of objects via `setunitdata` (except for NPC, but that could be added). The only difference is currently there's no visual refresh like you have. I guess what I'm getting at is if we should add another sub-script command or expand what we have. The visual portion could be added with a new script command (`unitspecialeffect`) since I think the 2 script commands for special effects only work on NPC and players.
> 
> Just my 2 cents!

@aleos89 the problem is that there are 2 sizes (1 for battle mechanics and 1 "special state" for display reasons).

Coming back to the PR itself:

1. I would like to see this being merged into the existing setunitdata command. 
Check clif.cpp::clif_spawn for type specific attributes and unit types that could be supported.

1. Also to me it is unclear why a periodic client update would be needed. This is only needed for the players that are in range once and then they should already get the updated unit information (as you can see in clif_spawn as well). 
Did you encounter some specific issues that made it necessary to re-send this in a timely manner and could you explaint that maybe (if it is really necessary)?

wel26:
@Lemongrass3110 The EP20 copy will utilize features like video.
image

Lemongrass3110:
> @Lemongrass3110 The EP20 copy will utilize features like video. https://upload3.inven.co.kr/upload/2023/03/29/bbs/i015505117499.gif

@wel26 I am pretty sure that EP20 uses the new SC and not a custom feature developed by us.

Balferian:
@wel26 They use new SC with EFST_CHANGE_SIZE \ EFST_CHANGE_SIZE_MONSTER that change sprite size by percentage.
image

Lemongrass3110:
> > 1. I would like to see this being merged into the existing setunitdata command.
> >    Check clif.cpp::clif_spawn for type specific attributes and unit types that could be supported.
> > 2. Also to me it is unclear why a periodic client update would be needed. This is only needed for the players that are in range once and then they should already get the updated unit information (as you can see in clif_spawn as well).
> >    Did you encounter some specific issues that made it necessary to re-send this in a timely manner and could you explaint that maybe (if it is really necessary)?
> 
> For the 1), I can make the transition. For the 2), I had to put on a timer because, for those who are on vision of the mob and this changes size, it gets like visually glitched. The first transformation it's always good, but, when you try to go from Big -> Small or Small -> Big it's when you get to see the visual glitch (which it's gone after you @refresh, ofc). You can definitely see it on the video I provided, after the first change, the next one looks weird. It would look even weirder if it wasn't by the refresh timer xD

@SapitoSucio For 2 I guess you could use the new packet to remove an effect from a unit, which is supported for clients >= 20181002 via clif_specialeffect_remove. For clients older than that you might need to find another way, like respawn the unit completely when you change the size. I dont really see it on the video, but I think the visual glitch might come from the client thinking the unit has both effects active and you have to work around that somehow.

PR #7781, opened 2023-05-24
Resolves players escaping jail early

### LINKED ISSUE #7771: Player can leave jail after relogin
### rAthena Hash

ccd7e5c14a432469b68f0bd731712887deb1224e

### Client Date

20220406

### Server Mode

Renewal

### Result

Player can leave jail after relogin

### Relevant Log Output

_No response_

### Expected Result

Players must not be able to leave jail by relogin

### How to Reproduce

1. gm type @jailfor 5mn Playername
2. Player relogin
3. After this player will leave jail and return last savepoint

### Official Information

.

_No response_

aleos89:
* **Addressed Issue(s)**: Fixes #7771

  * Adds a missing check to avoid warping players who are actively jailed out early after relogging.
Thanks to @fab10Carvalho!


src/map/pc.cpp
```diff
@@ -1926,6 +1926,11 @@ bool pc_lastpoint_special( map_session_data& sd ){
 		return true;
 	}
 
+	if (strcmpi(sd.status.last_point.map, MAP_JAIL) == 0) {
```
Lemongrass3110, line 1929: Doesnt that make normal "jail visitors" stuck on that map too?

src/map/status.cpp
```diff
@@ -13415,6 +13415,9 @@ int status_change_end(struct block_list* bl, enum sc_type type, int tid)
 			}
 			break;
 		case SC_JAILED:
+			if (tid == INVALID_TIMER)
```
Lemongrass3110, line 13418: This will break `@unjail`.

Lemongrass3110 (changes requested):

src/map/pc.cpp
aleos89: It shouldn't. Jail has the nosave mapflag so the player shouldn't have their lastpoint set to the jail map.

vstumpf: Unfortunately, I don't think there is another way

vstumpf (approved):

Lemongrass3110 (approved):

PR #7797, opened 2023-06-02
Fixes a party check for Weekend Dungeon

### LINKED ISSUE #7776: Weekend Dungeon bug
### rAthena Hash

latest

### Client Date

2021-11-03

### Server Mode

Renewal

### Result

The script is checking if everyone is a party leader, therefore, it doesn't let anyone enter unless they're a party leader, which shouldn't be the case... You should be able to enter with your party mates.

lines 45 through 49 state as follow:

	if (is_party_leader() == false) {
		mes "[Marry Jae]";
		mes "You need to form a party of one member or more.";
		close;
	}
	
It should check if there's already an instance open for your party and skip this part if there is.	

### Relevant Log Output

_No response_

### Expected Result

Should let party members enter the instance with you, but it doesn't.

### How to Reproduce

Open the instance with a party member and they can try to enter, they won't be allowed to, however.

### Official Information

It lets your party members enter the instance with you.

N/A

aleos89:
* **Addressed Issue(s)**: Fixes #7776

* **Server Mode**: Renewal

  * Rearranged the party checks to account for leader versus member.
Thanks to @grimexile!

### DIFF
--- a/npc/re/instances/WeekendDungeon.txt
+++ b/npc/re/instances/WeekendDungeon.txt
@@ -42,14 +42,32 @@ pay_arche,44,124,5	script	Marry Jae	4_F_JOB_BLACKSMITH,{
 	}
 	.@md_name$ = "Weekend Dungeon";
 
-	if (is_party_leader() == false) {
-		mes "[Marry Jae]";
-		mes "You need to form a party of one member or more.";
-		close;
-	}
 	switch( checkquest(12378,PLAYTIME) ) {
 	case -1:
-		break;
+		if (is_party_leader() == true) {
+			mes "[Marry Jae]";
+			mes "I think you're just about ready.";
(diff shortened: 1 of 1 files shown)


npc/re/instances/WeekendDungeon.txt
```diff
@@ -73,41 +93,27 @@ pay_arche,44,124,5	script	Marry Jae	4_F_JOB_BLACKSMITH,{
...
-				setquest 12377;
-			if (isbegin_quest(12378) == 0)
-				setquest 12378;
-			end;
-		}
+	mes "You need to form a party of one member or more.";
+	close;
+
+L_Enter:
+	switch( instance_enter(.@md_name$) ) {
```
Atemo, line 100: `.@md_name$` is undefined here

Atemo (approved):

PR #7688, opened 2023-04-07
Follow up to 65200eb

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Pre-renewal and Renewal

* **Description of Pull Request**: 
  * Introduces the s_mob_db constructor.
  * Fixes clones damagetaken flag getting set too late.
Thanks to @Lemongrass3110 and @mplaten!

PR #7799, opened 2023-06-02
Implement s_mob_db constructor

aleos89:
* **Addressed Issue(s)**: N/A

  * Follow up to 65200eb.
  * Removes a lot of else cases when parsing monster data.
  * Fixes an issue where the damagetaken value of monsters would be reset after issuing a mob_spawn().
Thanks to @Lemongrass3110 and @mplaten!

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -4215,7 +4215,6 @@ int mob_clone_spawn(map_session_data *sd, int16 m, int16 x, int16 y, const char 
 	if (!md) return 0; //Failed?
 
 	md->special_state.clone = 1;
-	md->damagetaken = 100; // Avoid Green Aura reduction calculation.
 
 	if (master_id || flag || duration) { //Further manipulate crafted char.
 		if (flag&1) //Friendly Character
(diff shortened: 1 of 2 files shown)


vstumpf:
Isn't this a duplicate of #7688?

aleos89:
Well, color me shocked and completely oblivious to what I've previously worked on!

I had this stashed and was wondering why I hadn't made a PR yet!

vstumpf:
I'll copy my comments from the other one:

Instead of putting it in a constructor, can we just put default values in the struct?
Something like:
```
struct s_mob_db {
	uint32 id{};
	std::string sprite{}, name{}, jname{};
	t_exp base_exp{};
	t_exp job_exp{};
	t_exp mexp{};
	uint16 range2, range3;
	std::vector<e_race2> race2{};	// celest
	uint16 lv;
	s_mob_drop dropitem[MAX_MOB_DROP_TOTAL], mvpitem[MAX_MVP_DROP_TOTAL];
	status_data status;
	view_data vd;
	uint32 option;
	std::vector<std::shared_ptr<s_mob_skill>> skill{};
	uint16 damagetaken{100};

	e_mob_bosstype get_bosstype();
};
```
The advantage to this is if we ever add a second constructor, we won't have to repeat the default constructor arguments everywhere.

src/map/mob.cpp
```diff
@@ -4361,6 +4360,28 @@ bool MobDatabase::parseDropNode(std::string nodeName, const ryml::NodeRef& node,
...
+	status.agi = 1;
+	status.vit = 1;
+	status.int_ = 1;
+	status.dex = 1;
+	status.luk = 1;
+	status.ele_lv = 1;
+	status.speed = DEFAULT_WALK_SPEED;
+	status.adelay = cap_value(0, battle_config.monster_max_aspd * 2, 4000);
+	status.amotion = cap_value(0, battle_config.monster_max_aspd, 2000);
+	status.mode = static_cast<e_mode>(MONSTER_TYPE_06);
```
vstumpf, line 4380: if only we had [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization), need c++20 though.

vstumpf (approved):

PR #7800, opened 2023-06-02
Fix Illusion of Underwater

mazvi:
Temp fix Illusion of Underwater because prt_elib is not implemented yet

* **Addressed Issue(s)**: N/A

* **Server Mode**: RENEWAL

### DIFF
--- a/npc/re/quests/quests_16_2_illusion.txt
+++ b/npc/re/quests/quests_16_2_illusion.txt
@@ -13,6 +13,7 @@
 //= 1.4 Added Illusion of Turtle [Capuche]
 //= 1.5 Added Illusion of Luanda [Capuche]
 //= 1.6 Added Illusion of Underwater [Capuche]
+//= 1.6.1 Temp fix Illusion of Underwater because prt_elib is not implemented yet [mazvi]
 //============================================================
 
 //============================================================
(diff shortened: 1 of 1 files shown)


Atemo:
> prt_elib is not implemented yet

If you mean the warps, add them
```
prt_elib,14,46,0	warp2	Libraryo	10,10,prontera,192,264
prontera,192,267,0	warp2	Libraryi	2,2,prt_elib,18,46
```

mazvi:
> > prt_elib is not implemented yet
> 
> If you mean the warps, add them
> 
> ```
> prt_elib,14,46,0	warp2	Libraryo	1,2,prontera,192,264
> prontera,192,267,0	warp2	Libraryi	2,2,prt_elib,18,46
> ```

The rathena default packet now is 20211103 this is not support prt_lib

Atemo (changes requested):
We can leave the old coordinates commented if you want.
Although the default packet is 2021, the server can support newer clients. As such there is no point to change the map in the script.

mazvi:
Okay :)

mazvi:
> We can leave the old coordinates commented if you want. Although the default packet is 2021, the server can support newer clients. As such there is no point to change the map in the script.

After I thought about it, the things I made earlier were good because all NPCs like books and library girls were still in prt_in not prt_elib.

OptimusM:
To be honest, i think the best for now it is leave it on the old prontera library, since new prt lib has alot more npc's, also a bunch of instances related to Fantasy VR books, so, don't think rA will develop that on some near future, because yeah, those are mostly "lore content" and not really important stuff.

Atemo:
The goal is to be as close as possible to the original content. The project is configurable enough to allow you to modify the npcs in your server to your liking.

PR #7828, opened 2023-06-20
Added new SC with EFST_CHANGE_SIZE \ EFST_CHANGE_SIZE_MONSTER

SapitoSucio:
* **Addressed Issue(s)**: 
None

* **Server Mode**: 
Both

 Added new SC with EFST_CHANGE_SIZE \ EFST_CHANGE_SIZE_MONSTER that change sprite size by percentage

You can use this NPC to test it

```cpp
prontera,156,183,4    script    NPC_Testing    10164,{
    for(.@i = 0; .@i < 20; .@i++) {
        monster "prontera",156,180,"Poring",1002,1;
        setbodysize($@mobid[0], rand(10, 300));
    }
}
```

image

Credits goes to @Balferian 🙏

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -19593,16 +19593,16 @@ BUILDIN_FUNC(setunitname)
 
 BUILDIN_FUNC(setbodysize)
 {
-	struct block_list* bl = NULL;
-	struct view_data* vd;
-	TBL_MOB* md = NULL;
-	TBL_NPC* nd = NULL;
+	block_list* bl;
 
 	if(!script_rid2bl(2,bl))
 	{
 		script_pushconststr(st, "Unknown");
 		return SCRIPT_CMD_FAILURE;
 	}
+
+	TBL_MOB* md;
+	TBL_NPC* nd;
 
 	switch (bl->type) {
 		case BL_MOB:  md = map_id2md(bl->id); break;


src/map/clif.cpp
```diff
@@ -6584,6 +6591,16 @@ void clif_status_change(struct block_list *bl, int type, int flag, t_tick tick,
...
 }
 
+void clif_body_size(struct block_list *bl, int val1) {
+	map_session_data *sd = NULL;
+
+	nullpo_retv(bl);
+
+	sd = BL_CAST(BL_PC, bl);
+
+	clif_status_change_sub(bl, bl->id, 1421, 1, 9999, val1, 0, 0, ((sd ? (pc_isinvisible(sd) ? SELF : AREA) : AREA_WOS)));
```
Balferian, line 6601: Add check to use right EFST
`clif_status_change_sub(bl, bl->id, (bl->type == BL_MOB) ? EFST_CHANGE_SIZE_MONSTER : EFST_CHANGE_SIZE, 1, 9999, val1, 0, 0, ((sd ? (pc_isinvisible(sd) ? SELF : AREA) : AREA_WOS)));`

doc/script_commands.txt
```diff
@@ -8950,6 +8950,21 @@ Valid types are:
    0: Elemental ID
    1: Elemental Game ID
 
+
+*setbodysize <GID>, <size value>;
```
Atemo, line 8954: it may be a matter of preference but the parameters are similar. How about adding a new constant for setunitdata?
Example: `setunitdata $@mobid[0], UMOB_BODYSIZE, rand(10, 300);`

new constant : `UMOB_BODYSIZE`

src/map/script.cpp
```diff
@@ -19591,6 +19591,53 @@ BUILDIN_FUNC(setunitname)
...
+	struct view_data* vd;
+	TBL_MOB* md = NULL;
+	TBL_NPC* nd = NULL;
+
+	if(!script_rid2bl(2,bl))
+	{
+		script_pushconststr(st, "Unknown");
+		return SCRIPT_CMD_FAILURE;
+	}
+
```
Atemo, line 19606: maybe add a range check / warning?

`Size value ranges from 1 to 400`

doc/script_commands.txt
```diff
@@ -8950,6 +8950,21 @@ Valid types are:
...
+
+This command allows you to change the size of a mob/NPC.
+Size value ranges from 1 to 400
+
+Example
+
+	prontera,156,183,4	script	SetBodySizeNPC	10164,{
```
aleos89, line 8961:
```suggestion
	// Spawn 20 Poring at random sizes ranging from 10~300.
	prontera,156,183,4	script	SetBodySizeNPC	10164,{
```

src/map/clif.cpp
```diff
@@ -1686,6 +1686,7 @@ static inline bool clif_npc_mayapurple(block_list *bl) {
  **/
 int clif_spawn( struct block_list *bl, bool walking ){
 	struct view_data *vd;
+	struct unit_data* ud = unit_bl2ud(bl);
```
aleos89, line 1689:
```suggestion
```

src/map/clif.cpp
```diff
@@ -1701,6 +1702,8 @@ int clif_spawn( struct block_list *bl, bool walking ){
 		clif_spawn_unit( bl, AREA_WOS );
 	}
```
aleos89, line 1703:
```suggestion
	}

	unit_data* ud = unit_bl2ud(bl);
```
Just moved this down since there's the possibility above for checks to fail.

src/map/clif.cpp
```diff
@@ -6584,6 +6591,16 @@ void clif_status_change(struct block_list *bl, int type, int flag, t_tick tick,
 	clif_status_change_sub(bl, bl->id, type, flag, tick, val1, val2, val3, ((sd ? (pc_isinvisible(sd) ? SELF : AREA) : AREA_WOS)));
 }
 
+void clif_body_size(struct block_list *bl, int val1) {
+	map_session_data *sd = NULL;
+
+	nullpo_retv(bl);
+
+	sd = BL_CAST(BL_PC, bl);
```
aleos89, line 6599:
```suggestion
	nullpo_retv(bl);

	map_session_data *sd = BL_CAST(BL_PC, bl);
```

src/map/script.cpp
```diff
@@ -19591,6 +19591,53 @@ BUILDIN_FUNC(setunitname)
...
+	struct block_list* bl = NULL;
+	struct view_data* vd;
+	TBL_MOB* md = NULL;
+	TBL_NPC* nd = NULL;
+
+	if(!script_rid2bl(2,bl))
+	{
+		script_pushconststr(st, "Unknown");
+		return SCRIPT_CMD_FAILURE;
+	}
```
aleos89, line 19605:
```suggestion
	block_list* bl;

	if(!script_rid2bl(2,bl))
	{
		script_pushconststr(st, "Unknown");
		return SCRIPT_CMD_FAILURE;
	}

	TBL_MOB* md;
	TBL_NPC* nd;
```

aleos89: Do we want to return "Unknown"? We don't return anything on success, so either we return 0/1 on fail/success or nothing at all. Should also document it in the script command doc if we decide on either.

src/map/unit.hpp
```diff
@@ -46,6 +46,7 @@ struct unit_data {
 	uint8 dir;
 	unsigned char walk_count;
 	unsigned char target_count;
+	uint8 body_size;
```
vstumpf, line 49: if the range is 1~400, a uint8 is not big enough to fit this, you would need uint16.

Balferian:
There missed packet version date in docs / script command.
I don't know the exact date, but it was first used in an event Dimensional Rift
https://www.inven.co.kr/board/ro/1952/314647?p=22
Should be like 2022-06-15.

Lemongrass3110:
In my opinion this should be a normal SC, just as the title of the PR suggests.

laziem:
@SapitoSucio please update to latest rAthena

PR #7829, opened 2023-06-22
Fix multi-hit skills activation priority

eppc0330:
* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

 If a single character is learning more than one multi-hit skills, all skills' success chance should be applied.
Also ```RA_FEARBREEZE``` has the highest priority among these skills.

Reference : https://www.inven.co.kr/board/ro/1941/13678


vstumpf (approved):
I see, so if one of the checks pass, we change wd->div_ to > 1, and every check has a check if it's == 1.
So only one can possibly be activated, but all are checked.

Issue #7826, opened 2023-06-18
RODEX mail map crash


AYauder:
### rAthena Hash

03ab0b68e

### Client Date

2021-11-03_Ragexe

### Server Mode

Pre-Renewal

### Result

Map Crash related to RODEX Mail

No modification
![Screenshot_1102](https://github.com/rathena/rathena/assets/119885282/5c82c78f-2737-452b-9b08-13f6c0ac4c30)

![Screenshot_1104](https://github.com/rathena/rathena/assets/119885282/c859a8d6-5256-439a-b440-825db4188969)

### Expected Result

I am not sure but it is seems to related with check weight on rodex attachment

### How to Reproduce

I am not sure but it is seems to related with check weight on rodex attachment

### Official Information

I am not sure but it is seems to related with check weight on rodex attachment

### Modifications that may affect results

no modification

dev1-juan:
(if you have AndRO Client)

It is confirmed that it is related to AndRO Clients causing these. You need to report it to the developer of AndRO for fixing and video of how to reproduce it on your server.

aleos89:
Fixed in [458a9d8](https://github.com/rathena/rathena/commit/458a9d8950c65ab279c8658c36e751f1a47e2fde).

PR #7833, opened 2023-06-30
Fixed potential crash in RODex

secretdataz:
* **Addressed Issue(s)**: #7826 

* **Server Mode**: Both


vstumpf (approved):
Looks good

But how does this even happen? We might need to rethink how mail attachments are handled.

aleos89 (approved):

PR #7850, opened 2023-07-20
Fix dynamic NPC behavior

eppc0330:
* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

Now dynamic NPC will be spawn, not set idle. This is official behavior.

Before
dynamicbefore

After
dynamicafter

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -1695,7 +1695,8 @@ int clif_spawn( struct block_list *bl, bool walking ){
 	if (clif_npc_mayapurple(bl))
 		return 0;
 
-	if( bl->type == BL_NPC && !vd->dead_sit ){
+	TBL_NPC* nd = ((TBL_NPC*)bl);
+	if( bl->type == BL_NPC && !vd->dead_sit && nd->dynamicnpc.owner_char_id == 0 ){
 		clif_set_unit_idle( bl, walking, AREA_WOS, bl );
 	}else{
 		clif_spawn_unit( bl, AREA_WOS );


src/map/clif.cpp
```diff
@@ -1695,7 +1695,8 @@ int clif_spawn( struct block_list *bl, bool walking ){
 	if (clif_npc_mayapurple(bl))
 		return 0;
 
-	if( bl->type == BL_NPC && !vd->dead_sit ){
+	TBL_NPC* nd = ((TBL_NPC*)bl);
```
Lemongrass3110, line 1698:
```suggestion
	npc_data* nd = BL_CAST( BL_NPC, bl );
```

src/map/clif.cpp
```diff
@@ -1695,7 +1695,8 @@ int clif_spawn( struct block_list *bl, bool walking ){
...
+	if( bl->type == BL_NPC && !vd->dead_sit && nd->dynamicnpc.owner_char_id == 0 ){
```
Lemongrass3110, line 1699:
```suggestion
	if( nd != nullptr && !vd->dead_sit && nd->dynamicnpc.owner_char_id == 0 ){
```

src/map/npc.cpp
```diff
@@ -1123,12 +1123,6 @@ TIMER_FUNC(npc_secure_timeout_timer){
 
 	if( DIFF_TICK(cur_tick,sd->npc_idle_tick) > (timeout*1000) ) {
 		pc_close_npc(sd,1);
-	} else if(sd->st && (sd->st->state == END || sd->st->state == CLOSE)){
```
Lemongrass3110, line 1126: Removing this part looks wrong.

src/map/pc.cpp
```diff
@@ -9534,6 +9534,13 @@ TIMER_FUNC(pc_close_npc_timer){
 void pc_close_npc(map_session_data *sd,int flag)
 {
 	nullpo_retv(sd);
+	
+	for (const auto &it : sd->npc_id_dynamic) {
```
Lemongrass3110, line 9538: This only closes a single NPC, why do we update all?

src/map/npc.cpp
eppc0330: Any other roles in this part? Because in KRo secure timer doesn't being stopped on ```close``` or ```end```

src/map/pc.cpp
eppc0330: Actually I'm not sure about this part. I wanted to initialize dynamic timer when NPC secure time expires. Please suggest if there is a better way.

aleos89: I would assume if you're using the NPC Secure Timer then that should only apply to the NPC the player is talking to from the `npc_id_dynamic` list.

eppc0330: @aleos89 Thanks!

PR #7861, opened 2023-07-28
instance_npcname to return existing duplicate

Singe-Horizontal:
* **Addressed Issue(s)**: 
N/A

* **Server Mode**: 
Both

The command `instance_npcname` can easily be misused on an npc which is already duplicated in an instance, for example by calling it in such way `instance_npcname(strnpcinfo(3))`.
The proposed modification adds a warning and returns existing name, since it is the most expected behavior.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -21473,8 +21473,13 @@ BUILDIN_FUNC(instance_npcname)
 
 	if( instance_id > 0 && (nd = npc_name2id(str)) != NULL ) {
 		static char npcname[NAME_LENGTH];
-		snprintf(npcname, sizeof(npcname), "dup_%d_%d", instance_id, nd->bl.id);
-		script_pushconststr(st,npcname);
+		if(nd->instance_id > 0){
+			strncpy_s(npcname,str,NAME_LENGTH);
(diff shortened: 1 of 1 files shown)


src/map/script.cpp
```diff
@@ -21473,8 +21473,13 @@ BUILDIN_FUNC(instance_npcname)
 
 	if( instance_id > 0 && (nd = npc_name2id(str)) != NULL ) {
 		static char npcname[NAME_LENGTH];
-		snprintf(npcname, sizeof(npcname), "dup_%d_%d", instance_id, nd->bl.id);
-		script_pushconststr(st,npcname);
+		if(nd->instance_id > 0){
+			strncpy(npcname,str,NAME_LENGTH);
+			ShowWarning("buildin_instance_npcname: npc is already in instance and its name has a duplicate format, returning input name (instance_id: %d, NPC name: \"%s\".)\n", instance_id, str);
```
Lemongrass3110, line 21478: I would rather be for SCRIPT_CMD_FAILURE and telling users to learn how to use the script command.
But this script command could be improved in general x_x

Atemo (approved):

PR #7873, opened 2023-07-31
Small changes in spelling and removed duplicated warp portal.

elanosilveira:
* **Addressed Issue(s)**: 
The first course of the test had two warp portals very close to each other warping to the same place for no reason. Probably from the very first version of this NPC. The correct warp sends a message to the current map, the one removed only warps the player to the next room but it is not reachable.

* **Server Mode**: 

### DIFF
--- a/npc/pre-re/jobs/1-1/swordman.txt
+++ b/npc/pre-re/jobs/1-1/swordman.txt
@@ -127,7 +127,7 @@ izlude_in,74,172,4	script	Swordman#swd_1	119,{
 				}
 				else {
 					mes "[Swordman]";
-					mes "Alright then. Feel free to come back whenever you are ready. All you ahve to do now is meet our requirements. Good luck to you.";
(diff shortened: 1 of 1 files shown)


npc/pre-re/jobs/1-1/swordman.txt
```diff
@@ -460,7 +460,7 @@ job_sword1,180,238,0	duplicate(SwdTrap1)	1_blank_14_a	-1,0,1
 job_sword1,183,238,0	duplicate(SwdTrap1)	1_blank_14_b	-1,0,1
 job_sword1,181,239,0	duplicate(SwdTrap1)	1_blank_14_c	-1,1,0
 
-job_sword1,193,245,0	warp	1_rest	1,3,job_sword1,215,244
+//job_sword1,193,245,0	warp	1_rest	1,3,job_sword1,215,244
```
Lemongrass3110, line 463: Just for anyone else looking:
The other one is located at job_sword1,192,244

Issue #7875, opened 2023-08-02
addrid forceflag has an undefined default behavior


Singe-Horizontal:
@### rAthena Hash

4ce3c460e

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

addrid will sometimes fail to apply default force behavior

### Expected Result

When using addrid(0), it should always imply addrid(0,0);

### How to Reproduce
Engage dialog  with npc with both characters.
Sample script:
`prontera,156,170,1 script  test    876,{`
`   while(true){`
`       select("loop");`
`       .@test=1;`
`       addrid(0);` 
`       if(!.@test){`
`           debugmes "added";` //<------- this won't display with `addrid(0)`, only with `addrid(0,0)`.
`           end;                    `
`       }`
`   }`
`   end;`
`}`
``

### Official Information

N/A

PR #7876, opened 2023-08-02
Call script_hasdata before script_getnum in addrid

Singe-Horizontal:
* **Addressed Issue(s)**: 
 #7875

* **Server Mode**: 
Both

Removes an undefined behavior that happens when calling script_getnum before checking if data is indeed present.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -13212,17 +13212,20 @@ BUILDIN_FUNC(addrid)
 		bl = map_id2bl(st->rid); //if run without rid it'd error,also oid if npc, else rid for map
 	iter = mapit_getallusers();
 
-	switch(script_getnum(st,2)) {
+	int type = script_getnum(st,2);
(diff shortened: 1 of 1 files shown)


Atemo (approved):

src/map/script.cpp
```diff
@@ -13212,17 +13212,20 @@ BUILDIN_FUNC(addrid)
 		bl = map_id2bl(st->rid); //if run without rid it'd error,also oid if npc, else rid for map
 	iter = mapit_getallusers();
 
-	switch(script_getnum(st,2)) {
+	int type = script_getnum(st,2);
+	int forceflag = (script_hasdata(st,3) ? script_getnum(st,3) : 0);
```
aleos89, line 13216:
```suggestion
	bool forceflag = (script_hasdata(st,3) ? script_getnum(st,3) != 0 : 0);
```

aleos89 (approved):

PR #7880, opened 2023-08-03
Refactor out a lot of hard coded hit rates

aleos89:
* **Addressed Issue(s)**: N/A

  * Moves skills that adjust hit rate into the skill database.

### DIFF
--- a/doc/skill_db.txt
+++ b/doc/skill_db.txt
@@ -3,7 +3,7 @@
 //===== By: ==================================================
 //= rAthena Dev Team
 //===== Last Updated: ========================================
-//= 20220126
+//= 20230803
 //===== Description: =========================================
 //= Explanation of the skill_db.yml file and structure.
 //============================================================
@@ -154,6 +154,27 @@ Sequence Map Form
       Count: 8
     - Level: 5
       Count: 10
+
+---------------------------------------
+
+HitRate: Skill hit rate. Rate is 100 by default.
+
+Can be defined in scalar form or sequence map form:
+Scalar Form
+  HitRate: 120
+
+Sequence Map Form
+  HitRate:
+    - Level: 1
+      Count: 120
+    - Level: 2
+      Count: 140
+    - Level: 3
+      Count: 160
+    - Level: 4
+      Count: 180
+    - Level: 5
+      Count: 200
 
 ---------------------------------------
 
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -34,6 +34,8 @@
 #include "pc_groups.hpp"
 #include "pet.hpp"
 
+using namespace rathena;
+
 struct Battle_Config battle_config;
 static struct eri *delay_damage_ers; //For battle delay damage structures.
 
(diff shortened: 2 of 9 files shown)


doc/skill_db.txt
```diff
@@ -157,6 +157,31 @@ Sequence Map Form
...
+      Count: 120
+    - Level: 2
+      Count: 140
+    - Level: 3
+      Count: 160
+    - Level: 4
+      Count: 180
+    - Level: 5
+      Count: 200
+    Skills: # Requires both NV_BASIC and SM_BASH to be learned.
```
Atemo, line 178: Maybe this description should be more explicit about the consequences when one of the skills is not learned (no hitrate bonus)

src/map/battle.cpp
```diff
@@ -3097,98 +3099,49 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
...
+
+				if (dist > 3) {
+					// Reduce n hitrate for each cell after initial 3 cells.
+					hitrate -= (skill_get_hitrate(RL_SLUGSHOT, skill_lv) * (dist - 3));
+				}
+			}
+			break;
+			default:
+			{
+				struct s_skill_hitrate skill_hitrate = skill_db.find(skill_id)->hitrate;
```
Lemongrass3110, line 3127:
```suggestion
				s_skill_hitrate& skill_hitrate = skill_db.find(skill_id)->hitrate;
```

src/map/battle.cpp
```diff
@@ -3097,98 +3099,49 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
...
+
+				if (skill_hitrate.rate[skill_lv] != 100) { // Hit skill modifiers
+					bool req_fail = false;
+
+					// List of skills required to be learned before adjusting hitrate (players only).
+					for (const auto &skill_req : skill_hitrate.skills) {
+						if (sd && pc_checkskill(sd, skill_req.first) < skill_req.second) {
```
Lemongrass3110, line 3134: sd check should be outside of the loop, because the whole loop is only for players

src/map/battle.cpp
```diff
@@ -3097,98 +3099,49 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
...
+							req_fail = true;
+							break;
+						}
 					}
+
+					if (!req_fail)
```
Lemongrass3110, line 3140:
```suggestion
					// If there were no requirements or the requirements were fulfilled successfully, apply the changed hitrate
					if (!req_fail)
```
Just to make it easier to read. Somehow to me if you dont see the ! (not) it looks like it failed the requirements and still gets the change.

src/map/battle.cpp
```diff
@@ -3097,98 +3099,49 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
...
-						// Reduce n hitrate for each cell after initial 3 cells. Different each level
-						// -10:-9:-8:-7:-6
-						dist -= 3;
-						hitrate -= ((11 - skill_lv) * dist);
+			{
+				uint32 dist = distance_bl(src, target);
+
+				if (dist > 3) {
+					// Reduce n hitrate for each cell after initial 3 cells.
+					hitrate -= (skill_get_hitrate(RL_SLUGSHOT, skill_lv) * (dist - 3));
```
Lemongrass3110, line 3121: Rate is an int32, so you could put the "minus" of the formula into the yaml file?

Lemongrass3110 (approved):
Did not compare all database values and did not do any ingame tests, but looks good with some small suggestions for further improvement.

src/map/battle.cpp
aleos89: I forgot to cast the value, but the reason it's int32 is because of the `parseNode()` function for loading the skill database into memory defaults the values to int32. The `parseNode()` template could use another refactor to utilize a typedef so we don't have to define all of the skill database values as int32.

Issue #7881, opened 2023-08-03
Random generator has module bias and has an slow implementation


Leemonn:
### rAthena Hash

40b2d0ff13fe7df8d32d20c9ce0ae6887e6e21e9

### Client Date

-

### Server Mode

Pre-Renewal, Renewal

### Result

This ticket has been created so that I can link it to my PR.

Current `rnd()` implementation generates a number between 0 and `SINT32_MAX`. Then to check for chance a module is applied (for example: `rnd() % 100`) this introduces module bias.

At the same time this implementation is way slower than it should because it uses signed integers and involves calculating the module everytime.

### Expected Result

.

### How to Reproduce

.

### Official Information

.

### Modifications that may affect results

.

PR #7882, opened 2023-08-03
Improved randomness (part 1)

Leemonn:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/7881

* **Server Mode**:  Both

## Description of Pull Request

Current rnd() implementation generates a number between `0` and `SINT32_MAX`. Then a module is applied to check for chance (for example: `rnd() % 100`) this introduces module bias.

At the same time this implementation is way slower than it should because it uses signed integers and involves calculating the module everytime.

In this PR I'm changing the way these random values are calculated to remove the module bias, make code simpler and also make it way faster.

## Comparison
Here is a comparison using [https://quick-bench.com/q/cL2SnlW8Zj7YuyUz-VQxbppGNck](https://quick-bench.com/q/cL2SnlW8Zj7YuyUz-VQxbppGNck)
image

Tested in Linux with no optimizations:
image

Tested in linux with -O2:
image

Tested in linux with -O3:
image

## Important reminder

```c++
rnd()%n == new_random(n)

rnd()%n+x == (rnd()%n) + x
```
## Why is it called `new_random`? Why didn't you replace all `rnd()`?
I decided to split this into 2 or 3 PRs, otherwise we would have 1 big PR with more than 40 files and 500 lines changed.

The name **new_random** is temporary, it allows us to make the transition easier keeping the old implementation of `rnd()` on files that have not been updated yet.

We can remove `rnd()` and improve the other functions on the last PR.

### DIFF
--- a/src/login/loginclif.cpp
+++ b/src/login/loginclif.cpp
@@ -377,8 +377,7 @@ static int logclif_parse_reqauth(int fd, struct login_session_data *sd, int comm
 static int logclif_parse_reqkey(int fd, struct login_session_data *sd){
 	RFIFOSKIP(fd,2);
 	{
-		memset(sd->md5key, '\0', sizeof(sd->md5key));
-		sd->md5keylen = rnd_value(12u, 15u);
+		sd->md5keylen = sizeof( sd->md5key );
 		MD5_Salt(sd->md5keylen, sd->md5key);
 
 		WFIFOHEAD(fd,4 + sd->md5keylen);
--- a/src/login/loginclif.cpp
+++ b/src/login/loginclif.cpp
@@ -378,7 +378,7 @@ static int logclif_parse_reqkey(int fd, struct login_session_data *sd){
 	RFIFOSKIP(fd,2);
 	{
 		memset(sd->md5key, '\0', sizeof(sd->md5key));
-		sd->md5keylen = (uint16)(12 + rnd() % 4);
+		sd->md5keylen = (uint16)(12 + new_random(4));
 		MD5_Salt(sd->md5keylen, sd->md5key);
 
 		WFIFOHEAD(fd,4 + sd->md5keylen);
--- a/src/login/loginclif.cpp
+++ b/src/login/loginclif.cpp
@@ -378,7 +378,7 @@ static int logclif_parse_reqkey(int fd, struct login_session_data *sd){
 	RFIFOSKIP(fd,2);
 	{
 		memset(sd->md5key, '\0', sizeof(sd->md5key));
-		sd->md5keylen = (uint16)(12 + new_random(4));
+		sd->md5keylen = (uint16)(12 + rnd(4));
 		MD5_Salt(sd->md5keylen, sd->md5key);
 
 		WFIFOHEAD(fd,4 + sd->md5keylen);
--- a/src/login/loginclif.cpp
+++ b/src/login/loginclif.cpp
@@ -378,7 +378,7 @@ static int logclif_parse_reqkey(int fd, struct login_session_data *sd){
 	RFIFOSKIP(fd,2);
 	{
 		memset(sd->md5key, '\0', sizeof(sd->md5key));
-		sd->md5keylen = (uint16)(12 + rnd(4));
+		sd->md5keylen = rnd_value(12u, 15u);
 		MD5_Salt(sd->md5keylen, sd->md5key);
 
 		WFIFOHEAD(fd,4 + sd->md5keylen);
--- a/src/map/elemental.cpp
+++ b/src/map/elemental.cpp
@@ -622,7 +622,7 @@ static int elemental_ai_sub_timer(s_elemental_data *ed, map_session_data *sd, t_
 			return 1;
 		}
 
-		if( battle_check_range(&ed->bl,target,view_range) && rnd()%100 < 2 ) { // 2% chance to cast attack skill.
+		if( battle_check_range(&ed->bl,target,view_range) && new_random(100) < 2 ) { // 2% chance to cast attack skill.
 			if(	elemental_action(ed,target,tick) )
 				return 1;
 		}
--- a/src/map/elemental.cpp
+++ b/src/map/elemental.cpp
@@ -622,7 +622,7 @@ static int elemental_ai_sub_timer(s_elemental_data *ed, map_session_data *sd, t_
 			return 1;
 		}
 
-		if( battle_check_range(&ed->bl,target,view_range) && new_random(100) < 2 ) { // 2% chance to cast attack skill.
+		if( battle_check_range(&ed->bl,target,view_range) && rnd(100) < 2 ) { // 2% chance to cast attack skill.
 			if(	elemental_action(ed,target,tick) )
 				return 1;
 		}
--- a/src/map/elemental.cpp
+++ b/src/map/elemental.cpp
@@ -622,7 +622,7 @@ static int elemental_ai_sub_timer(s_elemental_data *ed, map_session_data *sd, t_
 			return 1;
 		}
 
-		if( battle_check_range(&ed->bl,target,view_range) && rnd(100) < 2 ) { // 2% chance to cast attack skill.
+		if( battle_check_range(&ed->bl,target,view_range) && rnd_chance(2, 100) ) { // 2% chance to cast attack skill.
 			if(	elemental_action(ed,target,tick) )
 				return 1;
 		}
--- a/src/map/itemdb.cpp
+++ b/src/map/itemdb.cpp
@@ -2872,7 +2872,7 @@ std::shared_ptr<s_item_group_entry> get_random_itemsubgroup(std::shared_ptr<s_it
 	for (size_t j = 0, max = random->data.size() * 3; j < max; j++) {
 		std::shared_ptr<s_item_group_entry> entry = util::umap_random(random->data);
 
-		if (entry->rate == 0 || rnd(random->total_rate) < entry->rate)	// always return entry for rate 0 ('must' item)
+		if (entry->rate == 0 || rnd_chance<uint32>(entry->rate, random->total_rate))	// always return entry for rate 0 ('must' item)
 			return entry;
 	}
 
@@ -2956,9 +2956,9 @@ static void itemdb_pc_get_itemgroup_sub(map_session_data *sd, bool identify, std
 			if( data->refineMinimum > 0 && data->refineMaximum > 0 ){
 				tmp.refine = rnd_value( data->refineMinimum, data->refineMaximum );
 			}else if( data->refineMinimum > 0 ){
-				tmp.refine = rnd_value( data->refineMinimum, MAX_REFINE );
+				tmp.refine = rnd_value<uint16>( data->refineMinimum, MAX_REFINE );
 			}else if( data->refineMaximum > 0 ){
-				tmp.refine = rnd_value( 1, data->refineMaximum );
+				tmp.refine = rnd_value<uint16>( 1, data->refineMaximum );
 			}else{
 				tmp.refine = 0;
 			}
@@ -4441,7 +4441,7 @@ void s_random_opt_group::apply( struct item& item ){
 		for( size_t j = 0, max = this->slots[static_cast<uint16>(i)].size() * 3; j < max; j++ ){
 			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->slots[static_cast<uint16>(i)] );
 
-			if( rnd(10000) < option->chance ){
+			if ( rnd_chance<uint16>(option->chance, 10000) ) {
 				apply_sub( item.option[i], option );
 				break;
 			}
@@ -4466,7 +4466,7 @@ void s_random_opt_group::apply( struct item& item ){
 
 			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->random_options );
 
-			if( rnd(10000) < option->chance ){
+			if ( rnd_chance<uint16>(option->chance, 10000) ){
 				apply_sub( item.option[i], option );
 			}
 		}
--- a/src/map/itemdb.cpp
+++ b/src/map/itemdb.cpp
@@ -2872,7 +2872,7 @@ std::shared_ptr<s_item_group_entry> get_random_itemsubgroup(std::shared_ptr<s_it
 	for (size_t j = 0, max = random->data.size() * 3; j < max; j++) {
 		std::shared_ptr<s_item_group_entry> entry = util::umap_random(random->data);
 
-		if (entry->rate == 0 || rnd() % random->total_rate < entry->rate)	// always return entry for rate 0 ('must' item)
+		if (entry->rate == 0 || new_random(random->total_rate) < entry->rate)	// always return entry for rate 0 ('must' item)
 			return entry;
 	}
 
@@ -4441,7 +4441,7 @@ void s_random_opt_group::apply( struct item& item ){
 		for( size_t j = 0, max = this->slots[static_cast<uint16>(i)].size() * 3; j < max; j++ ){
 			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->slots[static_cast<uint16>(i)] );
 
-			if( rnd() % 10000 < option->chance ){
+			if( new_random(10000) < option->chance ){
 				apply_sub( item.option[i], option );
 				break;
 			}
@@ -4466,7 +4466,7 @@ void s_random_opt_group::apply( struct item& item ){
 
 			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->random_options );
 
-			if( rnd() % 10000 < option->chance ){
+			if( new_random(10000) < option->chance ){
 				apply_sub( item.option[i], option );
 			}
 		}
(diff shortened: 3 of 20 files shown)


vstumpf:
Can we use the same signature as the current `rnd()` and just overload it?
```
uint32 rnd(uint32 low, uint32 high) {
   // actual impl
}
uint32 rnd(uint32 high) {
   rnd(0, high);
}
uint32 rnd(void) {
   rnd (0, INT32_MAX);
}
```

this way we don't have to have `rnd_new()`, we can just change `rnd() % X` to `rnd(X)`

Leemonn:
> Can we use the same signature as the current `rnd()` and just overload it?
> 
> ```
> uint32 rnd(uint32 low, uint32 high) {
>    // actual impl
> }
> uint32 rnd(uint32 high) {
>    rnd(0, high);
> }
> uint32 rnd(void) {
>    rnd (0, INT32_MAX);
> }
> ```
> 
> this way we don't have to have `rnd_new()`, we can just change `rnd() % X` to `rnd(X)`

I wanted a different name to avoid any kind of confusion during the update process. But I will change it to `rnd`, I'm already making smaller pull requrests to make it easier to review so having the same name shouldn't be a problem.

Lemongrass3110:
I dont really like calling it rnd() again. The possibility of using rand() instead of rnd() was already too high.
Quite a few of the references that you changed should also not be replaced in the way you did, but improved as well, if we already touch them.
Just commenting this now, before you have to edit it over and over again.

Lemongrass3110:
I will comment it, so that you see what I mean.

src/map/elemental.cpp
```diff
@@ -622,7 +622,7 @@ static int elemental_ai_sub_timer(s_elemental_data *ed, map_session_data *sd, t_
 			return 1;
 		}
 
-		if( battle_check_range(&ed->bl,target,view_range) && rnd()%100 < 2 ) { // 2% chance to cast attack skill.
+		if( battle_check_range(&ed->bl,target,view_range) && rnd(100) < 2 ) { // 2% chance to cast attack skill.
```
Lemongrass3110, line 625: Should use rnd_chance

src/map/itemdb.cpp
```diff
@@ -2872,7 +2872,7 @@ std::shared_ptr<s_item_group_entry> get_random_itemsubgroup(std::shared_ptr<s_it
 	for (size_t j = 0, max = random->data.size() * 3; j < max; j++) {
 		std::shared_ptr<s_item_group_entry> entry = util::umap_random(random->data);
 
-		if (entry->rate == 0 || rnd() % random->total_rate < entry->rate)	// always return entry for rate 0 ('must' item)
+		if (entry->rate == 0 || rnd(random->total_rate) < entry->rate)	// always return entry for rate 0 ('must' item)
```
Lemongrass3110, line 2875: Should use rnd_chance

src/map/itemdb.cpp
```diff
@@ -4441,7 +4441,7 @@ void s_random_opt_group::apply( struct item& item ){
 		for( size_t j = 0, max = this->slots[static_cast<uint16>(i)].size() * 3; j < max; j++ ){
 			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->slots[static_cast<uint16>(i)] );
 
-			if( rnd() % 10000 < option->chance ){
+			if( rnd(10000) < option->chance ){
```
Lemongrass3110, line 4444: Should use rnd_chance

src/map/itemdb.cpp
```diff
@@ -4466,7 +4466,7 @@ void s_random_opt_group::apply( struct item& item ){
 
 			std::shared_ptr<s_random_opt_group_entry> option = util::vector_random( this->random_options );
 
-			if( rnd() % 10000 < option->chance ){
+			if( rnd(10000) < option->chance ){
```
Lemongrass3110, line 4469: Should use rnd_chance

src/map/map.cpp
```diff
@@ -1736,8 +1736,8 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
 	}
 
 	while(tries--) {
-		*x = (rx >= 0)?(rnd()%rx2-rx+bx):(rnd()%(mapdata->xs-2)+1);
-		*y = (ry >= 0)?(rnd()%ry2-ry+by):(rnd()%(mapdata->ys-2)+1);
+		*x = (rx >= 0) ? (rnd(rx2) - rx + bx):(rnd(mapdata->xs - 2) + 1);
+		*y = (ry >= 0) ? (rnd(ry2) - ry + by):(rnd(mapdata->ys - 2) + 1);
```
Lemongrass3110, line 1740: Should be a new rnd_value [bx - rx, bx + rx] and a new rnd_value [1, mapdata->xs[
Same for y.

src/map/map.cpp
```diff
@@ -1870,7 +1870,7 @@ int map_addflooritem(struct item *item, int amount, int16 m, int16 x, int16 y, i
 
 	if (!map_searchrandfreecell(m,&x,&y,flags&2?1:0))
 		return 0;
-	r = rnd();
+	r = rnd(UINT32_MAX);
```
Lemongrass3110, line 1873: I am too lazy too figure out the math below.
But it should be changed as well and additionally it should also be moved down.

src/map/map.cpp
```diff
@@ -3089,8 +3089,8 @@ int map_random_dir(struct block_list *bl, int16 *x, int16 *y)
 	if (dist < 1) dist =1;
 
 	do {
-		short j = 1 + 2*(rnd()%4); //Pick a random diagonal direction
-		short segment = 1+(rnd()%dist); //Pick a random interval from the whole vector in that direction
+		short j = 1 + 2*(rnd(4)); //Pick a random diagonal direction
```
Lemongrass3110, line 3092: Should use DIR_* constant.
Also I would prefer to convert that into a static array/vector and pick a random entry from there.

src/map/mercenary.cpp
```diff
@@ -421,7 +421,7 @@ bool mercenary_dead(s_mercenary_data *md) {
 void mercenary_killbonus(s_mercenary_data *md) {
 	std::vector<sc_type> scs = { SC_MERC_FLEEUP, SC_MERC_ATKUP, SC_MERC_HPUP, SC_MERC_SPUP, SC_MERC_HITUP };
 
-	sc_start(&md->bl,&md->bl, util::vector_random(scs), 100, rnd() % 5, 600000);
+	sc_start(&md->bl,&md->bl, util::vector_random(scs), 100, rnd(5), 600000);
```
Lemongrass3110, line 424: Should use scs.size()

src/map/pet.cpp
```diff
@@ -710,7 +710,7 @@ int pet_attackskill(struct pet_data *pd, int target_id)
 	if (DIFF_TICK(pd->ud.canact_tick, gettick()) > 0)
 		return 0;
 
-	if (rnd()%100 < (pd->a_skill->rate +pd->pet.intimate*pd->a_skill->bonusrate/1000)) { // Skotlex: Use pet's skill
+	if (rnd(100) < (pd->a_skill->rate +pd->pet.intimate*pd->a_skill->bonusrate/1000)) { // Skotlex: Use pet's skill
```
Lemongrass3110, line 713: Should use rnd_chance

src/map/pet.cpp
```diff
@@ -782,8 +782,8 @@ int pet_target_check(struct pet_data *pd,struct block_list *bl,int type)
 			rate = 1;
 	}
 
-	if(rnd()%10000 < rate) {
-		if(pd->target_id == 0 || rnd()%10000 < pet_db_ptr->change_target_rate)
+	if(rnd(10000) < rate) {
```
Lemongrass3110, line 785: Should use rnd_chance

src/map/pet.cpp
```diff
@@ -782,8 +782,8 @@ int pet_target_check(struct pet_data *pd,struct block_list *bl,int type)
...
+		if(pd->target_id == 0 || rnd(10000) < pet_db_ptr->change_target_rate)
```
Lemongrass3110, line 786: Should use rnd_chance

src/map/pet.cpp
```diff
@@ -942,7 +942,7 @@ static int pet_performance(map_session_data *sd, struct pet_data *pd)
 		val = 1;
 
 	pet_stop_walking(pd,2000<<8);
-	clif_pet_performance(pd, rnd()%val + 1);
+	clif_pet_performance(pd, rnd(val) + 1);
```
Lemongrass3110, line 945: Should be a new rnd_value [1,val]

src/map/pet.cpp
```diff
@@ -1315,7 +1315,7 @@ int pet_catch_process2(map_session_data* sd, int target_id)
 	if(battle_config.pet_catch_rate != 100)
 		pet_catch_rate = (pet_catch_rate*battle_config.pet_catch_rate)/100;
 
-	if(rnd()%10000 < pet_catch_rate) {
+	if(rnd(10000) < pet_catch_rate) {
```
Lemongrass3110, line 1318: Should use rnd_chance

src/map/pet.cpp
```diff
@@ -1702,7 +1702,7 @@ static int pet_randomwalk(struct pet_data *pd,t_tick tick)
 				c += pd->status.speed;
 		}
 
-		pd->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME+c;
+		pd->next_walktime = tick+rnd(1000)+MIN_RANDOMWALKTIME+c;
```
Lemongrass3110, line 1705: Should be a new rnd_value [tick+MIN_RANDOMWALKTIME+c,MIN_RANDOMWALKTIME+c+1000[

src/map/quest.cpp
```diff
@@ -775,7 +775,7 @@ void quest_update_objective(map_session_data *sd, struct mob_data* md)
 		for (const auto &it : qi->dropitem) {
 			if (it->mob_id != 0 && it->mob_id != md->mob_id)
 				continue;
-			if (it->rate < 10000 && rnd()%10000 >= it->rate)
+			if (it->rate < 10000 && rnd(10000) >= it->rate)
```
Lemongrass3110, line 778: Should use rnd_chance

src/map/script.cpp
```diff
@@ -5592,7 +5592,7 @@ BUILDIN_FUNC(rand)
 	if( range <= 1 )
 		script_pushint(st, min);
 	else
-		script_pushint(st, rnd()%range + min);
+		script_pushint(st, rnd(range) + min);
```
Lemongrass3110, line 5595: Should be a new rnd_value [min, min+range[

src/map/script.cpp
```diff
@@ -12545,7 +12545,7 @@ BUILDIN_FUNC(homunculus_mutate)
 	if(script_hasdata(st,2))
 		homun_id = script_getnum(st,2);
 	else
-		homun_id = 6048 + (rnd() % 4);
+		homun_id = 6048 + rnd(4);
```
Lemongrass3110, line 12548: Should be a new rnd_value [MER_EIRA,MER_ELEANOR]

src/map/unit.cpp
```diff
@@ -2625,7 +2625,7 @@ int unit_calc_pos(struct block_list *bl, int tx, int ty, uint8 dir)
 			int i;
 
 			for( i = 0; i < 12; i++ ) {
-				int k = rnd()%8; // Pick a Random Dir
+				int k = rnd(8); // Pick a Random Dir
```
Lemongrass3110, line 2628: Should use DIR_* constant.

Lemongrass3110:
I did not comment all of them, but quite many I guess.

Leemonn:
> I did not comment all of them, but quite many I guess.

Yep those are the same that I saw. I don't want to include that refactor in this PR.
I think it's better to do:
`6048 + (rnd() % 4)` --(PR 1)--> `6048 + rnd(4)` --(PR 2)--> `rnd_value(MER_EIRA,MER_ELEANOR)`
Than:
`6048 + (rnd() % 4)` --(PR 1)--> `rnd_value(MER_EIRA,MER_ELEANOR)`

I find it easier to review and detect errors when doing it step by step, also I don't mind making another PR for those refactors. I prefer smaller PRs.

Lemongrass3110:
The problem is that we dont see what usecases we have, if we do not touch it.
Like you wrote yesterday in Discord that rnd_value can be removed.
Just by reviewing your PR I can tell you that we still need it.

So my approach would be to apply the correct functions to the correct places and then improve what is behind that function.
Know what I mean?

Leemonn:
> The problem is that we dont see what usecases we have, if we do not touch it. Like you wrote yesterday in Discord that rnd_value can be removed. Just by reviewing your PR I can tell you that we still need it.
> 
> So my approach would be to apply the correct functions to the correct places and then improve what is behind that function. Know what I mean?

What I meant was that the current `rnd_value` is not needed since we can use `rnd(min, max)`. But yes I understand what you mean.
If you don't mind all those changes in the PR then I will make them, there are A TON of improvements that can be made, some might have a big impact (like the random walk part on mobs, it does a lot of calculations just to randomize the directions).

Do we agree on the new name being `nrand`?

Leemonn:
> I know even for myself there were times I confused the difference between `rand` and `rnd`. Maybe we could stick to something like `ra_rand` or `athena_rand` or something that distinguishes itself from the default `rand` function.

I will replace all calls to `rnd` using other functions. For example:
`rnd%10` or `rnd(10)` ----> `rnd_value(0, 9)`
`rnd%1000 < rate` ----> `rnd_chance(rate, 1000)`

Leemonn:
New commit pushed. Also addresses:
https://github.com/rathena/rathena/issues/7883
https://github.com/rathena/rathena/issues/7884
https://github.com/rathena/rathena/issues/7885

src/login/loginclif.cpp
```diff
@@ -378,7 +378,7 @@ static int logclif_parse_reqkey(int fd, struct login_session_data *sd){
 	RFIFOSKIP(fd,2);
 	{
 		memset(sd->md5key, '\0', sizeof(sd->md5key));
-		sd->md5keylen = (uint16)(12 + rnd() % 4);
+		sd->md5keylen = rnd_value(12u, 15u);
```
Leemonn, line 381: Is this intended to be in range [12, 15]? or was it meant to be [12, 16] but they implemented it wrong?

src/map/map.cpp
```diff
@@ -1894,8 +1890,8 @@ int map_addflooritem(struct item *item, int amount, int16 m, int16 x, int16 y, i
 
 	memcpy(&fitem->item,item,sizeof(*item));
 	fitem->item.amount = amount;
-	fitem->subx = (r&3)*3+3;
-	fitem->suby = ((r>>2)&3)*3+3;
+	fitem->subx = rnd_value(1, 4) * 3;
+	fitem->suby = rnd_value(1, 4) * 3;
```
Leemonn, line 1894: suby and subx are not properly randomized, previously the options were very limited. Now they can freely be 3, 6, 9 or 12 independent of each other's value.

src/map/pet.cpp
```diff
@@ -1672,10 +1672,10 @@ static int pet_randomwalk(struct pet_data *pd,t_tick tick)
 			d = 5;
 
 		for(i = 0; i < retrycount; i++) {
-			int r = rnd(), x, y;
+			int x, y;
 
-			x = pd->bl.x+r%(d*2+1)-d;
-			y = pd->bl.y+r/(d*2+1)%(d*2+1)-d;
+			x = pd->bl.x + rnd_value(-d, d);
+			y = pd->bl.y + rnd_value(-d, d);
```
Leemonn, line 1678: x and y are now properly randomized, previously the options were very limited. Now they are independent of each other's value.

src/map/pet.cpp
```diff
@@ -1702,7 +1702,7 @@ static int pet_randomwalk(struct pet_data *pd,t_tick tick)
 				c += pd->status.speed;
 		}
 
-		pd->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME+c;
+		pd->next_walktime = tick + MIN_RANDOMWALKTIME + c + rnd_value(0, 999);
```
Leemonn, line 1705: Is this code meant to be `tick + MIN_RANDOMWALKTIME + c + rnd_value(0, 999)`?

Leemonn:
@Lemongrass3110 pinging you just in case you didn't see it

Lemongrass3110 (approved):

src/map/mercenary.cpp
```diff
@@ -421,7 +421,7 @@ bool mercenary_dead(s_mercenary_data *md) {
 void mercenary_killbonus(s_mercenary_data *md) {
 	std::vector<sc_type> scs = { SC_MERC_FLEEUP, SC_MERC_ATKUP, SC_MERC_HPUP, SC_MERC_SPUP, SC_MERC_HITUP };
 
-	sc_start(&md->bl,&md->bl, util::vector_random(scs), 100, rnd() % 5, 600000);
+	sc_start(&md->bl,&md->bl, util::vector_random(scs), 100, rnd_value(1, 5), 600000);
```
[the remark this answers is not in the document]
aleos89: I'm afraid I'm not well versed with the Mercenary system. Do you know if a bug report was created prior?

PR #6204, opened 2021-08-19
Fixed Giant Fly Wing behavior

Daegaladh:
* **Addressed Issue(s)**: #6160 

* **Server Mode**: Both

* **Description of Pull Request**:  Adds new RandomAll warpparty mode that first finds a random cell first and then teleports all party members in the map to that position at once, fixing Giant Fly Wing behavior.

PR #7887, opened 2023-08-06
Fix Giant Fly Wing bugs

kaninhot004:
* **Addressed Issue(s)**: -

* **Server Mode**: Renewal

* **Description of Pull Request**: Fix Giant Fly Wing bugs, the normal one (Not unlimited) not consume when used.

This is how it look when bugged.

https://github.com/rathena/rathena/assets/9493525/7ac5c2db-38c7-431b-af60-a529896ec2bb

### DIFF
--- a/db/re/item_db_usable.yml
+++ b/db/re/item_db_usable.yml
@@ -47044,7 +47044,7 @@ Body:
       NoMail: true
       NoAuction: true
     Script: |
-      rentitem 23023,432000;
+      rentitem 12887,432000;
   - Id: 17475
     AegisName: LC_Max_Weight_Up_10Box
     Name: (Limited)Gym Membership Card Box(10)
--- a/db/re/item_db_usable.yml
+++ b/db/re/item_db_usable.yml
@@ -6682,7 +6682,7 @@ Body:
   - Id: 12212
     AegisName: Giant_Fly_Wing
     Name: Giant Fly Wing
-    Type: DelayConsume
+    Type: Usable
     Buy: 2
     Weight: 10
     Trade:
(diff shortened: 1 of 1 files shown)


db/re/item_db_usable.yml
```diff
@@ -47035,7 +47035,7 @@ Body:
       NoMail: true
       NoAuction: true
     Script: |
-      rentitem 12212,432000;
+      rentitem 23023,432000;
```
Atemo, line 47038: This box should give a rental fly wing for 5d (id 12887 I guess) according to kRO description.
Also our english item name is currently wrong.

kaninhot004: Thanks I will check and fix later.

Atemo (approved):

aleos89:
It looks like these were changed in #5859 to support the script function. I can't remember if we changed functionality of Useable and DelayConsume in the past 2 years now. Just leaving this note before it's merged.

Daegaladh:
> It looks like these were changed in #5859 to support the script function. I can't remember if we changed functionality of Useable and DelayConsume in the past 2 years now. Just leaving this note before it's merged.

Maybe related to my PR too https://github.com/rathena/rathena/pull/6204

aleos89:
So it seems like the solution is to add that sleep back in to the function.

Daegaladh:
> So it seems like the solution is to add that sleep back in to the function.

Not at all, the sleep was necessary because the itemskill+warpparty, but now it's an if/else, so changing it back to Usable would work, or we can just move the entire function to the item script.

aleos89:
Gotcha, makes sense!

aleos89 (approved):

PR #7892, opened 2023-08-10
Add custom item bonus bJumprange

eppc0330:
* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

This is a custom item bonus.
A character will jump to the target when a target is farther than the character's normal attack range.
`bonus5 bJumprange,range,splash,atkrate,penalty,mobid;`
 - When you set `splash`, normal attack will be splashed on the first attack after jump as a given splash range.
 - When you set `atkrate`, the splashed damage percent will be changed.
 - Jumping without any penalty or cooltime looks little overpower, so you can set penalty by `penalty`.
A given percent of aspd rate will be reduced on first attack after jump.
 - A character will transform to given mob id on first attack after jump. This will not work when your character is already in transform status. Also will not be shown clearly if your aspd is too high, because a transforming time is affected by your aspd.

 If you want to make character shout like a skill name when jump, uncomment belows in unit.cpp
```
		//char temp[70];
		//snprintf(temp, sizeof(temp), "edit here");
		//clif_showscript(src, temp, AREA);
```

and edit `"edit here"` as what you want


eppc0330:
Sorry Ive opened in wrong branch

PR #7896, opened 2023-08-15
Rewriting reg_db struct with STL container.

inhyositsu:

* **Server Mode**: Both

* **Description of Pull Request**: Rewriting scripting engine with c++ STL container, using std::unordered_map.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -6220,6 +6220,8 @@ BUILDIN_FUNC(deletearray)
 	start = reference_getindex(data);
 	name = reference_getname(data);
 
+	bool is_string = is_string_variable( name );
+
 	if( not_server_variable(*name) ) {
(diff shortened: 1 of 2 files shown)


src/map/intif.cpp
```diff
@@ -393,11 +393,8 @@ int intif_wis_message_to_gm(char *wisp_name, int permission, char *mes)
  * @param sd : Player to save registry
  * @return 1=msg sent, -1=error
  */
-int intif_saveregistry(map_session_data *sd)
+int intif_saveregistry(struct map_session_data *sd)
```
Pokye, line 396:
```suggestion
int intif_saveregistry(map_session_data *sd)
```

inhyositsu: Thanks for recommending!

PR #7902, opened 2023-08-21
Dont exit the process during a crash

inhyositsu:
…em will make core dump correctly in Linux.

* **Addressed Issue(s)**: None.

* **Server Mode**: Both

* **Description of Pull Request**: Fix the problem that server won't make core dump when crashing in Linux environments.

### DIFF
--- a/src/common/core.cpp
+++ b/src/common/core.cpp
@@ -475,10 +475,6 @@ void Core::signal_crash(){
 		this->handle_crash();
 	}
 
-#ifdef WIN32
-	// Now stop the process
-	exit( EXIT_FAILURE );
-#endif
 }
 
 void Core::signal_shutdown(){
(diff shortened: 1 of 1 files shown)


aleos89 (approved):

src/common/core.cpp
```diff
@@ -475,8 +475,10 @@ void Core::signal_crash(){
 		this->handle_crash();
 	}
 
+#ifdef WIN32
 	// Now stop the process
 	exit( EXIT_FAILURE );
+#endif
```
Lemongrass3110, line 481:
```suggestion
```

Should be removed on all OS.
Was mistakenly introduced by me in #7303

Lemongrass3110 (changes requested):

Lemongrass3110 (approved):

Issue #7904, opened 2023-08-22
Marine Sphere movement is not working properly


Leemonn:
### rAthena Hash

eadfa053f6db830e0e769b7241aab5b8df651c92

### Client Date

2015-11-04

### Server Mode

Pre-Renewal

### Result

Tested in latest hash without modifications.

![project_resurrection_Jx3rfHGcpA](https://github.com/rathena/rathena/assets/47922762/af9ce385-0863-4931-b933-8d23e4c689e0)

As you can see in the video everytime you hit a Marine Sphere it starts moving but it randomly stops after moving between 0 and 7 cells.
It happens in any direction and with any type of attack.

### Expected Result

1. Hit Marine Sphere any number of times
2. Marine Spheres moves 7 cells away from attacker

### How to Reproduce

1. Hit Marine Sphere any number of times
2. Marine Spheres moves between 0 and 7 cells away from attacker and stops

### Official Information

.

Singe-Horizontal:
I traced the bug up to TIMER_FUNC(unit_walktoxy_timer), where `!(ud->walk_count%WALK_SKILL_INTERVAL)` often equals false which prevents `clif_move(ud);`.

PR #7905, opened 2023-08-22
Set marine sphere walk_count before escape

Singe-Horizontal:
* **Addressed Issue(s)**: 
#7904

* **Server Mode**: 
Both

This sets the walk_count of the marine sphere when fleeing to make it start the cast of NPC_EXPLOSION at the begining of its movement.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -9757,8 +9757,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case NPC_RANDOMMOVE:
 		if (md) {
 			md->next_walktime = tick - 1;
-			if (md->special_state.ai == AI_SPHERE)
+			if (md->special_state.ai == AI_SPHERE){
+				md->ud.walk_count = WALK_SKILL_INTERVAL - 1;
 				unit_escape(&md->bl, bl, 7, 2);
+			}
 			else
 				mob_randomwalk(md,tick);
 		}


src/map/skill.cpp
```diff
@@ -9757,8 +9757,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 	case NPC_RANDOMMOVE:
 		if (md) {
 			md->next_walktime = tick - 1;
-			if (md->special_state.ai == AI_SPHERE)
+			if (md->special_state.ai == AI_SPHERE){
+				md->ud.walk_count = WALK_SKILL_INTERVAL - 1;
```
aleos89, line 9761: Just curious why we are using the value 4 in this case.

Singe-Horizontal: but I don't like it either

Lemongrass3110: Wouldnt ud->state.force_walk be the cleaner option?

Leemonn:
NPC_RUN seems to be affected by this bug as well. I just tested it on a Ghoul. Setting it to do NPC_RUN lv. 5 on any state will only make the Ghoul run 1 cell then come back and attack the player.
Tried with `ud->state.force_walk = true` but the behavior became really buggy. The Ghoul ran 5 cells, then froze (visual bug) and teleported 5 cells away (for a total of 10 cells). The client wasn't informed of the Ghoul moving.

Leemonn:
Tested with:
`1036,ghoul@NPC_RUN,any,354,10,10000,0,0,no,self,always,,,,,,,26,`
and
`1036,ghoul@NPC_RANDOMMOVE,any,331,1,10000,0,0,no,self,always,,,,,,,26,`

Obviously I didn't test with both at the same time. And yes, I did @reload mobdb between each change.

Vanilla rAthena:
- `NPC_RANDOMMOVE` (non marine-sphere): Nothing. The Ghoul just keeps attacking the player.
- `NPC_RANDOMMOVE` (marine-sphere): erratic behavior shown on gifs of linked issue
- `NPC_RUN`: The Ghoul starts running, after moving 1 cell it goes back and attacks the player. Seems like the running is being cut by the monster attacking the player.

With this PR (adapted to affect all 3 scenarios):
- `NPC_RANDOMMOVE` (non marine-sphere): Nothing. The Ghoul just keeps attacking the player.
- `NPC_RANDOMMOVE` (marine-sphere): Behavior is better but there are still times where the Marine Sphere stops before reaching destination.
- `NPC_RUN`: The Ghoul starts running, after moving 2 cells it goes back and attacks the player. Seems like the running is being cut by the monster attacking the player. It's the same as before but runs a little bit longer before attacking the player.

With `ud->state.force_walk = true`:
- `NPC_RANDOMMOVE` (non marine-sphere): The ghoul get's stuck, it doesn't move, it doesn't attack. Even after the effect wears off.
- `NPC_RANDOMMOVE` (marine-sphere): Erratic behavior. Self destruction casting starts after movement ends which is not right. Also if you attack them while they move they will stop moving and they will stop casting auto destruct.
- `NPC_RUN`: The Ghoul starts running, after reaching destination it will go back 1-2 cells to attack the player then it will stop moving (visual bug) and then it teleports like 10 cells away.


There seems to be a deeper issue here.

Daegaladh:
I thought we fixed this in https://github.com/rathena/rathena/pull/5236 and https://github.com/rathena/rathena/pull/7333. It seems something's really broken on the unit walk functions :/

Leemonn:
Well, I was testing something else and found out that Wink of Charm is also broken. It seems that some change broke mob ai.

This is vanilla rAthena up to date. (I did a pull from master and rebuild right before making this video).

In case you need double confirmation, yes, everything is vanilla, my `monster_ai` is set to `0`.

(Ignore the different skill name, that's a client-side thing)

project_resurrection_BcY3ighaRZ

aleos89:
Please see #7927. Thanks!

Issue #7921, opened 2023-09-05
RG_PLAGIARISM gives too much aspd bonus


eppc0330:
### rAthena Hash

master

### Client Date

20221207

### Server Mode

Renewal

### Result

 Screenshots belows are each from KRo and rA, with no equipments but only bonuses from ```Berserk_Potion``` and ```RG_PLAGIARISM```. (Found issue during testing aspd bonus with berserk potion)

KRo
![plagiarism KRo](https://github.com/rathena/rathena/assets/47050704/0da8119d-8dbe-4075-8024-609be7dc441e)

rA
![plagiarism rA](https://github.com/rathena/rathena/assets/47050704/1893930f-9bcd-46cf-923e-5b4d1615ebc3)

 We can check ```RG_PLAGIARISM``` gives too much aspd bonus.

I didn't capture more screenshots but aspd without berserk potion in KRo goes 178->179,
while aspd goes 178->183 in rA

### Expected Result

```RG_PLAGIARISM``` should give less aspd bonus than now.

### How to Reproduce

Learn ```RG_PLAGIARISM``` with Rogue character

### Official Information

Check aboves.

PR #7922, opened 2023-09-05
Fix RG_PLAGIARISM aspd bonus

eppc0330:
* **Addressed Issue(s)**: #7921

* **Server Mode**: Renewal

Fix RG_PLAGIARISM giving too much aspd bonus.

   Todo: I'm afraid if all other aspd bonuses from skills have same issue. Need check.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -2271,8 +2271,6 @@ int status_base_amotion_pc(map_session_data* sd, struct status_data* status)
 		val += 1 + skill_lv;
 	if ((skill_lv = pc_checkskill(sd,GS_SINGLEACTION)) > 0 && (sd->status.weapon >= W_REVOLVER && sd->status.weapon <= W_GRENADE))
 		val += ((skill_lv + 1) / 2);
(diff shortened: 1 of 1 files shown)


aleos89 (approved):

src/map/status.cpp
```diff
@@ -8218,8 +8216,12 @@ static short status_calc_aspd(struct block_list *bl, status_change *sc, bool fix
 		map_session_data* sd = BL_CAST(BL_PC, bl);
 		uint8 skill_lv;
 
-		if (sd && (skill_lv = pc_checkskill(sd, BA_MUSICALLESSON)) > 0)
-			bonus += skill_lv;
+		if (sd) {
```
chavychaze, line 8219: why you don't make something like this:
`if (sd && ((skill_lv = pc_checkskill(sd, BA_MUSICALLESSON)) > 0 || (skill_lv = pc_checkskill(sd, RG_PLAGIARISM)) > 0) bonus += skill_lv` ?
Just curious.

PR #7927, opened 2023-09-13
Fixes monster escape AI issues

### LINKED ISSUE #7904: Marine Sphere movement is not working properly

### LINKED ISSUE #7916: Mob ai seems broken
### rAthena Hash

0f1886927675a73e34636af4ed33caf7a13361ce

### Client Date

2015-11-04

### Server Mode

Pre-Renewal

### Just a few examples of mob ai being broken

Marine Sphere's movement being unpredictable
project_resurrection_Jx3rfHGcpA

`NPC_RUN` and `NPC_RANDOMNMOVE` don't work properly. Mobs can bypass them and start attacking.

Wink of Charm is broken too 
project_resurrection_BcY3ighaRZ

All of this has been tested again with latest rAthena, no modifications (other than setting packet version and credentials). For people that need the doublecheck, yes `monster_ai` was set to 0.

### Related PRs and Issues
https://github.com/rathena/rathena/issues/7904
https://github.com/rathena/rathena/pull/7905
https://github.com/rathena/rathena/pull/5236
https://github.com/rathena/rathena/pull/7333
https://github.com/rathena/rathena/issues/6371
https://github.com/rathena/rathena/pull/7264

aleos89:
* **Addressed Issue(s)**: Fixes #7904 and fixes #7916

  * Resolves NPC_RUN and NPC_RANDOMMOVE monsters not always reaching their intended escape distance.
  * Resolves Marine Sphere calling the run skill too early resulting in the distance escaped not always being the same.
  * Resolves Wink of Charm not forcing the target to fail skill use on the caster.
  * Renames the monster alchemist state to can_escape to fit multiple uses.
Thanks to @Leemonn!

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2417,8 +2417,7 @@ void mob_log_damage(struct mob_data *md, struct block_list *src, int damage)
 void mob_damage(struct mob_data *md, struct block_list *src, int damage)
 {
 	if( src != nullptr && md->special_state.ai == AI_SPHERE && !md->dmglog[0].id ) {//LOne WOlf explained that ANYONE can trigger the marine countdown skill. [Skotlex]
-		md->state.alchemist = 1;
-		mobskill_use(md, gettick(), MSC_ALCHEMIST);
+		md->state.can_escape = 1;
 	}
 
 	if (src && damage > 0) { //Store total damage...
@@ -3828,7 +3827,7 @@ int mobskill_use(struct mob_data *md, t_tick tick, int event, int64 damage)
 				case MSC_MASTERATTACKED:
 					flag = (md->master_id > 0 && (fbl=map_id2bl(md->master_id)) && unit_counttargeted(fbl) > 0); break;
 				case MSC_ALCHEMIST:
-					flag = (md->state.alchemist); break;
+					flag = (md->state.can_escape); break;
 				case MSC_MOBNEARBYGT:
 					flag = (map_foreachinallrange(mob_count_sub, &md->bl, AREA_SIZE, BL_MOB) > c2 ); break;
 			}
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -482,11 +482,8 @@ static TIMER_FUNC(unit_walktoxy_timer)
 #endif
 
 		// Remove any possible escape states present for mobs once they stopped moving.
-		if (bl->type == BL_MOB) {
-			mob_data *md = BL_CAST(BL_MOB, bl);
-
-			if (md != nullptr)
-				md->state.can_escape = 0;
+		if (md != nullptr) {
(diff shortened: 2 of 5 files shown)


src/map/mob.cpp
```diff
@@ -2417,8 +2417,7 @@ void mob_log_damage(struct mob_data *md, struct block_list *src, int damage)
 void mob_damage(struct mob_data *md, struct block_list *src, int damage)
 {
 	if( src != nullptr && md->special_state.ai == AI_SPHERE && !md->dmglog[0].id ) {//LOne WOlf explained that ANYONE can trigger the marine countdown skill. [Skotlex]
-		md->state.alchemist = 1;
-		mobskill_use(md, gettick(), MSC_ALCHEMIST);
```
Lemongrass3110, line 2421: If we remove this call, then MSC_ALCHEMIST is never used at all.

src/map/unit.cpp
```diff
@@ -481,6 +481,14 @@ static TIMER_FUNC(unit_walktoxy_timer)
 		}
 #endif
 
+		// Remove any possible escape states present for mobs once they stopped moving.
+		if (bl->type == BL_MOB) {
+			mob_data *md = BL_CAST(BL_MOB, bl);
```
Lemongrass3110, line 486: No need to BL_CAST anymore, since you already checked the type in the line above.

src/map/mob.cpp
aleos89: It's a slight trick. It is used in the switch case where because it checks `ms[i]->cond` since NPC_RUN uses alchemist as the condition for Marine Sphere. We want to call NPC_RUN only when the value of can_escape is 1. Before we were making the MSC_ALCHEMIST call before the mob's state was changed/saved resulting in the flag=1 and the skill firing off too early.

src/map/unit.cpp
aleos89: If I read up a little more in the function I would have seen `md` is already defined too. 🙃

Lemongrass3110 (approved):

Issue #7844, opened 2023-07-10
KRo not add item log on item refine?


eppc0330:
### Is your request related to a missing official feature?

No

### Describe the solution you'd like

![refine kro 01](https://github.com/rathena/rathena/assets/47050704/d0adfa21-a7b3-4a7a-9cfe-02b6b9204776)
This is from KRo, today from my test of ticket refiener.
There's no item obtain message on refining.
Only unequip, refine success, and equip.

KRo does not delete and add item anymore on refine

![refine rA](https://github.com/rathena/rathena/assets/47050704/3686e157-1d78-42f1-b0e1-63364057d609)
And this is what current rA does when we refine.

PR #7933, opened 2023-09-17
Remove pick log on item refine

eppc0330:
* **Addressed Issue(s)**: #7844

* **Server Mode**: renewal

   Actually I'm not really sure what's the intent of deleting and adding item on refining. I'm just guessing it's for picklog.
One certain is that KRo does not doing them anymore.


Badarosk0:
The registration log for refining is necessary to monitor players. If any player reports that they lost an item or that the item went missing and they are actually lying because they broke their item, then we will have that in the picklog. However, it needs to be kept in the log and should not be removed.

mazvi:
picklog refine is very important to find out if there are gaps that players are using. After all, the log is a backend for the admin from the player side, they have no influence there and there are no logs. but it really helps admins in monitoring items if there are bugs. I reject this pull.

aleos89:
Agreed, leaving the log in place will help admins. It might also be important to make this an ifdef based on renewal mode since this behavior is still expected for prerenewal.

eppc0330:
Thanks for feedback all. I reverted pick log but I will just close the pull if deleting/adding process on refine is still necessary.

PR #7957, opened 2023-10-04
Fixes Longing for Freedom making players stuck

### LINKED ISSUE #7777: Longing for Freedom
### rAthena Hash

Latest

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

The skill is supposed to give Clowns/Gypsies the freedom to move and use skills during Ensembles. Currently, the skill does not work at all. It worked correctly before, meaning a recent update must have broken it.
Using the skill does allow the character to use skills and attack but movement is not restored. Furthermore, the character is stuck in place indefinitely even after the Ensemble ends. You need to relog to fix this.

### Relevant Log Output

_No response_

### Expected Result

Characters should be able to move freely inside Ensembles.

### How to Reproduce

Just get a pair of Clown/Gypsy and use the skill, the issue is immediately apparent.

### Official Information

.

_No response_

aleos89:
* **Addressed Issue(s)**: Fixes #7777

  * Fixes Longing for Freedom making players not be able to walk after songs have ended.
Thanks to @XanKriegor1!


Daraen1:
I have asked on rathena discord if the behavior is correct, but no answer after 3 days. 
So if one of our contributors can confirm or deny if still buggy, it would be cool for everyone.
Video show how it works for now. 
You can't attack but you can skill, I think it's not correct, you should able to attack too.
Thank you.
https://www.youtube.com/watch?v=uOXRHjYip2k

Daraen1:
Sorry to bump this, but I think it is a important case for Pre-renewal (I don't know how it works on renewal)

src/map/status.cpp
```diff
@@ -5412,6 +5412,11 @@ void status_calc_state( struct block_list *bl, status_change *sc, std::bitset<SC
 				  || (sc->getSCE(SC_CRYSTALIZE) && bl->type != BL_MOB)
  				 )
 				 sc->cant.move += (start ? 1 : ((sc->cant.move) ? -1 : 0));
+#ifndef RENEWAL
+		// Remove movement restriction when Longing for Freedom becomes active with an Ensemble skill.
+		if (start && sc->getSCE(SC_DANCING) && sc->getSCE(SC_DANCING)->val4 && sc->getSCE(SC_LONGING))
+			sc->cant.move = 0;
```
Playtester, line 5418: Shouldn't it be just substracting "1" in case you also can't move out of other reasons?

Lemongrass3110: If I was not mistaken this would also be fixed with #8834 btw.
There I fixed the counting.

Playtester: True, but it can take a long time until that can go live, no? Because of the memory access issue.
Wouldn't it be better to fix this one for now?
Though the build doesn't seem to be working anymore...

Daraen1: Well, i comment just to give my experience last year when i tried this pull request on my own server.
If it's outdated with the new src of rAthena does not take into account my comment because I have made a huge pause on RO, As I said, when I tried this one, my character has able to move inside the song, the only thing you can't is to attacked in close combat, you can use others skill like Arrow vulcan but if you want to attack in close combat with normal attacks, you have to wait for the song ending.
There is a video on my old comment if you want what is going on.
Sorry for my bad english, I hope it will be understandable for you guys !

db/pre-re/status.yml
```diff
@@ -1706,6 +1706,8 @@ Body:
     CalcFlags:
       Speed: true
       Aspd: true
+    States:
+      NoMove: true
```
Playtester, line 1710: This also requires "NoMoveCond: true"

Lemongrass3110: Added in 954d7dc

db/pre-re/status.yml
```diff
@@ -1706,6 +1706,9 @@ Body:
...
+      NoMoveCond: true
```
Playtester, line 1711: There is one more issue with Longing for Freedom. Now it enables movement correctly, but it doesn't allow normal attacks yet.

Lemongrass3110: Fixed in 948001e

src/map/status.cpp
```diff
@@ -5627,6 +5628,10 @@ void status_calc_state( block_list& bl, status_change& sc, std::shared_ptr<s_sta
 					}
 					break;
 
+				case SC_LONGING:
```
Playtester, line 5631: One thing I wonder is... what's the point of having configurable flags when we need to add all SCs here anyway that have these flags defined?

Also, shouldn't this then be for pre-re only?

Lemongrass3110: This is only necessary to define the conditional ones. Not all of them.

Playtester (approved):
Not fully convinced on the lambda style, but seems to fully work now.

PR #7963, opened 2023-10-10
Added missing sd check for clif_skill_fail

Atemo:
* **Addressed Issue(s)**: #8239

* **Server Mode**: both


Lemongrass3110 (commented):
clif_skill_fail has a nullpo check inside already.
Maybe it would be better to change it in there (one location) instead of all over src?
We only would have to change the nullpo check to a normal if sd == nullptr return.
The other option would be to turn sd into a reference.
At the same time we could think of introducing default arguments to that clif function.

aleos89 (approved):

PR #7971, opened 2023-10-15
Add autoloot script command

Emistry:
* **Addressed Issue(s)**: 
NPC script unable to adjust the autoloot settings unless it use something like `atcmd "@autoloot";` which create unnecessary atcommandlog at server side.

* **Server Mode**: 
any

enable NPC script to configure the autoloot setting of players without need use the `@autoloot` commands which could create unnecessary atcommandlog if configured.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11362,7 +11362,15 @@ Returns true if the player is dead else false.
 
 *autoloot(<rate>{, <char_id>});
 
-This command set the rate of autoloot.
-Returns autoloot value on success.
+This command sets the rate of autoloot (default: 10000 = 100%).
+Returns true on success and false on failure.
+Note: It will toggle existing autoloot settings if rate value aren't provided.
+
+---------------------------------------
+
+*has_autoloot({<char_id>});
+
+This command checks whether a player configured autoloot.
+Returns current autoloot value on success.
 
 ---------------------------------------
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11359,3 +11359,10 @@ Note: Only works with classes that use the ranking system.
 Returns true if the player is dead else false.
 
 ---------------------------------------
+
+*autoloot(<rate>{, <char_id>});
+
+This command set the rate of autoloot.
+Returns autoloot value on success.
+
+---------------------------------------
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -26941,18 +26941,47 @@ BUILDIN_FUNC(macro_detector) {
 
 // ===================================
 // *autoloot(<rate>{, <char_id>});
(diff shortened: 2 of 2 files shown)


Atemo:
How about a parameter on the script command `atcommand` to configure the log instead?

Emistry:
> How about a parameter on the script command `atcommand` to configure the log instead?

this is also another way. But doesn't to execute npc script using atcommand since it doesn't look like a proper way to do it even if it does work.

src/map/script.cpp
```diff
@@ -26939,6 +26939,25 @@ BUILDIN_FUNC(macro_detector) {
...
+// ===================================
+BUILDIN_FUNC(autoloot) {
+	map_session_data *sd;
+
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, 0);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	sd->state.autoloot = cap_value(script_getnum(st, 2), 0, 10000);
```
Lemongrass3110, line 26955: Please change this so that wrong usage (< 0 and > 10000) get reported and result in SCRIPT_CMD_FAILURE

doc/script_commands.txt
```diff
@@ -11359,3 +11359,10 @@ Note: Only works with classes that use the ranking system.
 Returns true if the player is dead else false.
 
 ---------------------------------------
+
+*autoloot(<rate>{, <char_id>});
```
Lemongrass3110, line 11363: Shouldnt "rate" also be optional and default should be 10000?

doc/script_commands.txt
```diff
@@ -11359,3 +11359,10 @@ Note: Only works with classes that use the ranking system.
...
+
+This command set the rate of autoloot.
```
Lemongrass3110, line 11365:
```suggestion
This command sets the rate of autoloot.
```

doc/script_commands.txt
```diff
@@ -11359,3 +11359,10 @@ Note: Only works with classes that use the ranking system.
...
+Returns autoloot value on success.
```
Lemongrass3110, line 11366: If you do it like this, if you disable autoloot with this command, you will not be able to check success.
I would rather return true/false.

Lemongrass3110 (approved):

src/map/script.cpp
Emistry: I added warning if value are invalid, but I think it shouldn't stop the script if wrong value since we could just cap it and continue the script, the capped value doesn't affect the result compare to the value provided by users.

Emistry:
Hmm I think I merged unfinish conflicts and caused it went wrong. Recreated https://github.com/rathena/rathena/pull/8032

PR #7972, opened 2023-10-15
Add has_autoloot script command

Emistry:
* **Addressed Issue(s)**: 
NPC script unable to retrieve what is the current configured autoloot settings of players.
This is troublesome when players access event hunting/farming maps with pre-configured autoloot.
NPC unable to prevent it.

* **Server Mode**: 
any 

Enable NPC script to access to current configured autoloot settings of players.
Useful when we need to check if users enabled/disable it before access any event maps.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11360,7 +11360,7 @@ Returns true if the player is dead else false.
 
 ---------------------------------------
 
-*is_autoloot({<char_id>});
+*has_autoloot({<char_id>});
 
 This command checks whether a player configured autoloot.
 Returns current autoloot value on success.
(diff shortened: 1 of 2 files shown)


doc/script_commands.txt
```diff
@@ -11359,3 +11359,11 @@ Note: Only works with classes that use the ranking system.
 Returns true if the player is dead else false.
 
 ---------------------------------------
+
+*is_autoloot({<char_id>});
```
aleos89, line 11363: Maybe it's me, but `is_autoloot` seems a bit confusing since this almost sounds like you want to check if the item is an autoloot type or something. Maybe more like `is_autolooting` or `has_autoloot`?

Lemongrass3110 (approved):

PR #7734, opened 2023-04-22
Correction of the clone mode

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/7689, https://github.com/rathena/rathena/issues/7733

* **Server Mode**: -

* **Description of Pull Request**: 

[Removed MD_NOCAST and MD_NORANDOMWALK from default clone mode
Added MD_NORANDOMWALK for slaveclone

PR #7973, opened 2023-10-16
fix clone skill behavior

Aristeus09:
* **Addressed Issue(s)**: 
Clone are not able to use skill and it cast self/support skill to his target enemy which is wrong behavior. i already mentioned this to @aleos89 but it seems he was too busy right now.

* **Server Mode**: 
any

I add in base status mode exception the "Nocast", "NoRandomWalk", "RandomTarget" to correct the behavior of clone.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -3667,7 +3667,7 @@ int status_calc_pc_sub(map_session_data* sd, uint8 opt)
 
 	// !FIXME: Most of these stuff should be calculated once, but how do I fix the memset above to do that? [Skotlex]
 	// Give them all modes except these (useful for clones)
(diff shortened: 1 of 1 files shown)


Aristeus09:
this is what happen when not added the norandomtarget in exception.
https://github.com/rathena/rathena/assets/34960872/91269c78-dac8-4f7f-9baf-40b859853624



and this is the behavior when norandomtarget added in exception.
https://github.com/rathena/rathena/assets/34960872/eb4edd5c-7347-4881-a7e0-d440ccdda672

Atemo:
Duplicate of https://github.com/rathena/rathena/pull/7734

PR #7977, opened 2023-10-20
Implement NPC_WIDECRITICALWOUND

eppc0330:
* **Addressed Issue(s)**: #1525

* **Server Mode**: Renewal

 Implement mob skill `NPC_WIDECRITICALWOUND` to fix item option of `XM_Celine_Kimi_Card`(ID:31023)


src/map/status.cpp
```diff
@@ -11445,8 +11445,6 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 				val1 = rnd()%ELE_ALL;
 			break;
 		case SC_CRITICALWOUND:
```
aleos89, line 11447: Was this case supposed to be removed too? Right now it's going to fall into SC_MAGICMIRROR right below.

db/re/skill_db.yml
```diff
@@ -17953,6 +17953,41 @@ Body:
...
+      - Level: 6
+        Area: 14
+      - Level: 7
+        Area: 14
+      - Level: 8
+        Area: 14
+      - Level: 9
+        Area: 14
+      - Level: 10
+        Area: 2
```
Lemongrass3110, line 17988: Looks like a copy paste issue?

src/map/status.cpp
Lemongrass3110: Agree with aleos here. This looks incorrect.
If it should not decrease heal effectiveness, this case needs to be removed completely and also the references in the calculations in pc.cpp and skill.cpp need to be removed.

Lemongrass3110 (approved):

Lemongrass3110:
@eppc0330 you updated the branch meanwhile. Did you have time to check the comments? Otherwise we cannot merge this PR.

eppc0330:
> meanwhile

Sorry what should I do?

eppc0330:
@Lemongrass3110 Hmm.. Im query. I have commented explaining your both two questions and you are asking again. So it means my answers are not explaining enough right? Or is there something other problem? Like my comments are not visible for other?

Lemongrass3110:
> @Lemongrass3110 Hmm.. Im query. I have commented explaining your both two questions and you are asking again. So it means my answers are not explaining enough right? Or is there something other problem? Like my comments are not visible for other?

If you did not click "submit review" those comments/answers are marked "pending" and are not visible for others.

eppc0330:
@Lemongrass3110 Ohhh I didn't know it for real. Thank you and sorry. So thats' why people not replying my comments...
Now I submit review. Please check my reviews.

eppc0330:
@Lemongrass3110 
https://www.divine-pride.net/database/skill/749

About skill range, actually I don't know the exact reason either but just followed DP

src/map/status.cpp
Lemongrass3110: Can you please change the code like you copied it above?
```
		case SC_CRITICALWOUND:
			// Level 1 ~ 5 & 6 ~ 10 has different duration
			// Level 6 ~ 10 use effect of level 1 ~ 5
			val1 = 1 + ((val1-1)%5);
			val2 = 20*val1;
			break;
```
This way it is more readable and additionally it fits our plans for the future better.
Thank you!

Lemongrass3110 (approved):

src/map/status.cpp
eppc0330: Thanks for review. I ll do it soon when Im home

aleos89 (approved):

Issue #6885, opened 2022-04-25
Player-skill-summoned-mob weird behavior


eppc0330:
* **rAthena Hash**: master

* **Client Date**: 20211117

* **Server Mode**: Renewal

* **Description of Issue**: 
 *  Skill summoned monster, ex)Biolo mobs, does not get targeted from other monster.
    ![biolosummon03](https://user-images.githubusercontent.com/47050704/165068593-60d73102-00ba-4fd8-837b-a0d2286dcbe4.gif)
     Tested on my server, not targeted, gets no damage.

    ![biolosummon01](https://user-images.githubusercontent.com/47050704/165068637-c9e49b05-6dea-49b9-9873-49e9cfef63df.gif)
     From official KRo. It does get targeted and get damage.

  * The mob is not removed on logout.
     ![biolosummon02](https://user-images.githubusercontent.com/47050704/165068663-d18f0c27-c0f8-4009-b26e-0e00d62945aa.gif)
      I didn't put it all in video, but I have tried relogin almost 10 times to see this.
      The summoned mob randomly remains or disappears. Sometimes it goes away right after first relogin and sometimes doesn't.

      ![biolosummon04](https://user-images.githubusercontent.com/47050704/165070596-2b159d86-bc95-4877-bebb-8051f3889c6f.gif)
      If there's other user in map, the mob disappears right away.

  * Result: 
  * Expected Result: 
  * How to Reproduce: 
  * Official Information: 
  

* **Modifications that may affect results**:

Akkarinage:
Five months later - can anyone confirm if this is still occurring randomly?

We could do with some more feedback on this issue.

Daegaladh:
Could be related to #5064, I'd like to confirm no longer happens after [fa363b9](https://github.com/rathena/rathena/commit/fa363b972d730c249bc6a3ebcc1b4b081abb164b)

eppc0330:
@Akkarinage @Daegaladh  Sorry i didn't know I got replies.
Same, I have checked this today and confirmed that we still have all problems above.
One of my friends said Biolo and Meister's creatures must treated as separate kind from monster,
like Homunculus of Genetic or Element of Sorcerer do.

eppc0330:
Issue update.
![biolosummon05](https://github.com/rathena/rathena/assets/47050704/d7827915-c6e5-4341-894a-161641ce59c0)
When logout and change charcter to one in same account
The slave recognize new character as a master.

PR #7978, opened 2023-10-23
Fix slavemob bahavior

eppc0330:
* **Addressed Issue(s)**: #6885

* **Server Mode**: Renewal

 Fix slavemobs with special ai `AI_ABR` and `AI_BIONIC` not getting targeted from normal ai mobs.

  ~~Todo:~~
    ~~Still need to fix slavemob not removed on logout.~~ Solved. Thank you @aleos89 

  Add:
     Adjust slavemobs to walk back if master is too far. It is an official behavior, confirmed ones are Biolo's slaves.
    But need check how much distance is allowed for slave. Looks `MOB_SLAVEDISTANCE`(2) is too short. The distance I saw in video about Biolo using `Wooden Warrior` is at least 5.


aleos89 (changes requested):
Just some reference and function type changes/cleanups.

```cpp
/**
 * Remove slaves if master logs off.
 * @param bl: Mob data
 * @param ap: List of arguments
 * @return 1 on removal, otherwise 0
 */
int mob_removeslave_sub(block_list *bl, va_list ap) {
	int id = va_arg(ap, int);
	mob_data *md = (mob_data *)bl;

	if (md != nullptr && md->master_id == id) {
		unit_free(bl, CLR_OUTSIGHT);
		return 1;
	}

	return 0;
}

/**
 * Remove slaves on a map.
 * @param bl: Player data
 * @return 1 on removal, otherwise 0
 */
int mob_removeslave(block_list *bl) {
	return map_foreachinmap(mob_removeslave_sub, bl->m, BL_MOB, bl->id);
}
```
For the removal, you could try something like this, there are a few `mob_countslave` functions to throw it next to. Then in unit.cpp::unit_free for the BL_PC type, you can make the call for `mob_removeslave(sd->bl)`.

src/map/mob.cpp
```diff
@@ -1444,8 +1444,15 @@ static int mob_ai_sub_hard_slavemob(struct mob_data *md,t_tick tick)
 			}
 		}
 
-		if(md->target_id) //Slave is busy with a target.
-			return 0;
+		if(md->target_id) {//Slave is busy with a target.
```
Lemongrass3110, line 1447:
```suggestion
		// Slave is busy with a target.
		if(md->target_id) {
```

src/map/mob.cpp
```diff
@@ -1444,8 +1444,15 @@ static int mob_ai_sub_hard_slavemob(struct mob_data *md,t_tick tick)
...
+			if (bl->type == BL_PC && md->master_dist > 5) {//Player's slave should come back when master's too far, even if it is doing with a target.
```
Lemongrass3110, line 1448:
```suggestion
			// Player's slave should come back when master's too far, even if it is doing with a target.
			if (bl->type == BL_PC && md->master_dist > 5) {
```
Besides that a question: Should this be `( MOB_SLAVEDISTANCE * 2 + 1 )` instead of hardcoding 5? Or even a battle config?

src/map/mob.cpp
```diff
@@ -3477,6 +3484,33 @@ int mob_countslave(struct block_list *bl)
...
+
+	return 0;
+}
+
+/**
+ * Remove slaves on a map.
+ * @param bl: Player data
+ * @return 1 on removal, otherwise 0
+ */
+int mob_removeslave(block_list *bl) {
```
Lemongrass3110, line 3510:
```suggestion
int mob_removeslaves(block_list *bl) {
```

src/map/mob.hpp
```diff
@@ -536,6 +536,7 @@ int mobskill_event(struct mob_data *md,struct block_list *src,t_tick tick, int f
 int mob_summonslave(struct mob_data *md2,int *value,int amount,uint16 skill_id);
 int mob_countslave(struct block_list *bl);
 int mob_count_sub(struct block_list *bl, va_list ap);
+int mob_removeslave(block_list* bl);
```
Lemongrass3110, line 539:
```suggestion
int mob_removeslaves(block_list* bl);
```

src/map/mob.cpp
```diff
@@ -3477,6 +3484,33 @@ int mob_countslave(struct block_list *bl)
...
+	return map_foreachinmap(mob_removeslave_sub, bl->m, BL_MOB, bl->id);
```
Lemongrass3110, line 3511:
```suggestion
	return map_foreachinmap(mob_removeslaves_sub, bl->m, BL_MOB, bl->id);
```

src/map/mob.cpp
```diff
@@ -3477,6 +3484,33 @@ int mob_countslave(struct block_list *bl)
 	return map_foreachinmap(mob_countslave_sub, bl->m, BL_MOB,bl->id);
 }
 
+/**
+ * Remove slaves if master logs off.
+ * @param bl: Mob data
+ * @param ap: List of arguments
+ * @return 1 on removal, otherwise 0
+ */
+int mob_removeslave_sub(block_list *bl, va_list ap) {
```
Lemongrass3110, line 3493:
```suggestion
int mob_removeslaves_sub(block_list *bl, va_list ap) {
```

Lemongrass3110 (approved):

src/map/mob.cpp
eppc0330: Guess battle config sounds good?

Playtester:
Hint: Removed my assignment as this is referring to Renewal which I can't test and my change to make the slave AI not active the whole time is independent of this PR.

aleos89 (approved):

PR #7985, opened 2023-10-27
Implemented DK_DRAGONIC_BREATH

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal

https://github.com/rathena/rathena/pull/7024 split for easier review.
All credit to @datawulf 

Dragonic Breath
--------------------------------------------------------------
- Deals normal long ranged physical damage to the target.
- Can only be used when riding dragon.
- Deals additional damage depends on user's MaxHP and MaxSP.
- Affected by Hit/Flee calculation.
- Ignores physical defense.
- Damage property depends on weapon property.
- Affected by attack modifiers (such as race modifier).
- Recovers 2 AP.
- When using while under Dragonic Aura buff, increases influence of MaxHP and MaxSP in skill damage.
- Fixed casting time : 0.5 seconds (based on level 10).
- Variable Casting time : 2 seconds (based on level 10).
- Cooldown: 0.5 seconds (based on level 10).
- Global cool time : 0.15 seconds (based on level 10). 

Source
-----------------------------------------------------------------------
https://www.divine-pride.net/database/skill/6001
https://www.divine-pride.net/forum/index.php?/topic/3723-kro-jobs-improvement-project/page/13/

Additional note
-----------------------------------------------------------------------
The skill requires a client 20220602+.


src/map/battle.cpp
```diff
@@ -5405,6 +5405,14 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 			if (sc && sc->getSCE(SC_GIANTGROWTH) && rnd()%100 < 30)
 				skillratio *= 2;
 			break;
+		case DK_DRAGONIC_BREATH:
+			//TODO: needs official HP/SP scaling [Muh]
+			skillratio += -100 + 50 + 350 * skill_lv + 5 * sstatus->pow;
+			skillratio += sstatus->max_hp / 500 + status_get_max_sp(src) / 40;
+			if (sc && sc->getSCE(SC_DRAGONIC_AURA))
+				skillratio += sstatus->max_hp / 500 + status_get_max_sp(src) / 40;
```
Lemongrass3110, line 5413:
```
Damage is increased or reduced depends on user's MaxHP and MaxSP.
When using under Dragonic Aura buff, increases damage rate from user's MaxHP and MaxSP.
```
This sounds for me like it would be adjusted twice.

Lemongrass3110 (approved):

Kuyaraikub:
https://github.com/rathena/rathena/assets/147054311/d63710fd-9717-438f-bdb6-6e5c3225ba74

src/map/battle.cpp
aleos89: Looking at the current format of the skill, it looks like this is adjusted twice, so I think this is good now?

datawulf: Just a little note. I still don't know exact HP / SP scaling but:
- it's much stronger than here
- the extra boost from dragonic aura is not as strong as the base scaling (<50% of the base effect)

InFiNiTe90X:
Just a question, how can I make this work, I updated but it is not working...  And if it needs a newer client, where to get it? thx..

InFiNiTe90X:
> _2023_10_29_23_37_40_373.mp4

Can i ask how did you solve it?

PR #8005, opened 2023-11-05
Implemented script command opentips

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: -

Opens the tip box UI for the attached player or the given character ID.

Example
-------------------------------------------
```
bl_ice,198,268,1	script	#ba_d1_ice	GATE_SKYBLUE,{
	opentips( 8366 );	// open a tip box about depth biosphere
	end;
}
```


vstumpf (approved):

src/map/script.cpp
```diff
@@ -26939,6 +26940,22 @@ BUILDIN_FUNC(macro_detector) {
...
+	ShowError( "buildin_opentips: This command requires PACKETVER 20151202 or newer.\n" );
+	return SCRIPT_CMD_FAILURE;
+#else
+	map_session_data* sd;
+
+	if (!script_charid2sd(3, sd)) {
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	clif_ui_open( *sd, OUT_UI_TIP, script_getnum(st, 2) );
```
aleos89, line 26954: Is there an ID range we need to check or is that done in the clif/client side?

aleos89 (approved):

src/map/script.cpp
Atemo: Not that I know. Unknown ID on client just opens a random tip.

Lemongrass3110 (approved):

PR #8013, opened 2023-11-16
Refactoring of chat struct

Singe-Horizontal:
### Refactoring of chat struct
Converted C constructs to modern cpp
 * Conversion of regular arrays to std::array, that allows iterating and holds size info
 * Conversion of functions to class methods when appropriate
 * Conversion of Macros to inline constexpr for easier:
   * Scoping
   * Rebuild
   * Debugging
 * Declaration and definition on the same line, the closest to where it's used
 * Removed class-specifiers when possible
 Converted unused int return types to void

Added namespace chats::
Applied google guidelines for naming conventions
 * CamelCase for structs, classes, typedefs, templates and functions
 * Underscore for local variables and namespaces
 * A trailing _ for private members

### DIFF
--- a/src/map/chat.cpp
+++ b/src/map/chat.cpp
@@ -21,27 +21,36 @@
 #include "pc.hpp"
 #include "pc_groups.hpp"
 
-int chat_triggerevent(struct chat_data *cd); // forward declaration
+namespace chats{
 
 /// Initializes a chatroom object (common functionality for both pc and npc chatrooms).
-/// Returns a chatroom object on success, or NULL on failure.
-static struct chat_data* chat_createchat(struct block_list* bl, const char* title, const char* pass, int limit, bool pub, int trigger, const char* ev, int zeny, int minLvl, int maxLvl)
-{
-	struct chat_data* cd;
-	nullpo_retr(NULL, bl);
+/// Returns a chatroom object on success, or nullptr on failure.
+ChatData* ChatData::CreateChat(
+		block_list* bl,
+		const char* title,
+		const char* pass,
+		int limit,
+		bool pub,
+		int trigger,
+		const char* ev,
+		int zeny,
+		int min_level,
+		int max_level) {
 
-	cd = (struct chat_data *) aMalloc(sizeof(struct chat_data));
+	nullpo_retr(nullptr, bl);
+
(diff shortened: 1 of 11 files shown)


src/map/chat.hpp
```diff
@@ -5,42 +5,45 @@
...
+	block_list bl;            // data for this map object
 	char title[CHATROOM_TITLE_SIZE]; // room title 
 	char pass[CHATROOM_PASS_SIZE];   // password
 	bool pub;                        // private/public flag
-	uint8 users;                     // current user count
-	uint8 limit;                     // join limit
-	uint8 trigger;                   // number of users needed to trigger event
+	uint32 trigger;                   // number of users needed to trigger event
+	uint32 users;                     // current user count
+	uint32 limit;                     // join limit
```
vstumpf, line 21: Is there a reason why to increase the storage of these?
if it's to keep in line with MAX_CHAT_USERS, we can either change the storage type of MAX_CHAT_USERS, or have some static_assert comparing with `std::numeric_limit<decltype(ChatData::users)>::max()`

src/map/chat.cpp
```diff
@@ -115,51 +119,47 @@ int chat_createpcchat(map_session_data* sd, const char* title, const char* pass,
...
- * @return 0
  */
-int chat_joinchat(map_session_data* sd, int chatid, const char* pass)
-{
-	struct chat_data* cd;
+void JoinChat(map_session_data* sd, int chatid, const char* pass){
+	nullpo_retv(sd);
 
-	nullpo_ret(sd);
+	ChatData* cd = (ChatData*)map_id2bl(chatid);
```
vstumpf, line 134: we should replace every usage of `(ChatData*)map_id2bl(chatid)` with `map_id2cd(chatid)`; we get the BL_CAST for free

vstumpf (changes requested):
Havent checked most of the chat.cpp changes, will do in the future

src/map/chat.hpp
AoShinRO: client use uint16 here

Singe-Horizontal:
Closing this as this PR was mostly about changing nomenclature style which isn't much needed. Converting MACRO that aren't used compiler flags to const expr would be interesting.

Issue #8024, opened 2023-11-30
Retrieving Guild Bounded Items from Offline Members


micosanityyy:
### rAthena Hash

Tried with both latest and older rAthena version

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

When kicking an offline member with guild bounded items, the items are being left in the character instead of being returned to guild storage. This only happens when the member is offline, kicking them while they are online works fine and returns the guild bounded items in guild storage.

I'm getting this error when kicking an offline member with guild bounded items.
![qweewq](https://github.com/rathena/rathena/assets/68556725/b58373ca-404c-4dad-ac7e-d3e033a745c2)

### Expected Result

Retrieving guild bounded items should word regardless if the guild member is online or offline

### How to Reproduce

Create a guild, invite a player in the guild, give them a guild bounded item then try kicking them from the guild while online, this will return the items in your guild storage.

Now try doing it again, but kick the member while the character is offline. The items will not be retrieved and will stay in the character, guild bound status if the items still remain but it's now on a character with no guild.

### Official Information

N/A

### Modifications that may affect results

None

gidzdlcrz:
This is related to https://github.com/rathena/rathena/issues/7966

micosanityyy:
> This is related to #7966

Not really. I've found that and applied it before posting this. There's a bug in the option indexes.

gidzdlcrz:
why you close @micosanityyy You already have solution? Can you share?

PR #8031, opened 2023-12-09
Restores a correct option index

Singe-Horizontal:
* **Addressed Issue(s)**: 
#8024

* **Server Mode**: 
Both

Restores a correct option index when retrieving guild bound items for offline characters.


aleos89 (approved):

gidzdlcrz:
To add. The member you will kick should have storage permit usage on. If not the guild bounded item will be forever lost.

PR #8032, opened 2023-12-10
Add autoloot script command

Emistry:
* **Addressed Issue(s)**: 
NPC script unable to adjust the autoloot settings unless it use something like `atcmd "@autoloot";` which create unnecessary atcommandlog at server side.

* **Server Mode**: 
any

enable NPC script to configure the autoloot setting of players without need use the `@autoloot` commands which could create unnecessary atcommandlog if configured.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11390,3 +11390,10 @@ Returns current autoloot value on success.
 
 ---------------------------------------
 
+*autoloot(<rate>{, <char_id>});
+
+This command sets the rate of autoloot (default: 10000 = 100%).
+Returns true on success and false on failure.
+Note: It will toggle existing autoloot settings if rate value aren't provided.
+
+---------------------------------------
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11396,4 +11396,11 @@ This command sets the rate of autoloot (default: 10000 = 100%).
 Returns true on success and false on failure.
 Note: It will toggle existing autoloot settings if rate value aren't provided.
 
+Example:
+    autoloot();      // toggle on/off depend on existing autoloot
+    autoloot(0);     //   0.00% or off
+    autoloot(100);   //   1.00%
+    autoloot(3333);  //  33.33%
+    autoloot(10000); // 100.00%
+
 ---------------------------------------
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11390,11 +11390,12 @@ Returns current autoloot value on success.
 
 ---------------------------------------
 
-*autoloot(<rate>{, <char_id>});
+*autoloot({<rate>{, <char_id>}});
 
-This command sets the rate of autoloot (default: 10000 = 100%).
+This command sets the rate of autoloot.
+If no rate is provided and the user has autoloot disabled it will default to 10000 = 100% (enabled) or
+if the user has autoloot enabled it will default to 0 = 0% (disabled).
 Returns true on success and false on failure.
-Note: It will toggle existing autoloot settings if rate value aren't provided.
 
 Example:
     autoloot();      // toggle on/off depend on existing autoloot
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -27037,7 +27037,7 @@ BUILDIN_FUNC(autoloot) {
 		rate = script_getnum(st, 2);
 
 		if (rate < 0 || rate > 10000) {
-			ShowWarning("buildin_autoloot: invalid rate value %d, shall be between 0 ~ 10000.\n", rate);		
+			ShowWarning("buildin_autoloot: Invalid rate value %d, should be between 0 ~ 10000.\n", rate);		
 			script_pushint(st, false);
 			return SCRIPT_CMD_FAILURE;
 		}
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -27016,6 +27016,37 @@ BUILDIN_FUNC(has_autoloot) {
 	return SCRIPT_CMD_SUCCESS;
 }
 
+// ===================================
+// *autoloot(<rate>{, <char_id>});
+// This command sets the rate of autoloot (10000 = 100%).
+// Returns true on success and false on failure.
+// Note: It will toggle existing autoloot settings if rate value aren't provided.
+// ===================================
+BUILDIN_FUNC(autoloot) {
+	map_session_data *sd;
+
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, false);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	int rate = 0;
+	if (script_hasdata(st, 2))
+		rate = script_getnum(st, 2);
+	else if (sd->state.autoloot <= 0)
+		rate = 10000;
+
+	if (rate < 0 || rate > 10000) {
+		ShowWarning("buildin_autoloot: invalid rate value %d, shall be between 0 ~ 10000.\n", rate);
+		rate = cap_value(rate, 0, 10000);
+	}
+
+	sd->state.autoloot = rate;
+	script_pushint(st, true);
+
+	return SCRIPT_CMD_SUCCESS;
+}
+
 BUILDIN_FUNC(opentips){
 #if PACKETVER < 20171122
 	ShowError( "buildin_opentips: This command requires PACKETVER 20171122 or newer.\n" );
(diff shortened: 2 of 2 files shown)


doc/script_commands.txt
```diff
@@ -11375,3 +11375,10 @@ Returns current autoloot value on success.
 
 ---------------------------------------
 
+*autoloot(<rate>{, <char_id>});
```
Lemongrass3110, line 11378:
```suggestion
*autoloot({<rate>{, <char_id>}});
```

src/map/script.cpp
```diff
@@ -26957,6 +26957,37 @@ BUILDIN_FUNC(has_autoloot) {
 	return SCRIPT_CMD_SUCCESS;
 }
 
+// ===================================
+// *autoloot(<rate>{, <char_id>});
```
Lemongrass3110, line 26961:
```suggestion
// *autoloot({<rate>{, <char_id>}});
```

src/map/script.cpp
```diff
@@ -26957,6 +26957,37 @@ BUILDIN_FUNC(has_autoloot) {
...
+// This command sets the rate of autoloot (10000 = 100%).
+// Returns true on success and false on failure.
+// Note: It will toggle existing autoloot settings if rate value aren't provided.
+// ===================================
+BUILDIN_FUNC(autoloot) {
+	map_session_data *sd;
```
Lemongrass3110, line 26967:
```suggestion
	map_session_data *sd = nullptr;
```

src/map/script.cpp
```diff
@@ -26957,6 +26957,37 @@ BUILDIN_FUNC(has_autoloot) {
...
+
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, false);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	int rate = 0;
```
Lemongrass3110, line 26974:
```suggestion
	int rate = 10000;
```

src/map/script.cpp
```diff
@@ -26957,6 +26957,37 @@ BUILDIN_FUNC(has_autoloot) {
...
+	if (script_hasdata(st, 2))
+		rate = script_getnum(st, 2);
+	else if (sd->state.autoloot <= 0)
+		rate = 10000;
```
Lemongrass3110, line 26978:
```suggestion
```

src/map/script.cpp
```diff
@@ -26957,6 +26957,37 @@ BUILDIN_FUNC(has_autoloot) {
...
+
+	if (rate < 0 || rate > 10000) {
+		ShowWarning("buildin_autoloot: invalid rate value %d, shall be between 0 ~ 10000.\n", rate);
+		rate = cap_value(rate, 0, 10000);
```
Lemongrass3110, line 26982:
```suggestion
		script_pushint(st, false);
		return SCRIPT_CMD_FAILURE;
```

Lemongrass3110 (approved):

src/map/script.cpp
Emistry: i think this way it wont be able to toggle to off if user already set autoloot before this script command executed.

src/map/script.cpp
Emistry: It was meant to turn off the autoloot if user already autoloot turned on.

src/map/script.cpp
Emistry: In my opinion, if it doesn't cause major issues, it should be safe to continue execute the script. Capped the value, but the result are no different than what the user initially wanted, so it just throw a warning and cap it and proceed.

Lemongrass3110: This way the script being executed (path) gets reported to the user.
Otherwise you only get the warning message and do not know which script causes it.

And they should fix their scripts and not rely on being able to code badly.

src/map/script.cpp
Lemongrass3110: I think what you really mean is the exact opposite. This means if it was turned off, then it should be turned on.

Emistry: yeah
if existing setting is off, then it should turn on (100%)
if existing setting is on, then it should turn off (0%)

```
autoloot(); // toggle on/off depend on existing autoloot
autoloot(0); // 0% / off
autoloot(100); // 1%
autoloot(5000); // 50%
autoloot(10000); // 100%
```

src/map/script.cpp
Emistry: > This way the script being executed (path) gets reported to the user.
Otherwise you only get the warning message and do not know which script causes it.

I see, that's a good point.

src/map/script.cpp
Atemo: How about
```suggestion
	else
		rate = (sd->state.autoloot > 0 ? 0 : 10000);      // toggle on/off depend on existing autoloot
```

Lemongrass3110: @Atemo i dont mind this, but it should match the documentation.
Either the default value is 100% or the default value is 100% (if currently off) or 0% (if currently on).

Emistry: updated

doc/script_commands.txt
```diff
@@ -11375,3 +11375,17 @@ Returns current autoloot value on success.
...
+
+This command sets the rate of autoloot (default: 10000 = 100%).
+Returns true on success and false on failure.
+Note: It will toggle existing autoloot settings if rate value aren't provided.
```
Lemongrass3110, line 11382:
```suggestion
This command sets the rate of autoloot.
If no rate is provided and the user has autoloot disabled it will default to 10000 = 100% (enabled) or
if the user has autoloot enabled it will default to 0 = 0% (disabled).
Returns true on success and false on failure.
```

Lemongrass3110 (changes requested):
With your force pushing it seems that the previous fixes have been reverted again.

See: https://github.com/rathena/rathena/compare/238906668b56ac6a659e4d7986da823e7770a997..b59317c4f0636296d7ab4b1ad0fa9898a1690693

src/map/script.cpp
```diff
@@ -26957,6 +26957,35 @@ BUILDIN_FUNC(has_autoloot) {
...
+// *autoloot(<rate>{, <char_id>});
+// This command sets the rate of autoloot (10000 = 100%).
+// Returns true on success and false on failure.
+// Note: It will toggle existing autoloot settings if rate value aren't provided.
+// ===================================
+BUILDIN_FUNC(autoloot) {
+	map_session_data *sd;
+
+	if (!script_charid2sd(3, sd)) {
+		script_pushint(st, false);
```
Atemo, line 26970: The script shouldn't continue when the command does not work as expected.

src/map/script.cpp
```diff
@@ -27016,6 +27016,37 @@ BUILDIN_FUNC(has_autoloot) {
...
+
+	int rate = (sd->state.autoloot > 0 ? 0 : 10000);
+	if (script_hasdata(st, 2))
+		rate = script_getnum(st, 2);
+
+	if (rate < 0 || rate > 10000) {
+		ShowWarning("buildin_autoloot: invalid rate value %d, shall be between 0 ~ 10000.\n", rate);		
+		script_pushint(st, false);
+		return SCRIPT_CMD_FAILURE;
+	}
```
Lemongrass3110, line 27042:
```suggestion
	int rate;

	if (script_hasdata(st, 2)) {
		rate = script_getnum(st, 2);

		if (rate < 0 || rate > 10000) {
			ShowWarning("buildin_autoloot: invalid rate value %d, shall be between 0 ~ 10000.\n", rate);		
			script_pushint(st, false);
			return SCRIPT_CMD_FAILURE;
		}
	} else {
		rate = (sd->state.autoloot > 0 ? 0 : 10000);
	}
```

Lemongrass3110 (approved):

src/map/script.cpp
```diff
@@ -27016,6 +27016,41 @@ BUILDIN_FUNC(has_autoloot) {
...
+
+	int rate;
+
+	if (script_hasdata(st, 2)) {
+		rate = script_getnum(st, 2);
+
+		if (rate < 0 || rate > 10000) {
+			ShowWarning("buildin_autoloot: invalid rate value %d, shall be between 0 ~ 10000.\n", rate);
```
aleos89, line 27040:
```suggestion
			ShowWarning("buildin_autoloot: Invalid rate value %d, should be between 0 ~ 10000.\n", rate);		
```

aleos89 (approved):

PR #8040, opened 2023-12-20
Refactored script command rand

Lemongrass3110:
* **Addressed Issue(s)**: #8041

* **Server Mode**: Both

Added support for int64, so the range of the random numbers that can be generated is much wider now.

Changed the logic to be more restrictive:
	(1) The range version is intended for positive numbers only.
	~(2) Removed automatic swapping of minimum and maximum, if the minimum is bigger than the maximum.~
	(3) If minimum and maximum are equal, no random value can be calculated.
Triggering the more restrictive behavior will trigger the script source (file) to be reported in the map-server.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -5567,7 +5567,7 @@ BUILDIN_FUNC(return)
 }
 
 /// Returns a random number.
-/// rand(<range>) -> <int64> in the mathematical range [0, <range>[
+/// rand(<range>) -> <int64> in the mathematical range [0, <range - 1>]
 /// rand(<min>,<max>) -> <int64> in the mathematical range [<min>, <max>]
 BUILDIN_FUNC(rand)
 {
@@ -5578,6 +5578,11 @@ BUILDIN_FUNC(rand)
 	if( script_hasdata( st, 3 ) ){
 		minimum = script_getnum64( st, 2 );
 		maximum = script_getnum64( st, 3 );
+
+		if( minimum > maximum ){
+			ShowWarning( "buildin_rand: minimum (%" PRId64 ") is bigger than maximum (%" PRId64 ").\n", minimum, maximum );
+			// rnd_value already fixes this by swapping minimum and maximum automatically
+		}
 	// range
 	}else{
 		minimum = 0;
@@ -5591,12 +5596,12 @@ BUILDIN_FUNC(rand)
 
 		// The range version is exclusive maximum
 		maximum -= 1;
-	}
 
-	if( minimum > maximum ){
-		ShowError( "buildin_rand: minimum (%" PRId64 ") is bigger than maximum (%" PRId64 ").\n", minimum, maximum );
-		st->state = END;
-		return SCRIPT_CMD_FAILURE;
+		if( maximum < 1 ){
+			ShowError( "buildin_rand: range (%" PRId64 ") is too small. No randomness possible.\n", maximum );
+			st->state = END;
+			return SCRIPT_CMD_FAILURE;
+		}
 	}
 
 	if( minimum == maximum ){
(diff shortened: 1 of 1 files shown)


src/map/script.cpp
```diff
@@ -5566,33 +5566,43 @@ BUILDIN_FUNC(return)
...
+		minimum = script_getnum64( st, 2 );
+		maximum = script_getnum64( st, 3 );
+	// range
+	}else{
+		minimum = 0;
+		maximum = script_getnum64( st, 2 );
+
+		if( maximum < 0 ){
+			ShowError( "buildin_rand: range (%" PRId64 ") is negative.\n", maximum );
+			return SCRIPT_CMD_FAILURE;
```
Atemo, line 5588: Push a value or stop the script (`SCRIPT_CMD_FAILURE` only report error, the script continue)
```suggestion
			st->state = END;
			return SCRIPT_CMD_FAILURE;
```

src/map/script.cpp
```diff
@@ -5566,33 +5566,43 @@ BUILDIN_FUNC(return)
...
+		maximum -= 1;
 	}
-	else
-	{// range
-		min = 0;
-		range = script_getnum( st, 2 ) - 1;
+
+	if( minimum > maximum ){
+		ShowError( "buildin_rand: minimum (%" PRId64 ") is bigger than maximum (%" PRId64 ").\n", minimum, maximum );
+		return SCRIPT_CMD_FAILURE;
```
Atemo, line 5597:
```suggestion
		st->state = END;
		return SCRIPT_CMD_FAILURE;
```

src/map/script.cpp
```diff
@@ -5566,33 +5566,43 @@ BUILDIN_FUNC(return)
...
 	}
-	if( range <= 1 )
-		script_pushint(st, min);
-	else
-		script_pushint( st, rnd_value( min, range ) );
+
+	if( minimum == maximum ){
+		ShowError( "buildin_rand: minimum (%" PRId64 ") and maximum (%" PRId64 ") are equal. No randomness possible.\n", minimum, maximum );
+		return SCRIPT_CMD_FAILURE;
```
Atemo, line 5602:
```suggestion
		st->state = END;
		return SCRIPT_CMD_FAILURE;
```

Atemo (commented):
Possible to keep the old behavior?
```
(2) Removed automatic swapping of minimum and maximum, if the minimum is bigger than the maximum.
(3) If minimum and maximum are equal, no random value can be calculated.
```
or to remove the warning for (3)
Many official and custom scripts will break (and some are a pain to check)

src/map/script.cpp
```diff
@@ -5566,33 +5566,46 @@ BUILDIN_FUNC(return)
 	return SCRIPT_CMD_SUCCESS;
 }
 
-/// Returns a random number from 0 to <range>-1.
-/// Or returns a random number from <min> to <max>.
-/// If <min> is greater than <max>, their numbers are switched.
-/// rand(<range>) -> <int>
-/// rand(<min>,<max>) -> <int>
+/// Returns a random number.
+/// rand(<range>) -> <int64> in the mathematical range [0, <range>[
```
aleos89, line 5570:
```suggestion
/// rand(<range>) -> <int64> in the mathematical range [0, <range>]
```

aleos89 (approved):

src/map/script.cpp
Lemongrass3110: Its exclusive <range>.
If we want to write it with inclusive definition we would have to put <range - 1>.

vstumpf: Then it'd be `[0-range)`
Unless there's a regional difference

Issue #8043, opened 2023-12-23
Can not login anymore (invalid password) since loginclif.cpp


LasherasGH:
### rAthena Hash

7c153416bc2b69bbb515c0a3e7a90bcb216d5e4a

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

Login gives always invalid password since the last changes at loginclif.cpp (rollback or update it with new rnd_value function works fine)

loginclif.cpp Line: 380:
`sd->md5keylen = (uint16)(12 + rnd() % 4);`
`sd->md5keylen = sizeof( sd->md5key );`

### Expected Result

Login correctly to the game

### How to Reproduce

Try to login with 2022-04-06 client to local server (tried with passwordencrypt/passwordencrypt2 and without it)

### Official Information

Nothing from KRO

LasherasGH:
Hello,
i'm tested more it, and it feels like the md5keylen needs to be less or equal 19 to works fine, i don't know the all behind behavior, but feels like client don't hash it correctly with more than 19 size, i'm not sure, but maybe it needs to end with 0x00?, so, thats why maybe memset to \0 still need for it.

Anyways, i'm working fine with it at this moment, but i don't know the all behind behavior 😄 



```
static int logclif_parse_reqkey(int fd, struct login_session_data *sd){
	RFIFOSKIP(fd,2);
	{
		memset(sd->md5key, '\0', sizeof(sd->md5key));
		sd->md5keylen = (uint16)(12 + rnd() % 4);
		MD5_Salt(sd->md5keylen, sd->md5key);

		WFIFOHEAD(fd,4 + sd->md5keylen);
		WFIFOW(fd,0) = 0x01dc;
		WFIFOW(fd,2) = 4 + sd->md5keylen;
		memcpy(WFIFOP(fd,4), sd->md5key, sd->md5keylen);
		WFIFOSET(fd,WFIFOW(fd,2));
	}
	return 1;
}
```



Greetings.

PR #8052, opened 2023-12-26
Fixed login_check_password

Lemongrass3110:
* **Addressed Issue(s)**: #8043

* **Server Mode**: Both

The server did not properly take the salt length into account and the array for combination of password and salt was never properly initialized. Rewrote the function and removed login_check_encrypted.

Thanks to @LasherasGH

### DIFF
--- a/src/common/md5calc.cpp
+++ b/src/common/md5calc.cpp
@@ -235,10 +235,11 @@ void MD5_String(const char * string, char * output)
 }
 
 /** output is a sequence of non-zero characters to be used as password salt. */
-void MD5_Salt(unsigned int len, char * output)
-{
-	unsigned int i;
-	for( i = 0; i < len; ++i )
-		output[i] = (char)(1 + rnd() % 255);
+void MD5_Salt( size_t len, char* output ){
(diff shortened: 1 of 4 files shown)


vstumpf:
I'm not sure this is the issue, before 7c153416bc2b69bbb515c0a3e7a90bcb216d5e4a we didn't null-terminate the string.
It was just a char buffer of random data with length between 12 and 15. 
https://github.com/rathena/rathena/blob/ad3a2e99eb76f883f02376211c15dd9f87c0fe95/src/login/loginclif.cpp#L380-L388

LasherasGH:
i'm not sure about networking client but... i only wanna clarify, it always had 0x00 because never fill up more than 15 char, the other remainder it's cleared by the memset to 0x00, also like i said, in my tests, when i filled the last char (20) always got issues with password, with 19, fine atleast with 2022-04-06. 

Greetings.

vstumpf:
Im guessing the client does something like
``` 
char md5key[20] = {0};
memcpy(md5key, packet->buff, packet->buff_len);
```
so when we send a 20 len key, it doesn't null-terminate, the client overruns the buffer

vstumpf (approved):

Issue #8018, opened 2023-11-23
Fourth class jobs improvement bundle 2022.12.21


Atemo:
### rAthena Hash

https://github.com/rathena/rathena/commit/8740574ae349e516c5d3914bfaf13db13539327e

### Client Date

N/A

### Server Mode

Renewal

### Result

The changes of the Fourth class jobs improvement bundle have not been implemented.

### Expected Result

Updated skill behavior

- [x] Shadow cross
- [x] Windhawk
- [x] Meister
- [x] Cardinal
- [x] Inquisitor
- [x] Biolo (done when the new skills are implemented,  https://github.com/rathena/rathena/pull/7987)
- [x] Abyss chaser
- [x] Elemental master (no change)
- [x] Archmage
- [x] Dragon knight
- [x] Imperial guard
- [x] Troubadour & Trouvere
- [x] Shinkiro & Shiranui
- [x] Night Watch
- [x] Sky Emperor
- [x] Soul Ascetic
- [x] Hyper Novice
- [x] Spirit Handler

### How to Reproduce

Use the skills and compare them with the patch notes.

### Official Information

https://www.divine-pride.net/forum/index.php?/topic/3723-kro-jobs-improvement-project/page/14/
https://ro.gnjoy.com/news/devnote/View.asp?category=1&seq=4157430&curpage=1
https://ro.gnjoy.com/news/devnote/View.asp?category=1&seq=4159311&curpage=1
https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7670&curpage=1

PR #8068, opened 2023-12-30
Rebalance Troubadour & Trouvere 20221221

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8018

* **Server Mode**: renewal

Troubadour & Trouvere
---------------------------------

1. [Rhythm Shooting](https://www.divine-pride.net/database/skill/5355)
	- Increases cooldown from 0.15 seconds to 0.35 seconds.
	- Increases base damage from 800%/1400%Atk to 3700%/4550%Atk per hit based on level 5.
	- Increases factor weight of CON in skill formula from 3 to 5 (7 on targets marked with sound brand).

2. [Metallic Fury](https://www.divine-pride.net/database/skill/5356)
	- Reworks skill to dealing damage to surrounding enemies around the target (9 x 9 cells based on level 5).
	- Increases cooldown from 0.3 seconds to 0.4 seconds.
	- Increases global cool time from 0.3 seconds to 0.5 seconds.
	- No longer remove sound brand debuff from target. (doesn't mention it changelog but it works)
	- Increases base damage from 11000%Matk to 13000%/18000%(targets marked with sound brand)Matk based on level 5.
	- Increases factor weight of SPL in skill formula (on targets marked with sound brand) from 1.5x of Stage Manner skill level to 2x of Stage Manner skill level. 

Additionally
-----------------------------
Rhythm Shooting
- No longer remove sound brand debuff from target. (from skill description)

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -8039,12 +8039,13 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 						skillratio += -100 + ( 570 + 20 * pc_checkskill( sd, ABC_MAGIC_SWORD_M ) ) * skill_lv + 5 * sstatus->spl;
 						RE_LVL_DMOD(100);
 						break;
-					case TR_METALIC_FURY: {
-							int area = skill_get_splash(skill_id, skill_lv);
-							int count = map_forcountinarea(skill_check_bl_sc,target->m,target->x - area,target->y - area,target->x + area,target->y + area,5,BL_MOB,SC_SOUNDBLEND);
-							skillratio += -100 + (2200 + 300 * count) * skill_lv + 5 * sstatus->spl;
-							RE_LVL_DMOD(100);
-						}
+					case TR_METALIC_FURY:
+						skillratio += -100 + 2600 * skill_lv;
+						if (tsc && tsc->getSCE(SC_SOUNDBLEND))
+							skillratio += 1000 * skill_lv;
+						if (sd)
+							skillratio += 2 * pc_checkskill(sd, TR_STAGE_MANNER) * sstatus->spl;
+						RE_LVL_DMOD(100);
 						break;
 					case TR_SOUNDBLEND:
 						skillratio += -100 + 120 * skill_lv + 5 * sstatus->spl;
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -8047,8 +8047,7 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 						// !Todo: skill affected by SPL (without SC_SOUNDBLEND) as well?
 						if (tsc && tsc->getSCE(SC_SOUNDBLEND)) {
 							skillratio += 1000 * skill_lv;
-							if (sd)
-								skillratio += 2 * pc_checkskill(sd, TR_STAGE_MANNER) * sstatus->spl;
+							skillratio += 2 * pc_checkskill(sd, TR_STAGE_MANNER) * sstatus->spl;
 						}
 						RE_LVL_DMOD(100);
 						break;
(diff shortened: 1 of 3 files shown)


src/map/battle.cpp
```diff
@@ -5734,13 +5734,15 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
...
+			skillratio += -100 + 450 + 650 * skill_lv;
 
 			if (sd && pc_checkskill(sd, TR_STAGE_MANNER) > 0)
-				skillratio += 3 * sstatus->con;
+				skillratio += 5 * sstatus->con;
 
-			if (tsc && tsc->getSCE(SC_SOUNDBLEND))
-				skillratio += 100 + 100 * skill_lv;
+			if (tsc && tsc->getSCE(SC_SOUNDBLEND)) {
+				skillratio += 350 + 100 * skill_lv;
```
Lemongrass3110, line 5743: Not sure if this is a typo in the skill description, but level 4 does not match the formula.
Pretty sure that they have a typo though.

src/map/skill.cpp
```diff
@@ -2183,8 +2183,6 @@ int skill_additional_effect( struct block_list* src, struct block_list *bl, uint
 		[[fallthrough]];
 	case WM_METALICSOUND:
 	case WM_REVERBERATION:
-	case TR_RHYTHMSHOOTING:
```
Lemongrass3110, line 2186: Not mentioned in the changelog?
If you are sure about this, just ignore this comment.

src/map/battle.cpp
```diff
@@ -8037,12 +8039,13 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
...
-							int count = map_forcountinarea(skill_check_bl_sc,target->m,target->x - area,target->y - area,target->x + area,target->y + area,5,BL_MOB,SC_SOUNDBLEND);
-							skillratio += -100 + (2200 + 300 * count) * skill_lv + 5 * sstatus->spl;
-							RE_LVL_DMOD(100);
-						}
+					case TR_METALIC_FURY:
+						skillratio += -100 + 2600 * skill_lv;
+						if (tsc && tsc->getSCE(SC_SOUNDBLEND))
+							skillratio += 1000 * skill_lv;
+						if (sd)
+							skillratio += 2 * pc_checkskill(sd, TR_STAGE_MANNER) * sstatus->spl;
```
Lemongrass3110, line 8047:
```suggestion
						if (tsc && tsc->getSCE(SC_SOUNDBLEND)){
							skillratio += 1000 * skill_lv;
							skillratio += 2 * pc_checkskill(sd, TR_STAGE_MANNER) * sstatus->spl;
						}
```
The description (of Sigma) sounds like this should only apply to targets that are marked with SC_SOUNDBLEND.

Lemongrass3110 (approved):

src/map/battle.cpp
datawulf: Before it should have been 1.5 x SPL against any target. Now I read it as 1.5 x SPL without SC_SOUNDBLEND and 2 x SPL with SC_SOUNDBLEND

src/map/battle.cpp
Atemo: Possible. I guess we should follow the official value until we have more information.

src/map/skill.cpp
Atemo: This is not mentioned in the official changelog (like `TR_METALIC_FURY`). But the skill description has been changed as well

src/map/battle.cpp
Atemo: Possible, the part without SC_SOUNDBLEND is not clear.

PR #8101, opened 2024-01-12
Initial implementation of Hyper Novice

Lemongrass3110:

* **Server Mode**: Renewal

Taken from #7024

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -1844,31 +1844,31 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 	if (sc && sc->count) {
 		if (sc->getSCE(SC_BREAKINGLIMIT)) {
 			switch (skill_id) {
-			case HN_SHIELD_CHAIN_RUSH:
-			case HN_DOUBLEBOWLINGBASH:
-				damage += damage / 2; 
-				break;
-			case HN_MEGA_SONIC_BLOW:
-			case HN_SPIRAL_PIERCE_MAX:
-				damage += damage * 7 / 10;
-				break;
+				case HN_SHIELD_CHAIN_RUSH:
+				case HN_DOUBLEBOWLINGBASH:
+					damage += damage / 2; 
+					break;
+				case HN_MEGA_SONIC_BLOW:
+				case HN_SPIRAL_PIERCE_MAX:
+					damage += damage * 70 / 100;
+					break;
 			}
 		}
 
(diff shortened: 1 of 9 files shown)


src/map/battle.cpp
```diff
@@ -1569,6 +1575,11 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
 			damage *= 4;
 #endif
 
+		if (tsc->getSCE(SC_GROUNDGRAVITY) && flag&(BF_MAGIC|BF_WEAPON) && !status_bl_has_mode(bl, MD_STATUSIMMUNE))
```
Lemongrass3110, line 1578:
```suggestion
		if (tsc->getSCE(SC_GROUNDGRAVITY) && flag&(BF_MAGIC|BF_WEAPON))
```
Do we really need to check this again? It cannot be started on bosses anyway.

src/map/battle.cpp
```diff
@@ -1569,6 +1575,11 @@ int64 battle_calc_damage(struct block_list *src,struct block_list *bl,struct Dam
...
+			damage += damage * 15 / 100;
+		if (tsc->getSCE(SC_SHIELDCHAINRUSH) && !status_bl_has_mode(bl, MD_STATUSIMMUNE))
```
Lemongrass3110, line 1580:
```suggestion
		if (tsc->getSCE(SC_SHIELDCHAINRUSH))
```
Do we really need to check this again? It cannot be started on bosses anyway.

src/map/skill.cpp
```diff
@@ -12792,6 +12862,11 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		}
 		break;
 
+	case HN_HELLS_DRIVE:
```
Lemongrass3110, line 12865: Should this really be in skill_castend_nodamage_id?
Or should it be moved to skill_castend_damage_id?

aleos89: Agreed, this can be moved to the generic splash skill list!

src/map/battle.cpp
aleos89: Agreed, the status has `BossResist` so it should not have applied to a boss type monster.

src/map/battle.cpp
aleos89: Agreed, the status has `BossResist` so it should not have applied to a boss type monster.

PR #8116, opened 2024-01-26
Extended TimeLimit format (quest DB)

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

Extended `TimeLimit` format to support expiration time on a given day of the week.

Example
------------------------
```
  - Id: 9419
    Title: Attack Sky Fortress Invading Prontera
    # The quest expires 3 days after being taken at 4am.
    TimeLimit: 3d 4h
  - Id: 5965
    Title: "[Standby] Devil's Special"
    # The quest expires every Monday at 4am.
    TimeLimit: Monday 4h
```

Note
------------------------
`TimeLimit` should be reworked to be more user friendly - but this is not the meaning of this PR.

### DIFF
--- a/doc/quest_db.txt
+++ b/doc/quest_db.txt
@@ -42,7 +42,7 @@ Examples:
     TimeLimit: 3d 4h
   - Id: 5965
     Title: "[Standby] Devil's Special"
-    # The quest expires every Monday at 4am.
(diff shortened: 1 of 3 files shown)


aleos89 (approved):
Agreed, this concept could use a revitalization since new formats keep being introduced!

PR #8119, opened 2024-01-29
Multithreaded loading of YAML databases

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**:  Both

This small modification made reading YAML databases approximately 5% faster, I don't know if it will make any difference on extremely powerful computers, maybe difference can be greater, but on my humble phenom it made a considerable difference

my benchmark:
in ssd:
without async
image

with async
image

in harddisk
without async
image

with async
image

debug mode without async:
image

debug mode with async:
image

undefine loading output debug  without async:
image

undefine loading output debug  with async:
image

If you test it, let me know your times

accepting suggestions


AoShinRO:
results could be better, but are limited by rathena's current memory management

aleos89:
While I can appreciate having some aspects of the emulator support multi threading, I'm not sure if the boot process is that necessary. As with compiling in debug mode vs release, running in debug mode will take longer. There's a lot of content that is being parsed so it makes sense that over the years load time would increase for renewal mode.

Have you tried disabling [this define](https://github.com/rathena/rathena/blob/c43f7b369e797e8f765fef89d082eed1a3eb1b24/src/config/core.hpp#L109)? This should drastically improve the load time as it won't spit out all the x/y parsing messages.

AoShinRO:
> While I can appreciate having some aspects of the emulator support multi threading, I'm not sure if the boot process is that necessary. As with compiling in debug mode vs release, running in debug mode will take longer. There's a lot of content that is being parsed so it makes sense that over the years load time would increase for renewal mode.
> 
> Have you tried disabling [this define](https://github.com/rathena/rathena/blob/c43f7b369e797e8f765fef89d082eed1a3eb1b24/src/config/core.hpp#L109)? This should drastically improve the load time in debug mode as it won't spit out all the x/y parsing messages.

This DETAILED READING OUTPUT option does not reduce boot time outside of debug mode, this change is here for testing purposes only as to implement it to its maximum potential it would be necessary to change the memory management cap of the emulator, as I don't understand much well about that I won't even try, but maybe someone will find a solution in the future and we can really take advantage of it without gaps between the heavier databases, the reduction was around 3 seconds before I noticed the memory errors, giving time off to the processes needed to mitigate them by and lose boot time, and now the difference are really lower than i expected

PR #8127, opened 2024-02-09
Refactoring the A* walkpath algorithm to use c++ features instead of magic words

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**:  Both

This function comes from Aegis, right?


src/map/navi.cpp
```diff
@@ -203,7 +181,8 @@ bool navi_path_search(struct navi_walkpath_data *wpd, const struct navi_pos *fro
 	tp[i].flag = SET_OPEN;
 	tpused[i] = 1 + (from->x << 16 | from->y);
 
-	heap_push_node(&g_open_set, &tp[i]); // Put start node to 'open' set
+	g_open_set.push_back(&tp[i]);
```
vstumpf, line 184: we should use push_heap here too

src/map/navi.cpp
```diff
@@ -218,12 +197,16 @@ bool navi_path_search(struct navi_walkpath_data *wpd, const struct navi_pos *fro
...
 
-		current = BHEAP_PEEK(g_open_set); // Look for the lowest f_cost node in the 'open' set
-		BHEAP_POP2(g_open_set, NODE_MINTOPCMP); // Remove it from 'open' set
+		// Look for the lowest f_cost node in the 'open' set
+		auto it = std::min_element(g_open_set.begin(), g_open_set.end(), [](const path_node* a, const path_node* b) {
+			return a->f_cost < b->f_cost;
+		});
+
+		current = *it;
+		g_open_set.erase(it); // Remove it from 'open' set
```
vstumpf, line 209: here we can just std::pop_heap/vector.back/vector.pop_back

src/map/navi.cpp
```diff
@@ -118,9 +87,15 @@ static int add_path(struct node_heap *heap, int16 x, int16 y, int g_cost, struct
 			tp[i].parent = parent;
 			tp[i].f_cost = g_cost + h_cost;
 			if (tp[i].flag == SET_CLOSED) {
-				heap_push_node(heap, &tp[i]); // Put it in open set again
+				heap.push_back(&tp[i]);
+				std::push_heap(heap.begin(), heap.end(), [](const path_node* a, const path_node* b) {
+					return a->f_cost > b->f_cost;
```
vstumpf, line 92: in order to prevent rewriting the same comparator, we can create a std::function and use it everywhere
somehting like
```c++
auto heap_comp = [](const path_node* a, const path_node* b) {
    return a->f_cost > b->f_cost;
};
```

and then `std::push_heap(heap.begin(), heap.end(), heap_comp)`

src/map/navi.cpp
```diff
@@ -203,7 +181,8 @@ bool navi_path_search(struct navi_walkpath_data *wpd, const struct navi_pos *fro
 	tp[i].flag = SET_OPEN;
 	tpused[i] = 1 + (from->x << 16 | from->y);
 
-	heap_push_node(&g_open_set, &tp[i]); // Put start node to 'open' set
+	g_open_set.push_back(&tp[i]);
+	current = g_open_set.front();
```
vstumpf, line 185: do we need to set current? it should be done in the infinite loop below

vstumpf (changes requested):

src/map/navi.cpp
AoShinRO: just to escape warning from visual studio saying that I am referencing an uninitialized variable

src/map/navi.cpp
AoShinRO: this is really needed? g_open_set are cleared at this point, push_heap will change [0] to [0]?

src/map/navi.cpp
```diff
@@ -218,12 +197,16 @@ bool navi_path_search(struct navi_walkpath_data *wpd, const struct navi_pos *fro
...
+		if (g_open_set.size() == 0)
 			return false;
-		}
 
-		current = BHEAP_PEEK(g_open_set); // Look for the lowest f_cost node in the 'open' set
-		BHEAP_POP2(g_open_set, NODE_MINTOPCMP); // Remove it from 'open' set
+		// Look for the lowest f_cost node in the 'open' set
+		auto it = std::min_element(g_open_set.begin(), g_open_set.end(), [](const path_node* a, const path_node* b) {
+			return a->f_cost < b->f_cost;
+		});
```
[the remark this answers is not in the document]
vstumpf: It's something like
pop_heap() //moves to end
it = g_open_set.back()
g_open_set.pop_back()
current = *it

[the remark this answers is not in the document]
AoShinRO: worked, thank you
<3

src/map/navi.cpp
AoShinRO: can you help with replacing the similar macros in timer.cpp? I was unable to successfully apply it there, if it is changed there we can remove the BHEAP macros from the emulator, discord: aoshinho

src/map/navi.cpp
```diff
@@ -139,7 +112,8 @@ static int add_path(struct node_heap *heap, int16 x, int16 y, int g_cost, struct
 	tp[i].f_cost = g_cost + h_cost;
 	tp[i].flag = SET_OPEN;
 	tpused[i] = 1 + (x << 16 | y);
-	heap_push_node(heap, &tp[i]);
+	heap.push_back(&tp[i]);
+	std::push_heap(heap.begin(), heap.end(), max_heap_comp);
```
vstumpf, line 116: minheapcomp

src/map/navi.cpp
```diff
@@ -75,39 +72,13 @@ static enum directions walk_choices [3][3] =
 /// @name A* pathfinding related functions
 /// @{
 
-/// Pushes path_node to the binary node_heap.
-/// Ensures there is enough space in array to store new element.
-
-static void heap_push_node(struct node_heap *heap, struct path_node *node)
```
vstumpf, line 81: we could keep these functions and just put the repeated lines here:
```
static void heap_push_node(std::vector<path_node *>& heap, path_node *node) {
    heap.push_back(node);
	std::push_heap(heap.begin(), heap.end(), min_heap_comp);
}
```
then we just call heap_push_node(heap, &tp[i]);

src/map/navi.cpp
```diff
@@ -118,9 +89,11 @@ static int add_path(struct node_heap *heap, int16 x, int16 y, int g_cost, struct
 			tp[i].parent = parent;
 			tp[i].f_cost = g_cost + h_cost;
 			if (tp[i].flag == SET_CLOSED) {
-				heap_push_node(heap, &tp[i]); // Put it in open set again
+				heap.push_back(&tp[i]); // Put it in open set again
+				std::push_heap(heap.begin(), heap.end(), max_heap_comp);
 			}
-			else if (heap_update_node(heap, &tp[i])) {
+			else if (std::find(heap.begin(), heap.end(), &tp[i]) != heap.end()) {
+				std::make_heap(heap.begin(), heap.end(), max_heap_comp);
```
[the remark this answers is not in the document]
AoShinRO: I arrived at this in an attempt to reevaluate an existing node
/// Updates path_node in the binary node_heap.
static void heap_update_node(std::vector<path_node*>& heap, struct path_node* node, int g_cost)
{
	node->g_cost = node->parent->g_cost + heuristic(node->parent->x, node->parent->y, node->x, node->y); // Reavalia o custo g do nó
	node->f_cost = node->g_cost + g_cost; // Atualiza o custo f do nó
	heap.push_back(node);
	std::push_heap(heap.begin(), heap.end(), min_heap_comp);
}

src/map/navi.cpp
AoShinRO: idk why, but not work if i put min heap comp at addpath

src/map/navi.cpp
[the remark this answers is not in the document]
AoShinRO: i tried something to keep tracking of the nodes by indexing, tell me if you have an better solution

PR #8145, opened 2024-02-19
Fixed a Zeny variable overflow in mails

inhyositsu:
* **Addressed Issue(s)**: 
When sending a mail, if the attached Zeny is too high, calculating the handling fee will cause the total Zeny to become negative, resulting in situations where Zeny should be deducted turning into gaining Zeny.

* **Server Mode**: 
Both

Fix the Zeny fee calculation, and break the function if Zeny is negative.

### DIFF
--- a/src/map/mail.cpp
+++ b/src/map/mail.cpp
@@ -106,14 +106,28 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
 	if( sd->mail.zeny > 0 ){
 		//Zeny send
 		if( flag ){
-			// To avoid negative values, we need to calculate the fee before the subtraction
-			int Zeny = sd->mail.zeny + (int) ((double)sd->mail.zeny * battle_config.mail_zeny_fee / 100);
+			int64 fee;
 
-			if (Zeny < 0)
+			if( util::safe_multiplication( static_cast<decltype(fee)>( sd->mail.zeny ), static_cast<decltype(fee)>( battle_config.mail_zeny_fee ), fee ) ){
 				return false;
+			}
+
+			if( fee < 0 ){
+				return false;
+			}
+
+			fee /= 100;
+
+			if( fee > MAX_ZENY ){
+				return false;
+			}
 
 			// It's possible that we don't know what the dest_id is, so it will be 0
-			if (pc_payzeny(sd, Zeny, LOG_TYPE_MAIL, sd->mail.dest_id)) {
+			if( pc_payzeny( sd, static_cast<int32>( fee ), LOG_TYPE_MAIL, sd->mail.dest_id ) ){
+				return false;
+			}
+			// It's possible that we don't know what the dest_id is, so it will be 0
+			if( pc_payzeny( sd, static_cast<int32>( fee ), LOG_TYPE_MAIL, sd->mail.dest_id ) ){
 				return false;
 			}
 		}else{
--- a/src/map/mail.cpp
+++ b/src/map/mail.cpp
@@ -7,6 +7,7 @@
 #include <common/showmsg.hpp>
 #include <common/strlib.hpp>
 #include <common/timer.hpp>
+#include <common/utilities.hpp>
 
 #include "atcommand.hpp"
 #include "battle.hpp"
@@ -17,6 +18,8 @@
 #include "log.hpp"
 #include "pc.hpp"
 #include "pet.hpp"
+
+using namespace rathena;
 
 void mail_clear(map_session_data *sd)
 {
--- a/src/map/mail.cpp
+++ b/src/map/mail.cpp
@@ -106,8 +106,14 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
 	if( sd->mail.zeny > 0 ){
 		//Zeny send
 		if( flag ){
+			// To avoid negative values, we need to calculate the fee before the subtraction
+			int Zeny = sd->mail.zeny + (int) ((double)sd->mail.zeny * battle_config.mail_zeny_fee / 100);
+
+			if (Zeny < 0)
+				return false;
+
 			// It's possible that we don't know what the dest_id is, so it will be 0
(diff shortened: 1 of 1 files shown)


src/map/mail.cpp
```diff
@@ -106,8 +106,14 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
 	if( sd->mail.zeny > 0 ){
 		//Zeny send
 		if( flag ){
+			// To avoid negative values, we need to calculate the fee before the subtraction
+			int Zeny = sd->mail.zeny + (int) ((double)sd->mail.zeny * battle_config.mail_zeny_fee / 100);
```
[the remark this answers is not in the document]
inhyositsu: sure, it works.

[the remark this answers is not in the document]
iberryRO: sure, it works. confirm

src/map/mail.cpp
```diff
@@ -106,8 +106,14 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
...
+
+			if (Zeny < 0)
+				return false;
+
```
Lemongrass3110, line 114:
```suggestion
			int64 fee;

			if( util::safe_multiplication( static_cast<decltype(fee)>( sd->mail.zeny ), static_cast<decltype(fee)>( battle_config.mail_zeny_fee ), fee ) ){
				return false;
			}

			if( fee < 0 ){
				return false;
			}

			fee /= 100;

			if( fee > MAX_ZENY ){
				return false;
			}

			// It's possible that we don't know what the dest_id is, so it will be 0
			if( pc_payzeny( sd, static_cast<int32>( fee ), LOG_TYPE_MAIL, sd->mail.dest_id ) ){
				return false;
			}
```

src/map/mail.cpp
```diff
@@ -106,8 +106,14 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
...
 			// It's possible that we don't know what the dest_id is, so it will be 0
-			if (pc_payzeny(sd, sd->mail.zeny + sd->mail.zeny * battle_config.mail_zeny_fee / 100, LOG_TYPE_MAIL, sd->mail.dest_id)) {
+			if (pc_payzeny(sd, Zeny, LOG_TYPE_MAIL, sd->mail.dest_id)) {
```
Lemongrass3110, line 116:
```suggestion
			if( pc_payzeny( sd, static_cast<int32>( fee ), LOG_TYPE_MAIL, sd->mail.dest_id ) ){
```

Lemongrass3110 (changes requested):
In addition to my requested changes it needs the following:

At the end of the common headers:
```cpp
#include <common/utilities.hpp>
```

After the last include:
```cpp
using namespace rathena;
```

inhyositsu (commented):
thanks for the suggestion, but @Lemongrass3110 just push another suggestion, and I'll use that change.

inhyositsu (commented):
thanks for the suggestion, but @Lemongrass3110 just push another suggestion, and I'll use that change.

inhyositsu (commented):
Thanks for suggestion. I've added these code.

src/map/mail.cpp
```diff
@@ -106,8 +109,28 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
...
+
+			if( fee < 0 ){
+				return false;
+			}
+
+			fee /= 100;
+
+			if( fee > MAX_ZENY ){
+				return false;
+			}
```
Lemongrass3110, line 126:
```suggestion
			int64 zeny = sd->mail.zeny;

			if( battle_config.mail_zeny_fee > 0 ){
				int64 fee;

				if( util::safe_multiplication( zeny, static_cast<decltype(fee)>( battle_config.mail_zeny_fee ), fee ) ){
					return false;
				}

				if( fee < 0 ){
					return false;
				}

				fee /= 100;

				if( fee > MAX_ZENY ){
					return false;
				}

				if( util::safe_addition( zeny, fee, zeny ) ){
					return false;
				}

				if( zeny > MAX_ZENY ){
					return false;
				}
			}
```
Cleaned it up a little further. Also now summing up zeny+fee.

src/map/mail.cpp
```diff
@@ -106,8 +109,28 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
...
+
+			// It's possible that we don't know what the dest_id is, so it will be 0
+			if( pc_payzeny( sd, static_cast<int32>( fee ), LOG_TYPE_MAIL, sd->mail.dest_id ) ){
+				return false;
+			}
```
Lemongrass3110, line 131:
```suggestion
```
Somehow this check was duplicated?

src/map/mail.cpp
```diff
@@ -106,8 +109,28 @@ bool mail_removezeny( map_session_data *sd, bool flag ){
...
 			// It's possible that we don't know what the dest_id is, so it will be 0
-			if (pc_payzeny(sd, sd->mail.zeny + sd->mail.zeny * battle_config.mail_zeny_fee / 100, LOG_TYPE_MAIL, sd->mail.dest_id)) {
+			if( pc_payzeny( sd, static_cast<int32>( fee ), LOG_TYPE_MAIL, sd->mail.dest_id ) ){
```
Lemongrass3110, line 133:
```suggestion
			if( pc_payzeny( sd, static_cast<int32>( zeny ), LOG_TYPE_MAIL, sd->mail.dest_id ) ){
```

Lemongrass3110 (approved):

gidzdlcrz:
How do you reproduce the bug prior this commit?

inhyositsu:
1. Having a charactor which has 1.6 billion zeny.
2. Sending a mail with 1.5 billion zeny and the battle_config.mail_zeny_fee is 2 or more.
3. And the charactor sending the mail will increase zeny instead of giving it away, but the mail itself still have zeny with it, so the reciever can get zeny too.

PR #8149, opened 2024-02-26
Refactored safe math to templates

Lemongrass3110:

* **Server Mode**: Both

This way it can also be used for smaller numeric types.


Lemongrass3110:
Requires C++20 standard like this.
Closing.

PR #8151, opened 2024-02-27
Add an all-in-one devcontainer and make use of -j on production builder

Danil0v3s:
* **Addressed Issue(s)**: 

There was some confusion around the readme of the docker part, we did not remove the wording saying the containers are not suited for production when we split each server to its container.

* **Server Mode**: 
N/D

This PR aims to create two different folders of docker-compose files, mostly restore the old one and divide them into "production" and "development", being "production" the one with separate containers and the latter the one with an all-in-one container.


tools/docker/development/README.md
```diff
@@ -0,0 +1,29 @@
...
+### How to setup :computer:
+
+1. `docker-compose up -d` to spin up dev container and database (ensure port `3306` is free)
+2. `docker exec -it rathena-dev bash` to connect to dev container
+3. All rAthena development commands can be executed inside the dev container, such as compiling (`./configure`, `make clean server`) and starting the server (`./athena-start`, `gdb map-server`, etc ...)
+4. `docker-compose down` outside the dev container when done to close database and free resources
+
+#### Tips & tricks :beginner:
+
+- Ensure you don't have a database running locally and listening on port `3306` this will cause the database container to fail starting up.
```
aleos89, line 14:
```suggestion
- Ensure you don't have a database running locally and listening on port `3306` as this will cause the database container to fail to start up.
```

tools/docker/development/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+
+services:
+    db:
+        image: "mariadb:bionic"
+        container_name: "rathena-dev_db"
+        ports:
+            - "3306:3306" # allow DB connections from host
+        volumes:
+            - "rathenadb:/var/lib/mysql" # save database to local disk
+            - "../../sql-files/:/docker-entrypoint-initdb.d" # initialize db with ./sql-files
```
aleos89, line 11: Is this a little dangerous with the tools folder in sql-files? The conversion of engine types and the MD5 password setting (if the server config isn't setup for this password type).

Unless I'm hyping this up and it doesn't crawl to sub-directories, then ignore me!

Danil0v3s: It does map the directory to the container but i think the container only queries for sql files. The change I needed to do was because the dir is now a level below

Issue #8171, opened 2024-03-15
Asprika damage reduction not working as official


Playtester:
### rAthena Hash

4e64623

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

On rAthena, Asprika reduces only melee damage in both PVP and PVM.
It should actually reduce all damage in PVM and physical damage in PVP.

The root cause are two mistakes:

1. The command "bonus3 bSubEle,Ele_All,30,BF_SHORT;" should actually be "bonus3 bSubEle,Ele_All,30,BF_WEAPON;".
2. The battle flag should only apply to player damage, not to monster damage.

### Expected Result

Asprika reduces all monster damage by 30%. It does not matter whether melee, ranged, magic or physical.
Asprika reduces all **physical** player damage by 30%. Magic damage is never reduced. It does not matter whether melee or ranged.

### How to Reproduce

Test damage in PVP and PVM with and without Asprika.

### Official Information

Verified via official server tests.

Playtester:
Upon further investigation, it appears that all monster spells should be considered as "weapon" damage, even if the attack type is magic.

Need to find a better solution.

PR #8173, opened 2024-03-16
Fixed Pre-Renewal Asprika damage reduction (#8171)

Playtester:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8171

- Monster damage is now always reduced by items that reduce BF_WEAPON damage
- Monster damage is now never reduced by items that reduce BF_MAGIC damage
- Asprika now reduces all damage from monsters by 30%
- Asprika now reduces all physical damage from players by 30%
- Asprika now reduces damage regardless of range of the attack
- Shaman Hat will no longer reduce damage from monsters, but will still reduce magic damage from players
- Added a config to battle/items.conf for returning to old behavior

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -727,8 +727,7 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 					for (const auto &it : tsd->subele2) {
 						if (it.ele != ELE_ALL && it.ele != rh_ele)
 							continue;
-						if ((sd || !(it.flag&BF_PLAYER)) &&
-							!(((it.flag)&flag)&BF_WEAPONMASK &&
+						if (!(((it.flag)&flag)&BF_WEAPONMASK &&
 							((it.flag)&flag)&BF_RANGEMASK &&
 							((it.flag)&flag)&BF_SKILLMASK))
 							continue;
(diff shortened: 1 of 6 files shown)


src/map/battle.cpp
```diff
@@ -689,6 +689,12 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 	t_race2 = status_get_race2(target);
 	s_defele = (tsd) ? (enum e_element)status_get_element(src) : ELE_NONE;
 
+	// When the attacker is a monster, then all bonuses on BF_WEAPON will work and no bonuses on BF_MAGIC
+	if (src->type == BL_MOB) {
```
Lemongrass3110, line 693: Since this has been wrong forever: Shouldnt we have a config for it?

src/map/battle.cpp
```diff
@@ -689,6 +689,12 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
...
+		flag |= BF_WEAPON;
+		flag &= ~BF_MAGIC;
```
Lemongrass3110, line 695: I know you probably have a reasoning, but it would be nice to also know about why you only change the flag and not the attack_type itself (maybe document that in an inline comment).

Lemongrass3110 (approved):

src/map/battle.cpp
Playtester: Yeah, the reason I didn't finish this PR is that I was started to think about that too.

src/map/battle.cpp
Playtester: I just tried to be minimal invasive. We don't know how the Aegis code works, but I know it only impacts the damage modifiers.

aleos89 (approved):

PR #8182, opened 2024-03-23
Implemented the permission "TRADE_UNCONDITIONAL"

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: both

Suggestion to implement the permission `trade_unconditional` for `conf/groups.yml`.

Currently, the default group level in `item_db_*.yml` for `Trade` is 100, which means that by default no groups can bypass trade restrictions (currently the maximum group level is 99).

(this field) :
```
#   Trade:                  Trade restrictions. (Default: null)
#     Override              Group level to override these conditions. (Default: 100)
#     NoDrop                If the item can not be dropped. (Default: false)
#     NoTrade               If the item can not be traded. (Default: false)
#     TradePartner          If the item can not be traded to the player's partner. (Default: false)
#     NoSell                If the item can not be sold. (Default: false)
#     NoCart                If the item can not be put in a cart. (Default: false)
#     NoStorage             If the item can not be put in a storage. (Default: false)
#     NoGuildStorage        If the item can not be put in a guild storage. (Default: false)
#     NoMail                If the item can not be put in a mail. (Default: false)
#     NoAuction             If the item can not be put in an auction. (Default: false)
```

With the new permission, a group will be able to ignore the trade conditions (drop, trade, sell, cart, storage/gstorage, mail and auction).

### DIFF
--- a/doc/permissions.txt
+++ b/doc/permissions.txt
@@ -125,6 +125,12 @@ item delay, etc).
 
 ---------------------------------------
 
+*trade_unconditional
+
+Player ignore the trade conditions of items (drop, trade, sell, cart, storage/gstorage, mail and auction).
+
+---------------------------------------
+ 
 ======================
 | 3. Command-related |
 ======================
--- a/doc/permissions.txt
+++ b/doc/permissions.txt
@@ -127,7 +127,7 @@ item delay, etc).
 
 *trade_unconditional
 
-Player ignore the trade conditions of items (drop, trade, sell, cart, storage/gstorage, mail and auction).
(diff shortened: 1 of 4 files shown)


doc/permissions.txt
```diff
@@ -125,6 +125,12 @@ item delay, etc).
 
 ---------------------------------------
 
+*trade_unconditional
+
+Player ignore the trade conditions of items (drop, trade, sell, cart, storage/gstorage, mail and auction).
```
aleos89, line 130:
```suggestion
Allows player to ignore the trade conditions of items (drop, trade, sell, cart, storage/gstorage, mail and auction).
```

aleos89 (approved):

Issue #8187, opened 2024-03-25
Enchant Deadly Poison / Magnum Break Damage Bonus and Durations


Playtester:
### rAthena Hash

42bd87d

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

In pre-renewal, Enchant Deadly Poison is missing the additional 25% Poison damage effect (same as Magnum Break except it is 25% instead of 20% and POISON instead of FIRE).
In renewal, the additional damage is working, but the duration is off (using Duration2 which is reserved for DPOISON duration instead of using Duration1).

### Expected Result

In both pre-renewal and renewal Enchant Deadly Poison applies an additional 25% Poison damage effect for the duration of EDP (same effect as Magnum Break, but beware that the effect in rAthena is different for each).

### How to Reproduce

I found this by searching for occurrences of ASC_EDP in the source code.

### Official Information

https://ro.kokotewa.com/db/skl_info?id=ASC_EDP

Playtester:
```
		case SC_SUB_WEAPONPROPERTY:
			if (sd && val3 == ASC_EDP) {
				uint16 poison_level = pc_checkskill(sd, GC_RESEARCHNEWPOISON);

				if (poison_level > 0) {
					tick += 30000; // Base of 30 seconds
					tick += poison_level * 15 * 1000; // Additional 15 seconds per level
				}
			}
			break;
```
There's actually some cute that overwrites the duration.

This might need some more tests to verify.

Playtester:
```
  - Id: 378
    Name: ASC_EDP
    Description: Enchant Deadly Poison
    MaxLevel: 5
    Type: Weapon
    TargetType: Self
    DamageFlags:
      NoDamage: true
    Hit: Single
    HitCount: 1
    Element: Poison
    Duration1:
      - Level: 1
        Time: 40000
      - Level: 2
        Time: 45000
      - Level: 3
        Time: 50000
      - Level: 4
        Time: 55000
      - Level: 5
        Time: 120000
```
Duration for EDP in renewal is strangely defined I assume it should be 40000, 60000, 80000, 100000, 120000.

Playtester:
Poison chance should be:
3% - Level 1+2
4% - Level 3+4
5% - Level 5

Playtester:
Poison and DPoison both reduce HardDEF by 25% as well. But they also shouldn't stack (currently they stack).

Playtester:
Verifying other SCs that increase ATK:

True Sight
[...]

PR #8191, opened 2024-03-27
Enchant Deadly Poison / Magnum Break Damage Bonus and Durations

Playtester:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8187

* **Server Mode**: Pre-Renewal (Renewal only slightly affected)

Major pre-renewal damage handling rework on everything regarding the Magnum Break / EDP damage bonus based on in-depth testing (see linked issue).

Notable changes:
- Added missing EDP 25% poison damage bonus (it was already fixed in renewal but applies to pre-re too), that works identical to the 20% fire damage bonus from Magnum Break; both buffs overwrite each other
- Added "basedamage" and "basedamage2" to the damage object - this is the damage on which EDP/Magnum Break is based on; it now considers defense reduction, passive mastery, refine and true sight (there might be more interactions, will continue with this in a follow-up branch, don't want this to blow up too much)
- The EDP/Magnum Break damage bonus now considers spirit sphere damage, but the spirit sphere part is non-elemental; both are added together before they get rounded down
- The EDP/Magnum Break damage bonus can now become negative through attribute table and is dispelled on logout
- The EDP/Magnum Break damage bonus no longer applies when damage is already 0 or lower beforehand
- EDP is now applied after defense, refine and mastery
- Refine bonus now is applied before mastery and between both the damage is capped to 1
- Envenom's fixed damage bonus is now a mastery bonus (was already fixed in renewal but applies to pre-re too) and thus increased by EDP
- Fixed damage interaction between MO_FINGEROFFENSIVE (Throw Spirit Sphere) with Magnum Break, refine and spirit spheres
- Fixed Magnum Break costing HP (created a function that is used for both mercenaries and players, so you don't have to manually code it into their respective functions anymore; can easily add more skills if we find more that require HP/SP/AP but do not consume them)
- Fixed Sonic Blow's damage formula and how the bonus from Sonic Acceleration works
- Poison/DPoison now reduce hard DEF vs. monsters by 25%, but the DEF reduction no longer stacks
- Fixed DPoison's duration (60s base duration on all levels)
- Fixed Enchant Deadly Poison duration in Renewal
- Fixed chance of DPoison to be inflicted
- Fixed negative resistances not increasing the chance of a status change to occur (it's only capped to 0 in renewal)

All damage results in the linked issue that are not striked through are fully verified.
Will continue testing further interactions in the next branch.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -6486,16 +6486,12 @@ static void battle_calc_defense_reduction(struct Damage* wd, struct block_list *
  */
 static void battle_min_damage(struct Damage* wd, struct block_list* src, uint16 skill_id, int64 min) {
 	if (is_attack_right_handed(src, skill_id)) {
-		if (wd->damage < min)
-			wd->damage = min;
-		if (wd->basedamage < min)
-			wd->basedamage = min;
+		cap_value(wd->damage, min, INT64_MAX);
+		cap_value(wd->basedamage, min, INT64_MAX);
 	}
 	if (is_attack_left_handed(src, skill_id)) {
-		if (wd->damage2 < min)
-			wd->damage2 = min;
-		if (wd->basedamage2 < min)
-			wd->basedamage2 = min;
+		cap_value(wd->damage2, min, INT64_MAX);
+		cap_value(wd->basedamage2, min, INT64_MAX);
 	}
 }
 
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -421,10 +421,10 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
  * @param atk_elem
  * @param def_type
  * @param def_lv
- * @param flag
+ * @param flag 0x1 = allow to return negative values even if config for healing through negative resist is disabled
  * @return damage
  */
-int64 battle_attr_fix(struct block_list *src, struct block_list *target, int64 damage,int atk_elem,int def_type, int def_lv)
+int64 battle_attr_fix(struct block_list *src, struct block_list *target, int64 damage,int atk_elem,int def_type, int def_lv, int flag)
 {
 	status_change *sc = NULL, *tsc = NULL;
 	int ratio;
@@ -631,7 +631,7 @@ int64 battle_attr_fix(struct block_list *src, struct block_list *target, int64 d
 #endif
 	}
 
-	if (battle_config.attr_recover == 0 && ratio < 0)
+	if (battle_config.attr_recover == 0 && !(flag & 1) && ratio < 0)
 		ratio = 0;
 
 #ifdef RENEWAL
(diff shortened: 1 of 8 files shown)


src/map/battle.cpp
```diff
@@ -6474,12 +6491,42 @@ static void battle_calc_attack_post_defense(struct Damage* wd, struct block_list
...
+		ATK_ADD2(wd->damage, wd->damage2, sstatus->rhw.atk2, sstatus->lhw.atk2);
+		ATK_ADD2(wd->basedamage, wd->basedamage2, sstatus->rhw.atk2, sstatus->lhw.atk2);
+	}
+
+	//After DEF reduction, damage can be negative, refine bonus works against that value
+	//After refinement bonus was applied, damage is capped to 1, then masteries are applied
+	if (is_attack_right_handed(src, skill_id) && wd->damage < 1) wd->damage = 1;
+	if (is_attack_left_handed(src, skill_id) && wd->damage2 < 1) wd->damage2 = 1;
+	if (is_attack_right_handed(src, skill_id) && wd->basedamage < 1) wd->basedamage = 1;
+	if (is_attack_left_handed(src, skill_id) && wd->basedamage2 < 1) wd->basedamage2 = 1;
```
aleos89, line 6506:
```suggestion
	if (is_attack_right_handed(src, skill_id)) {
		if (wd->damage < 1)
			wd->damage = 1;
		if (wd->basedamage < 1)
			wd->basedamage = 1;
	}
	if (is_attack_left_handed(src, skill_id)) {
		if (wd->damage2 < 1)
			wd->damage2 = 1;
		if (wd->basedamage2 < 1)
			wd->basedamage2 = 1;
	}
```
Is it easier this way? We also only make singular calls to `is_attack_<left/right>_handed()`.

src/map/battle.cpp
```diff
@@ -6488,32 +6535,20 @@ static void battle_calc_attack_post_defense(struct Damage* wd, struct block_list
...
-		} else {
-			ATK_ADD2(wd->damage, wd->damage2, sstatus->rhw.atk2, sstatus->lhw.atk2);
-		}
-	}
-#endif
 	//Set to min of 1
 	if (is_attack_right_handed(src, skill_id) && wd->damage < 1) wd->damage = 1;
 	if (is_attack_left_handed(src, skill_id) && wd->damage2 < 1) wd->damage2 = 1;
+	if (is_attack_right_handed(src, skill_id) && wd->basedamage < 1) wd->basedamage = 1;
+	if (is_attack_left_handed(src, skill_id) && wd->basedamage2 < 1) wd->basedamage2 = 1;
```
aleos89, line 6542:
```suggestion
	if (is_attack_right_handed(src, skill_id)) {
		if (wd->damage < 1)
			wd->damage = 1;
		if (wd->basedamage < 1)
			wd->basedamage = 1;
	}
	if (is_attack_left_handed(src, skill_id)) {
		if (wd->damage2 < 1)
			wd->damage2 = 1;
		if (wd->basedamage2 < 1)
			wd->basedamage2 = 1;
	}
```
Same as before.

aleos89 (approved):

src/map/battle.cpp
Playtester: Stretches out the code a bit, maybe we should bundle this into a battle_min_damage function.

PR #8204, opened 2024-03-29
Cleaned up CZ_PARTY_CONFIG

Lemongrass3110:

* **Server Mode**: Both

Renamed the status variable from allow_party to refuse_partyinvite so it better shows what is really does.
Implemented a TODO.
Removed the unused function clif_partytickack.
Changed CZ_PARTY_CONFIG to struct.

### DIFF
--- a/src/common/mmo.hpp
+++ b/src/common/mmo.hpp
@@ -608,7 +608,7 @@ struct mmo_charstatus {
 	struct hotkey hotkeys[MAX_HOTKEYS_DB];
 #endif
 	bool show_equip, disable_call;
-	bool refuse_partyinvite;
+	bool disable_partyinvite;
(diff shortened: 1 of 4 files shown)


src/common/mmo.hpp
```diff
@@ -607,7 +607,8 @@ struct mmo_charstatus {
 #ifdef HOTKEY_SAVING
 	struct hotkey hotkeys[MAX_HOTKEYS_DB];
 #endif
-	bool show_equip,allow_party, disable_call;
+	bool show_equip, disable_call;
+	bool refuse_partyinvite;
```
aleos89, line 611: You define it as refuse but in clif call it disable.

aleos89 (approved):

PR #8210, opened 2024-03-31
ATKpercent, Base Damage, Occult Impaction, Freezing Trap, Acid Terror, Paladin skills

### LINKED ISSUE #8193: Pre-Renewal Damage Rework (Part 2): ATKPERCENT, SkillRatio
### rAthena Hash

2633599

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

To be verified

Note: I'm using this issue to document further official damage tests that are not included in https://github.com/rathena/rathena/pull/8191. This is to prevent a single PR from getting too big.

### Relevant Log Output

_No response_

### Expected Result

1.
True Sight, Spear Dynamo, Curse and Gospel use the same bonus on official servers, we should try to unify this in rAthena also. As True Sight was proven to impact base damage (relevant for Magnum Break bonus) we should check that this is applied correctly. From what I see except for True Sight, it directly goes into batk, so it's included automatically.
However, there is a linear damage bonus applied between base attack and true sight (right after the skill ratio - which base attack seems to ignore). We need to check if this kind of bonus should be applied before or after that.

2.
~We should investigate the Star Crumb and Spiritball bonus:~
```
		if (skill_id != CR_SHIELDBOOMERANG) //Only Shield boomerang doesn't takes the Star Crumbs bonus.
			ATK_ADD2(wd.damage, wd.damage2, ((wd.div_ < 1) ? 1 : wd.div_) * sd->right_weapon.star, ((wd.div_ < 1) ? 1 : wd.div_) * sd->left_weapon.star);
[...]
			ATK_ADD(wd.damage, wd.damage2, ((wd.div_ < 1) ? 1 : wd.div_) * sd->spiritball * 3);

```
~It getting multiplied with div here seems wrong. Also, need to test if it's included in the Magnum Break bonus or not (similar to Spirit Spheres).~

Moved into separate issue: https://github.com/rathena/rathena/issues/8211

3.
I verified damage with Sonic Blow and Throw Spirit Sphere and those did not include damage ratio in base damage. Should verify this again with a more simple skill that only has one hit like Bash.

### How to Reproduce

Deal damage on official servers and compare damage

### Official Information

"GetAttackFinalDamage seems to take the damage calced from teh skill ratio (adds the overthrust value to the ratio), subtracts the values as necessary towards def, then adds refine atk/mastery atk/star crumbs/spirit spheres, does element mult, does berserk mult"

"True sight attack bonus raises atkpercent, so that would apply to any skill that boosts that stat
Spear dynamo, provoke, battle chant, and curse"

(Source: mrj)

https://rathena.org/board/topic/112343-various-status-effects-off-from-officials-and-info-from-aegis/#comment-331275

_No response_

Playtester:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8193

* **Server Mode**: Both

- A handful of skills grant status changes that actually are calculated together into a value "ATKpercent" rather than increasing BATK/WATK, this value is then applied to most weapon skills
  * Pre-re - Damage after skill ratio but before DEF reduction is multiplied with ATKpercent
  * Renewal - Is a linear addition to skill ratio instead (ATKpercent of 200 means +100% to any skill ratio, so 100% becomes 200%, but 500% only becomes 600%)
- Fixed base damage not being calculated correctly for certain skills
- MO_INVESTIGATE is no longer influenced by Spirit Spheres except when you have the WATK_ELEMENT effect (Magnum Break)
- Freezing trap now deals 100% damage on all levels and can miss
- Acid Terror now ignores DEF, but its base attack is reduced by VIT
- Acid Terror skill ratio is now 50%+50%*level
- Acid Terror and Martyr's Reckoning (PA_SACRIFICE) are no longer affected by refine and mastery bonus
- Magnum Break/EDP bonus no longer affects Acid Terror and unit skills
- Fixed damage formula of Shield Boomerang and Rapid Smiting
- Code improvements
- Fixes #8193

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3364,7 +3364,7 @@ static bool battle_skill_stacks_masteries_vvs(uint16 skill_id)
 {
 	if (
 #ifndef RENEWAL
-		skill_id == PA_SHIELDCHAIN || skill_id == CR_SHIELDBOOMERANG ||
+		skill_id == PA_SHIELDCHAIN || skill_id == CR_SHIELDBOOMERANG || skill_id == AM_ACIDTERROR ||
 #endif
 		skill_id == MO_INVESTIGATE || skill_id == MO_EXTREMITYFIST ||
 		skill_id == RK_DRAGONBREATH || skill_id == RK_DRAGONBREATH_WATER || skill_id == NC_SELFDESTRUCTION ||
@@ -3658,7 +3658,9 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
 
 #ifndef RENEWAL
 		if (sd && (wd->damage > 0 || wd->damage2 > 0)) { // Applies only to player damage, monsters and mercenaries don't get this damage boost
-			if (sc && sc->getSCE(SC_WATK_ELEMENT)) { // Descriptions indicate this means adding a percent of a normal attack in another element [Skotlex]
+			// This adds a percentual damage bonus based on the damage you would deal with a normal attack
+			// Does not apply to unit skills or skills that have their own base damage formula such as AM_ACIDTERROR
+			if (sc && sc->getSCE(SC_WATK_ELEMENT) && !skill_get_unit_id(skill_id) && skill_id != AM_ACIDTERROR) {
 				int64 damage = wd->basedamage * sc->getSCE(SC_WATK_ELEMENT)->val2;
 				damage = battle_attr_fix(src, target, damage, sc->getSCE(SC_WATK_ELEMENT)->val1, tstatus->def_ele, tstatus->ele_lv, 1);
 				//Spirit Sphere bonus damage is not affected by element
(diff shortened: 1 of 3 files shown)


src/map/battle.cpp
```diff
@@ -4287,6 +4296,69 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
...
+ * This bonus works as a separate unit to the rest (e.g., if one of these is not applied to a skill, then we know none are)
+ * Do not add additional status changes here unless they are confirmed to use ATKpercent
+ * @param bl: Object to calc atkpercent for
+ * @param skill_id: Skill used by object
+ * @param sc: Object's status change information
+ * @return atkpercent with cap_value(watk,0,USHRT_MAX)
+ */
+static unsigned short battle_get_atkpercent(struct block_list* bl, uint16 skill_id, status_change* sc)
+{
+	int atkpercent = 100;
```
aleos89, line 4312: I'd move this after the switch below. No need to define if it's never utilized.

Playtester: In C you needed to define all variables at the beginning of a {} section, but I guess it's C++ now, ok sure.

Playtester: Fixed.

aleos89 (approved):

PR #8219, opened 2024-04-02
Star Crumbs, Spirit Spheres, EDP, Guillotine Fist, Sand Attack, Rapid Smiting, Clashing Spiral

### LINKED ISSUE #1823: Asura damage should not be affected by Provoke, Curse and INCATKRATE
Provoke, Curse and INCATKRATE (e.g. from Gospel or Tarot Card) do not change the Asura damage at all.
BATKFOOD and ATKPOTION however do increase Asura damage.

This would indicate that Provoke, Curse and INCATKRATE should not actually be part of the base attack calculation but instead add / substract damage in the damage calculation.

This applies to both pre-re and re.

### LINKED ISSUE #7750: Shield Chain discrepancies
### rAthena Hash

latest

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

I was comparing rAthena's implementation of Shield Chain against iROwiki's article. That website states:
"ONLY stat-derived ATK contributes to damage. [STR](https://irowiki.org/classic/STR) greatly affects damage; [DEX](https://irowiki.org/classic/DEX) and [LUK](https://irowiki.org/classic/LUK) have a minor effect."
"Any other [ATK](https://irowiki.org/classic/ATK) - weapon ATK, weapon refinement, ATK-increasing cards or items - have NO effect on the damage."
While I was able to confirm by in-game testing that Weapon ATK is indeed ignored by the skill, when I tried using other +ATK sources such as Hillslion Card and Giant Whisper, those increased damage.
"The attack bonus gained from [Provoke](https://irowiki.org/classic/Provoke) or [Berserk](https://irowiki.org/classic/Berserk) does not affect the damage."
This was another discrepancy. I tried Shield Chain before and after having Lv10 Provoke cast upon myself and damage increased.

Of course, iROwiki is full of mistakes but I figured I'd post this bug report to see if perhaps this skill isn't working wrong.

### Relevant Log Output

_No response_

### Expected Result

iROwiki article should match implementation

### How to Reproduce

Tested in PvP against players only.

### Official Information

https://irowiki.org/classic/Rapid_Smiting

_No response_

### LINKED ISSUE #8211: Pre-Renewal Damage Rework (Part 3): Star Crumbs / Spirit Spheres
### rAthena Hash

8a10907

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

Star Crumb bonus damage is not increased by EDP.
It's also likely that Star Crumb damage and Spirit Sphere damage is incorrectly multiplied for multi-hits.
Also the damage seems to be applied to almost all skills.

### Relevant Log Output

_No response_

### Expected Result

Star Crumb bonus damage is increased by EDP, so it should added before the EDP multiplier, similar to other mastery bonuses.
Need further investigation on Star Crumb / Spirit Sphere bonus damage on multi-hit skills, but it should not be multiplied by number of hits twice.
Need further investigation which skills consider Star Crumb damage.

### How to Reproduce

Test damage of EDP and multi-hit skills using any VVVS weapon and having Spirit Spheres.

### Official Information

Assassin Cross 94/69, DEX 99 vs. Golem
Main Gauche - 7
Main Gauche+5 - 17
VVVS Main Gauche+5 - 47

EDP:
Main Gauche - 29
Main Gauche+5 - 72
VVVS Main Gauche+5 - 199 (rAthena 69)

Monk 98/50, DEX 99 vs. Golem, Ironhand 5
Studded Knuckles - 26
Studded Knuckles+4 - 38
VVVS Studded Knuckles - 66

With Magnum Bonus:
Studded Knuckles - 31
Studded Knuckles+4 - 45
VVVS Studded Knuckles - 79

Throw Spirit Sphere 5:
Studded Knuckles - 132x5
Studded Knuckles+4 - 144x5
VVVS Studded Knuckles - 172x5

Throw Spirit Sphere 5 + Magnum Bonus:
Studded Knuckles - 140x5
Studded Knuckles+4 - 154x5
VVVS Studded Knuckles - 188x5

Occult Impaction 3 with VVVS Studded Knuckles:
Normal - 257
Magnum Bonus - 273/272/272/271/270 (5-1 spirit spheres)

_No response_

### LINKED ISSUE #8218: Skills affected by EDP/Magnum Break should be the same
### rAthena Hash

4f82aff

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

We have individual lists of skills that affect EDP and skills that affect the Magnum Break bonus in pre-renewal.
Also EDP affects offhand damage.

### Relevant Log Output

_No response_

### Expected Result

All skills that are not affected by EDP should also not affect the Magnum Break bonus and vice-versa.
EDP does not affect offhand damage.

### How to Reproduce

Use a skill with no damage variance (high DEX) - Check damage.
Use Magnum Break and use the skill again within 10 seconds - Check damage.
Use EDP and use the skill again - Check damage.
Use Katar or Dual Wield and test offhand damage with EDP.

### Official Information

I didn't take the damage number but this is what I tested:

Sprinkle Sand - Both not affected (**wrong in rAthena**)
Meteor Assault - Both not affected
Grimtooth - Both affected
Venom Splasher - Both not affected
Venom Knife - Both not affected (**wrong in rAthena**)
Soul Destroyer - Both not affected
Sonic Blow - Both affected
Throw Stone - Both not affected
Envenom - Both affected

Note: Secret confirmed that Soul Destroyer IS affected in renewal. 
Sprinkle Sand is not affected in renewal either (also **needs to be fixed**!)

From previous tests of Magnum Break damage we know that the following skills do not affect it:
- Acid Terror
- All unit skills

Skills that ignore element are also not affected:
Gloria Domini
Throw Coins
etc.

_No response_

Playtester:
* **Addressed Issue(s)**: #8211

* **Server Mode**: Both (changes to the order of damage processing only applies to pre-renewal)

- ATKpercent, Spirit Spheres and other mastery bonuses, now all affect the exactly same skills with some exceptions for refine and Star Crumbs, see battle_skill_stacks_masteries_vvs
- EDP and elemental bonus damage now apply to the exact same list of skills, see battle_skill_stacks_edp_element
- Star Crumbs and Spirit Spheres no longer multiply the damage by number of hits twice
- Star Crumbs, Spirit Spheres and similar mastery bonuses that apply even on MISS are now applied after the attribute table but before EDP and elemental bonus damage in pre-renewal
- Attribute table can now make damage go negative which is weighted against any of the above mastery bonuses and elemental bonus damage before being capped to 0
- EDP and elemental bonus damage no longer apply to the left hand; removed "basedamage2" as it is no longer needed
- Left-hand damage is now always capped to 0 instead of 1
- Fixed Guillotine Fist ignoring size modifiers
- Fixed inaccuracies in Guillotine Fist's damage due to rounding issues
- Fixed bonus damage from Magnum Break when using Guillotine Fist
- Sand Attack is no longer influenced by EDP
- Rapid Smiting now has a fixed 20% hit bonus (it hits at least to 25%)
- Rapid Smiting no longer deals damage on MISS
- Clashing Spiral damage is no longer influenced by ATKpercent and any mastery except Star Crumbs
- Further code improvements / compiler warning fix
- Fixes #8211 
- Fixes #8218 
- Fixes #7750 
- Fixes #1823

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3665,14 +3665,14 @@ int battle_get_misc_element(struct block_list* src, struct block_list* target, u
  * @param skill_id: Skill ID of the skill used by source
  * @param min: Minimum value to which damage should be capped
  */
-static void battle_min_damage(struct Damage* wd, struct block_list* src, uint16 skill_id, int64 min) {
-	if (is_attack_right_handed(src, skill_id)) {
-		wd->damage = cap_value(wd->damage, min, INT64_MAX);
-		wd->basedamage = cap_value(wd->basedamage, min, INT64_MAX);
+static void battle_min_damage(struct Damage &wd, struct block_list &src, uint16 skill_id, int64 min) {
+	if (is_attack_right_handed(&src, skill_id)) {
+		wd.damage = cap_value(wd.damage, min, INT64_MAX);
+		wd.basedamage = cap_value(wd.basedamage, min, INT64_MAX);
 	}
 	// Left-hand damage is always capped to 0
-	if (is_attack_left_handed(src, skill_id)) {
-		wd->damage2 = cap_value(wd->damage2, 0, INT64_MAX);
+	if (is_attack_left_handed(&src, skill_id)) {
+		wd.damage2 = cap_value(wd.damage2, 0, INT64_MAX);
 	}
 }
 
@@ -3685,11 +3685,13 @@ static void battle_min_damage(struct Damage* wd, struct block_list* src, uint16 
  * @param skill_id: Skill ID of the skill used by source
  * @param min: Minimum value to which damage should be capped
  */
-static int battle_get_spiritball_damage(struct Damage* wd, struct block_list* src, uint16 skill_id) {
+static int battle_get_spiritball_damage(struct Damage& wd, struct block_list& src, uint16 skill_id) {
 
-	map_session_data* sd = BL_CAST(BL_PC, src);
+	map_session_data* sd = BL_CAST(BL_PC, &src);
 
-	nullpo_ret(sd); // Returns 0 for non-players
+	// Return 0 for non-players
+	if (!sd)
+		return 0;
 
 	int damage = 0;
 
@@ -3699,7 +3701,7 @@ static int battle_get_spiritball_damage(struct Damage* wd, struct block_list* sr
 	case MO_FINGEROFFENSIVE:
 #endif
 		// These skills used as many spheres as they do hits
-		damage = (wd->div_ + sd->spiritball) * 3;
+		damage = (wd.div_ + sd->spiritball) * 3;
 		break;
 	case MO_EXTREMITYFIST:
 		// These skills store the number of spheres the player had before cast
@@ -3791,7 +3793,7 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
 	}
 	if (battle_skill_stacks_masteries_vvs(skill_id, 0)) {
 		// Spirit Sphere bonus damage
-		ATK_ADD(wd->damage, wd->damage2, battle_get_spiritball_damage(wd, src, skill_id));
+		ATK_ADD(wd->damage, wd->damage2, battle_get_spiritball_damage(*wd, *src, skill_id));
 
 		// Skill-specific bonuses
 		if (skill_id == TF_POISON)
@@ -3816,7 +3818,7 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
 			// Star Crumb bonus damage
 			wd->basedamage += sd->right_weapon.star;
 			// Spirit Sphere bonus damage
-			wd->basedamage += battle_get_spiritball_damage(wd, src, skill_id);
+			wd->basedamage += battle_get_spiritball_damage(*wd, *src, skill_id);
 			// Add percent of the base damage to the damage
 			wd->damage += (wd->basedamage * sc->getSCE(SC_WATK_ELEMENT)->val2) / 100;
 		}
@@ -3824,7 +3826,7 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
 #endif
 	// Cap damage to 0
 	if (battle_config.attr_recover == 0)
-		battle_min_damage(wd, src, skill_id, 0);
+		battle_min_damage(*wd, *src, skill_id, 0);
 }
 
 /*==================================
@@ -3871,7 +3873,7 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
 		// Star Crumb bonus damage
 		ATK_ADD2(wd->masteryAtk, wd->masteryAtk2, sd->right_weapon.star, sd->left_weapon.star);
 		// Spirit Sphere bonus damage
-		ATK_ADD(wd->masteryAtk, wd->masteryAtk2, battle_get_spiritball_damage(wd, src, skill_id));
+		ATK_ADD(wd->masteryAtk, wd->masteryAtk2, battle_get_spiritball_damage(*wd, *src, skill_id));
 #endif
 
 		if (skill_id == NJ_SYURIKEN && (skill = pc_checkskill(sd,NJ_TOBIDOUGU)) > 0) { // !TODO: Confirm new mastery formula
@@ -4430,33 +4432,33 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
  * @param sc: Object's status change information
  * @return atkpercent with cap_value(watk,0,USHRT_MAX)
  */
-static unsigned short battle_get_atkpercent(struct block_list* bl, uint16 skill_id, status_change* sc)
+static unsigned short battle_get_atkpercent(struct block_list& bl, uint16 skill_id, status_change& sc)
 {
 	if (!battle_skill_stacks_masteries_vvs(skill_id, 0))
 		return 100;
 
 	int atkpercent = 100;
 
-	if (sc->getSCE(SC_CURSE))
+	if (sc.getSCE(SC_CURSE))
 		atkpercent -= 25;
-	if (sc->getSCE(SC_PROVOKE))
-		atkpercent += sc->getSCE(SC_PROVOKE)->val2;
-	if (sc->getSCE(SC_STRIPWEAPON) && bl->type != BL_PC)
-		atkpercent -= sc->getSCE(SC_STRIPWEAPON)->val2;
-	if (sc->getSCE(SC_CONCENTRATION))
-		atkpercent += sc->getSCE(SC_CONCENTRATION)->val2;
-	if (sc->getSCE(SC_TRUESIGHT))
-		atkpercent += 2 * sc->getSCE(SC_TRUESIGHT)->val1;
-	if (sc->getSCE(SC_JOINTBEAT) && sc->getSCE(SC_JOINTBEAT)->val2 & BREAK_WAIST)
+	if (sc.getSCE(SC_PROVOKE))
+		atkpercent += sc.getSCE(SC_PROVOKE)->val2;
+	if (sc.getSCE(SC_STRIPWEAPON) && bl.type != BL_PC)
+		atkpercent -= sc.getSCE(SC_STRIPWEAPON)->val2;
+	if (sc.getSCE(SC_CONCENTRATION))
+		atkpercent += sc.getSCE(SC_CONCENTRATION)->val2;
+	if (sc.getSCE(SC_TRUESIGHT))
+		atkpercent += 2 * sc.getSCE(SC_TRUESIGHT)->val1;
+	if (sc.getSCE(SC_JOINTBEAT) && sc.getSCE(SC_JOINTBEAT)->val2 & BREAK_WAIST)
 		atkpercent -= 25;
-	if (sc->getSCE(SC_INCATKRATE))
-		atkpercent += sc->getSCE(SC_INCATKRATE)->val1;
-	if (sc->getSCE(SC_SKE))
+	if (sc.getSCE(SC_INCATKRATE))
+		atkpercent += sc.getSCE(SC_INCATKRATE)->val1;
+	if (sc.getSCE(SC_SKE))
 		atkpercent += 300;
-	if (sc->getSCE(SC_BLOODLUST))
-		atkpercent += sc->getSCE(SC_BLOODLUST)->val2;
-	if (sc->getSCE(SC_FLEET))
-		atkpercent += sc->getSCE(SC_FLEET)->val3;
+	if (sc.getSCE(SC_BLOODLUST))
+		atkpercent += sc.getSCE(SC_BLOODLUST)->val2;
+	if (sc.getSCE(SC_FLEET))
+		atkpercent += sc.getSCE(SC_FLEET)->val3;
 
 	/* Only few selected skills should use this function, DO NOT ADD any that are not caused by the skills listed below
 	* TODO:
@@ -4494,7 +4496,7 @@ static int battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list *
 	if(sc && skill_id != PA_SACRIFICE) {
 #ifdef RENEWAL
 		//ATK percent modifier (in renewal, it's applied before the skillratio)
-		skillratio = battle_get_atkpercent(src, skill_id, sc);
+		skillratio = battle_get_atkpercent(*src, skill_id, *sc);
 #endif
 		if(sc->getSCE(SC_OVERTHRUST))
 			skillratio += sc->getSCE(SC_OVERTHRUST)->val3;
@@ -6343,8 +6345,8 @@ static void battle_attack_sc_bonus(struct Damage* wd, struct block_list *src, st
 			ATK_ADD(wd->equipAtk, wd->equipAtk2, sc->getSCE(SC_GATLINGFEVER)->val3);
 #else
 		//ATK percent modifier (in pre-renewal, it's applied multiplicatively after the skill ratio)
-		ATK_RATE(wd->damage, wd->damage2, battle_get_atkpercent(src, skill_id, sc));
-		ATK_RATER(wd->basedamage, battle_get_atkpercent(src, 0, sc));
+		ATK_RATE(wd->damage, wd->damage2, battle_get_atkpercent(*src, skill_id, *sc));
+		ATK_RATER(wd->basedamage, battle_get_atkpercent(*src, 0, *sc));
 #endif
 		if (sc->getSCE(SC_SPIRIT)) {
 			if (skill_id == AS_SONICBLOW && sc->getSCE(SC_SPIRIT)->val2 == SL_ASSASIN) {
@@ -6691,7 +6693,7 @@ static void battle_calc_attack_post_defense(struct Damage* wd, struct block_list
 
 	//After DEF reduction, damage can be negative, refine bonus works against that value
 	//After refinement bonus was applied, damage is capped to 1, then masteries are applied
-	battle_min_damage(wd, src, skill_id, 1);
+	battle_min_damage(*wd, *src, skill_id, 1);
 
 	battle_calc_attack_masteries(wd, src, target, skill_id, skill_lv);
 #endif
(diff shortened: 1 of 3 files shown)


src/map/battle.cpp
```diff
@@ -3353,25 +3356,67 @@ static bool attack_ignores_def(struct Damage* wd, struct block_list *src, struct
...
- *	Refined and optimized by helvetica
+/**
+ * This function lists which skills are unaffected by refine bonus, masteries, Star Crumbs and Spirit Spheres
+ * This function is also used to determine if atkpercent applies
+ * @param skill_id: Skill being used
+ * @param type 1 - Checking refine bonus; 2 - Checking Star Crumb bonus
+ * @return true = bonus applies; false = bonus does not apply
  */
-static bool battle_skill_stacks_masteries_vvs(uint16 skill_id)
+static bool battle_skill_stacks_masteries_vvs(uint16 skill_id, int type)
```
Lemongrass3110, line 3366: May be just me, but this whole function sounds like a flag that should be put into skill_db.yml

src/map/battle.cpp
```diff
@@ -3353,25 +3356,67 @@ static bool attack_ignores_def(struct Damage* wd, struct block_list *src, struct
...
+	return true;
+}
+
+/**
+ * This function lists which skills are unaffected by EDP and the elemental bonus from Magnum Break / EDP
+ * Unit skills (e.g. Bomb and Freezing Trap) are never affected.
+ * @param skill_id: Skill being used
+ * @return true = bonus applies; false = bonus does not apply
+ */
+static bool battle_skill_stacks_edp_element(uint16 skill_id)
```
Lemongrass3110, line 3404: May be just me, but this whole function sounds like a flag that should be put into skill_db.yml

src/map/battle.cpp
```diff
@@ -3592,6 +3637,85 @@ int battle_get_misc_element(struct block_list* src, struct block_list* target, u
...
+
+#ifdef RENEWAL
+	#define RE_ALLATK_ADD(wd, a) do { int64 a_ = (a); ATK_ADD((wd)->statusAtk, (wd)->statusAtk2, a_); ATK_ADD((wd)->weaponAtk, (wd)->weaponAtk2, a_); ATK_ADD((wd)->equipAtk, (wd)->equipAtk2, a_); ATK_ADD((wd)->masteryAtk, (wd)->masteryAtk2, a_); } while(0);
+	#define RE_ALLATK_RATE(wd, a) do { int64 a_ = (a); ATK_RATE((wd)->statusAtk, (wd)->statusAtk2, a_); ATK_RATE((wd)->weaponAtk, (wd)->weaponAtk2, a_); ATK_RATE((wd)->equipAtk, (wd)->equipAtk2, a_); ATK_RATE((wd)->masteryAtk, (wd)->masteryAtk2, a_); } while(0);
+	#define RE_ALLATK_ADDRATE(wd, a) do { int64 a_ = (a); ATK_ADDRATE((wd)->statusAtk, (wd)->statusAtk2, a_); ATK_ADDRATE((wd)->weaponAtk, (wd)->weaponAtk2, a_); ATK_ADDRATE((wd)->equipAtk, (wd)->equipAtk2, a_); ATK_ADDRATE((wd)->masteryAtk, (wd)->masteryAtk2, a_); } while(0);
+#else
+	#define RE_ALLATK_ADD(wd, a) {;}
+	#define RE_ALLATK_RATE(wd, a) {;}
+	#define RE_ALLATK_ADDRATE(wd, a) {;}
+#endif
```
Lemongrass3110, line 3659: If you already touch those:
Do you mind C++ing them and making them typesafe?

src/map/battle.cpp
```diff
@@ -3592,6 +3637,85 @@ int battle_get_misc_element(struct block_list* src, struct block_list* target, u
...
+		return 0;
+
+	int damage = 0;
+
+	switch (skill_id) {
+	case MO_INVESTIGATE:
+#ifndef RENEWAL
+	case MO_FINGEROFFENSIVE:
+#endif
+		// These skills used as many spheres as they do hits
```
Lemongrass3110, line 3703:
```suggestion
		// These skills use as many spheres as they do hits
```

src/map/battle.cpp
```diff
@@ -3652,63 +3776,58 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
...
-		}
-#endif
+	// These mastery bonuses are non-elemental and should apply even if the attack misses
+	// They are still increased by the EDP/Magnum Break bonus damage (WATK_ELEMENT)
+	// In renewal these bonuses do not apply when the attack misses
+	if (sd && battle_skill_stacks_masteries_vvs(skill_id, 2)) {
+		// Star Crumb bonus damage
+		ATK_ADD2(wd->damage, wd->damage2, sd->right_weapon.star, sd->left_weapon.star);
+	}
+	if (battle_skill_stacks_masteries_vvs(skill_id, 0)) {
```
Lemongrass3110, line 3794:
```suggestion

	// Comment why you call it again and with different parameters
	if (battle_skill_stacks_masteries_vvs(skill_id, 0)) {
```

src/map/battle.cpp
```diff
@@ -3652,63 +3776,58 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
...
-#define ATK_ADDRATE2(damage, damage2, a , b) do { int64 rate_ = (a), rate2_ = (b); (damage) += (damage) * rate_ / 100; if(is_attack_left_handed(src, skill_id)) (damage2) += (damage2) * rate2_ / 100; } while(0);
-//Adds an absolute value to damage. 100 = +100 damage
-#define ATK_ADD(damage, damage2, a) do { int64 rate_ = (a); (damage) += rate_; if(is_attack_left_handed(src, skill_id)) (damage2) += rate_; } while(0);
-#define ATK_ADD2(damage, damage2, a , b) do { int64 rate_ = (a), rate2_ = (b); (damage) += rate_; if(is_attack_left_handed(src, skill_id)) (damage2) += rate2_; } while(0);
+	// These bonuses do not apply to skills that ignore element, unit skills and skills that have their own base damage formula
+	// If damage was reduced below 0 and was not increased again to a positive value through mastery bonuses, these bonuses are ignored
+	// Any of these are only applied to your right hand weapon in pre-renewal
+	if (!nk[NK_IGNOREELEMENT] && (wd->damage > 0 || wd->damage2 > 0) && sc && battle_skill_stacks_edp_element(skill_id)) {
```
Lemongrass3110, line 3806: Why is the EDP check already in the outer if and not the inner one, where it really checks for EDP?

src/map/battle.cpp
```diff
@@ -3754,12 +3870,10 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
...
-		if (skill_id != CR_SHIELDBOOMERANG)
-			ATK_ADD2(wd->masteryAtk, wd->masteryAtk2, ((wd->div_ < 1) ? 1 : wd->div_) * sd->right_weapon.star, ((wd->div_ < 1) ? 1 : wd->div_) * sd->left_weapon.star);
-		ATK_ADD(wd->masteryAtk, wd->masteryAtk2, ((wd->div_ < 1) ? 1 : wd->div_) * sd->spiritball * 3);
-#else
-		if (skill_id == TF_POISON)
-			ATK_ADD(wd->damage, wd->damage2, 15 * skill_lv);
+		// Star Crumb bonus damage
+		ATK_ADD2(wd->masteryAtk, wd->masteryAtk2, sd->right_weapon.star, sd->left_weapon.star);
+		// Spirit Sphere bonus damage
+		ATK_ADD(wd->masteryAtk, wd->masteryAtk2, battle_get_spiritball_damage(*wd, *src, skill_id));
```
Lemongrass3110, line 3876: I think if we already have it, then we should check the bonuses with battle_skill_stacks_masteries_vvs here?

src/map/battle.cpp
```diff
@@ -3652,63 +3776,58 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
...
+
+		// This adds a percentual damage bonus based on the damage you would deal with a normal attack
+		// Applies only to player damage; monsters and mercenaries don't get this damage boost
+		if (sd && sc->getSCE(SC_WATK_ELEMENT)) {
+			// Pretend the normal attack was of the element stored in the status change
+			wd->basedamage = battle_attr_fix(src, target, wd->basedamage, sc->getSCE(SC_WATK_ELEMENT)->val1, tstatus->def_ele, tstatus->ele_lv, 1);
+			// Star Crumb bonus damage
+			wd->basedamage += sd->right_weapon.star;
+			// Spirit Sphere bonus damage
+			wd->basedamage += battle_get_spiritball_damage(*wd, *src, skill_id);
```
Lemongrass3110, line 3821: Shouldnt we check battle_skill_stacks_masteries_vvs here?

src/map/battle.cpp
```diff
@@ -4316,49 +4432,33 @@ static void battle_calc_multi_attack(struct Damage* wd, struct block_list *src,s
...
-	case AM_ACIDTERROR:
-	case CR_GRANDCROSS:
-	case NPC_GRANDDARKNESS:
-	case MO_INVESTIGATE:
-	case MO_EXTREMITYFIST:
-	case PA_SACRIFICE:
-	case NPC_DRAGONBREATH:
-	case RK_DRAGONBREATH:
-	case RK_DRAGONBREATH_WATER:
+	if (!battle_skill_stacks_masteries_vvs(skill_id, 0))
```
Lemongrass3110, line 4437: Was the old implementation wrong or do we abuse a "wrong named" function here?

src/map/battle.cpp
```diff
@@ -2263,7 +2263,7 @@ int64 battle_addmastery(map_session_data *sd,struct block_list *target,int64 dmg
 			damage += damage * 30 / 100;
 	}
 
-	if(type == 0)
+	if (type == 0)
```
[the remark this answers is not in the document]
Playtester: Space behind if is the automatic standard formatting we have to get to eventually, but if you don't want this to count as a change because it makes using Blame harder, I can revert this part.

src/map/battle.cpp
```diff
@@ -3284,6 +3283,10 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
 	}
 
 	hitrate = cap_value(hitrate, battle_config.min_hitrate, battle_config.max_hitrate);
+
+	if(skill_id == PA_SHIELDCHAIN)
```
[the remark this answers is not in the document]
Playtester: No, this is intentional, please see the comment.

If your min_hitrate is 5% then Shield Chain should have a minimum hit rate of 20.

src/map/battle.cpp
Playtester: Eventually yes, but I won't mess with YML myself.

src/map/battle.cpp
```diff
@@ -3353,25 +3356,62 @@ static bool attack_ignores_def(struct Damage* wd, struct block_list *src, struct
...
-	if (
-#ifndef RENEWAL
-		skill_id == PA_SHIELDCHAIN || skill_id == CR_SHIELDBOOMERANG ||
-		skill_id == PA_SACRIFICE || skill_id == AM_ACIDTERROR ||
-#endif
-		skill_id == MO_INVESTIGATE || skill_id == MO_EXTREMITYFIST ||
-		skill_id == RK_DRAGONBREATH || skill_id == RK_DRAGONBREATH_WATER || skill_id == NC_SELFDESTRUCTION ||
-		skill_id == LG_SHIELDPRESS || skill_id == LG_EARTHDRIVE)
+	switch (skill_id) {
+	case PA_SHIELDCHAIN:
```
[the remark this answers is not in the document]
Playtester: This is automatically done by VS formatting, it's annoying to go against that because you have to keep reverting automatic formatting.

src/map/battle.cpp
Playtester: Same as above, I won't mess with YML myself, but it's something that can be done later, yes.

src/map/battle.cpp
```diff
@@ -3353,25 +3356,62 @@ static bool attack_ignores_def(struct Damage* wd, struct block_list *src, struct
...
+/**
+ * This function lists which skills are unaffected by EDP and the elemental bonus from Magnum Break / EDP
+ * Unit skills (e.g. Bomb and Freezing Trap) are never affected.
+ * @param skill_id: Skill being used
+ * @return true = bonus applies; false = bonus does not apply
+ */
+static bool battle_skill_stacks_edp_element(uint16 skill_id)
+{
+	switch (skill_id) {
+	case TF_SPRINKLESAND:
```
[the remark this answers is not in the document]
Playtester: See above.

src/map/battle.cpp
Playtester: Technically they already used them but both phrasings work.

src/map/battle.cpp
Playtester: This is not an EDP check, this is a check for this whole section that includes EDP and other elemental damage bonuses (right now only WATK_ELEMENT).

src/map/battle.cpp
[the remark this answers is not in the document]
Lemongrass3110: I thought this is a custom *Athena configuration and not official behavior?
If I set that maximum to 10 I wouldnt expect a skill to exceed the value of 10.

src/map/battle.cpp
Playtester: No need because basedamage is the damage a normal attack would do and masteries always apply to normal attacks.

src/map/battle.cpp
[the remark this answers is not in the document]
Lemongrass3110: Its for blaming, because you didnt change anything here.

src/map/battle.cpp
Playtester: It already checks for it around 10 lines above.

src/map/battle.cpp
Lemongrass3110: Okay, please open an issue for these two after merging then, so that someone can take it up.

src/map/battle.cpp
Playtester: I updated the function to also handle ATKpercent (see comment of the function).
Skills that don't apply masteries/VVS also won't apply all other things (unless Aegis manually coded it into the skill code as with the refine bonus for Grand Cross), so it's better to channel it all through one function rather than having switches everywhere.
It prevents devs from adding a special skill exception only to one thing they just found instead of making it a general rule.

src/map/battle.cpp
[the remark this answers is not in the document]
Playtester: But if all other skills are officially 5% minimum and Shield Chain is 25% minimum hitrate. If someone changes it to 10% (+5% from official value), I think they would expect Shield Chain to becomes 30% and not 10%. Do you think otherwise?

src/map/battle.cpp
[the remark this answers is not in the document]
Playtester: Ok!

src/map/battle.cpp
Playtester: Could already do this now (then I can mark the comment as resolved).

But it's a bit more complex because there are a few special skills with hard-coded custom behaviors that go against the norm such as Grand Cross (considers refine but nothing else) and Spiral Pierce (considers refine and Star Crumbs, but nothing else).

Also some of the existing flags probably also require some rework. I think the "IgnoreElement" flag for example is kinda weird, because it makes you ignore element cards but not other cards. But the one weapon skill that even uses this flag "Venom Splasher" actually ignores all card modifiers but shouldn't even ignore the attribute table (I created an issue on that one already).

It would be better to have more test data on all skills to see the full picture first. But I guess I can just set the "Needs more info" tag on the issue and explain it in the issue.

src/map/battle.cpp
```diff
@@ -3353,25 +3356,67 @@ static bool attack_ignores_def(struct Damage* wd, struct block_list *src, struct
...
- *------------------------------------------------
- * Credits:
- *	Original coder Skotlex
- *	Initial refactoring by Baalberith
- *	Refined and optimized by helvetica
+/**
+ * This function lists which skills are unaffected by refine bonus, masteries, Star Crumbs and Spirit Spheres
+ * This function is also used to determine if atkpercent applies
+ * @param skill_id: Skill being used
+ * @param type 1 - Checking refine bonus; 2 - Checking Star Crumb bonus
```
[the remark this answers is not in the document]
Playtester: Note: Agreed with Lemon to leave it like this for now. We might eventually turn the function into a YML option anyway.

src/map/battle.cpp
Playtester: Feature request added: https://github.com/rathena/rathena/issues/8226

src/map/battle.cpp
Playtester: Feature request added: https://github.com/rathena/rathena/issues/8226

aleos89 (approved):

PR #8237, opened 2024-04-08
Envenom Element Modifier Adjustments

### LINKED ISSUE #8212: Envenom damage on earth monsters
### rAthena Hash

23787bd

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

Envenom damage on rAthena is lower against enemies that are vulnerable to poison.

### Relevant Log Output

_No response_

### Expected Result

Envenom damage is higher against enemies that are vulnerable to poison (I assume attribute table is partially applied twice).

### How to Reproduce

Use Envenom on Stapo.

### Official Information

Assassin 99 DEX all Masteries vs. Stapo
Katar+5+Envenom - 325 (306 in rAthena)
Magnum Bonus+Katar+5+Envenom - 346 (327 in rAthena)

Katar of Quaking+Envenom - 282
Magnum Bonus+Katar of Quaking+Envenom - 294

_No response_

Playtester:
* **Addressed Issue(s)**: #8212 

* **Description of Pull Request**: Envenom now applies the attribute table to the final damage again (fixes #8212)

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3739,11 +3739,11 @@ static void battle_calc_element_damage(struct Damage* wd, struct block_list *src
 	status_change* sc = status_get_sc(src);
 	struct status_data* sstatus = status_get_status_data(src);
 	struct status_data* tstatus = status_get_status_data(target);
+	int right_element = battle_get_weapon_element(wd, src, target, skill_id, skill_lv, EQI_HAND_R, true);
 
 	// Elemental attribute fix
 	if(!nk[NK_IGNOREELEMENT] && (wd->damage > 0 || wd->damage2 > 0)) {
 		int left_element = battle_get_weapon_element(wd, src, target, skill_id, skill_lv, EQI_HAND_L, true);
-		int right_element = battle_get_weapon_element(wd, src, target, skill_id, skill_lv, EQI_HAND_R, true);
 
 		switch (skill_id) {
 			case PA_SACRIFICE:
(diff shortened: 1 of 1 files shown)


aleos89 (approved):
I'm semi-skeptical of exposing the right_element out of the if-statement but I guess it's not computationally expensive. The skill_db shows there are only 17~ skills with the `IgnoreElement` flag, so not many more skills to look up additionally.

Playtester:
> I'm semi-skeptical of exposing the right_element out of the if-statement but I guess it's not computationally expensive. The skill_db shows there are only 17~ skills with the `IgnoreElement` flag, so not many more skills to look up additionally.

I moved it out because I'm using it in the TF_POISON code which is not inside the "IgnoreElement" section. I could also hardcode it to ELE_POISON instead of using right_element.

It only matters for people who customize the skill:
Say someone says he wants Envenom to be of Undead element instead so he changes the YML file. Should it then be fully Undead or half Undead half Poison?
What if someone wants Envenom to ignore element? Should it then apply the attribute table once instead of twice or not at all?

Depending on those answers, the implementation would be different. Just hardcoding it to ELE_POISON would be most simple from development viewpoint, if we didn't care about customization.

Lemongrass3110 (approved):

aleos89:
> > I'm semi-skeptical of exposing the right_element out of the if-statement but I guess it's not computationally expensive. The skill_db shows there are only 17~ skills with the `IgnoreElement` flag, so not many more skills to look up additionally.
> 
> I moved it out because I'm using it in the TF_POISON code which is not inside the "IgnoreElement" section. I could also hardcode it to ELE_POISON instead of using right_element.
> 
> It only matters for people who customize the skill: Say someone says he wants Envenom to be of Undead element instead so he changes the YML file. Should it then be fully Undead or half Undead half Poison? What if someone wants Envenom to ignore element? Should it then apply the attribute table once instead of twice or not at all?
> 
> Depending on those answers, the implementation would be different. Just hardcoding it to ELE_POISON would be most simple from development viewpoint, if we didn't care about customization.

Don't hard code it, we want to move away from that! It was just a thought I had when reviewing, but as stated, 99% of the skills already fall into the `!IgnoreElement` check, so this wouldn't make any difference to the remaining 17~ that avoid this. These changes work just fine!

Issue #8239, opened 2024-04-09
HP_BASILICA causes crash when cast by Non-Players 


Singe-Horizontal:
### rAthena Hash

03f07ab8c

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

Server is crashing when a mob uses HP_BASILICA

### Expected Result

No crash should happen

### How to Reproduce

Add basilica skill to a mob skillset

### Official Information

N/A

PR #8240, opened 2024-04-09
Fixes a crash when a mob uses HP_BASILICA

Singe-Horizontal:
A check for sd pointer was missing

* **Addressed Issue(s)**: 
#8239

* **Server Mode**: 
Both

Restores a missing sd pointer check before send a clif_fail packet


aleos89 (approved):

Atemo:
https://github.com/rathena/rathena/pull/7963 fixed some clif_skill_fail, unfinished.
As said, maybe a better solution should be to just replace the nullpo in clif_skill_fail with if sd == nullptr return ?

Lemongrass3110:
Already covered by the PR mentioned by @Atemo.

PR #8243, opened 2024-04-09
NPC_POWERUP, NPC_AGIUP (vs. Quagmire) and Dispel

### LINKED ISSUE #904: Quagmire vs NPC_AGIUP
On official servers, Quagmire works against NPC_AGIUP. It does not dispell it, but it reduces flee and speed (see here: http://forums.irowiki.org/showthread.php?p=301145#post301145).

In rAthena there is a problem that NPC_AGIUP uses SC_SPEEDUP1 which is also used for Speed Potion. Speed Potion effect is actually the only effect that can NOT be reduced officially. This was also implemented, but the problem is that this affected NPC_AGIUP too.

``` c
    // Check for immunities / sc fails
    switch (type) {
    case SC_DECREASEAGI:
    case SC_QUAGMIRE:
    case SC_DONTFORGETME:
    case SC_ADORAMUS:
        if(sc->data[SC_SPEEDUP1])
            return 0;
        break;
```

Also note the FIXME comment here:

``` c
        if( sc->data[SC_SPEEDUP1] ) // !FIXME: used both by NPC_AGIUP and Speed Potion script
            val = max( val, sc->data[SC_SPEEDUP1]->val1 );
```

So, either we do as the comment says and make a status change for NPC_AGIUP or make it use SC_SPEEDUP0 instead (MVPs can't use items so there shouldn't be a conflict).

Like this:

``` c
    case NPC_AGIUP:
        sc_start(src,bl,SC_SPEEDUP0,100,50,skill_get_time(skill_id, skill_lv));
```

### LINKED ISSUE #7346: SC_EXTREMITYFIST2 and statuses

### LINKED ISSUE #8242: NPC_POWERUP and NPC_AGIUP duration
### rAthena Hash

03f07ab

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

NPC_POWERUP and NPC_AGIUP have a duration 5s+5s*level.

### Relevant Log Output

_No response_

### Expected Result

NPC_POWERUP has a duration of `5s*level` (max level 5 = 25s)
NPC_AGIUP has a duration of `10s*level` (max level 10 = 100s)

Note: The effect of NPC_POWERUP displays always for 25s 

### How to Reproduce

Have a monster use NPC_POWERUP and test how long it takes until the damage becomes normal again.
Have a monster use NPC_AGIUP and test how long it takes until it becomes slower again.

Note: The NPC_POWERUP effect always lasts for 25s even if a lower level is used, but you can see it already ended by the damage being lower (and the monster also missing more often).

### Official Information

Verified on official servers.

_No response_

Playtester:
* **Addressed Issue(s)**: #8242, #904, #7346 

* **Server Mode**: Both

- NPC_POWERUP and NPC_AGIUP now grant their own unique status changes
- Duration of NPC_POWERUP is now 5s*level
- Duration of NPC_AGIUP is now 10s*level and max level is back to 10
- Quagmire, Slow Grace and Adoramus now work even when the target used NPC_AGIUP
- NPC_AGIUP will remove Decrease Agi on start but will not prevent it
- NPC_AGIUP can be dispelled, but NPC_POWERUP cannot
- Dispel will now make a monster unlock its target
- Removed custom code to display the NPC_POWERUP effect, it's now correctly linked to EFST_POWERUP
- Removed the ifdef around EXTREMITYFIST2, so we no longer constantly change its ID
- Fixes #8242 
- Fixes #904 
- Fixes #7346 

PS: If you are reading this and didn't restart your Discord yet. Discord kinda crashed globally (but it doesn't tell you) and you need to restart it to send or receive messages again.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -7404,7 +7404,7 @@ static signed short status_calc_hit(struct block_list *bl, status_change *sc, in
 	if(sc->getSCE(SC_INCHITRATE))
 		hit += hit * sc->getSCE(SC_INCHITRATE)->val1/100;
 	if (sc->getSCE(SC_POWERUP))
-		hit *= hit * sc->getSCE(SC_POWERUP)->val2 / 100;
+		hit += hit * sc->getSCE(SC_POWERUP)->val2 / 100;
 	if(sc->getSCE(SC_BLIND))
 		hit -= hit * 25/100;
 	if(sc->getSCE(SC_HEAT_BARREL))
@@ -7522,7 +7522,7 @@ static signed short status_calc_flee(struct block_list *bl, status_change *sc, i
 	if(sc->getSCE(SC_INCFLEERATE))
 		flee += flee * sc->getSCE(SC_INCFLEERATE)->val1/100;
 	if (sc->getSCE(SC_AGIUP))
-		flee *= flee * sc->getSCE(SC_AGIUP)->val2 / 100;
+		flee += flee * sc->getSCE(SC_AGIUP)->val2 / 100;
 	if(sc->getSCE(SC_SPIDERWEB) || sc->getSCE(SC_WIDEWEB))
 		flee -= flee * 50/100;
 	if(sc->getSCE(SC_BERSERK))


db/pre-re/status.yml
```diff
@@ -7062,3 +7060,25 @@ Body:
...
+    DurationLookup: NPC_AGIUP
+    CalcFlags:
+      Speed: true
+      Flee: true
+    Flags:
+      NoClearbuff: true
+      NoBanishingBuster: true
+      NoClearance: true
+    EndOnStart:
+      Decreaseagi: true
```
Lemongrass3110, line 7084:
```suggestion
      Decreaseagi: true

```

db/re/status.yml
```diff
@@ -8897,3 +8895,25 @@ Body:
...
+    DurationLookup: NPC_AGIUP
+    CalcFlags:
+      Speed: true
+      Flee: true
+    Flags:
+      NoClearbuff: true
+      NoBanishingBuster: true
+      NoClearance: true
+    EndOnStart:
+      Decreaseagi: true
```
Lemongrass3110, line 8919:
```suggestion
      Decreaseagi: true

```

src/map/skill.cpp
```diff
@@ -9648,9 +9648,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			if(status_isimmune(bl))
 				break;
 
-			//Remove bonus_script by Dispell
+			// Remove bonus_script by Dispell
```
Lemongrass3110, line 9651: No change

src/map/skill.cpp
```diff
@@ -9648,9 +9648,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 			if (dstsd)
-				pc_bonus_script_clear(dstsd,BSF_REM_ON_DISPELL);
+				pc_bonus_script_clear(dstsd, BSF_REM_ON_DISPELL);
```
Lemongrass3110, line 9653: No change

src/map/skill.cpp
```diff
@@ -9997,13 +10000,11 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		break;
 
 	case NPC_POWERUP:
-		sc_start(src,bl,SC_INCATKRATE,100,200,skill_get_time(skill_id, skill_lv));
 		clif_skill_nodamage(src,bl,skill_id,skill_lv,
 			sc_start(src,bl,type,100,100,skill_get_time(skill_id, skill_lv)));
```
Lemongrass3110, line 10004: Val1 changes from 200 -> 100 and is not used in general.

src/map/status.cpp
```diff
@@ -7403,6 +7403,8 @@ static signed short status_calc_hit(struct block_list *bl, status_change *sc, in
 		hit += sc->getSCE(SC_MTF_HITFLEE)->val1;
 	if(sc->getSCE(SC_INCHITRATE))
 		hit += hit * sc->getSCE(SC_INCHITRATE)->val1/100;
+	if (sc->getSCE(SC_POWERUP))
+		hit *= 2; // +100%
```
Lemongrass3110, line 7407: Consider using val1 instead

src/map/skill.cpp
```diff
@@ -9997,13 +10000,11 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 		break;
 
 	case NPC_AGIUP:
-		sc_start(src,bl,SC_SPEEDUP1,100,50,skill_get_time(skill_id, skill_lv));
 		clif_skill_nodamage(src,bl,skill_id,skill_lv,
 			sc_start(src,bl,type,100,100,skill_get_time(skill_id, skill_lv)));
```
Lemongrass3110, line 10009: Val1 change from 50 -> 100 and is not used in general.

src/map/status.cpp
```diff
@@ -7519,6 +7521,8 @@ static signed short status_calc_flee(struct block_list *bl, status_change *sc, i
 	// Rate value
 	if(sc->getSCE(SC_INCFLEERATE))
 		flee += flee * sc->getSCE(SC_INCFLEERATE)->val1/100;
+	if (sc->getSCE(SC_AGIUP))
+		flee *= 2; // +100%
```
Lemongrass3110, line 7525: Consider using val1 instead

src/map/status.cpp
```diff
@@ -8012,8 +8016,10 @@ static unsigned short status_calc_speed(struct block_list *bl, status_change *sc
 			speed_rate = 150;
 
 		// GetMoveHasteValue1()
-		if( sc->getSCE(SC_SPEEDUP1) ) // !FIXME: used both by NPC_AGIUP and Speed Potion script
+		if( sc->getSCE(SC_SPEEDUP1) )
 			val = max( val, sc->getSCE(SC_SPEEDUP1)->val1 );
+		if (sc->getSCE(SC_AGIUP))
+			val = max(val, 50);
```
Lemongrass3110, line 8022: Consider using val2 instead

Lemongrass3110 (commented):
Good work, but a lot mixed again!

src/map/skill.cpp
Playtester: Just wanted to bring it into a consistent format since I was working on this part anyway.

src/map/skill.cpp
Playtester: See above.

src/map/skill.cpp
Playtester: Value not needed. The increase it coded into battle.cpp.

src/map/status.cpp
Playtester: I was wondering about that while developing already. Is it really better to use val1 / val2, etc.?

I feel like those are only needed if the value is different per level, not when the status change has a fixed effect and only the duration is variable.

Because you are basically just either adding redundant code in status.cpp where you set val1/val2, etc. and explain what they mean, even though you could just have it only at the place where SC is actually applied. Or you just set them directly in sc_start but then at the place of the call you wouldn't understand what val1/val2 are for and need to look up the code where they are applied anyway.

After those considerations I decided to not use val1/val2. But I'll bring it up on Discord again before finishing this PR.

src/map/skill.cpp
Playtester: Value not needed, the speed increase is directly coded into status_calc_speed.

src/map/status.cpp
Playtester: See above

src/map/status.cpp
Playtester: See above

src/map/status.cpp
```diff
@@ -7519,6 +7521,8 @@ static signed short status_calc_flee(struct block_list *bl, status_change *sc, i
 	// Rate value
 	if(sc->getSCE(SC_INCFLEERATE))
 		flee += flee * sc->getSCE(SC_INCFLEERATE)->val1/100;
+	if (sc->getSCE(SC_AGIUP))
+		flee *= flee * sc->getSCE(SC_AGIUP)->val2 / 100;
```
Lemongrass3110, line 7525:
```suggestion
		flee += flee * sc->getSCE(SC_AGIUP)->val2 / 100;
```

src/map/status.cpp
```diff
@@ -7403,6 +7403,8 @@ static signed short status_calc_hit(struct block_list *bl, status_change *sc, in
 		hit += sc->getSCE(SC_MTF_HITFLEE)->val1;
 	if(sc->getSCE(SC_INCHITRATE))
 		hit += hit * sc->getSCE(SC_INCHITRATE)->val1/100;
+	if (sc->getSCE(SC_POWERUP))
+		hit *= hit * sc->getSCE(SC_POWERUP)->val2 / 100;
```
Lemongrass3110, line 7407:
```suggestion
		hit += hit * sc->getSCE(SC_POWERUP)->val2 / 100;
```

Lemongrass3110 (approved):

PR #8252, opened 2024-04-14
Updated/Simplified Damage Bonus Stacking

### LINKED ISSUE #8250: Skill with IgnoreAtkCard / IgnoreEDPMagnum should be identical
### rAthena Hash

81894ee

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

There is a separate check for whether EDP / Magnum Break applies and whether +% damage cards apply.

### Relevant Log Output

_No response_

### Expected Result

The list of skills to which EDP / Magnum Break and +% damage cards apply should be exactly the same.

### How to Reproduce

Get uncarded and carded weapon, compare damage of skill with both weapon. Then use Magnum Break and test damage again.

### Official Information

Tested on official servers, the following skills are the ones that ignore both +% damage cards and EDP/Magnum:

Venom Splasher - Both ignored
Bomb - Both ignored
Acid Terror - Both ignored
Grand Cross - Both ignored
Meteor Assault - Both ignored
Soul Destroyer - Both ignored
High Speed Cart Ram - Both ignored
Bull's Eye - Both ignored
Venom Knife - Both ignored
Sand Attack - Both ignored
Freezing Trap - Both ignored
Gunslinger Mine - Both ignored
Throw Shuriken - Both ignored
Throw Kunai - Both ignored
Flip Tatami - Both ignored
Excrutiating Palm - Both ignored

All other skills are affected by both.

I thought all unit skills are not affected, but I found two that are affected by both:

Heat - Both affected
Desperado - Both affected

I assume the same probably also applies to renewal, but it would need a lot of updating and testing to confirm.

_No response_

Playtester:
* **Addressed Issue(s)**: #8250

- Freezing Trap, Sand Attack, Gunslinger Mine, Flip Tatami and Excruciating Palm no longer are affected by +% dmg cards
- The skills that ignore +% dmg cards are now exactly those skills that also ignore EDP and the EDP/Magnum Break elemental damage bonus
- Non-players are now always affected by ATKpercent
- Simplified the code accordingly
- Fixes #8250

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3364,7 +3364,7 @@ static bool attack_ignores_def(struct Damage* wd, struct block_list *src, struct
  * @param chk_flag: The bonus that is currently being checked for, see e_bonus_chk_flag
  * @return true = bonus applies; false = bonus does not apply
  */
-static bool battle_skill_stacks_masteries_vvs(struct block_list &src, uint16 skill_id, e_bonus_chk_flag chk_flag)
+static bool battle_skill_stacks_masteries_vvs(uint16 skill_id, e_bonus_chk_flag chk_flag)
 {
 	switch (skill_id) {
 		// PC skills that are unaffected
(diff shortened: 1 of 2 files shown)


src/map/battle.cpp
```diff
@@ -3364,7 +3364,7 @@ static bool attack_ignores_def(struct Damage* wd, struct block_list *src, struct
  * @param chk_flag: The bonus that is currently being checked for, see e_bonus_chk_flag
  * @return true = bonus applies; false = bonus does not apply
  */
-static bool battle_skill_stacks_masteries_vvs(struct block_list &src, uint16 skill_id, e_bonus_chk_flag chk_flag)
+static bool battle_skill_stacks_masteries_vvs(uint16 skill_id, e_bonus_chk_flag chk_flag)
```
aleos89, line 3367: You can remove src from the comment parameter list.

aleos89 (approved):

Lemongrass3110 (approved):

PR #8258, opened 2024-04-18
Script correction of Lux Anima Runestone IDs 22540

idk-whoami:
* **Addressed Issue(s)**: 

* **Server Mode**: re


db/re/item_db_usable.yml
```diff
@@ -56781,7 +56781,9 @@ Body:
       NoMail: true
       NoAuction: true
     Script: |
-      itemskill "RK_LUXANIMA",1;
+      if (strcharinfo(3) == "job3_rune02") {
```
Atemo, line 56784: Source? From what I know, runstone 12388, 12389 and 12390 are used when changing to RK job, so it makes sense to restrict them. This one is not used when during the job change, so logically speaking I don't see why there should be a restriction.

idk-whoami: runestones can be storage

so, any class or jobs can use it :)

datawulf: Looks like there's a mix-up with trial and regular rune stones

aleos89 (approved):

PR #8259, opened 2024-04-18
Modes correction of Temple Of Demon God monsters

idk-whoami:
* **Addressed Issue(s)**: 

* **Server Mode**: re

### DIFF
--- a/db/re/mob_db.yml
+++ b/db/re/mob_db.yml
@@ -78190,6 +78190,9 @@ Body:
     DamageMotion: 768
     Ai: 10
     Class: Boss
+    Modes:
+      CanMove: true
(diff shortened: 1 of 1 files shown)


db/re/mob_db.yml
```diff
@@ -78190,6 +78190,9 @@ Body:
     DamageMotion: 768
     Ai: 10
     Class: Boss
+    Modes:
+      CanMove: true
+      Detector: true
```
Atemo, line 78195: Class 'boss' already include detector. I guess the 'Ai 'should be 21 lika all other bosses? (same for other corrections)
```suggestion
    Ai: 21
    Class: Boss
```

PR #8260, opened 2024-04-18
* fixes wrong double damage calculation for skills in renewal mode

datawulf:
* **Addressed Issue(s)**: Skill Damage calculation was doubled in Renewal Mode

* **Server Mode**: Renewal

Fixes skill damage calculation in Renewal.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -7335,11 +7335,13 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 		wd.dmg_lv = ATK_FLEE;
 	else if(!(infdef = is_infinite_defense(target, wd.flag))) { //no need for math against plants
 
+#ifndef RENEWAL
 		// First call function with skill_id 0 to get base damage of a normal attack
 		battle_calc_skill_base_damage(&wd, src, target, 0, 0); // base damage
 		wd.basedamage = wd.damage;
 		// Now get actual skill damage
 		if (skill_id != 0)
+#endif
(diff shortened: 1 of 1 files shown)


Playtester:
It's not really doubled. The first call is just to get the damage of a normal attack.

But it's not yet used in Renewal because someone would need to run tests on how the bonus damage for EDP and Magnum Break works exactly.

So I guess we could remove this part for renewal to save some calculation time.

datawulf:
> It's not really doubled. The first call is just to get the damage of a normal attack.
> 
> But it's not yet used in Renewal because someone would need to run tests on how the bonus damage for EDP and Magnum Break works exactly.
> 
> So I guess we could remove this part for renewal to save some calculation time.

This way all the renewal atk types calculation happens twice for skills. Once in the first call and again in the second. 
Can't say anything about pre-re but for renewal this breaks the damage.

Playtester:
> > It's not really doubled. The first call is just to get the damage of a normal attack.
> > But it's not yet used in Renewal because someone would need to run tests on how the bonus damage for EDP and Magnum Break works exactly.
> > So I guess we could remove this part for renewal to save some calculation time.
> 
> This way all the renewal atk types calculation happens twice for skills. Once in the first call and again in the second and they add up. Can't say anything about pre-re but for renewal this breaks the damage.

I see, it doesn't reset after the first call but adds the values together.

I think it would be best to define basedamage only for pre-re then and remove the parts from renewal that still access it.

Playtester (approved):

PR #8268, opened 2024-04-21
Cylinder Boots [1] fix

kaninhot004:
* **Addressed Issue(s)**:  -

* **Server Mode**:  Renewal

* **Description of Pull Request**: Cylinder Boots [1] has exploit bonuses.


idk-whoami:
confirm

Playtester:
That's not the only exploitable part on this item.

If you look at the script you can see the refine bonus is already applied, so the whole lower section for refining adds a duplicate bonus. For example HP/SP is +37% (20% + 7% + 10%) on refine+7 instead of +20% (3% + 7% + 10%).
And variable cast rate is reduced by 100% on refine+5 instead of by 50% (and by 200% on refine+7 instead of 100%).

Also would be nice to have the unslotted script to be consistent with the slotted one.

kaninhot004:
Done using [0] script.

PR #8275, opened 2024-04-23
Script correction of NPC ammo_boxes

idk-whoami:
* **Addressed Issue(s)**: 

* **Server Mode**: re


npc/re/merchants/ammo_boxes.txt
```diff
@@ -262,7 +262,7 @@ S_Slug_Exchange:
 				case 1:
 					mes "[Kenny]";
 					mes "All done.";
-					delitem .@slug_id, .@new_count * 500;
+					delitem .@slug_id, .@new_count * .@req_count;
```
Lemongrass3110, line 265: Should also be done above, where it is output via "mes" or not?
There `.@count` is used directly.

npc/re/merchants/ammo_boxes.txt
```diff
@@ -167,8 +167,8 @@ S_Sphere_Exchange:
 				case 1:
 					mes "[Kenny]";
 					mes "All done.";
-					delitem .@slug_id, .@new_count * 10;
-					getitem .@new_slug_id, .@old_count;
+					delitem .@old_id, .@new_count * 10;
```
Lemongrass3110, line 170: Should also be done above, where it is output via "mes" or not?
There `.@old_count` is used directly.

npc/re/merchants/ammo_boxes.txt
idk-whoami: if do that, output via "mes" is use .@count countitem, which doesn't match char inventory

what should I do? please advise

Atemo (approved):
Look good to me!

npc/re/merchants/ammo_boxes.txt
Atemo: About that, I don't know how it is on official, I'm in favor of leaving it like that for now

PR #8281, opened 2024-04-25
Small optimization on change sex

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Small optimization when changing sex from sex specific classes.

### DIFF
--- a/src/char/char_logif.cpp
+++ b/src/char/char_logif.cpp
@@ -91,7 +91,7 @@ void chlogif_pincode_start(int fd, struct char_session_data* sd){
  * @param tick : Scheduled tick
  * @param id : GID linked to that timered call
  * @param data : data transmited for delayed function
- * @return 
+ * @return
  */
 TIMER_FUNC(chlogif_send_acc_tologin){
 	if ( chlogif_isconnected() ){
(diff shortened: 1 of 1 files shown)


Atemo (approved):

aleos89 (approved):

aleos89 (approved):
A little longer but I see no issue in that since now it's much safer and less assumptions about the expected class.

SapitoSucio:
Could be done like this too, to shrink it a lil bit
```
switch (class_) {
	case JOB_BARD: class_ = JOB_DANCER; break;
	case JOB_DANCER: class_ = JOB_BARD; break;
	case JOB_CLOWN: class_ = JOB_GYPSY; break;
	...
```

Daegaladh:
> Could be done like this too, to shrink it a lil bit
> 
> ```
> switch (class_) {
> 	case JOB_BARD: class_ = JOB_DANCER; break;
> 	case JOB_DANCER: class_ = JOB_BARD; break;
> 	case JOB_CLOWN: class_ = JOB_GYPSY; break;
> 	...
> ```

I could, but I thought it would be clearer this way

aleos89:
> Could be done like this too, to shrink it a lil bit
> 
> ```
> 
> switch (class_) {
> 
> 	case JOB_BARD: class_ = JOB_DANCER; break;
> 
> 	case JOB_DANCER: class_ = JOB_BARD; break;
> 
> 	case JOB_CLOWN: class_ = JOB_GYPSY; break;
> 
> 	...
> 
> ```

As my professor always said, white space is free when developing!

PR #8286, opened 2024-04-28
Fix unlimited use of potion.

kaninhot004:

* **Server Mode**: Renewal

* **Description of Pull Request**: Fix unlimited item use exploit.


gmragnarok:
Is the correct script like this?

kaninhot004:
Not sure but at least it fix the unlimited use bug.

aleos89 (approved):
The only comment I have is if officially it's supposed to show the Kyrie skill effect or not.

db/re/item_db_usable.yml
```diff
@@ -14426,14 +14426,14 @@ Body:
...
-    Type: Delayconsume
+    Type: Usable
     Weight: 30
     Delay:
       Duration: 1000
       Status: Reuse_Limit_F
     Script: |
       percentheal 100,100;
-      itemskill "PR_KYRIE",10;
+      sc_start SC_KYRIE,100000,10;
```
Atemo, line 14436:
```suggestion
      sc_start SC_KYRIE,100000,10;
      skilleffect "PR_KYRIE",0;
```

Atemo (commented):
(the item was consumed after using the skill, not sure we can call it an exploit)

kaninhot004:
More explaination: Use this item will heal you 100% HP and SP then cancel the cast. Repeat this forever.

PR #8287, opened 2024-04-29
Coma Rework, Exp Calc Type, Fiber Lock Range

### LINKED ISSUE #8279: Coma effect and trigger conditions
### rAthena Hash

2963e52

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

Coma sets SP to 0.
Coma does not trigger on Provoke.
An attack that causes Coma still deals damage.

### Relevant Log Output

_No response_

### Expected Result

Coma sets SP to 1.
Coma can trigger on Provoke. It even triggers on yourself on Self Provoke, so using Aloevera can make you Coma yourself. It does not trigger on any other nodamage skills.
An attack that causes Coma will show normal damage but does not actually deal damage. If you had 100% Coma chance, you can effectively never kill anything.

### How to Reproduce

Set Coma chance on Lord of the Death Card to 100%. Then use Provoke / Aloevera or other nodamage skills.

Use Grand Cross at 100% Coma chance (you will not die from self damage).

### Official Information

Verified by various tests.

_No response_

Playtester:
* **Addressed Issue(s)**: #8279 

* **Server Mode**: Both

- Coma item bonus is no longer an additional effect / status change, but instead replaces the attack damage with setting HP/SP to 1
  * That means Coma attacks will never kill someone, even if the original attack would be lethal
  * When doing a Coma attack, the attack displays the original damage and this is also what counts towards the mob damage log
- Coma is checked for every single event that applies damage, including reflect damage, and for Provoke / Self Provoke
  * Some Renewal-only effects couldn't be tested and were left unchanged (e.g. damage from White Imprison)
- Coma now sets SP to 1 instead of 0
- bComaRace, bWeaponComaRace and bWeaponComaEle won't work on bosses, GVG and battlefield objects
- bComaClass and bWeaponComaClass no longer work on GVG and battlefield objects, but can work on bosses if you specify Class_Boss or Class_All
- Edge is now defined to only work on normal monster
- The default exp_calc_type 0 no longer counts the first attacker twice and instead works like on official servers
- Added a new exp_calc_type 2 that counts the first attacker twice as previously
- The damage logged will now include overkill damage on exp_calc_type 0 and 2, but not count Coma damage (it will count the damage you would have originally dealt instead)
- Fiber Lock now has a range of 9 instead of 7
- Fixes #8279

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -34,6 +34,8 @@
 #include "pc_groups.hpp"
 #include "pet.hpp"
 
+using namespace rathena;
+
 struct Battle_Config battle_config;
 static struct eri *delay_damage_ers; //For battle delay damage structures.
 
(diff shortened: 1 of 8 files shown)


src/map/battle.cpp
```diff
@@ -254,33 +256,42 @@ struct block_list* battle_getenemyarea(struct block_list *src, int x, int y, int
...
 * @param isspdamage: If the damage is done to SP
 * @param tick: Current tick
-*------------------------------------------*/
-void battle_damage(struct block_list *src, struct block_list *target, int64 damage, t_tick delay, uint16 skill_lv, uint16 skill_id, enum damage_lv dmg_lv, unsigned short attack_type, bool additional_effects, t_tick tick, bool isspdamage) {
+* @return HP+SP+AP (0 if HP/SP/AP remained unchanged)
+*/
+int battle_damage(struct block_list *src, struct block_list *target, int64 damage, t_tick delay, uint16 skill_lv, uint16 skill_id, enum damage_lv dmg_lv, unsigned short attack_type, bool additional_effects, t_tick tick, bool isspdamage) {
+	int dmg_change;
+	map_session_data* sd = nullptr;
+	if (src)
```
aleos89, line 280: Can src be null? If not just move this into the sd define. It's not checked elsewhere so it would crash for any other case.

aleos89 (approved):

src/map/mob.cpp
```diff
@@ -2675,14 +2674,16 @@ int mob_dead(struct mob_data *md, struct block_list *src, int type)
...
+			if (battle_config.exp_calc_type == 1 && md->tdmg) {
+				// eAthena's exp formula based on max hp
+				per = (double)md->dmglog[i].dmg / (double)status->max_hp;
+			}
 			else {
-				//eAthena's exp formula based on max hp.
-				per = (double)md->dmglog[i].dmg/(double)status->max_hp;
-				if (per > 2) per = 2; // prevents unlimited exp gain
+				// Aegis's exp formula based on total damage
+				per = (double)md->dmglog[i].dmg / (double)md->tdmg;
```
Atemo, line 2683: `md->tdmg` is no longer checked against 0 here

src/map/battle.cpp
Playtester: src can be NULL if you don't want the attack to log the damage. There were actually a few calls that passed NULL into status_damage before, but I guess I replaced them now.

I think the function should work if src is NULL though, so it shouldn't crash.

src/map/mob.cpp
Playtester: Good spot, will fix that.

PR #8290, opened 2024-05-01
Fixed warning about favorite item

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: renewal

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -2810,7 +2810,7 @@ void clif_additem( map_session_data *sd, int n, int amount, unsigned char fail )
 #if PACKETVER >= 20150226
 		clif_add_random_options( p.option_data, sd->inventory.u.items_inventory[n] );
 #if PACKETVER >= 20160921
-		p.favorite = sd->inventory.u.items_inventory[n].favorite;
+		p.favorite = ( sd->inventory.u.items_inventory[n].favorite != 0 ) ? 1 : 0;
 		p.look = sd->inventory_data[n]->look;
 #if PACKETVER_MAIN_NUM >= 20200916 || PACKETVER_RE_NUM >= 20200724
 		p.grade = sd->inventory.u.items_inventory[n].enchantgrade;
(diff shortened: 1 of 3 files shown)


aleos89 (approved):

src/map/clif.cpp
```diff
@@ -15832,7 +15832,7 @@ void clif_Mail_setattachment( map_session_data* sd, int index, int amount, uint8
 
 			p.weight += sd->mail.item[i].amount * ( sd->inventory_data[sd->mail.item[i].index]->weight / 10 );
 		}
-		p.favorite = item->favorite;
+		p.favorite = ( item->favorite != 0 ) ? 1 : 0;
```
Daegaladh, line 15835:
```suggestion
		p.favorite = ( item->favorite != 0 ) ? true : false;
```

src/map/clif.cpp
```diff
@@ -2810,7 +2810,7 @@ void clif_additem( map_session_data *sd, int n, int amount, unsigned char fail )
 #if PACKETVER >= 20150226
 		clif_add_random_options( p.option_data, sd->inventory.u.items_inventory[n] );
 #if PACKETVER >= 20160921
-		p.favorite = sd->inventory.u.items_inventory[n].favorite;
+		p.favorite = ( sd->inventory.u.items_inventory[n].favorite != 0 ) ? 1 : 0;
```
Daegaladh, line 2813:
```suggestion
		p.favorite = ( sd->inventory.u.items_inventory[n].favorite != 0 ) ? true : false;
```

Daegaladh (changes requested):
Since p.favorite is bool, i'd prefer this style to differentiate from the other char and int types

src/map/clif.cpp
Atemo: That can't be, favorite from `PACKET_ZC_ITEM_PICKUP_ACK` is uint8
https://github.com/rathena/rathena/blob/master/src/map/packets_struct.hpp#L572

src/map/clif.cpp
Atemo: Same as above, favorite from `PACKET_ZC_ACK_ADD_ITEM_RODEX` is uint8

src/map/clif.cpp
Daegaladh: you're right sorry, I mistook then with PACKET_ZC_INVENTORY_TAB p.favorite

Daegaladh (approved):

PR #8305, opened 2024-05-04
Synchronize Damage Feature

### LINKED ISSUE #259: Movement desync
Client Date: **2012-04-10**
Src Mods: **NONE**
General settings: 
- **Full Renewal**
- **1/1/1/1/1 rate**

How to reproduce & Description:

The dev guys already corrected a lot of things regarding the movement lag/desync issue, but still there is at least one issue:

When I get hit and in my desire to break free I click somewhere to move, the character displays the flintch movement for it's whole duration, but in the meantime the server receives my movement commands and starts to move my character on the server side. On the client side this results in my character jumping from one location to another.

So, the client we usually use is an official one, with some minor modifications. On the other hand, the server side was written from sketch. So I think it is logical to say, that we should follow the client's "guide" if we want to mimic the official behav. If the client says "you shouldn't move during hitlock", then why do we allow the server side to move the player? Or perhaps we should allow the player to move, because the client still sends the move commands during hitlock, but we miss a packet that tells the client to cancel the flintch movement. Either way, it is pretty annoying that if a single mob hits us during movement, we desync in movement.

Another theory I have is that in case of monsters, the hitlock works better, becuause they simply won't issue a new movetoxy() command during their hitlock/flintch time, but since players can click whenever they wish, they do issue a move command in the urge to break free, and maybe this case is not handled correctly.

Playtester:
* **Addressed Issue(s)**: #259 

* **Server Mode**: Both

- Added a new monster stat "ClientAttackMotion" to mob_db.yml which is the time from when a monster attacks until which the damage shows on the client at 1x speed
- Added a new config synchronize_damage; when set to "yes", the client will display the damage of normal monster attacks at the exact time it is applied on the server, removing position lag (fixes #259)

### DIFF
--- a/conf/battle/battle.conf
+++ b/conf/battle/battle.conf
@@ -130,9 +130,18 @@ equip_self_break_rate: 100
 // This affects the behaviour of skills like acid terror and meltdown
 equip_skill_break_rate: 100
 
-// Do weapon attacks have a attack speed delay before actual damage is applied? (Note 1)
-// NOTE: The official setting is yes, even thought it degrades performance a bit.
+// Should damage have a delay before it is applied? (Note 1)
+// Some skills might not have a delay by default regardless of this setting.
+// The official setting is yes, even thought it degrades performance a bit.
 delay_battle_damage: yes
+
+// Should the damage timing be synchronized between the client and server? (Note 1)
+// This is not official behavior, but it removes the position lag after being hit.
+// This setting only affects normal monster attacks and takes priority over "delay_battle_damage".
+// Many skills show their damage immediately, so setting "delay_battle_damage" to "no" at the same
+// time might improve the experience further, but will not work for all skills.
+// Tired of Dark Illusion hitting you 5 seconds too late? Then turn this on.
+synchronize_damage: no
 
 // Are arrows/ammo consumed when used on a bow/gun?
 // 0 = No
--- a/conf/battle/battle.conf
+++ b/conf/battle/battle.conf
@@ -136,7 +136,7 @@ equip_skill_break_rate: 100
 delay_battle_damage: yes
 
 // Should the damage timing be synchronized between the client and server? (Note 1)
-// This is not official behavior, but it removes the position lag after being hit.
+// This is not official behavior, but it should remove the position lag after being hit by a monster.
 // This setting only affects normal monster attacks and takes priority over "delay_battle_damage".
 // Many skills show their damage immediately, so setting "delay_battle_damage" to "no" at the same
 // time might improve the experience further, but will not work for all skills.
--- a/conf/battle/client.conf
+++ b/conf/battle/client.conf
@@ -169,3 +169,9 @@ macro_detection_punishment: 0
 // Amount of time in minutes that the punishment type is active for. Use 0 for infinite.
 // Official: 0
 macro_detection_punishment_time: 0
+
+// Should the damage timing be synchronized between the client and server?
+// This is not official behavior, but it removes the position lag after being hit.
+// Right now it only works for being hit by normal monster attacks.
+// Tired of Dark Illusion hitting you 5 seconds too late? Then turn this on. (Note 1)
+synchronize_damage: no
--- a/conf/battle/client.conf
+++ b/conf/battle/client.conf
@@ -169,10 +169,3 @@ macro_detection_punishment: 0
 // Amount of time in minutes that the punishment type is active for. Use 0 for infinite.
 // Official: 0
 macro_detection_punishment_time: 0
-
-// Should the damage timing be synchronized between the client and server?
-// This is not official behavior, but it removes the position lag after being hit.
-// This setting only affects normal monster attacks, unless you set "delay_battle_damage" in battle.conf to "no".
-// If you do, most skills will also be in-sync as skills usually show their damage immediately.
-// Tired of Dark Illusion hitting you 5 seconds too late? Then turn this on. (Note 1)
-synchronize_damage: no
--- a/conf/battle/client.conf
+++ b/conf/battle/client.conf
@@ -172,6 +172,7 @@ macro_detection_punishment_time: 0
 
 // Should the damage timing be synchronized between the client and server?
 // This is not official behavior, but it removes the position lag after being hit.
-// Right now it only works for being hit by normal monster attacks.
+// This setting only affects normal monster attacks, unless you set "delay_battle_damage" in battle.conf to "no".
+// If you do, most skills will also be in-sync as skills usually show their damage immediately.
 // Tired of Dark Illusion hitting you 5 seconds too late? Then turn this on. (Note 1)
 synchronize_damage: no
--- a/doc/mob_db.txt
+++ b/doc/mob_db.txt
@@ -205,15 +205,14 @@ AttackMotion: Attack animation motion. Low value means monster's attack will be 
 ---------------------------------------
 
 ClientAttackMotion: The time from the start of a normal attack until the damage frame shows on client. At the same time you also get stopped on the client.
-This value is only needed if you use the "synchronize_damage" feature (battle/client.conf).
-If you created a custom sprite, you want to set this value to the timing of the damage frame in your *.act file.
+This value is only needed if you use the "synchronize_damage" feature (battle/battle.conf).
 
-Background:
-When a clif_damage packet is sent to the client it will also send "sdelay" (AttackMotion) as value.
-The client will ignore all values above 432, and 432 stands for 1.0x animation speed. 216 for example means play the animation at double the speed.
-Each monster has an attack animation and can define the frame in the attack animation on which it displays the damage and makes the target flinch / stop.
-If the damage frame is undefined, it instead displays the damage / flinch / stop at the very end of the animation.
-Knowing when the damage frame happens in the animation allows us to synchronize the timing between client and server.
+If you created a custom sprite, you want to set this value to the timing of the damage frame in your *.act file.
+In Act Editor you can set the damage frame by setting the frame sound to "atk". If you don't define a damage frame, it will default to the second to last
+frame. Also keep in mind that the Act Editor displays slightly inaccurate speed. Every 25ms in Act Editor is 24ms in reality.
+
+Example: Drops has a animation speed of 24ms per frame and the 13th frame (frame #12) is the damage frame. This means the damage shows after 12 frames.
+That's why Drops has a ClientAttackMotion of 24*12 = 288.
 
 ---------------------------------------
 
--- a/doc/mob_db.txt
+++ b/doc/mob_db.txt
@@ -204,6 +204,14 @@ AttackMotion: Attack animation motion. Low value means monster's attack will be 
 
 ---------------------------------------
 
+ClientAttackMotion: When a clif_damage packet is sent to the client it will also send "sdelay" (AttackMotion) as value.
+
+The client will ignore all values above 432, and 432 stands for 1.0x animation speed. 216 for example means play the animation at double the speed.
+Each monster has an attack animation and can define the frame in the attack animation on which it displays the damage and makes the target flinch / stop.
+If the damage frame is undefined, it instead displays the damage / flinch / stop at the very end of the animation.
+
+---------------------------------------
+
 DamageMotion: Damage animation motion, same as aMotion but used to display the "I am hit" animation. Coincidentally, this same value is used to determine how long it is before the monster/player can move again. Endure is dMotion = 0, obviously.
 
 ---------------------------------------
--- a/doc/mob_db.txt
+++ b/doc/mob_db.txt
@@ -204,11 +204,16 @@ AttackMotion: Attack animation motion. Low value means monster's attack will be 
 
 ---------------------------------------
 
-ClientAttackMotion: When a clif_damage packet is sent to the client it will also send "sdelay" (AttackMotion) as value.
+ClientAttackMotion: The time from the start of a normal attack until the damage frame shows on client. At the same time you also get stopped on the client.
+This value is only needed if you use the "synchronize_damage" feature (battle/client.conf).
+If you created a custom sprite, you want to set this value to the timing of the damage frame in your *.act file.
 
+Background:
+When a clif_damage packet is sent to the client it will also send "sdelay" (AttackMotion) as value.
 The client will ignore all values above 432, and 432 stands for 1.0x animation speed. 216 for example means play the animation at double the speed.
 Each monster has an attack animation and can define the frame in the attack animation on which it displays the damage and makes the target flinch / stop.
 If the damage frame is undefined, it instead displays the damage / flinch / stop at the very end of the animation.
+Knowing when the damage frame happens in the animation allows us to synchronize the timing between client and server.
 
 ---------------------------------------
 
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -416,7 +416,7 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
 	dat->additional_effects = additional_effects;
 	dat->src_type = src->type;
 	dat->isspdamage = isspdamage;
-	if (skill_id == 0 && src->type == BL_MOB && battle_config.synchronize_damage && amotion > status_get_clientamotion(src)) {
+	if (battle_config.synchronize_damage && skill_id == 0 && src->type == BL_MOB && amotion > status_get_clientamotion(src)) {
 		// The client refuses to display animations slower than 1x speed
 		// So we need to shorten AttackMotion to be in-sync with the client in this case
 		amotion = status_get_clientamotion(src);
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -393,8 +393,12 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
 		damage = 0;
 	}
 
-	// If the synchronize damage feature is activated then disabling delay battle damage should only affect skills
-	if ((!battle_config.delay_battle_damage && (!battle_config.synchronize_damage || skill_id != 0)) || amotion <= 1) {
+	// If the synchronize damage feature is activated then disabling delay battle damage should only affect skills (for monsters)
+	// If clientamotion is 1 or lower for monsters, we can apply damage directly for normal attacks and don't need to create a timer
+	// Same applies for all types, normal attacks and skills if amotion is 1 or lower
+	if ((!battle_config.delay_battle_damage && (src->type != BL_MOB || !battle_config.synchronize_damage || skill_id != 0))
+		|| (skill_id == 0 && src->type == BL_MOB && battle_config.synchronize_damage && status_get_clientamotion(src) <= 1)
+		|| amotion <= 1) {
 		//Deal damage
 		battle_damage(src, target, damage, ddelay, skill_lv, skill_id, dmg_lv, attack_type, additional_effects, gettick(), isspdamage);
 		return 0;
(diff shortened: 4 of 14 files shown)


Playtester:
Note: The value for Anubis in the mob_db.yml file seems to be incorrect, @aleos89 is currently investigating this.

src/map/status.hpp
```diff
@@ -3172,7 +3172,7 @@ struct status_data {
 #endif
 		matk_min, matk_max,
 		speed,
-		amotion, adelay, dmotion;
+		amotion, clientamotion, adelay, dmotion;
```
aleos89, line 3175:
```suggestion
		amotion, client_amotion, adelay, dmotion;
```
If `clientamotion` is too hard to read without snake-case then feel free to adjust it!

src/map/battle.cpp
```diff
@@ -411,7 +411,14 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
 	dat->additional_effects = additional_effects;
 	dat->src_type = src->type;
 	dat->isspdamage = isspdamage;
-	if (src->type != BL_PC && amotion > 1000)
+	if (skill_id == 0 && src->type == BL_MOB && battle_config.synchronize_damage) {
+		mob_data* md = BL_CAST(BL_MOB, src);
+		// The client refuses to display animations slower than 1x speed
+		// So we need to shorten AttackMotion to be in-sync with the client in this case
+		if (amotion > md->status.clientamotion)
```
Atemo, line 418:
```suggestion
		if (md != nullptr && amotion > md->status.clientamotion)
```

src/map/clif.cpp
```diff
@@ -5242,6 +5242,13 @@ int clif_damage(struct block_list* src, struct block_list* dst, t_tick tick, int
 		}
 	}
 
+	// Calculate what sdelay to send to the client so it applies damage at the same time as the server
+	if (src->type == BL_MOB && battle_config.synchronize_damage) {
+		mob_data* md = BL_CAST(BL_MOB, src);
+		// The client only accepts values between 0 and 432; 432 stands for 1x animation speed
+		sdelay = sdelay * 432 / md->status.clientamotion;
```
Atemo, line 5249:
```suggestion
		if (md != nullptr && md->status.clientamotion != 0)
			sdelay = sdelay * 432 / md->status.clientamotion;
```

src/map/mob.cpp
```diff
@@ -4811,6 +4811,18 @@ uint64 MobDatabase::parseBodyNode(const ryml::NodeRef& node) {
...
+	if (this->nodeExists(node, "ClientAttackMotion")) {
+		uint16 speed;
+
+		if (!this->asUInt16(node, "ClientAttackMotion", speed))
+			return 0;
+
+		mob->status.clientamotion = cap_value(speed, 1, USHRT_MAX);
+	} else {
+		if (!exists)
+			mob->status.clientamotion = cap_value(mob->status.amotion, 1, USHRT_MAX);
```
Atemo, line 4823: How about putting a default in s_mob_db constructor for cases like player clone?

SapitoSucio:
This seems nice :O

src/map/mob.cpp
Playtester: @Atemo I assume that would only work for fixed values, but we could put 432 as default there I guess?

src/map/battle.cpp
Playtester: I was wondering if md can be nullptr at all because I check for bl->type beforehand.

src/map/status.hpp
Playtester: I think it's fine.

src/map/clif.cpp
Lemongrass3110: Here it cannot be null, as "src" was already accessed before and "type" was already verified

src/map/battle.cpp
```diff
@@ -411,7 +416,12 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
 	dat->additional_effects = additional_effects;
 	dat->src_type = src->type;
 	dat->isspdamage = isspdamage;
-	if (src->type != BL_PC && amotion > 1000)
+	if (skill_id == 0 && src->type == BL_MOB && battle_config.synchronize_damage && amotion > status_get_clientamotion(src)) {
```
Lemongrass3110, line 419:
```suggestion
	if (battle_config.synchronize_damage && skill_id == 0 && src->type == BL_MOB && amotion > status_get_clientamotion(src)) {
```
Feature check first, so that it fails as fast as possible.

src/map/battle.cpp
```diff
@@ -393,7 +393,12 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
 		damage = 0;
 	}
 
-	if ( !battle_config.delay_battle_damage || amotion <= 1 ) {
+	// If the synchronize damage feature is activated then disabling delay battle damage should only affect skills (for monsters)
+	// If clientamotion is 1 or lower for monsters, we can apply damage directly for normal attacks and don't need to create a timer
+	// Same applies for all types, normal attacks and skills if amotion is 1 or lower
+	if ((!battle_config.delay_battle_damage && (src->type != BL_MOB || !battle_config.synchronize_damage || skill_id != 0))
+		|| (skill_id == 0 && src->type == BL_MOB && battle_config.synchronize_damage && status_get_clientamotion(src) <= 1)
+		|| amotion <= 1) {
```
Lemongrass3110, line 401: This if is too long again.
It also changes existing behavor.

src/map/clif.cpp
```diff
@@ -5242,6 +5242,12 @@ int clif_damage(struct block_list* src, struct block_list* dst, t_tick tick, int
 		}
 	}
 
+	// Calculate what sdelay to send to the client so it applies damage at the same time as the server
+	if (src->type == BL_MOB && battle_config.synchronize_damage && status_get_clientamotion(src) != 0) {
```
Lemongrass3110, line 5246:
```suggestion
	if (battle_config.synchronize_damage && src->type == BL_MOB && status_get_clientamotion(src) != 0) {
```
Feature check first, so that it fails as fast as possible.

src/map/clif.cpp
```diff
@@ -5242,6 +5242,12 @@ int clif_damage(struct block_list* src, struct block_list* dst, t_tick tick, int
...
+		// The client only accepts values between 0 and 432; 432 stands for 1x animation speed
+		sdelay = sdelay * 432 / status_get_clientamotion(src);
```
Lemongrass3110, line 5248: Would be nice to define a constant somewhere.

src/map/mob.cpp
```diff
@@ -4811,6 +4812,18 @@ uint64 MobDatabase::parseBodyNode(const ryml::NodeRef& node) {
...
```
Lemongrass3110, line 4824:
```suggestion
			mob->status.clientamotion = cap_value(mob->status.amotion, 1, 432);
```
Or a newly introduced constant.

src/map/mob.cpp
```diff
@@ -4811,6 +4812,18 @@ uint64 MobDatabase::parseBodyNode(const ryml::NodeRef& node) {
 		mob->status.amotion = cap_value(speed, battle_config.monster_max_aspd, 2000);
 	}
 
+	if (this->nodeExists(node, "ClientAttackMotion")) {
+		uint16 speed;
+
+		if (!this->asUInt16(node, "ClientAttackMotion", speed))
+			return 0;
+
+		mob->status.clientamotion = cap_value(speed, 1, USHRT_MAX);
```
Lemongrass3110, line 4821: We should not use cap value here.
Give a proper warning for 0 or > 432 (or newly defined constant).

Lemongrass3110 (changes requested):

src/map/battle.cpp
Playtester: It does not change existing behavior. You'd need to enable the new feature for the behavior to change.

But maybe should just adjust amotion in more simple checks beforehand.

src/map/mob.cpp
Playtester: Needs to stay capped to USHRT_MAX or it won't work. There are plenty of official values higher than 432.

src/map/mob.cpp
Playtester: Those values do exist in the files. The reason I cap 0 to 1 is simply to prevent the risk of division by 0.
Values above 432 are valid.

src/map/clif.cpp
Playtester: Yeah we can do that, but this is the only place where it's used.

src/map/battle.cpp
Playtester: Split into 3 if() checks.

src/map/clif.cpp
Playtester: Added constant.

conf/battle/client.conf
```diff
@@ -169,3 +169,10 @@ macro_detection_punishment: 0
 // Amount of time in minutes that the punishment type is active for. Use 0 for infinite.
 // Official: 0
 macro_detection_punishment_time: 0
+
+// Should the damage timing be synchronized between the client and server?
+// This is not official behavior, but it removes the position lag after being hit.
+// This setting only affects normal monster attacks, unless you set "delay_battle_damage" in battle.conf to "no".
+// If you do, most skills will also be in-sync as skills usually show their damage immediately.
+// Tired of Dark Illusion hitting you 5 seconds too late? Then turn this on. (Note 1)
+synchronize_damage: no
```
Lemongrass3110, line 178: Should be moved to conf/battle/battle.conf right after delay_battle_damage

src/map/clif.cpp
```diff
@@ -5242,6 +5244,12 @@ int clif_damage(struct block_list* src, struct block_list* dst, t_tick tick, int
 		}
 	}
 
+	// Calculate what sdelay to send to the client so it applies damage at the same time as the server
+	if (battle_config.synchronize_damage && src->type == BL_MOB && status_get_clientamotion(src) != 0) {
```
Lemongrass3110, line 5248: status_get_clientamotion will always return != 0, since you cap it to 1

src/map/battle.cpp
```diff
@@ -393,7 +393,16 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
...
+	// Check for delay battle damage config
+	// If the synchronize damage feature is activated then disabling the config should only affect skills (for monsters)
+	if (!battle_config.delay_battle_damage && (!battle_config.synchronize_damage || skill_id != 0 || src->type != BL_MOB))
+		amotion = 1;
+	// If clientamotion is 1 or lower for monsters, we can apply damage directly for normal attacks and don't need to create a timer
+	else if (battle_config.synchronize_damage && skill_id == 0 && src->type == BL_MOB && status_get_clientamotion(src) <= 1)
+		amotion = 1;
+
+	// Skip creation of timer
+	if (amotion <= 1) {
```
Lemongrass3110, line 405:
```suggestion
	// The client refuses to display animations slower than 1x speed
	// So we need to shorten AttackMotion to be in-sync with the client in this case
	if( battle_config.synchronize_damage && skill_id == 0 && src->type == BL_MOB && amotion > status_get_clientamotion(src) ){
		amotion = status_get_clientamotion(src);
	}

	if ( !battle_config.delay_battle_damage || amotion <= 1 ) {
```

src/map/battle.cpp
```diff
@@ -411,7 +420,12 @@ int battle_delay_damage(t_tick tick, int amotion, struct block_list *src, struct
 	dat->additional_effects = additional_effects;
 	dat->src_type = src->type;
 	dat->isspdamage = isspdamage;
-	if (src->type != BL_PC && amotion > 1000)
+	if (battle_config.synchronize_damage && skill_id == 0 && src->type == BL_MOB && amotion > status_get_clientamotion(src)) {
+		// The client refuses to display animations slower than 1x speed
+		// So we need to shorten AttackMotion to be in-sync with the client in this case
+		amotion = status_get_clientamotion(src);
+	}
+	else if (src->type != BL_PC && amotion > 1000)
```
Lemongrass3110, line 428:
```suggestion
	if (src->type != BL_PC && amotion > 1000)
```

src/map/clif.cpp
Lemongrass3110, line 5248:
```suggestion
	if (battle_config.synchronize_damage && src->type == BL_MOB) {
```
status_get_clientamotion will always return != 0, since you cap it to 1

src/map/clif.cpp
```diff
@@ -5242,6 +5244,12 @@ int clif_damage(struct block_list* src, struct block_list* dst, t_tick tick, int
...
+		// The client only accepts values between 0 and 432; 432 stands for 1x animation speed
```
Lemongrass3110, line 5249: Your comment does not make sense to me.
I tried to understand it with the first monster (SCORPION).
AttackMotion 864 (=sdelay) * DEFAULT_ANIMATION_SPEED (=432) / ClientAttackMotion (=384) => 972
But 972 is out of the range of 0-432.

doc/mob_db.txt
```diff
@@ -204,6 +204,19 @@ AttackMotion: Attack animation motion. Low value means monster's attack will be
 
 ---------------------------------------
 
+ClientAttackMotion: The time from the start of a normal attack until the damage frame shows on client. At the same time you also get stopped on the client.
+This value is only needed if you use the "synchronize_damage" feature (battle/client.conf).
+If you created a custom sprite, you want to set this value to the timing of the damage frame in your *.act file.
+
+Background:
+When a clif_damage packet is sent to the client it will also send "sdelay" (AttackMotion) as value.
+The client will ignore all values above 432, and 432 stands for 1.0x animation speed. 216 for example means play the animation at double the speed.
```
Lemongrass3110, line 213: I know what you are trying to say, but reading this explanation here gives the impression that I should set ClientAttackMotion to a maximum of 432 and if I want to double its speed I should set it to 216.
But in reality you are talking about something totally different.
Please consider moving this (useful) documentation into clif, as it may be confusing for "normal" users here.

conf/battle/client.conf
```diff
@@ -169,3 +169,10 @@ macro_detection_punishment: 0
 // Amount of time in minutes that the punishment type is active for. Use 0 for infinite.
 // Official: 0
 macro_detection_punishment_time: 0
+
+// Should the damage timing be synchronized between the client and server?
+// This is not official behavior, but it removes the position lag after being hit.
+// This setting only affects normal monster attacks, unless you set "delay_battle_damage" in battle.conf to "no".
```
Lemongrass3110, line 175:
```
// Do weapon attacks have a attack speed delay before actual damage is applied? (Note 1)
// NOTE: The official setting is yes, even thought it degrades performance a bit.
delay_battle_damage: yes
```
delay_battle_damage is currently not matching the description, as it affects all attacks and not only "weapon attacks".

Lemongrass3110 (changes requested):

src/map/clif.cpp
Playtester: Should still check for it since we are making a division.

Playtester: Should still check for it as we divide by it.

src/map/clif.cpp
Playtester: It's fine to send a higher value, it will still just be 1x animation speed.
But maybe I should explain that here.

doc/mob_db.txt
Playtester: You want documentation to be in the source file rather than the documentation folder?

conf/battle/client.conf
Playtester: Yeah, it was wrong before already. I didn't really touch the functionality of this feature. Should we update the description as part of this PR?

src/map/battle.cpp
Playtester: I don't understand this change, doesn't this just remove the whole feature of making sychronize_damage work with delay_battle_damage?

src/map/battle.cpp
Playtester: We need to put the if(battle_config.synchronize_damage [...]" part here because otherwise you'd cap amotion to 1000 even if the feature applies.

src/map/battle.cpp
Lemongrass3110: Either you want to delay or you want to synchronize. You cannot have both.
If whoever enables your new setting, they want synchronized damage and not delayed damage.
So instead of hiding that we give the new feature priority, we just straight forward go for it.

(And additionally we dont touch the old feature - we may have to find out, when it was broken - or if it has always been for any damage, but outside of the scope of this PR)

src/map/clif.cpp
Lemongrass3110: Then put std::max into the formula, if you want to be safe.
But not checking here for something that cannot happen and hiding the actual reason behind the check (division by zero safeguard vs value not set).

src/map/battle.cpp
Lemongrass3110: Everything above 432 (so 1000) will be ignored by the client as you commented.
So everything is fine.

src/map/clif.cpp
Lemongrass3110: std::max

doc/mob_db.txt
Lemongrass3110: Well not 100% sure, but since it has nothing to do with "to what should i set that value" I was thinking that may be the better place.

conf/battle/client.conf
Playtester: Quick update: I'll rewrite it. Also the new config I move here. So makes more sense when read together.

conf/battle/client.conf
Playtester: Originally placed it here because I thought while delay_battle_damage affects more the functionality, the new feature more affects the client display. But considering that both can affect the amotion value and also have some synergy, it makes sense, I'll move it together.

src/map/battle.cpp
Playtester: Hmm, but it's important to give the player the option to combine the two settings. Basically battle_sychronize_damage only affects normal skills so of course if you enable this you want normal attacks to be synchronized, which should take priority over the other feature (I'll make it clearer in the description that it takes priority).

At the same time, we want to give the user the choice if they want to delay skill damage (and player damage) or not.

The current implementation from me does not affect the other feature as long as you keep it disabled, so it isn't a problem of backward compatibility.

src/map/battle.cpp
Playtester: I think you are missing that the amotion is something completely different from the value we want to send to the client.

Amotion is the time in milliseconds until which the damage shows (here it's only used to create the timer). This isn't limited to 432.

What is (technically) limited to 432 is the what we send to the client, but this is not a value in milliseconds but rather the inverted animation speed.

One monster where this is relevant would be Medusa:
```
    AttackMotion: 1320
    ClientAttackMotion: 1080
```
Here we want amotion in battle.cpp to be 1080.

The value we send to the client would be 432 (or higher), so it plays at 1x speed which results in 1080ms until the damage shows.

src/map/clif.cpp
Playtester: Okay can do that.

Playtester: Will accept suggestion and add max in next commit.

doc/mob_db.txt
Playtester: I improved the doc to focus on how to set the value rather than how it works internally here.

src/map/battle.cpp
Playtester: Actually after sleeping about it, why don't we move the 1000 limit also to the top? Then we can simplify the code as you suggested without breaking the feature.

If I look at the code "amotion" is not used anywhere except the <=1 check and for the timer at the end.

Playtester: Done!

conf/battle/battle.conf
```diff
@@ -130,10 +130,19 @@ equip_self_break_rate: 100
...
 
-// Do weapon attacks have a attack speed delay before actual damage is applied? (Note 1)
-// NOTE: The official setting is yes, even thought it degrades performance a bit.
+// Should damage have a delay before it is applied? (Note 1)
+// Some skills might not have a delay by default regardless of this setting.
+// The official setting is yes, even thought it degrades performance a bit.
 delay_battle_damage: yes
 
+// Should the damage timing be synchronized between the client and server? (Note 1)
+// This is not official behavior, but it removes the position lag after being hit.
```
Lemongrass3110, line 139:
```suggestion
// This is not official behavior, but it should remove the position lag after being hit by a monster.
```

conf/battle/battle.conf
```diff
@@ -130,10 +130,19 @@ equip_self_break_rate: 100
...
+// This setting only affects normal monster attacks and takes priority over "delay_battle_damage".
+// Many skills show their damage immediately, so setting "delay_battle_damage" to "no" at the same
+// time might improve the experience further, but will not work for all skills.
```
Lemongrass3110, line 142:
```suggestion
// Many skills show their damage immediately, so setting "delay_battle_damage" to "no" at the same
// time might improve the experience further, but will not work for all skills.
```
Could we not add a check for that as well? Maybe in another PR?
Basically we would just have to replace the skill_id == 0 check with a skill_id == 0 || skill_get_delay == 0 (pseudo) check?

Lemongrass3110 (approved):
Approved with the changes discussed in Discord.

conf/battle/battle.conf
Playtester: The correct way to implement this for skills is to define the clientamotion for each skill and then use this information when sychnronize_damage is enabled. Would be great, but currently we don't have this information available.

Issue #8307, opened 2024-05-05
clif_wis_message make server crash


mrpond:
### rAthena Hash

fd36fc04b2dcb3806ae130f5a89d09378327cb18

### Client Date

>20131120

### Server Mode

Pre-Renewal, Renewal

### Result

Lack of checking for nullptr.

### Expected Result

Fix
```
void clif_wis_message(map_session_data* sd, const char* nick, const char* mes, size_t mes_len, int gmlvl){

	nullpo_retv(sd);

	PACKET_ZC_WHISPER p;

	p.PacketType = HEADER_ZC_WHISPER;
	p.PacketLength = static_cast<decltype(p.PacketLength)>( mes_len );
	safestrncpy(p.sender, nick, NAME_LENGTH);
	safestrncpy(p.message, mes, mes_len);

#if PACKETVER >= 20091104
	map_session_data* ssd = map_nick2sd(nick, false);
	// If it is not a message from the server or a player from another map-server
	if (ssd) {
		gmlvl = pc_get_group_level(ssd);
#if PACKETVER_MAIN_NUM >= 20131204 || PACKETVER_RE_NUM >= 20131120 || defined(PACKETVER_ZERO)
		p.senderGID = ssd->bl.id;
	}
	else {
		p.senderGID = 0;
#endif
	}
	p.isAdmin = (gmlvl == 99) ? 1 : 0;
#endif

	clif_send( &p, sizeof( p ), &sd->bl, SELF );
}
```

### How to Reproduce

Just do whisper with some server whisper.

### Official Information

N/A

mrpond:
p->senderGID = ssd->bl.id; 
&sd->bl;
both sd and ssd can be nullptr server will crash. need put some check into it.

PR #8308, opened 2024-05-05
Fixed clif_wis_message

Lemongrass3110:
* **Addressed Issue(s)**: #8307

* **Server Mode**: Both

Packet was created in a static way, but it is actually a dynamic packet.
Thanks to @mrpond


mrpond:
p->senderGID = ssd->bl.id;
&sd->bl;
both sd and ssd can be nullptr server will crash. need put some check into it.

PR #8322, opened 2024-05-10
Attacking While Casting Crash Fix

### LINKED ISSUE #8319: Crash when mob cast skill
### rAthena Hash

13bc3e2bb765f64266ba56dc0cc06a1b99c09fc5

### Client Date

20131120

### Server Mode

Pre-Renewal

### Result

image

this is my temp fix
```
static int8 skill_castend_id_check(struct block_list *src, struct block_list *target, uint16 skill_id, uint16 skill_lv) {
	std::shared_ptr<s_skill_db> skill = skill_db.find(skill_id);

#ifdef 1
	if (nullptr == skill) // invalid skill id
		return USESKILL_FAIL_MAX;
#endif
	...
```

### Relevant Log Output

```Shell
See picture above
```

### Expected Result

It should not crash

### How to Reproduce

When mob cast character lock-on skill, map will crash
maybe unit.cpp line 2857 cause this.

### Official Information

N/A

#define RENEWAL_CAST

Playtester:
* **Addressed Issue(s)**: #8319 

* **Server Mode**: Both

- Fixed a server crash that occurs when a unit did a normal attack while casting a spell (e.g. Free Cast)
- Follow-up to ffe40de
- Fixes #8319

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -2855,7 +2855,10 @@ static int unit_attack_timer_sub(struct block_list* src, int tid, t_tick tick)
 			return 1;
 
 		ud->attackabletime = tick + sstatus->adelay;
-		ud->skill_id = 0;
+
+		// Only reset skill_id here if no skilltimer is currently ongoing
+		if (ud->skilltimer == INVALID_TIMER)
+			ud->skill_id = 0;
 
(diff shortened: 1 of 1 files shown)


aleos89 (approved):

rotloso:
Tested ok, Thank you

PR #8324, opened 2024-05-11
Spawn and Free Cell Search Behavior

### LINKED ISSUE #8300: Monsters should not spawn on unwalkable tiles
### rAthena Hash

30928a0

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

Monsters spawn on unwalkable tiles, for example the Hydras in iz_dun03.

### Relevant Log Output

_No response_

### Expected Result

Monsters never spawn on walkable tiles, if no walkable tile could be found after a while (I assume it does a number of attempts), the monster is just spawned on a random location on the map instead.

For example there are ~30 hydras that spawn randomly on iz_dun03 instead of just 10.

### How to Reproduce

Check spawns in iz_dun03.

### Official Information

If you check iz_dun03, you will see that those fixed position Hydras that rAthena has are not there on any official server.

_No response_

Playtester:
* **Addressed Issue(s)**: #8300 

* **Server Mode**: Both

- When searching for a map-wide free cell, the tiles 15 cells from the edge are no longer considered
  * Added a configuration to change the edge size to any value between 1 and 40
- When searching for a free cell, the tiles 4-5 cells from the edge are now considered invalid and trigger a retry
  * If you make the edge size smaller than this, it will use edge size instead
- Searching for a free cell now defaults to 50 tries, but if the "no spawn on player" option is active, those failed attempts are not counted towards the limit anymore
- When a monster spawns in a defined area there will now be 8 attempts to spawn it on a valid cell within the area and then one attempt on the center cell; if all 9 attempts fail, there will now be 50 tries to spawn it map-wide before it gives up
- When a monster has fixed spawn coordinates, but those coordinates are a wall, it will now spawn in a random location map-wide instead
  * This also applies to icewall blocking the cell unless the boss_monster command was used
- Each monster in an area spawn will now receive its own spawn center within the spawn area on server start
  * This results in the spawn area being larger but having a bias towards the center
  * Added a configuration to disable this behavior
- Fixed slave monsters always being active and constantly calling the "search freecell" function even though neither them nor their master have been spotted yet
- Fixed map server crash when setting no_spawn_on_player to 100 (follow-up to 33b2b02)
- Updated prontera field spawns to official episode 18+
- Updated all champion mob respawn times to 3 minutes and sorted them by map name
- Fixes #8300

### DIFF
--- a/conf/battle/misc.conf
+++ b/conf/battle/misc.conf
@@ -187,6 +187,6 @@ hide_fav_sell: no
 // When searching for a random position on the map, how much of the border of the map shall not be considered?
 // Officially the 15 tiles from the edge of the map on each side are not considered as target cells.
 // On some maps like in Pyramids this causes there to be very few monsters in the outer areas. This also
-// affects teleportation. Set this to 4 if you want it to be closer to the old emulator behavior.
-// Valid values: 4-40
+// affects teleportation. Set this to 1 if you want it to be closer to the old emulator behavior.
+// Valid values: 1-40
 map_edge_size: 15
--- a/conf/battle/misc.conf
+++ b/conf/battle/misc.conf
@@ -183,3 +183,10 @@ mail_delay: 1000
 
 // Hides items from the player's favorite tab from being sold to a NPC. (Note 1)
 hide_fav_sell: no
+
+// When searching for a random position on the map, how much of the border of the map shall not be considered?
+// Officially the 15 tiles from the edge of the map on each side are not considered as target cells.
+// On some maps like in Pyramids this causes there to be very few monsters in the outer areas. This also
+// affects teleportation. Set this to 4 if you want it to be closer to the old emulator behavior.
+// Valid values: 4-40
+map_edge_size: 15
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -138,11 +138,11 @@ Map name is the name of the map the monsters will spawn on. x,y are the
 coordinates where the mob should spawn. Putting zeros instead of these
 coordinates will spawn the monsters randomly.
 
-If xs and ys are above 1, each monster will spawn in a radius around x,y.
-At server start, each monster will get assigned its own center cell adding or
-substracting up to (xs-1),(ys-1) from the given x,y coordinates.
-Each time a monster respawns, it will spawn in a radius around its personal
-center cell by adding or substracting up to (xs-1),(ys-1) again.
+If the coordinates are non-zero and xs and ys are above 1, each monster will
+spawn in a radius around x,y. At server start, each monster will get assigned
+its own center cell adding or substracting up to (xs-1),(ys-1) from the given
+x,y coordinates. Each time a monster respawns, it will spawn in a radius around
+its personal center cell by adding or substracting up to (xs-1),(ys-1) again.
 This results in a total possible spawn area over all monsters of the spawn line
 of (xs,ys)*4-3, but with a strong bias towards the center of that area:
 2,2 - 5x5
@@ -153,6 +153,8 @@ etc.
 
 Note this is only the initial spawn zone, as mobs random-walk, they are free
 to move away from their specified spawn region.
+You can disable the picking of a random center cell for each monster in the
+battle config. (See /conf/battle/monster.conf::randomize_center_cell.)
 
 Monster name is the name the monsters will have on screen, and has no relation
 whatsoever to their names anywhere else. It's the mob id that counts, which
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -1,4 +1,4 @@
-//===== rAthena Documentation================================
+﻿//===== rAthena Documentation================================
 //= rAthena Script Commands
 //===== By:==================================================
 //= rAthena Dev Team
@@ -135,9 +135,22 @@ and 'doc/mapflags.txt'.
 <map name>{,<x>{,<y>{,<xs>{,<ys>}}}}%TAB%monster%TAB%<monster name>{,<monster level>}%TAB%<mob id>,<amount>{,<delay1>{,<delay2>{,<event>{,<mob size>{,<mob ai>}}}}}
 
 Map name is the name of the map the monsters will spawn on. x,y are the
-coordinates where the mob should spawn. If xs and ys are non-zero, they
-specify the 'radius' of a spawn-rectangle area centered at x,y.
-Putting zeros instead of these coordinates will spawn the monsters randomly.
+coordinates where the mob should spawn. Putting zeros instead of these
+coordinates will spawn the monsters randomly.
+
+If xs and ys are above 1, each monster will spawn in a radius around x,y.
+At server start, each monster will get assigned its own center cell adding or
+substracting up to (xs-1),(ys-1) from the given x,y coordinates.
+Each time a monster respawns, it will spawn in a radius around its personal
+center cell by adding or substracting up to (xs-1),(ys-1) again.
+This results in a total possible spawn area over all monsters of the spawn line
+of (xs,ys)*4-3, but with a strong bias towards the center of that area:
+2,2 - 5x5
+3,3 - 9x9
+4,4 - 13x13
+5,5 - 17x17
+etc.
+
 Note this is only the initial spawn zone, as mobs random-walk, they are free
 to move away from their specified spawn region.
 
@@ -188,6 +201,8 @@ Natural enemies for AI monsters are normal monsters.
 
 Alternately, a monster spawned using 'boss_monster' instead of 'monster' is able
 to be detected on the map with the SC_BOSSMAPINFO status (used by Convex Mirror).
+A boss monster spawn with fixed coordinates will always spawn at the given 
+coordinates, even if they are blocked (e.g. by Icewall).
 
 ** NPC names
 
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -11482,6 +11482,7 @@ static const struct _battle_data {
 	{ "mob_respawn_time",                   &battle_config.mob_respawn_time,                1000,   1000,   INT_MAX,        },
 	{ "mob_unlock_time",                    &battle_config.mob_unlock_time,                 2000,   0,      INT_MAX,        },
 	{ "map_edge_size",                      &battle_config.map_edge_size,                   15,     1,      40,             },
+	{ "randomize_center_cell",              &battle_config.randomize_center_cell,           1,      0,      1,              },
 
 	{ "feature.stylist",                    &battle_config.feature_stylist,                 1,      0,      1,              },
 	{ "feature.banking_state_enforce",      &battle_config.feature_banking_state_enforce,   0,      0,      1,              },
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -11458,7 +11458,7 @@ static const struct _battle_data {
 
 	{ "mob_respawn_time",                   &battle_config.mob_respawn_time,                1000,   1000,   INT_MAX,        },
 	{ "mob_unlock_time",                    &battle_config.mob_unlock_time,                 2000,   0,      INT_MAX,        },
-	{ "map_edge_size",                      &battle_config.map_edge_size,                   15,     4,      40,             },
+	{ "map_edge_size",                      &battle_config.map_edge_size,                   15,     1,      40,             },
 
 	{ "feature.stylist",                    &battle_config.feature_stylist,                 1,      0,      1,              },
 	{ "feature.banking_state_enforce",      &battle_config.feature_banking_state_enforce,   0,      0,      1,              },
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -11458,6 +11458,7 @@ static const struct _battle_data {
 
 	{ "mob_respawn_time",                   &battle_config.mob_respawn_time,                1000,   1000,   INT_MAX,        },
 	{ "mob_unlock_time",                    &battle_config.mob_unlock_time,                 2000,   0,      INT_MAX,        },
+	{ "map_edge_size",                      &battle_config.map_edge_size,                   15,     4,      40,             },
 
 	{ "feature.stylist",                    &battle_config.feature_stylist,                 1,      0,      1,              },
 	{ "feature.banking_state_enforce",      &battle_config.feature_banking_state_enforce,   0,      0,      1,              },
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -1731,10 +1731,10 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
 		return 0;
 	}
 
+	int16 edge = battle_config.map_edge_size;
 	// In most situations there are 50 tries officially (default value)
 	while(tries--) {
 		// For map-wide search, the cells 15 tiles from the edge are not considered
-		int16 edge = battle_config.map_edge_size;
 		*x = (rx >= 0) ? rnd_value(bx - rx, bx + rx) : rnd_value<int16>(edge, mapdata->xs - edge - 1);
 		*y = (ry >= 0) ? rnd_value(by - ry, by + ry) : rnd_value<int16>(edge, mapdata->ys - edge - 1);
 
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -1699,7 +1699,7 @@ static int map_count_sub(struct block_list *bl,va_list ap)
  * &2 = the target should be able to walk to the target tile.
  * &4 = there shouldn't be any players around the target tile (use the no_spawn_on_player setting)
  *------------------------------------------*/
-int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int16 rx, int16 ry, int flag, int32 tries)
+int map_search_freecell(struct block_list *src, int16 m, int16 *x, int16 *y, int16 rx, int16 ry, int flag, int32 tries)
 {
 	int spawn=0;
 	int bx, by;
@@ -1732,17 +1732,20 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
 	}
 
 	int16 edge = battle_config.map_edge_size;
+	int16 edge_valid = std::min(edge, (int16)5);
 	// In most situations there are 50 tries officially (default value)
 	while(tries--) {
-		// For map-wide search, the cells 15 tiles from the edge are not considered
+		// For map-wide search, the configured tiles from the edge are not considered (default: 15)
 		*x = (rx >= 0) ? rnd_value(bx - rx, bx + rx) : rnd_value<int16>(edge, mapdata->xs - edge - 1);
 		*y = (ry >= 0) ? rnd_value(by - ry, by + ry) : rnd_value<int16>(edge, mapdata->ys - edge - 1);
 
 		if (*x == bx && *y == by)
 			continue; //Avoid picking the same target tile.
 
-		// Cells outside the map or within 4-5 cells of the map edge are considered invalid
-		if (*x < 5 || *x > mapdata->xs - 5 || *y < 5 || *y > mapdata->ys - 5)
+		// Cells outside the map or within 4-5 cells of the map edge are considered invalid officially
+		// Note that unlike the other edge, this isn't symmetric (NE - 4 cells, SW - 5 cells)
+		// If for some reason edge size was set to below 5 cells, we consider them as valid
+		if (*x < edge_valid || *x > mapdata->xs - edge_valid || *y < edge_valid || *y > mapdata->ys - edge_valid)
 			continue;
 
 		if (map_getcell(m,*x,*y,CELL_CHKREACH))
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -1699,7 +1699,7 @@ static int map_count_sub(struct block_list *bl,va_list ap)
  * &2 = the target should be able to walk to the target tile.
  * &4 = there shouldn't be any players around the target tile (use the no_spawn_on_player setting)
  *------------------------------------------*/
-int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int16 rx, int16 ry, int flag, int tries)
+int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int16 rx, int16 ry, int flag, int32 tries)
 {
 	int spawn=0;
 	int bx, by;
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -1699,9 +1699,9 @@ static int map_count_sub(struct block_list *bl,va_list ap)
  * &2 = the target should be able to walk to the target tile.
  * &4 = there shouldn't be any players around the target tile (use the no_spawn_on_player setting)
  *------------------------------------------*/
-int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int16 rx, int16 ry, int flag)
+int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int16 rx, int16 ry, int flag, int tries)
 {
-	int tries, spawn=0;
+	int spawn=0;
 	int bx, by;
 
 	if( !src && (!(flag&1) || flag&2) )
@@ -1731,20 +1731,19 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
 		return 0;
 	}
 
-	if (rx >= 0 && ry >= 0) {
-		tries = (rx * 2 + 1) * (ry * 2 + 1);
-		if (tries > 100) tries = 100;
-	} else {
-		tries = mapdata->xs*mapdata->ys;
-		if (tries > 500) tries = 500;
-	}
-
+	// In most situations there are 50 tries officially (default value)
 	while(tries--) {
-		*x = (rx >= 0) ? rnd_value(bx - rx, bx + rx) : rnd_value<int16>(1, mapdata->xs - 1);
-		*y = (ry >= 0) ? rnd_value(by - ry, by + ry) : rnd_value<int16>(1, mapdata->ys - 1);
+		// For map-wide search, the cells 15 tiles from the edge are not considered
+		int16 edge = battle_config.map_edge_size;
+		*x = (rx >= 0) ? rnd_value(bx - rx, bx + rx) : rnd_value<int16>(edge, mapdata->xs - edge - 1);
+		*y = (ry >= 0) ? rnd_value(by - ry, by + ry) : rnd_value<int16>(edge, mapdata->ys - edge - 1);
 
 		if (*x == bx && *y == by)
 			continue; //Avoid picking the same target tile.
+
+		// Cells outside the map or within 4-5 cells of the map edge are considered invalid
+		if (*x < 5 || *x > mapdata->xs - 5 || *y < 5 || *y > mapdata->ys - 5)
+			continue;
 
 		if (map_getcell(m,*x,*y,CELL_CHKREACH))
 		{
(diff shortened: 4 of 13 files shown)


src/map/map.cpp
```diff
@@ -1731,21 +1731,20 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
...
-		tries = mapdata->xs*mapdata->ys;
-		if (tries > 500) tries = 500;
-	}
-
+	// In most situations there are 50 tries officially (default value)
 	while(tries--) {
-		*x = (rx >= 0) ? rnd_value(bx - rx, bx + rx) : rnd_value<int16>(1, mapdata->xs - 1);
-		*y = (ry >= 0) ? rnd_value(by - ry, by + ry) : rnd_value<int16>(1, mapdata->ys - 1);
+		// For map-wide search, the cells 15 tiles from the edge are not considered
+		int16 edge = battle_config.map_edge_size;
```
Atemo, line 1737: `edge` could be defined before 'while'

src/map/pc.cpp
```diff
@@ -7047,8 +7047,9 @@ char pc_randomwarp(map_session_data *sd, clr_type type, bool ignore_mapflag)
 		return 3;
 
 	do {
-		x = rnd()%(mapdata->xs-2)+1;
-		y = rnd()%(mapdata->ys-2)+1;
+		int edge = battle_config.map_edge_size;
```
aleos89, line 7050: I'd put this define right above the while loop like the other comment.

aleos89 (approved):

conf/battle/misc.conf
```diff
@@ -183,3 +183,10 @@ mail_delay: 1000
 
 // Hides items from the player's favorite tab from being sold to a NPC. (Note 1)
 hide_fav_sell: no
+
+// When searching for a random position on the map, how much of the border of the map shall not be considered?
+// Officially the 15 tiles from the edge of the map on each side are not considered as target cells.
+// On some maps like in Pyramids this causes there to be very few monsters in the outer areas. This also
+// affects teleportation. Set this to 4 if you want it to be closer to the old emulator behavior.
+// Valid values: 4-40
```
Lemongrass3110, line 191:
```suggestion
// Valid values: 1-40
```

src/map/battle.cpp
```diff
@@ -11458,6 +11458,7 @@ static const struct _battle_data {
 
 	{ "mob_respawn_time",                   &battle_config.mob_respawn_time,                1000,   1000,   INT_MAX,        },
 	{ "mob_unlock_time",                    &battle_config.mob_unlock_time,                 2000,   0,      INT_MAX,        },
+	{ "map_edge_size",                      &battle_config.map_edge_size,                   15,     4,      40,             },
```
Lemongrass3110, line 11461:
```suggestion
	{ "map_edge_size",                      &battle_config.map_edge_size,                   15,     1,      40,             },
```

src/map/mob.cpp
```diff
@@ -1084,11 +1084,12 @@ int mob_setdelayspawn(struct mob_data *md)
 int mob_count_sub(struct block_list *bl, va_list ap) {
```
Lemongrass3110, line 1084: Please apply tabs instead of spaces to the whole function, if you already touch it.

Also please apply usage of the ARRAYLENGTH macro, if we already touch it.

src/map/mob.cpp
```diff
@@ -1084,11 +1084,12 @@ int mob_setdelayspawn(struct mob_data *md)
...
     int mobid[10], i;
     ARR_FIND(0, 10, i, (mobid[i] = va_arg(ap, int)) == 0); //fetch till 0
-    if (mobid[0]) { //if there one let's check it otherwise go backward
-        TBL_MOB *md = BL_CAST(BL_MOB, bl);
+	TBL_MOB* md = BL_CAST(BL_MOB, bl);
```
Lemongrass3110, line 1087:
```suggestion
	mob_data* md = BL_CAST(BL_MOB, bl);
```

src/map/mob.cpp
```diff
@@ -1119,17 +1120,31 @@ int mob_spawn (struct mob_data *md)
...
-					delete_timer(md->spawn_timer, mob_delayspawn);
-				md->spawn_timer = add_timer(tick + battle_config.mob_respawn_time,mob_delayspawn,md->bl.id,0);
-				return 1;
+		{
+			// Officially the area is split into 4 squares, 4 lines and 1 dot and for each of those there is one attempt
+			// We simplify this to trying 8 times in the whole area and then at the center cell even though that's not fully accurate
+			// The search for a spawn location is called twice, meaning the area is doubled in size but with a strong bias towards the center
+			// The center cell has a 1/(xs*ys) chance to be picked, otherwise we call the search function to get the start coordinates
+			if ((md->spawn->xs > 1 || md->spawn->ys > 1) && !rnd_chance(1, md->spawn->xs * md->spawn->ys))
+				map_search_freecell(&md->bl, -1, &md->bl.x, &md->bl.y, md->spawn->xs-1, md->spawn->ys-1, 0, 8);
```
Lemongrass3110, line 1131: Somehow here the logic is missing, if this succeeds.

Also why is this one not using `battle_config.no_spawn_on_player?4:0`?

src/map/map.cpp
```diff
@@ -1731,21 +1731,20 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
...
-		tries = mapdata->xs*mapdata->ys;
-		if (tries > 500) tries = 500;
-	}
-
+	int16 edge = battle_config.map_edge_size;
+	// In most situations there are 50 tries officially (default value)
 	while(tries--) {
-		*x = (rx >= 0) ? rnd_value(bx - rx, bx + rx) : rnd_value<int16>(1, mapdata->xs - 1);
-		*y = (ry >= 0) ? rnd_value(by - ry, by + ry) : rnd_value<int16>(1, mapdata->ys - 1);
+		// For map-wide search, the cells 15 tiles from the edge are not considered
```
Lemongrass3110, line 1737:
```suggestion
		// For map-wide search, the cells the edge tiles are not considered
```

src/map/pc.cpp
```diff
@@ -7046,9 +7046,10 @@ char pc_randomwarp(map_session_data *sd, clr_type type, bool ignore_mapflag)
 	if (mapdata->getMapFlag(MF_NOTELEPORT) && !ignore_mapflag) //Teleport forbidden
 		return 3;
 
+	int edge = battle_config.map_edge_size;
 	do {
-		x = rnd()%(mapdata->xs-2)+1;
-		y = rnd()%(mapdata->ys-2)+1;
+		x = rnd()%(mapdata->xs - edge*2) + edge;
+		y = rnd()%(mapdata->ys - edge*2) + edge;
 	} while((map_getcell(sd->bl.m,x,y,CELL_CHKNOPASS) || (!battle_config.teleport_on_portal && npc_check_areanpc(1,sd->bl.m,x,y,1))) && (i++) < 1000);
```
Lemongrass3110, line 7053: Should we decrease from 1000 to 50 here too?

conf/battle/misc.conf
```diff
@@ -183,3 +183,10 @@ mail_delay: 1000
 
 // Hides items from the player's favorite tab from being sold to a NPC. (Note 1)
 hide_fav_sell: no
+
+// When searching for a random position on the map, how much of the border of the map shall not be considered?
+// Officially the 15 tiles from the edge of the map on each side are not considered as target cells.
+// On some maps like in Pyramids this causes there to be very few monsters in the outer areas. This also
+// affects teleportation. Set this to 4 if you want it to be closer to the old emulator behavior.
```
Lemongrass3110, line 190:
```suggestion
// affects teleportation. Set this to 1 if you want it to be closer to the old emulator behavior.
```

src/map/map.cpp
```diff
@@ -1731,21 +1731,20 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
...
+		*x = (rx >= 0) ? rnd_value(bx - rx, bx + rx) : rnd_value<int16>(edge, mapdata->xs - edge - 1);
+		*y = (ry >= 0) ? rnd_value(by - ry, by + ry) : rnd_value<int16>(edge, mapdata->ys - edge - 1);
```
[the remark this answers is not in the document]
Playtester: You need the "- 1" here so that it's symmetric (the valid cell check is not symmetric officially, so it's a special case).

src/map/map.cpp
Playtester: That doesn't make sense, do you mean "the edge tiles" without "the cells" in front?

src/map/pc.cpp
```diff
@@ -7046,9 +7046,10 @@ char pc_randomwarp(map_session_data *sd, clr_type type, bool ignore_mapflag)
 	if (mapdata->getMapFlag(MF_NOTELEPORT) && !ignore_mapflag) //Teleport forbidden
 		return 3;
 
+	int edge = battle_config.map_edge_size;
 	do {
-		x = rnd()%(mapdata->xs-2)+1;
-		y = rnd()%(mapdata->ys-2)+1;
+		x = rnd()%(mapdata->xs - edge*2) + edge;
+		y = rnd()%(mapdata->ys - edge*2) + edge;
```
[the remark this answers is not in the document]
Lemongrass3110: It should be the same as at the other location, whatever that value is in the end.

src/map/map.hpp
```diff
@@ -1124,7 +1124,7 @@ int map_count_oncell(int16 m,int16 x,int16 y,int type,int flag);
 struct skill_unit *map_find_skill_unit_oncell(struct block_list *,int16 x,int16 y,uint16 skill_id,struct skill_unit *, int flag);
 // search and creation
 int map_get_new_object_id(void);
-int map_search_freecell(struct block_list *src, int16 m, int16 *x, int16 *y, int16 rx, int16 ry, int flag);
+int map_search_freecell(struct block_list *src, int16 m, int16 *x, int16 *y, int16 rx, int16 ry, int flag, int tries = 50);
```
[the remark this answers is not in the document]
Playtester: The function Aegis uses for fully random locations always has 50 tries, this is also used by all skills that search for a map-wide random location.

I also think the previous calculation of tries didn't really make sense as the chance to find a free cell doesn't depend on the size of the area, so we can just use a fixed value for simplicity. We can still adjust the other calls later when we need or have more information on them (that's why it's a parameter so it's easy to adjust the calls later).

Defaulting it to a lower value has the risk of it failing (e.g. for item drops we basically don't realy want it to fail), so I went with the highest official value which is 50.

src/map/mob.cpp
Playtester: This goes against your philosophy to never change formatting of cells you didn't change! I left it untouched just for you.

src/map/mob.cpp
Playtester: We don't need the logic here because we don't care of it succeeds or not, if it fails x and y remain the same, if it succeeds x and y are already updated (it's call-by-reference).

It's also important not to consider "no_spawn_on_player" here because here we just calculate the "bridge cell" from which we create the real spawn area below (where we then consider the setting).

src/map/mob.cpp
```diff
@@ -2082,6 +2097,13 @@ static int mob_ai_sub_lazy(struct mob_data *md, va_list args)
 	md->last_thinktime=tick;
 
 	if (md->master_id) {
+		// Get mob data of master
```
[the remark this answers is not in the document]
Playtester: I know but I needed this change to be able to debug map_search_freecell to test if it's working, because otherwise it would have been called constantly.

src/map/mob.cpp
```diff
@@ -2082,6 +2097,13 @@ static int mob_ai_sub_lazy(struct mob_data *md, va_list args)
...
+		mob_data* mmd = map_id2md(md->master_id);
+
+		// If neither master nor slave have been spotted we don't have to execute the slave AI
+		if (mmd && !mob_is_spotted(mmd) && !mob_is_spotted(md))
```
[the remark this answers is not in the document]
Playtester: && checks are easier to read than || checks if you ask me. Especially if the order matters (in this case checking for mmd first).

src/map/npc.cpp
```diff
@@ -5294,20 +5294,29 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
 		}
 		mob.xs = 0;
 	}
```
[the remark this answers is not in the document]
Playtester: Further below there is a case where x and y are 0 it would just put xs and ys to -1 (map-wide).

src/map/npc.cpp
```diff
@@ -5294,20 +5294,29 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
...
+	else if (mob.xs == 0) {
+		// Both 0 and 1 result in fixed x-coordinate
+		mob.xs = 1;
```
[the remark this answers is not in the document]
Playtester: This is wrong, we need to eliminate 0 here. We substract 1 later and don't want it to go negative because that would mean map-wide. Maybe I need to improve the comment.

src/map/npc.cpp
```diff
@@ -5294,20 +5294,29 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
 		}
 		mob.xs = 0;
 	}
+	else if (mob.xs == 0) {
```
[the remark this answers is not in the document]
Playtester: We should only touch the value if it's 0. If it's -1 (map-wide) it should remain -1.

src/map/npc.cpp
```diff
@@ -5294,20 +5294,29 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
...
+		mob.xs = 1;
+	}
 
 	if (mob.ys < 0) {
 		if (w1count > 4) {
 			ShowWarning("npc_parse_mob: Negative y-span %hd for mob ID %d (file '%s', line '%d').\n", mob.ys, mob_id, filepath, strline(buffer, start - buffer));
 		}
 		mob.ys = 0;
 	}
+	else if (mob.ys == 0) {
```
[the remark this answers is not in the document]
Playtester: See above

src/map/npc.cpp
```diff
@@ -5294,20 +5294,29 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
...
+		// Both 0 and 1 result in fixed y-coordinate
+		mob.ys = 1;
```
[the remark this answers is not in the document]
Playtester: See above

src/map/pc.cpp
Playtester: We could, but it doesn't make a big difference since the chance to even reach 50 attempts is low.

src/map/mob.cpp
```diff
@@ -1119,17 +1120,31 @@ int mob_spawn (struct mob_data *md)
 		md->bl.x = md->spawn->x;
 		md->bl.y = md->spawn->y;
 
+		// Search can be skipped if spawn location is fixed
 		if( (md->bl.x == 0 && md->bl.y == 0) || md->spawn->xs || md->spawn->ys )
```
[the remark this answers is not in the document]
Playtester: Dangerous, can't do this as we need to consider -1 for map-wide as well.
Maybe != 1 works.

src/map/mob.cpp
Playtester: I added another comment already, is this sufficiently clear?

src/map/mob.cpp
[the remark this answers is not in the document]
Playtester: That's not true, "false &&" will stop there and not check the next condition.

src/map/npc.cpp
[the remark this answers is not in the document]
Playtester: Hmm before this version I had it with `if (xs >= 1) xs--;` but then when I realized I need to calculate (xs+1*ys+1) I wanted xs and ys to be one higher so it's easier to write.

One more reason I decided for this version is that later on it will be incredibly confusing for devs to figure out why the value is 1 lower than what they put into their spawn file.

Your version would make it so it's:
0 - 1x1 search - 1x1 spawn area
1 - usused
2 - 3x3 search - 5x5 spawn area
3 - 5x5 search - 9x9 spawn area

This makes it very hard to use because it has a gap and a jump of 2. We could need to add logic that if the value is >1 we would need to reduce it by 1 but otherwise not.

I think my version is easier to understand because it doesn't have a gap and doesn't jumps by 2:
0 - unused
1 - 1x1 search - 1x1 spawn area
2 - 3x3 search - 5x5 spawn area
3 - 5x5 search - 9x9 spawn area

src/map/npc.cpp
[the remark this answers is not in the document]
Playtester: See comment above.

src/map/npc.cpp
```diff
@@ -5294,20 +5294,29 @@ static const char* npc_parse_mob(char* w1, char* w2, char* w3, char* w4, const c
...
+	}
 
 	if (mob.num > 1 && battle_config.mob_count_rate != 100) {
 		if ((mob.num = mob.num * battle_config.mob_count_rate / 100) < 1)
 			mob.num = 1;
 	}
 
-	if (battle_config.force_random_spawn || (mob.x == 0 && mob.y == 0))
+	if (battle_config.force_random_spawn || (mob.x == 0 && mob.y == 0)
+		|| (mob.xs <= 1 && mob.ys <= 1 && !map_getcell(mob.m, mob.x, mob.y, CELL_CHKREACH)))
```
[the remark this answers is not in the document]
Playtester: I wouldn't be okay with having 35 warnings every time I start a vanilla rAthena installation.

src/map/pc.cpp
Playtester: I talked with Daegalagh about this before, I think long term we should unify all into using the same function(s). Even monsters using teleport should actually not land on portals, so it would be nice to have this check in the map_search_freecell function anyway. I just don't want to do this as part of this PR (if you change a lot of functionality and at the same time do a full refactoring, it's hard to see what functionality was changed).

src/map/npc.cpp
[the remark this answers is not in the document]
Lemongrass3110: It is fine. Those are just different approaches.
Please make sure to check if the default value for negative input (a little above), which is set to 0, is also adjusted to 1, if necessary.

Somehow your logic also makes it confusing for the devs, if they put in 0 and end up with 1.

src/map/npc.cpp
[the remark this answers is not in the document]
Lemongrass3110: It was not my intention to introduce warnings and you know that.
But okay, we leave away the check and dont further validate the input, because official data is garbage.

src/map/mob.cpp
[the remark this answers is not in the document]
Playtester: I guess that works, at least you save the resources from the memory allocation in most cases then.

src/map/npc.cpp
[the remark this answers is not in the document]
Playtester: You already get a warning in the log file in that case and it still results in a fixed coordinate as it did before. 

However, I personally think it would be better to set it to "-1" there, so instead of a fixed coordinate in case of error you have a map-wide one. Seems less exploitable at least to me. Should I go ahead with this?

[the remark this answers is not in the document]
Lemongrass3110: I also think it would be better to switch to map-wide.
Maybe we could include that in the warning somehow?
"... Defaulting to map-wide."
I think this is how we do it in the database files, if we cap the value or default to something, because it is out of range.

src/map/npc.cpp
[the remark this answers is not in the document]
Playtester: Hmm, the logic of this function is that of x and y is 0 then the coordinates are map-wide xs and ys don't matter in this case. There actually isn't a way to set xs and ys so that the coordinates are random (the default value and any given values will overwritten with 0), it just depends on x and y. 

Though I'm considering to changing the min cap to -1 instead based on your other comment. Unfortunately you cannot intentionally set it to -1 as there will be a warning issued then.

I guess we could issue a warning if the center is (0,0) and xs or ys are above 0 that the spawn will be random instead?

src/map/mob.cpp
```diff
@@ -1119,17 +1120,32 @@ int mob_spawn (struct mob_data *md)
...
+			// The center cell has a 1/(xs*ys) chance to be picked, otherwise we call the search function to get the start coordinates
+			// On this first iteration we don't care if the search succeeds or if it's visible to player as it's not the final spawn location
+			if ((md->spawn->xs > 1 || md->spawn->ys > 1) && !rnd_chance(1, md->spawn->xs * md->spawn->ys))
+				map_search_freecell(&md->bl, -1, &md->bl.x, &md->bl.y, md->spawn->xs-1, md->spawn->ys-1, 0, 8);
+
+			// Try to spawn monster in defined area (8 tries)
+			if (!map_search_freecell(&md->bl, -1, &md->bl.x, &md->bl.y, md->spawn->xs-1, md->spawn->ys-1, battle_config.no_spawn_on_player?4:0, 8))
+			{
+				// If area search failed and center cell not reachable, try to spawn the monster anywhere on the map (50 tries)
+				if (!map_getcell(md->bl.m, md->bl.x, md->bl.y, CELL_CHKREACH) && !map_search_freecell(&md->bl, -1, &md->bl.x, &md->bl.y, -1, -1, battle_config.no_spawn_on_player?4:0))
```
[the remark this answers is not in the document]
Playtester: Not useless. If you are implying it cannot currently be on an unwalkable cell, then the whole if() would be useless and not just the first part.

The first part of the if would only be useless if we are guaranteed to currently be on an unwalkable cell which is definitely not guaranteed.

Basically if the official spawn search cell algorithm fails and then current cell is walkable, then it should always spawn on that cell (even if it's the cell that was found by the first call of map_search_freecell), but if the cell is unwalkable, then the monster spawns in a random area of the map.

doc/script_commands.txt
```diff
@@ -135,9 +135,22 @@ and 'doc/mapflags.txt'.
 <map name>{,<x>{,<y>{,<xs>{,<ys>}}}}%TAB%monster%TAB%<monster name>{,<monster level>}%TAB%<mob id>,<amount>{,<delay1>{,<delay2>{,<event>{,<mob size>{,<mob ai>}}}}}
 
 Map name is the name of the map the monsters will spawn on. x,y are the
-coordinates where the mob should spawn. If xs and ys are non-zero, they
-specify the 'radius' of a spawn-rectangle area centered at x,y.
-Putting zeros instead of these coordinates will spawn the monsters randomly.
+coordinates where the mob should spawn. Putting zeros instead of these
+coordinates will spawn the monsters randomly.
+
+If xs and ys are above 1, each monster will spawn in a radius around x,y.
```
Lemongrass3110, line 141: Since you added a battle conf it should be mentioned here somehow

src/map/map.cpp
```diff
@@ -1754,10 +1756,11 @@ int map_search_freecell(struct block_list *src, int16 m, int16 *x,int16 *y, int1
 				if (spawn >= 100) return 0; //Limit of retries reached.
```
Lemongrass3110, line 1756:
```suggestion
```
The battle config is already limited [0,100]

src/map/mob.cpp
```diff
@@ -1116,20 +1119,33 @@ int mob_spawn (struct mob_data *md)
...
-		if( (md->bl.x == 0 && md->bl.y == 0) || md->spawn->xs || md->spawn->ys )
-		{	//Monster can be spawned on an area.
-			if( !map_search_freecell(&md->bl, -1, &md->bl.x, &md->bl.y, md->spawn->xs, md->spawn->ys, battle_config.no_spawn_on_player?4:0) )
-			{ // retry again later
-				if( md->spawn_timer != INVALID_TIMER )
-					delete_timer(md->spawn_timer, mob_delayspawn);
-				md->spawn_timer = add_timer(tick + battle_config.mob_respawn_time,mob_delayspawn,md->bl.id,0);
-				return 1;
+		md->bl.x = md->centerX;
+		md->bl.y = md->centerY;
```
Lemongrass3110, line 1123:
```suggestion
		if( battle_config.randomize_center_cell ){
			md->bl.x = md->centerX;
			md->bl.y = md->centerY;
		}else{
			md->bl.x = md->spawn->x;
			md->bl.y = md->spawn->y;
		}
```

src/map/mob.cpp
```diff
@@ -1116,20 +1119,33 @@ int mob_spawn (struct mob_data *md)
...
+
+		// Search can be skipped for boss monster spawns if spawn location is fixed
+		// We can't skip normal monsters as they should pick a random location if the cell is blocked (e.g. Icewall)
+		// The center cell has a 1/(xs*ys) chance to be picked if free
+		if ((!md->spawn->state.boss || (md->bl.x == 0 && md->bl.y == 0) || md->spawn->xs != 1 || md->spawn->ys != 1)
+			&& (md->spawn->xs + md->spawn->ys < 1 || !rnd_chance(1, md->spawn->xs * md->spawn->ys) || !map_getcell(md->bl.m, md->bl.x, md->bl.y, CELL_CHKREACH)))
```
Lemongrass3110, line 1129: Based on the complexity of this condition I would say that it may be better to put this into a function and make those conditions return early.
Otherwise you could create a pseudo scope, to be able to break out of it as well.

src/map/npc.cpp
```diff
@@ -5180,6 +5180,13 @@ void npc_parse_mob2(struct spawn_data* mob)
 	{
 		struct mob_data* md = mob_spawn_dataset(mob);
 		md->spawn = mob;
+		// Determine center cell for each mob in the spawn line
+		if (battle_config.randomize_center_cell) {
+			if (mob->xs > 1)
+				md->centerX = rnd_value(mob->x - mob->xs + 1, mob->x + mob->xs - 1);
```
Lemongrass3110, line 5186:
```suggestion
				md->centerX = rnd_value(mob->x - mob->xs + 1, mob->x + mob->xs - 1);
			else
				md->centerX = 0;
```

src/map/npc.cpp
```diff
@@ -5180,6 +5180,13 @@ void npc_parse_mob2(struct spawn_data* mob)
...
+			if (mob->ys > 1)
+				md->centerY = rnd_value(mob->y - mob->ys + 1, mob->y + mob->ys - 1);
```
Lemongrass3110, line 5188:
```suggestion
				md->centerY = rnd_value(mob->y - mob->ys + 1, mob->y + mob->ys - 1);
			else
				md->centerY = 0;
```

src/map/npc.cpp
```diff
@@ -5180,6 +5180,13 @@ void npc_parse_mob2(struct spawn_data* mob)
...
+		}
```
Lemongrass3110, line 5189:
```suggestion
		}else{
			md->centerX = 0;
			md->centerY = 0;
		}
```

Lemongrass3110 (approved):

src/map/map.cpp
Playtester: This will break the feature. It's different because if you set it to 100 the monster should not spawn at all, whereas if the value is lower than 100, it should spawn anyway.

src/map/mob.cpp
Playtester: Not needed, md->centerX is set to md->spawn->x during initialization of the spawnset.

src/map/mob.cpp
Playtester: It's not really that complex and the comment explains it too. Also too much structural change so close before pushing would just be more risky than not touching it anymore.

src/map/npc.cpp
Playtester: Let's just leave centerX as initialized, then we also don't need the other changes!

src/map/npc.cpp
Playtester: See above

PR #8333, opened 2024-05-14
Added Siege_Teleport_Scroll2 function

### LINKED ISSUE #7571: Item 12415 does not have script
### rAthena Hash

d93d4144eb599bf1656ab0e7fbdb7cb8edf0decf

### Client Date

2020-09-02

### Server Mode

Renewal

### Result

(Id: 14591) with **AegisName: Siege_Teleport_Scroll** on _item_db_usable.yml_ works as expected but the (Id: 12415) with **AegisName: Siege_Teleport_Scroll2**  does not works but it is usable ... the amount of item go down when being used but does not do anything .

`@item 14591 100`   <~ works as expected
`@item 12415 100` <~ item does not do anything( does not have script)

### Relevant Log Output

_No response_

### Expected Result

Item 12415 when use would show a window that give user area(s) selections where to teleport next.

### How to Reproduce

`@item 14591 100`   <~ works as expected
`@item 12415 100` <~ item does not do anything( does not have script)

### Official Information

N/A

_No response_

Akkarinage:
* **Addressed Issue(s)**: #7571 

* **Server Mode**: Renewal

* Adds NPC function `F_CashSiegeTele2`
* Given item Siege_Teleport_Scroll2 (item_id:12415) script in RE item_db_usable.yml. This item would not appear in Pre-RE so that file remains untouched.

This PR fixes #7571

### DIFF
--- a/npc/other/CashShop_Functions.txt
+++ b/npc/other/CashShop_Functions.txt
@@ -263,6 +263,26 @@ function	script	F_CashSiegeTele	{
 	return;
 }
 
+// Siege Teleport Scroll 2 (WoE 1.5)
+//============================================================ 
+// - Warp player to selected guild castle.
+// - No arguments.
+function	script	F_CashSiegeTele2	{
+	switch(select("Mardoll (arug_cas01):Cyr (arug_cas02):Horn (arug_cas03):Gefn (arug_cas04):Badanis (arug_cas05):Himinn (schg_cas01):Andlangr (schg_cas02):Vidblainn (schg_cas03):Hljod (schg_cas04):Schatirnil (schg_cas05)")) {
+	case 1: warp "aru_gld",159,272; end;
+	case 2: warp "aru_gld",86,50; end;
+	case 3: warp "aru_gld",63,157; end;
+	case 4: warp "aru_gld",307,354; end;
+	case 5: warp "aru_gld",298,110; end;
+	case 6: warp "sch_gld",300,100; end;
+	case 7: warp "sch_gld",284,249; end;
+	case 8: warp "sch_gld",102,196; end;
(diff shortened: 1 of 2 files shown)


npc/other/CashShop_Functions.txt
```diff
@@ -263,6 +263,26 @@ function	script	F_CashSiegeTele	{
 	return;
 }
 
+// Siege Teleport Scroll 2 (WoE 1.5)
+//============================================================ 
+// - Warp player to selected guild castle.
+// - No arguments.
+function	script	F_CashSiegeTele2	{
+	switch(select("Mardoll (arug_cas01):Cyr (arug_cas02):Horn (arug_cas03):Gefn (arug_cas04):Badanis (arug_cas05):Himinn (schg_cas01):Andlangr (schg_cas02):Vidblainn (schg_cas03):Hljod (schg_cas04):Schatirnil (schg_cas05)")) {
+	case 1: warp "aru_gld",159,272; end;
```
aleos89, line 272: Not sure if we should fix it here, but the end here stops the script. The return at the bottom will never be ran thus anything in the item script after the callfunc will also never be ran.

Since this is a mimic from the other items those are "broken" as well.

aleos89 (approved):
Looks good overall! Just need to determine if we want to fix the end issue here or in a follow up!

npc/other/CashShop_Functions.txt
aleos89: Ignore me. It's been a while since I worked on a script. I think we put ends after warps on purpose to detach the player from the script to avoid any possible issues.

Atemo (approved):

PR #8343, opened 2024-05-19
Consistent Dirx / Diry Array Usage

Playtester:
* **Addressed Issue(s)**: (Cleanup Task)

* **Server Mode**: Both

* **Description of Pull Request**: All skills now use the globally defined dirx/diry variables instead of defining their own

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -6625,11 +6625,9 @@ int skill_castend_damage_id (struct block_list* src, struct block_list *bl, uint
 		break;
 	case RA_WUGSTRIKE:
 		if( sd && pc_isridingwug(sd) ){
-			short x[8]={0,-1,-1,-1,0,1,1,1};
-			short y[8]={1,1,0,-1,-1,-1,0,1};
 			uint8 dir = map_calc_dir(bl, src->x, src->y);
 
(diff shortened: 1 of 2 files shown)


src/map/skill.cpp
```diff
@@ -17531,10 +17529,8 @@ bool skill_check_condition_castbegin( map_session_data& sd, uint16 skill_id, uin
 		{
 			if( skill_lv < 3 && ((sd.bl.type == BL_PC && battle_config.pc_cloak_check_type&1)
 			||	(sd.bl.type != BL_PC && battle_config.monster_cloak_check_type&1) )) { //Check for walls.
-				static int dx[DIR_MAX] = { 0, 1, 0, -1, -1,  1, 1, -1};
-				static int dy[DIR_MAX] = {-1, 0, 1,  0, -1, -1, 1,  1};
```
Lemongrass3110, line 17535: Is the change of values intended here? They are different to the ones in dirx/diry.
Rest looks good.

Playtester: I couldn't find a reason for it as the only thing is does it iterate through all directions to see if the skill should fail or not.

I also tested the skill ingame and it still works.

PR #8354, opened 2024-05-23
Updated npcspeed, npcwalkto and npcstop script commands

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

- Added the optional parameter `<npc name>` to `npcspeed` to change the walking speed of the npc with the said name.
Usage : `npcspeed( <speed value> {,"<npc name>"} );`

- Added the optional parameter `<npc name>` to `npcwalkto` to move the npc with the said name at the given coordinates.
Usage : `npcwalkto( <x>,<y> {,"<npc name>"} } );`

- Added the optional parameters `<npc name>` and `<flag>` to `npcstop` to stop the movement of the npc with the said name with options.
Usage : `npcstop( {"<npc name>", {"<flag>"}});`


doc/script_commands.txt
```diff
@@ -7961,17 +7961,15 @@ Return values:
...
-These commands will make the NPC object in question move around the map. As they
-currently are, they are a bit buggy and are not useful for much more than making
-an NPC move randomly around the map.
+These commands will make the NPC object in question move around the map.
 
 'npcspeed' will set the NPCs walking speed to a specified value. As in the
-@speed GM command, 200 is the slowest possible speed while 0 is the fastest
-possible (instant motion). 100 is the default character walking speed.
+@speed GM command, 1000 is the slowest possible speed while 20 is the fastest
+possible (instant motion). 200 is the default npc walking speed.
```
Lemongrass3110, line 7972: Should we export the constants and use them here?
And should we create a constant for the default walk npc walkspeed?

src/map/script.cpp
```diff
@@ -16064,49 +16064,106 @@ BUILDIN_FUNC(chatmes)
...
-	nd =(struct npc_data *)map_id2bl(st->oid);
+	if (script_hasdata(st, 3))
+		nd = npc_name2id(script_getstr(st, 3));
+	else
+		nd = map_id2nd(st->oid);
 
-	if( nd ) {
-		nd->speed = speed;
-		nd->ud.state.speed_changed = 1;
+	if (!nd) {
```
Lemongrass3110, line 16080:
```suggestion
	if (nd == nullptr) {
```

src/map/script.cpp
```diff
@@ -16064,49 +16064,106 @@ BUILDIN_FUNC(chatmes)
...
-// change npc walkspeed [Valaris]
+/**
+ * Change npc walkspeed.
+ * npcspeed <speed value> {,"<npc name>"};
+ */
 BUILDIN_FUNC(npcspeed)
 {
-	struct npc_data* nd;
-	int speed;
+	npc_data *nd;
```
Lemongrass3110, line 16073:
```suggestion
	npc_data* nd;
```

src/map/script.cpp
```diff
@@ -16064,49 +16064,106 @@ BUILDIN_FUNC(chatmes)
...
+
+/**
+ * Make an npc walk to a position.
+ * npcwalkto <x>,<y> {,"<speed>" {,"<npc name>"} };
+ */
 BUILDIN_FUNC(npcwalkto)
 {
-	struct npc_data *nd=(struct npc_data *)map_id2bl(st->oid);
-	int x=0,y=0;
+	npc_data *nd;
```
Lemongrass3110, line 16107:
```suggestion
	npc_data* nd;
```

src/map/script.cpp
```diff
@@ -16064,49 +16064,106 @@ BUILDIN_FUNC(chatmes)
...
-	y=script_getnum(st,3);
+	if (script_hasdata(st, 5))
+		nd = npc_name2id(script_getstr(st, 5));
+	else
+		nd = map_id2nd(st->oid);
 
-	if(nd) {
-		if (!nd->status.hp)
-			status_calc_npc(nd, SCO_FIRST);
+	if (!nd) {
```
Lemongrass3110, line 16114:
```suggestion
	if (nd == nullptr) {
```

src/map/script.cpp
```diff
@@ -16064,49 +16064,106 @@ BUILDIN_FUNC(chatmes)
...
+	if (script_hasdata(st, 4)) {
+		int speed = script_getnum(st, 4);
+		
+		if (speed < MIN_WALK_SPEED || speed > MAX_WALK_SPEED) {
+			ShowError("buildin_npcwalkto: invalid speed %d (min: %d, max: %d).\n", speed, MIN_WALK_SPEED, MAX_WALK_SPEED);
+			return SCRIPT_CMD_FAILURE;
+		}
+		nd->speed = speed;
+		nd->ud.state.speed_changed = 1;
+	}
```
Lemongrass3110, line 16131: Since this affects the speed of the NPC after any call to this function:
Why do we have to make it a parameter for this function and why can npcspeed not be called by the user before calling this function?

src/map/script.cpp
```diff
@@ -16064,49 +16064,106 @@ BUILDIN_FUNC(chatmes)
...
 
-// stop an npc's movement [Valaris]
+/**
+ * Stop an npc's movement.
+ * npcstop {"<npc name>"} };
+ */
 BUILDIN_FUNC(npcstop)
 {
-	struct npc_data *nd=(struct npc_data *)map_id2bl(st->oid);
+	npc_data *nd;
```
Lemongrass3110, line 16151:
```suggestion
	npc_data* nd;
```

src/map/script.cpp
```diff
@@ -16064,49 +16064,106 @@ BUILDIN_FUNC(chatmes)
...
+
+	if (script_hasdata(st, 2))
+		nd = npc_name2id(script_getstr(st, 2));
+	else
+		nd = map_id2nd(st->oid);
 
-	if(nd) {
-		unit_stop_walking(&nd->bl,1|4);
+	if (!nd) {
```
Lemongrass3110, line 16158:
```suggestion
	if (nd == nullptr) {
```

src/map/script.cpp
```diff
@@ -27584,9 +27641,9 @@ struct script_function buildin_func[] = {
 	BUILDIN_DEF(mobcount,"ss"),
 	BUILDIN_DEF(getlook,"i?"),
 	BUILDIN_DEF(getsavepoint,"i?"),
-	BUILDIN_DEF(npcspeed,"i"), // [Valaris]
-	BUILDIN_DEF(npcwalkto,"ii"), // [Valaris]
-	BUILDIN_DEF(npcstop,""), // [Valaris]
+	BUILDIN_DEF(npcspeed,"i?"), // [Valaris]
+	BUILDIN_DEF(npcwalkto,"ii??"), // [Valaris]
+	BUILDIN_DEF(npcstop,"?"), // [Valaris]
```
Lemongrass3110, line 27646:
```suggestion
	BUILDIN_DEF(npcspeed,"i?"),
	BUILDIN_DEF(npcwalkto,"ii??"),
	BUILDIN_DEF(npcstop,"?"),
```

src/map/script.cpp
Atemo: The "speed" parameter was to match aegis command. Maybe this parameter is overkill since we save the walking speed indefinitely. Ideally I would prefer to have it.

Should I remove the parameter?
How about
```
	if (nd->speed != speed) {
		if (speed < MIN_WALK_SPEED || speed > MAX_WALK_SPEED) {
			ShowError("buildin_npcwalkto: invalid speed %d (min: %d, max: %d).\n", speed, MIN_WALK_SPEED, MAX_WALK_SPEED);
		etc..
	}
```

doc/script_commands.txt
```diff
@@ -7961,17 +7961,15 @@ Return values:
...
-These commands will make the NPC object in question move around the map. As they
-currently are, they are a bit buggy and are not useful for much more than making
-an NPC move randomly around the map.
+These commands will make the NPC object in question move around the map.
 
 'npcspeed' will set the NPCs walking speed to a specified value. As in the
-@speed GM command, 200 is the slowest possible speed while 0 is the fastest
-possible (instant motion). 100 is the default character walking speed.
+@speed GM command, MAX_WALK_SPEED (1000) is the slowest possible speed while MIN_WALK_SPEED (20) is the fastest
+possible (instant motion). DEFAULT_NPC_WALK_SPEED (200) is the default npc walking speed.
```
aleos89, line 7972:
```suggestion
possible (instant motion). DEFAULT_NPC_WALK_SPEED (200) is the default NPC walking speed.
```

aleos89 (approved):

src/map/script.cpp
Lemongrass3110: If you prefer to have it, it is ok.
But we should document it somehow, because I think for the users it creates the impression that the speed will only be changed for "this" movement and not forever.
Thats why I would have removed it to not create confusion.

Atemo: The speed parameter was finally removed to avoid confusion

src/map/script.cpp
```diff
@@ -16071,48 +16071,94 @@ BUILDIN_FUNC(chatmes)
...
+			ShowError("buildin_npcwalkto: %s is a non-existing NPC.\n", script_getstr(st, 4));
 		else
-			status_calc_npc(nd, SCO_NONE);
-		unit_walktoxy(&nd->bl,x,y,0);
+			ShowError("buildin_npcwalkto: non-existing NPC.\n");
+		return SCRIPT_CMD_FAILURE;
 	}
+
+	int x = script_getnum(st, 2);
+	int y = script_getnum(st, 3);
```
Lemongrass3110, line 16129:
```suggestion
	if( nd->bl.m < 0 ){
		ShowError( "buildin_npcwalkto: NPC is not on a map.\n" );
		return SCRIPT_CMD_FAILURE;
	}

	struct map_data* mapdata = map_getmapdata( nd->bl.m );
	int x = script_getnum(st, 2);
	int y = script_getnum(st, 3);

	if( x < 0 || x >= mapdata->xs || y < 0 || y >= mapdata->ys ){
		ShowWarning( "buildin_npcwalkto: xcoordinates %d/%d are out of bounds in map %s(%dx%d).\n", x, y, mapdata->name, mapdata->xs, mapdata->ys );
		return SCRIPT_CMD_FAILURE;
	}
```

src/map/script.cpp
```diff
@@ -16071,48 +16071,94 @@ BUILDIN_FUNC(chatmes)
...
+		nd = map_id2nd(st->oid);
+
+	if (nd == nullptr) {
+		if (script_hasdata(st, 2))
+			ShowError("buildin_npcstop: %s is a non-existing NPC.\n", script_getstr(st, 2));
+		else
+			ShowError("buildin_npcstop: non-existing NPC.\n");
+		return SCRIPT_CMD_FAILURE;
 	}
+	unit_stop_walking( &nd->bl, USW_FIXPOS | USW_MOVE_FULL_CELL | USW_FORCE_STOP );
```
Lemongrass3110, line 16160: Just because I see that you added USW_FORCE_STOP here.
Do we want to make flag a parameter?

Lemongrass3110 (approved):

doc/script_commands.txt
```diff
@@ -7962,12 +7962,12 @@ Return values:
...
 
 *npcspeed( <speed value> {,"<npc name>"} );
-*npcwalkto( <x>,<y> {,"<speed value>" {,"<npc name>"} } );
+*npcwalkto( <x>,<y> {,"<npc name>"} } );
 *npcstop( {"<npc name>"} );
 
 These commands will make the NPC object in question move around the map.
 
-'npcspeed' will set the NPCs walking speed to a specified value. As in the
+'npcspeed' will perma set the NPCs walking speed to a specified value. As in the
```
secretdataz:
```suggestion
'npcspeed' will permanently set the NPCs walking speed to a specified value. As in the
```

Issue #8364, opened 2024-05-26
Illusion of under water daily quest not working


BotGigo:
### rAthena Hash

7ee7a25887dce884a66cb82cd4202a8f8f4bb913

### Client Date

2022-04-06

### Server Mode

Renewal

### Result

The players are reporting that they can´t take the daily quest fo said dungeon, the script is just letting complete the main quest, but don´t let players take the daily quest.

### Expected Result

Jerimon is supposed to give you the option to repeat the hunting mission inside the dungeon, but this is not happening.

### How to Reproduce

Complete the main quest of the Illusion Of Under Water dungeon, then try to take the daily quest. it will not work.

### Official Information

"Entrance to Illusion of Underwater is open now. Daily Quests are avalable at Jerimon NPC inside now."

This is taken from the step 11 of the main quest from iRo. [https://irowiki.org/wiki/Illusion_of_Underwater#Daily_Quests](url)

Atemo:
I didn't find any issues while testing.
Do you have the correct hash? A cooldown has been fixed in commit https://github.com/rathena/rathena/commit/fbd0478864990369111960e9b6c620160c1b7107

BotGigo:
Yes Sr, the has is correct, and yes, that fix is already in my emulator. ill post 

this is the msg that is giving me.
![imagen](https://github.com/rathena/rathena/assets/59173632/d835f801-9af6-4c2c-986e-7080116027af)

my impresion is that for some reason the script is not erasing the cool down quests, and is stuck since it needs to clear both quest in order to re take them.

BotGigo:
![imagen](https://github.com/rathena/rathena/assets/59173632/3528fd1f-7b22-465f-9c6f-d4a59eacfd0a)
this is the db status for both quests in the sql db

BotGigo:
[quests_illusion_dungeons.txt](https://github.com/rathena/rathena/files/15462045/quests_illusion_dungeons.txt)

i think the problem is in line 9598 and 9645 for quest 3414 and 3416. The state 1 in the quest leads to a dead end in the script.

BotGigo:
I found the solution. ill make a pull request and quot this issue.

gmragnarok:
Don't close this issue, leave it open until the pull request is merged. This is so that it becomes a common concern if this is an issue.

BotGigo:
> Don't close this issue, leave it open until the pull request is merged. This is so that it becomes a common concern if this is an issue.

understood

[...]

PR #8383, opened 2024-05-28
Illusion of underwater fix

BotGigo:
Add a variable update in order to correct a bug in the refresh of daily quest for Illusion of underwater dungeon.

Addressed issue: #8364

Server Mode:
Renewal

Description of Pull Request:

When you complete the main quest for illusion of underwater dungeon, the 2 quest goes to cooldown (3414 & 3416).
When the cooldown is completed for both quest (quest status 2), the script is unable to make the player to retake the quests. This happerd because when the script checks the quest status via "switch( checkquest(3414,PLAYTIME) ) {" in line 9545 and "switch( checkquest(3416,PLAYTIME) ) {" in line 9569, the case 2 fo each switch menu is only doing a break;


gmragnarok:
It seems I've also experienced a similar problem

PR #8405, opened 2024-06-06
Fixed clif_move to use the correct movement packets

Daegaladh:
* **Addressed Issue(s)**: n/a

* **Server Mode**: Both

 Fixes clif_move to use the correct movement packets (unit_walkingType) instead of the unused ZC_NOTIFY_MOVE (0x86) packet.
 Removes speed_changed unit state since is no longer needed
 Official servers never seemed to use ZC_NOTIFY_MOVE at all, so maybe is an alpha packet.
 Slightly fixes movement de-sync.

Packet ZC_NOTIFY_MOVE (0x86) was introduced in https://github.com/rathena/rathena/commit/96921544 as a way to reduce bandwidth usage even though the packet was never used on official servers, probably because it caused desync

speed_changed state was introduced in https://github.com/rathena/rathena/commit/d404a449a15d6f2372115d36609a606351d36a4c to use the (correct) walking packets when the speed is changed, since ZC_NOTIFY_MOVE cannot update the speed of the unit

In https://github.com/rathena/rathena/commit/ae7050eb1a00e7fb32faf70130468c696e5a3464 the use of this wrong packet was extended further without any need instead of using the correct packet.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -1991,104 +1991,64 @@ void clif_walkok( map_session_data& sd ){
 }
 
 
-static void clif_move2( struct block_list *bl, struct view_data *vd, struct unit_data *ud ){
-	status_change *sc = nullptr;
-
-	if ((sc = status_get_sc(bl)) && sc->option&(OPTION_HIDE|OPTION_CLOAK|OPTION_INVISIBLE|OPTION_CHASEWALK))
-		clif_ally_only = true;
-
-	clif_set_unit_walking( bl, nullptr, ud, AREA_WOS );
-
-	if(vd->cloth_color)
-		clif_refreshlook(bl,bl->id,LOOK_CLOTHES_COLOR,vd->cloth_color,AREA_WOS);
-	if(vd->body_style)
-		clif_refreshlook(bl,bl->id,LOOK_BODY2,vd->body_style,AREA_WOS);
-
-	switch(bl->type) {
-	case BL_PC:
-		{
-			TBL_PC *sd = ((TBL_PC*)bl);
-//			clif_movepc(sd);
-			if(sd->state.size==SZ_BIG) // tiny/big players [Valaris]
-				clif_specialeffect(&sd->bl,EF_GIANTBODY2,AREA);
-			else if(sd->state.size==SZ_MEDIUM)
-				clif_specialeffect(&sd->bl,EF_BABYBODY2,AREA);
-			if (sd->status.robe)
-				clif_refreshlook(bl,bl->id,LOOK_ROBE,sd->status.robe,AREA);
-		}
-		break;
-	case BL_MOB:
-		{
-			TBL_MOB *md = ((TBL_MOB*)bl);
-			if(md->special_state.size==SZ_BIG) // tiny/big mobs [Valaris]
-				clif_specialeffect(&md->bl,EF_GIANTBODY2,AREA);
-			else if(md->special_state.size==SZ_MEDIUM)
-				clif_specialeffect(&md->bl,EF_BABYBODY2,AREA);
-		}
-		break;
-	case BL_PET:
-		if(vd->head_bottom) // needed to display pet equip properly
-			clif_pet_equip_area((TBL_PET*)bl);
-		break;
-	}
-	clif_ally_only = false;
-}
-
-
 /// Notifies clients in an area, that an other visible object is walking.
-/// 0086 <id>.L <walk data>.6B <walk start time>.L (ZC_NOTIFY_MOVE)
 /// Note: unit must not be self
 void clif_move(struct unit_data *ud)
 {
-	struct view_data* vd;
 	struct block_list* bl = ud->bl;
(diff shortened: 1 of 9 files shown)


src/map/clif.cpp
```diff
@@ -1987,7 +1987,7 @@ void clif_walkok( map_session_data& sd ){
 
 /// Notifies clients in an area, that an other visible object is walking.
 /// Note: unit must not be self
-void clif_move(struct unit_data& ud)
+void clif_move( struct unit_data& ud )
```
Playtester, line 1990: Why do you now also start with this weird spaces inside parenthesis style from Lemon?

Daegaladh: Just for better readability and trying to unify styles, we can always change it later

Playtester: But if you unify it, why not unify it in a good format? Besides, we said we don't want to change styles unless we change the code line so that blame shows the person who actually changed it.

Daegaladh: What's wrong with it? And I've just changed it for the lines I actually modified

src/map/clif.cpp
```diff
@@ -1436,31 +1430,31 @@ static void clif_set_unit_walking( struct block_list *bl, map_session_data *tsd,
...
 		p.maxHP = -1;
 		p.HP = -1;
 	}
 
-	if( bl->type == BL_MOB ){
-		p.isBoss = ( (mob_data*)bl )->get_bosstype();
-	}else if( bl->type == BL_PET ){
-		p.isBoss = ( (pet_data*)bl )->db->get_bosstype();
+	if( bl.type == BL_MOB ){
+		p.isBoss = ( BL_CAST(BL_MOB, &bl) )->get_bosstype();
```
Lemongrass3110, line 1455:
```suggestion
		p.isBoss = static_cast<mob_data*>( &bl )->get_bosstype();
```
Type was already checked, no need to do it again.

src/map/clif.cpp
```diff
@@ -1436,31 +1430,31 @@ static void clif_set_unit_walking( struct block_list *bl, map_session_data *tsd,
...
+	}else if( bl.type == BL_PET ){
+		p.isBoss = ( BL_CAST(BL_PET, &bl) )->db->get_bosstype();
```
Lemongrass3110, line 1457:
```suggestion
		p.isBoss = static_cast<pet_data*>( &bl )->db->get_bosstype();
```
Type was already checked, no need to do it again.

src/map/clif.cpp
```diff
@@ -1991,104 +1985,64 @@ void clif_walkok( map_session_data& sd ){
...
-	//This performance check is needed to keep GM-hidden objects from being notified to bots.
-	else if( vd->class_ == JT_INVISIBLE ){
-		// If the player was disguised we still need to update the disguised unit, since the main unit will be updated through clif_walkok
-		if(disguised(bl)) {
-			packet.gid = disguised_bl_id(bl->id);
-			WBUFPOS2(packet.moveData, 0, bl->x, bl->y, ud->to_x, ud->to_y, 8, 8);
-			packet.moveStartTime = client_tick(gettick());
-
-			clif_send(&packet, sizeof(packet), bl, SELF);
-		}
```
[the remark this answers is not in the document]
Daegaladh: Yeah, because there's already a check in clif_set_unit_walking, that was just for packet 0x86

src/map/clif.cpp
```diff
@@ -1991,104 +1985,64 @@ void clif_walkok( map_session_data& sd ){
...
 	if (clif_npc_mayapurple(bl))
 		return;
 
-	if (ud->state.speed_changed) {
-		// Since we don't know how to update the speed of other objects,
-		// use the old walk packet to update the data.
-		ud->state.speed_changed = 0;
-		clif_move2(bl, vd, ud);
-		return;
-	}
```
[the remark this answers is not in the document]
Daegaladh: Now that you mention it, I think we can safely remove speed_changed from the code

src/map/clif.cpp
```diff
@@ -1991,104 +1985,64 @@ void clif_walkok( map_session_data& sd ){
...
+		{
+			map_session_data* sd = BL_CAST(BL_PC, bl);
+			if (sd != nullptr) {
+				if (sd->state.size == SZ_BIG) // tiny/big players [Valaris]
+					clif_specialeffect(&sd->bl, EF_GIANTBODY2, AREA);
+				else if (sd->state.size == SZ_MEDIUM)
+					clif_specialeffect(&sd->bl, EF_BABYBODY2, AREA);
+				if (sd->status.robe)
+					clif_refreshlook(bl, bl->id, LOOK_ROBE, sd->status.robe, AREA);
+			}
```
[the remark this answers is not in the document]
Daegaladh: Right, I forgot to remove those after changing to references

Lemongrass3110 (approved):

aleos89 (approved):

PR #8374, opened 2024-05-27
Follow up to 05ced985e33823201e592c198a52d9f3ec8feea6

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**:
-Replaces more constant names with the official ones
-Adds the official MSI_ prefix to the constants
-Fixes some constant line numbers
-Removes clif_mercenary_message
-Adds packetver checks for some constants that changes line numbers in some client dates

PR #8407, opened 2024-06-07
Fixes a couple of clif_message constant values

Aristeus09:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/pull/8374

2013-08-07 Client
20130807_msgstringtable

Base on 2013-08-07 Client the line for ```%d seconds left until you can use#``` is in ```1864```. if client process it by -1 it should be ```1863```

20211103 Client
20211103_msgstringtable

While in 2021-11-03 client the ```%d seconds left until you can use#``` is in ```1863```. if client process it by -1 it should be ```1862```

* **Server Mode**:  Both

Correcting ID numbers for MSI_ITEM_REUSE_LIMIT_SECOND base on msgstringtable.txt of client.
thanks to @llchrisll for helping me.

### DIFF
--- a/src/map/clif.hpp
+++ b/src/map/clif.hpp
@@ -577,7 +577,7 @@ enum clif_messages : uint16_t {
 	MSI_PARTY_MASTER_CHANGE_SAME_MAP = 2095,
 
 	// Merge items available does not exist.
-	MSI_NOT_EXIST_MERGE_ITEM = 2183,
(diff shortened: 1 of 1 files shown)


Daegaladh:
I'll re-check all lines again, because these doesn't seem to match with kRO

Aristeus09:
okay, thank you @Daegaladh.

PR #8409, opened 2024-06-08
Clean-up sc_display

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

Content
----------------------------------
* Statuses can now also be saved for monsters
* Updated the status.txt docs
* Added missing BL_NPC to BL_SCEFFECT
* Corrected conditions for saving statuses depending on the unit type affected
* Corrected conditions for displaying status depending on the unit type affected when the status is given

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -1721,7 +1721,7 @@ int clif_spawn( struct block_list *bl, bool walking ){
 				clif_spiritcharm(sd);
 			if (sd->status.robe)
 				clif_refreshlook(bl,bl->id,LOOK_ROBE,sd->status.robe,AREA);
-			clif_efst_status_change_sub(bl, bl, AREA);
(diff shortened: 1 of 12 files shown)


src/map/clif.cpp
```diff
@@ -6599,20 +6593,23 @@ void clif_efst_status_change_sub(struct block_list *tbl, struct block_list *bl,
...
 				break;
 			case SC_HELLS_PLANT:
-				if( sc && sc->getSCE(type) ){
-					tick = sc->getSCE(type)->val4;
-				}
+				tick = sce->val4;
 				break;
 		}
 
+		int dval1 = scdb->flag[SCF_SENDVAL1] ? sce->val1 : 0;
```
aleos89, line 6605: Are the val# not 0 by default? Could have sworn it was but maybe I'm blowing smoke!

aleos89 (approved):

Issue #8414, opened 2024-06-10
Attribute table should apply to the sum of weaponAtk and equipAtk


Playtester:
### rAthena Hash

17a6d03

### Client Date

2021-11-03

### Server Mode

Renewal

### Result

The elemental table is currently applied to weaponAtk and equipAtk individually and each value is rounded down.

### Expected Result

The elemental table is applied to the sum of weaponAtk and equipAtk and the final result is rounded down.

### How to Reproduce

Need to find a situation where the rounding would make a difference in the damage. You need to have equipAtk, so ideally tested with Bows or Guns.

### Official Information

From https://github.com/rathena/rathena/issues/8380

99/70 Gunslinger, 99 Dex, 99 Luk
Six Shooter vs. Golem

Bullet (25 ATK) - 225-227 (Crit 317)
Silver Bullet (15 ATK) - 217-219 (Crit 308)
Shell of Blood (30 ATK) - 228-230 (Crit 322)
GongBug (50 ATK) - 242-244 (Crit 341)

vs. Verit
Bullet (25 ATK) - 319-321 (Crit 449)
Silver Bullet (15 ATK) - 322-326 (Crit 456) <-- rAthena 455
Shell of Blood (30 ATK) - 323-326 (Crit 456)
GongBug (50 ATK) - 341-343 (Crit 480)

Silver Bullet:
vs. Poporing - 329-331 (Crit 463) <-- rAthena 464
vs. Zombie - 376-378 (Crit 532) <-- rAthena 530
vs. Drainliar - 348-352 (Crit 492)

Silver Bullet vs. Training Dummy:
vs. Neutral 21065 - 347-350 (Crit 490)
vs. Poison 21082 - 333-335 (Crit 469) <-- rAthena 333-336 (Crit 470)
vs. Holy 21083 - 288 (Crit 403)
vs. Dark 21084 - 361-365 (Crit 511) <-- rAthena 361-364 (Crit 509)
vs. Ghost 21085 - 315-318 (Crit 445)
vs. Undead 21086 - 361-365 (Crit 511) <-- rAthena 361-364 (Crit 509)

with Porcellio Card:
vs. Neutral 21065 - 372-375 (Crit 525)
vs. Poison 21082 - 351-354 (Crit 495)
vs. Holy 21083 - 288 (Crit 403)
vs. Dark 21084 - 393-396 (Crit 554)
vs. Ghost 21085 - 340-343 (Crit 480)
[...]

PR #8415, opened 2024-06-10
Cannon Ball Skills (P.ATK and Res)

### LINKED ISSUE #8380: PATK/Res Effect on Cart Cannon / Arm Cannon
### rAthena Hash

924d1dd

### Client Date

2021-11-03

### Server Mode

Renewal

### Result

PATK affects Cart Cannon's "Cannon Ball" ATK.

RES affects Cart Cannon.

### Relevant Log Output

_No response_

### Expected Result

PATK does not affect Cart Cannon's "Cannon Ball" ATK.

RES does not affect Cart Cannon.

### How to Reproduce

Use Cart Cannon with a character that has PATK against a character that has RES with only a Cannon Ball equipped. Compare damage.

### Official Information

grafik
Biolo 250/55
Cannon Ball -> Item ID 18004
Cart Boost 5
Cart Cannon 5

vs. Inquisitor 250/51
grafik
Damage: 114523

Damage vs. Dummy (21064): 114667

Normal Attack Ungeared -> 2096
With Cart Boost 5 -> 2146

With Weapon -> Item ID 590047

Normal Attack Crit -> 4391
With Cart Boost 5 -> 4479

Note: The previous tests on Guillotine Fist also show that Res is ignored: https://github.com/rathena/rathena/issues/8363
It is probably that all skills that use the "Simple Defense" formula also ignore Res. But this requires more tests and should be done in a separate PR, so I removed it from the scope.

The 250 ATK from Cannon Ball is apparently treated like MasteryATK rather than EquipATK. At first I thought EquipATK is not affected by PATK in general, but these countertests proof that EquipATK IS affected:
grafik
Regular Bow Critical Hits vs Dummy (21065):
Arrow - 3097
Iron Arrow - 3107
Elven Arrow - 3137
Oridecon Arrow - 3147

Elemental Test:
Crystal Arrow vs. Dummy (21080) - 3162

_No response_

Playtester:
* **Addressed Issue(s)**: #8380 

* **Server Mode**: Renewal

- When using Cart Cannon or Arm Cannon, the arrow attack you get from cannon balls is no longer affected by P.ATK
- Cart Cannon and Arm Cannon are no longer reduced by Res
- More consistent naming (PATK -> P.ATK, SMATK -> S.MATK)
- Fixes #8380

**Hints for Reviewers**:

- It seems that Cannon Balls are generally not adding any equip attack, but the damage of Cart Cannon and Arm Cannon is still increased by arrow attack, so my assumption is that this bonus is hard-coded into the two skills
- So far all tests indicate that all skills that use the simple defense formula also ignore Res, should be turned into a general rule later on after all skills have been tested
- Test numbers for Cart Cannon / Arm Cannon in the linked issue all match already, but if you run other tests you still might find small deviations when doing elemental tests, see #8414

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3420,7 +3420,9 @@ static int battle_calc_equip_attack(struct block_list *src, int skill_id)
 		struct status_data *status = status_get_status_data(src);
 		map_session_data *sd = BL_CAST(BL_PC, src);
 
-		if (sd) // add arrow atk if using an applicable skill
+		// Cart Cannon is an arrow skill but adds arrow attack as mastery damage instead
+		if (sd && skill_id != GN_CARTCANNON)
+			// add arrow atk if using an applicable skill
 			eatk += (is_skill_using_arrow(src, skill_id) ? sd->bonus.arrow_atk : 0);
 
 		return eatk + status->eatk;
@@ -3903,6 +3905,16 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
 #endif
 				}
 				break;
+#ifdef RENEWAL
+			case GN_CARTCANNON:
+				// These skills add arrow attack as mastery attack instead of equip attack
+				// The only way to determine this is by confirming that the arrow attack isn't influenced by P.ATK
+				if (sd) {
+					struct status_data* tstatus = status_get_status_data(target);
+					ATK_ADD(wd->masteryAtk, wd->masteryAtk2, battle_attr_fix(src, target, sd->bonus.arrow_atk, sd->bonus.arrow_ele, tstatus->def_ele, tstatus->ele_lv));
+				}
+				break;
+#endif
 		}
 
 		if (sc) { // Status change considered as masteries
@@ -7442,7 +7454,7 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 			if( is_attack_left_handed( src, skill_id ) ){
 				wd.damage2 = wd.statusAtk2 + wd.weaponAtk2 + wd.equipAtk2 + wd.percentAtk2;
 			}
-			// Apply PATK mod
+			// Apply P.ATK mod
 			// But for Dragonbreaths it only applies if Dragonic Aura is skilled
 			if( ( skill_id != RK_DRAGONBREATH && skill_id != RK_DRAGONBREATH_WATER ) || pc_checkskill( sd, DK_DRAGONIC_AURA ) > 0 ){
 				wd.damage = (int64)floor( (float)( wd.damage * ( 100 + sstatus->patk ) / 100 ) );
@@ -7493,7 +7505,7 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 		// Res reduces physical damage by a percentage and
 		// is calculated before DEF and other reductions.
 		// This should be the official formula. [Rytech]
-		if ((wd.damage + wd.damage2) && tstatus->res > 0 && skill_id != MO_EXTREMITYFIST) {
+		if ((wd.damage + wd.damage2) && tstatus->res > 0 && skill_id != MO_EXTREMITYFIST && skill_id != GN_CARTCANNON) {
 			short res = tstatus->res;
 			short ignore_res = 0;// Value used as a percentage.
 
@@ -8733,7 +8745,7 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 						skillratio += 25;
 				}
 #ifdef RENEWAL
-				// SMATK needs to be applied before the skill ratio to prevent rounding issues
+				// S.MATK needs to be applied before the skill ratio to prevent rounding issues
 				if (sd && sstatus->smatk > 0)
 					ad.damage += ad.damage * sstatus->smatk / 100;
 #endif
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3420,10 +3420,13 @@ static int battle_calc_equip_attack(struct block_list *src, int skill_id)
 		struct status_data *status = status_get_status_data(src);
 		map_session_data *sd = BL_CAST(BL_PC, src);
 
-		// Cart Cannon is an arrow skill but adds arrow attack as mastery damage instead
-		if (sd != nullptr && skill_id != GN_CARTCANNON)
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -3903,6 +3905,16 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
 #endif
 				}
 				break;
+#ifdef RENEWAL
```
aleos89, line 3908:
```suggestion
```
This skill is renewal only, so I think this is extra.

src/map/battle.cpp
```diff
@@ -3903,6 +3905,16 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
...
+			case GN_CARTCANNON:
+				// These skills add arrow attack as mastery attack instead of equip attack
+				// The only way to determine this is by confirming that the arrow attack isn't influenced by P.ATK
+				if (sd) {
+					struct status_data* tstatus = status_get_status_data(target);
+					ATK_ADD(wd->masteryAtk, wd->masteryAtk2, battle_attr_fix(src, target, sd->bonus.arrow_atk, sd->bonus.arrow_ele, tstatus->def_ele, tstatus->ele_lv));
+				}
+				break;
+#endif
```
aleos89, line 3917:
```suggestion
```

aleos89 (approved):

src/map/battle.cpp
```diff
@@ -3903,6 +3905,16 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
 #endif
 				}
 				break;
+#ifdef RENEWAL
+			case GN_CARTCANNON:
+				// These skills add arrow attack as mastery attack instead of equip attack
```
Lemongrass3110, line 3910:
```suggestion
				// This skill adds arrow attack as mastery attack instead of equip attack
```

src/map/battle.cpp
```diff
@@ -3903,6 +3905,16 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
...
+				// The only way to determine this is by confirming that the arrow attack isn't influenced by P.ATK
```
Lemongrass3110, line 3911: Is this a documentation of how you found out that it should be changed like this or a TODO for the future?

src/map/battle.cpp
```diff
@@ -3903,6 +3905,16 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
...
+				if (sd) {
```
Lemongrass3110, line 3912:
```suggestion
				if (sd != nullptr) {
```

src/map/battle.cpp
```diff
@@ -3420,7 +3420,9 @@ static int battle_calc_equip_attack(struct block_list *src, int skill_id)
 		struct status_data *status = status_get_status_data(src);
 		map_session_data *sd = BL_CAST(BL_PC, src);
 
-		if (sd) // add arrow atk if using an applicable skill
+		// Cart Cannon is an arrow skill but adds arrow attack as mastery damage instead
+		if (sd && skill_id != GN_CARTCANNON)
```
Lemongrass3110, line 3424:
```suggestion
		if (sd != nullptr && skill_id != GN_CARTCANNON)
```

src/map/battle.cpp
```diff
@@ -3420,7 +3420,9 @@ static int battle_calc_equip_attack(struct block_list *src, int skill_id)
 		struct status_data *status = status_get_status_data(src);
 		map_session_data *sd = BL_CAST(BL_PC, src);
 
-		if (sd) // add arrow atk if using an applicable skill
+		// Cart Cannon is an arrow skill but adds arrow attack as mastery damage instead
```
Lemongrass3110, line 3423:
```suggestion
		// Add arrow atk if using an applicable skill
		// Cart Cannon is an exceptional arrow skill and adds arrow attack as mastery damage instead
```

src/map/battle.cpp
```diff
@@ -3420,7 +3420,9 @@ static int battle_calc_equip_attack(struct block_list *src, int skill_id)
...
+		if (sd && skill_id != GN_CARTCANNON)
+			// add arrow atk if using an applicable skill
```
Lemongrass3110, line 3425:
```suggestion
```

Lemongrass3110 (approved):

src/map/battle.cpp
Playtester: I put it in plural so people add similar skills to this list rather than creating their own checks.

src/map/battle.cpp
Playtester: Both - so whenever someone wonders whether to add a skill to the list here or not they know how they can verify that.

src/map/battle.cpp
Lemongrass3110: But you merged your check to the outer and not the inner (short) if.

src/map/battle.cpp
Playtester: Rephrased so it reflects reality better.

src/map/battle.cpp
Playtester: Reworked this part

src/map/battle.cpp
```diff
@@ -3903,6 +3908,16 @@ static void battle_calc_attack_masteries(struct Damage* wd, struct block_list *s
...
+#ifdef RENEWAL
+			case GN_CARTCANNON:
+			case NC_ARMSCANNON:
+				// Arrow attack of these skills is not influenced by P.ATK so we add it as mastery attack
+				if (sd != nullptr) {
+					struct status_data* tstatus = status_get_status_data(target);
+					ATK_ADD(wd->masteryAtk, wd->masteryAtk2, battle_attr_fix(src, target, sd->bonus.arrow_atk, sd->bonus.arrow_ele, tstatus->def_ele, tstatus->ele_lv));
+				}
+				break;
+#endif
```
Lemongrass3110, line 3920: Since you made it generic for all cannonballs in battle_calc_equip_attack, shouldnt we make it generic here too?

Lemongrass3110 (approved):

src/map/battle.cpp
Playtester: As discussed on Discord, my assumption is that this part is in the skill-specific code on official servers, so I designed it on our side alike to be prepared for future skills.

PR #8442, opened 2024-06-17
Max RES/MRES and max RES/MRES ignored

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

According to the tipbox 106 and 107 :
`The physical resistance ignore option only applies up to 50%.`
`Ignore magic resistance option only applies up to 50%.`

There should be no limit to the max RES/MRES a unit can have. A unit can only ignore (pierce) 50% of the target's RES/MRES.

Content
--------------------------------------------
* Removed the limit of RES/MRES and remove `max_res_mres_reduction` battle config
* Added a limit of RES/MRES ignored by items and skills in `max_res_mres_ignored` battle config

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -7513,15 +7513,10 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 					ignore_res += sc->getSCE(SC_POTENT_VENOM)->val2;
 			}
 
-			ignore_res = min(ignore_res, 100);
+			ignore_res = min(ignore_res, battle_config.max_res_mres_ignored);
 
 			if (ignore_res > 0)
 				res -= res * ignore_res / 100;
-
-			// Max damage reduction from Res is officially 50%.
-			// That means 625 Res is needed to hit that cap.
-			if (res > battle_config.max_res_mres_reduction)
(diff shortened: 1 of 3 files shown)


src/map/battle.cpp
```diff
@@ -11499,7 +11489,7 @@ static const struct _battle_data {
 	{ "use_traitpoint_table",               &battle_config.use_traitpoint_table,            1,      0,      1,              },
 	{ "trait_points_job_change",            &battle_config.trait_points_job_change,         7,      1,      1000,           },
 	{ "max_trait_parameter",                &battle_config.max_trait_parameter,             100,    10,     SHRT_MAX,       },
-	{ "max_res_mres_reduction",             &battle_config.max_res_mres_reduction,          625,    1,      SHRT_MAX,       },
+	{ "max_res_mres_ignored",               &battle_config.max_res_mres_ignored,            50,     1,      100,            },
```
Lemongrass3110, line 11492:
```suggestion
	{ "max_res_mres_ignored",               &battle_config.max_res_mres_ignored,            50,     0,      100,            },
```
Otherwise cannot be disabled. Code also checks for this.

Lemongrass3110 (approved):

aleos89 (approved):

PR #8443, opened 2024-06-17
 Guillotine Fist SP Recovery Prevention Cleanup

### LINKED ISSUE #8377: Guillotine Fist SP Recovery Prevention
### rAthena Hash

924d1dd

### Client Date

2021-11-03

### Server Mode

Renewal

### Result

After you use Guillotine Fist you can't use SP recovery items for 3 seconds and you don't naturally recover SP for 300 seconds or until you relog.

Guillotine Fist gives two status changes: SC_EXTREMITYFIST and SC_EXTREMITYFIST2

### Relevant Log Output

_No response_

### Expected Result

After you use Guillotine Fist you can't use SP recovery items for 3 seconds and you don't naturally recover SP for 3 seconds.

Guillotine Fist only requires one status change that lasts 3 seconds and has both effects.

### How to Reproduce

Use Guillotine Fist, check if you recover SP after 3 seconds.
Check source code for SC_EXTREMITYFIST2 to find second status change that is not needed.

### Official Information

Tested on kRO.

That both effects only last 3 seconds is the case since a kRO update in 2018: https://www.divine-pride.net/forum/index.php?/topic/3453-kro-mass-skills-balance-1st-2nd-and-transcendent-classes-skills/

_No response_

Playtester:
* **Addressed Issue(s)**: #8377 

* **Server Mode**: Both

- Unified code for SC_EXTREMITYFIST
  * It stops natural SP recovery for 300s in pre-re
  * It stops natural SP recovery and recovery from items for 3s and shows icon in renewal
- Removed SC_EXTREMITYFIST2 as it's no longer needed
- Fixed an issue that caused a natural SP regen tick to happen right after using Guillotine Fist
- Fixes #8377 

**Hints for Reviewers**

I removed the constant from the enumeration instead of keeping the numbers of the status changes afterwards as none of those are preserved between server restarts anyway.


src/map/status.hpp
```diff
@@ -1307,7 +1307,6 @@ enum sc_type : int16 {
 	SC_PERIOD_RECEIVEITEM_2ND,
 	SC_PERIOD_PLUSEXP_2ND,
 
-	SC_EXTREMITYFIST2,
```
aleos89, line 1310: I thought this was already commented on so it's harder to load the rest of the code, but we need to be careful about erasing SC in the middle since this is the array index stored in SQL. If we need to remove it we should provide a SQL upgrade script to adjust index values above this -1.

src/map/skill.cpp
```diff
@@ -13487,15 +13478,9 @@ TIMER_FUNC(skill_castend_id){
 		//Consume SP/spheres
 		skill_consume_requirement(sd,ud->skill_id, ud->skill_lv,1);
 		status_set_sp(src, 0, 0);
-		sc = &sd->sc;
```
aleos89, line 13490: Reviewing on mobile, but you erase this here. Is it needed later in the function? Should we remove the `sc` variable completely?

src/map/status.hpp
Playtester: See my explanation in the reviewer notes. Only two SCs come after it and those are not stored beyond server restart.

src/map/skill.cpp
Playtester: It's also a mystery to me but sc is defined at the top of the function and used also just 5 lines above:

`status_change *sc = status_get_sc(src);`

`!(sc && sc->getSCE(SC_FOGWALL))`

I don't understand why it was overwritten with &sd->sc here. Doesn't that return the exactly same sc object?

aleos89: I looked over the file on the PC and can agree that the removal here is safe. The `sc` assignment here is a duplicate, most likely from a left over in a refactor.

src/map/status.hpp
Atemo: How about 
`SC_EXTREMITYFIST2,	// unused`

Playtester: If you guys are concerned about this, I'll just do it the normal way (give the next one a fixed number). Just wanted the code to be more clean and we kept moving this SC since like a month ago anyway, so I didn't think it will hurt.

Playtester: Fixed

Atemo (approved):

aleos89 (approved):

PR #8448, opened 2024-06-19
Configs for account and character names and passwords minimum length

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Adds a setting in char_athena.conf to allow you to easily change the 4-letter character name minimum length limit when using the 'Remove 4/6 letter Character Name limit' diff patch.
Also modifies and splits the current setting for account names and passwords minimum lenght, to be more configurable.

### DIFF
--- a/conf/char_athena.conf
+++ b/conf/char_athena.conf
@@ -144,6 +144,10 @@ unknown_char_name: Unknown
 // To log the character server?
 log_char: yes
 
+// Enable or disable the 4-letter character name minimum length limit
+// Must be set to 'yes' unless your client uses the 'Remove 4 letter Character Name limit' diff patch.
+char_name_length_limit: yes
+
 // Allow or not identical name for characters but with a different case (upper/lower):
 // example: Test-test-TEST-TesT; Value: 0 not allowed (default), 1 allowed
 name_ignoring_case: no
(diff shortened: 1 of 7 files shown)


Lemongrass3110 (commented):
I would store the limit as a number. Would simplify the code a little and allow even more customization.

aleos89:
I thought about using an int value for the config as well but then like Daegaladh stated, the diff completely removes the size limit (with no customization into what you'd want). That would mean a bool on our end is the best fit.

aleos89 (approved):

aleos89 (approved):

conf/char_athena.conf
```diff
@@ -144,6 +144,10 @@ unknown_char_name: Unknown
 // To log the character server?
 log_char: yes
 
+// Minimum length for a character name.
+// Must be set to '4' unless your client uses the 'Remove 4/6 letter Character Name limit' diff patch.
+char_name_min_length: 4
```
Lemongrass3110, line 149: Is that one not packetver dependent, like the rest?

Lemongrass3110 (approved):

conf/char_athena.conf
Daegaladh: Despite the name of the diff, no, it's still 4 letters in the new UI

Issue #8426, opened 2024-06-14
Instance Check Party Return Failure


Haydrich:
### rAthena Hash

master

### Client Date

20230905

### Server Mode

Renewal

### Result

When a try to talk to instance npc while i have no party, it shows
```
[Warning]: Script command 'instance_check_party' returned failure.
[Debug]: Source (NPC): Hugin at glast_01 (204,273)
[Debug]: Source (NPC): Hugin is located in: npc/re/instances/OldGlastHeim.txt
```
This happens for all instances not just Old Glast Hein

### Expected Result

It should be with no warning?

### How to Reproduce

1. Click on Old Glast Heim Entrance NPC while your character have no party
2. Check mapserv> it shows warning and debug

### Official Information

N/A

PR #8457, opened 2024-06-25
Updated instance_check_party script command

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8426

* **Server Mode**: both

The command now returns 0 if the party id does not exist.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -9843,7 +9843,7 @@ For details on the other parameters, see 'announce'.
 
 This function checks if a party meets certain requirements, returning 1 if all
 conditions are met and 0 otherwise. It will only check online characters.
-The command returns 0 is the party id does not exist.
+The command returns 0 is the party ID does not exist.
 
 amount - number of online party members (default is 1).
 min    - minimum level of all characters in the party (default is 1).
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -9843,6 +9843,7 @@ For details on the other parameters, see 'announce'.
 
 This function checks if a party meets certain requirements, returning 1 if all
 conditions are met and 0 otherwise. It will only check online characters.
+The command returns 0 is the party id does not exist.
 
 amount - number of online party members (default is 1).
 min    - minimum level of all characters in the party (default is 1).
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -21837,46 +21837,44 @@ BUILDIN_FUNC(instance_announce) {
  *------------------------------------------*/
 BUILDIN_FUNC(instance_check_party)
 {
-	int amount, min, max, i, party_id, c = 0;
-	struct party_data *p;
+	int32 min = script_hasdata(st, 4) ? script_getnum(st, 4) : 1; // Minimum Level needed to join the Instance.
+	int32 max  = script_hasdata(st, 5) ? script_getnum(st, 5) : MAX_LEVEL; // Maxium Level allowed to join the Instance.
 
-	amount = script_hasdata(st,3) ? script_getnum(st,3) : 1; // Amount of needed Partymembers for the Instance.
-	min = script_hasdata(st,4) ? script_getnum(st,4) : 1; // Minimum Level needed to join the Instance.
-	max  = script_hasdata(st,5) ? script_getnum(st,5) : MAX_LEVEL; // Maxium Level allowed to join the Instance.
-
-	if( min < 1 || min > MAX_LEVEL) {
+	if (min < 1 || min > MAX_LEVEL) {
 		ShowError("buildin_instance_check_party: Invalid min level, %d\n", min);
+		script_pushint(st, 0);
 		return SCRIPT_CMD_FAILURE;
-	} else if(  max < 1 || max > MAX_LEVEL) {
+	}
+	else if (max < 1 || max > MAX_LEVEL) {
 		ShowError("buildin_instance_check_party: Invalid max level, %d\n", max);
+		script_pushint(st, 0);
 		return SCRIPT_CMD_FAILURE;
 	}
 
-	if( script_hasdata(st,2) )
-		party_id = script_getnum(st,2);
-	else return SCRIPT_CMD_FAILURE;
+	int32 party_id = script_getnum(st, 2);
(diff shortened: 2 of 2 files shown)


src/map/script.cpp
```diff
@@ -21837,46 +21837,44 @@ BUILDIN_FUNC(instance_announce) {
...
+			if (sd->status.base_level < min || sd->status.base_level > max) {
+				script_pushint(st, 0);
+				return SCRIPT_CMD_SUCCESS;
 			}
+			count++;
+		}
 	}
 
-	if(c < amount)
+	int32 amount = script_hasdata(st, 3) ? script_getnum(st, 3) : 1; // Amount of needed Partymembers for the Instance.
```
Lemongrass3110, line 21875: Should be validated (>0 && <=MAX_PARTY) if script_hasdata(st, 3)

src/map/script.cpp
```diff
@@ -21837,46 +21837,44 @@ BUILDIN_FUNC(instance_announce) {
...
+
+	if (count < amount)
 		script_pushint(st, 0); // Not enough Members in the Party to join Instance.
 	else
 		script_pushint(st, 1);
```
Lemongrass3110, line 21880:
```suggestion
	script_pushint(st, count >= amount);
```
Directly push the bool

src/map/script.cpp
```diff
@@ -21837,46 +21837,44 @@ BUILDIN_FUNC(instance_announce) {
...
+
+	for( int32 i = 0; i < MAX_PARTY; i++ ) {
+		map_session_data *sd;
+		if ((sd = p->data[i].sd) && map_id2bl(sd->bl.id) && !sd->state.autotrade) {
+			if (sd->status.base_level < min || sd->status.base_level > max) {
+				script_pushint(st, 0);
+				return SCRIPT_CMD_SUCCESS;
 			}
+			count++;
+		}
```
Lemongrass3110, line 21872:
```suggestion
		map_session_data *sd = p->data[i].sd;

		if (sd == nullptr){
			continue;
		}
		
		if (sd->state.autotrade) {
			continue;
		}
	
		if (sd->status.base_level < min || sd->status.base_level > max) {
			script_pushint(st, 0);
			return SCRIPT_CMD_SUCCESS;
		}

		count++;
```

src/map/script.cpp
```diff
@@ -21837,46 +21837,44 @@ BUILDIN_FUNC(instance_announce) {
...
-					script_pushint(st, 0);
-					return SCRIPT_CMD_SUCCESS;
-				} else if(pl_sd->status.base_level > max) {
-					script_pushint(st, 0);
-					return SCRIPT_CMD_SUCCESS;
-				}
-					c++;
+	int32 count = 0;
+
+	for( int32 i = 0; i < MAX_PARTY; i++ ) {
```
Lemongrass3110, line 21864:
```suggestion
	for( size_t i = 0; i < MAX_PARTY; i++ ) {
```

src/map/script.cpp
```diff
@@ -21837,46 +21837,44 @@ BUILDIN_FUNC(instance_announce) {
...
-	else return SCRIPT_CMD_FAILURE;
+	int32 party_id = script_getnum(st, 2);
+	party_data *p = party_search( party_id );
 
-	if( !(p = party_search(party_id)) ) {
-		script_pushint(st, 0); // Returns false if party does not exist.
-		return SCRIPT_CMD_FAILURE;
+	if (p == nullptr) {
+		script_pushint(st, 0);
+		return SCRIPT_CMD_SUCCESS;
```
Lemongrass3110, line 21859: Do we really support people like that? :-(
This is a command intended to check "A PARTY".

src/map/script.cpp
```diff
@@ -21837,46 +21837,44 @@ BUILDIN_FUNC(instance_announce) {
...
-	amount = script_hasdata(st,3) ? script_getnum(st,3) : 1; // Amount of needed Partymembers for the Instance.
-	min = script_hasdata(st,4) ? script_getnum(st,4) : 1; // Minimum Level needed to join the Instance.
-	max  = script_hasdata(st,5) ? script_getnum(st,5) : MAX_LEVEL; // Maxium Level allowed to join the Instance.
+	int32 min = script_hasdata(st, 4) ? script_getnum(st, 4) : 1; // Minimum Level needed to join the Instance.
+	int32 max  = script_hasdata(st, 5) ? script_getnum(st, 5) : MAX_LEVEL; // Maxium Level allowed to join the Instance.
 
-	if( min < 1 || min > MAX_LEVEL) {
+	if (min < 1 || min > MAX_LEVEL) {
 		ShowError("buildin_instance_check_party: Invalid min level, %d\n", min);
+		script_pushint(st, 0);
```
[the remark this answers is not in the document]
Atemo: A value need to be returned or the script must be stopped, `SCRIPT_CMD_FAILURE` does not stop the script.
I will go with `st->state = END;`

src/map/script.cpp
Atemo: We currently are using it like that :
https://github.com/rathena/rathena/blob/master/npc/instances/EndlessTower.txt#L213
Imo we could leave it like this, the command fails when there is no party and the command is always used as a condition. Also this would add a duplicate message in official instances.
```
	if (.@party < 1) {
		mes "Make or join a party with more than 1 member and try again.";
		close;
	}
	if (!instance_check_party(.@party_id,2)) {
		mes "Make or join a party with more than 1 member and try again.";
		close;
	}
```

doc/script_commands.txt
```diff
@@ -9843,6 +9843,7 @@ For details on the other parameters, see 'announce'.
 
 This function checks if a party meets certain requirements, returning 1 if all
 conditions are met and 0 otherwise. It will only check online characters.
+The command returns 0 is the party id does not exist.
```
aleos89, line 9846:
```suggestion
The command returns 0 is the party ID does not exist.
```

aleos89 (approved):

src/map/script.cpp
Lemongrass3110: Ok you got me convinced. ;-)

Lemongrass3110 (approved):

Issue #8339, opened 2024-05-18
Merchants do not earn money from Vending


gmragnarok:
### rAthena Hash

aa62b2fa80043db0be831b30ab71a9a521133445

### Client Date

2022-04-06

### Server Mode

Renewal

### Result

Merchants do not earn money from Vending

if Merchant have 2,147,483,647 Zeny (Max Zeny)

Merchants will do not earn money from Vending

![image](https://github.com/rathena/rathena/assets/163844770/a8ed1d6c-cb33-40c3-89e3-329b027c929d)

### Expected Result

In Official players cannot buy from merchants whose Zeny is full.

### How to Reproduce

1. create item Poring Egg on Merchant
2. Make zeny to maximum on merchant 2,147,483,647
3. vending the poring egg with price 1.000.000
4. using `@autotrade`
5. buy it using another account

### Official Information

In Official players cannot buy from merchants whose Zeny is full.

jamesandrewww:
vending_over_max: no

Daegaladh:
Actually in official servers you can't use Vending if the total amount exceeds max zeny

Pokye:
An (unofficial) suggestion
When enabling the **vending_over_max option: yes**
The exceeded zeny value should be sent via Rodex to be redeemed later.

gmragnarok:
> Actually in official servers you can't use Vending if the total amount exceeds max zeny

Thanks for the correction

Daegaladh:
Thanks to @secretdataz I've learned that in kRO (and probably others) it allows you to put the items on the shop, but other players cannot buy them if you'll exceed max zeny.
But some servers like twRO throw a pop-up message if the total amount of the zeny in the shop plus your current zeny exceeds max zeny, so I changed vending_over_max to emulate that behavior.

PR #8469, opened 2024-06-28
Fixed vending_over_max behavior to match official

Daegaladh:
* **Addressed Issue(s)**: #8339 

* **Server Mode**: Both

* **Description of Pull Request**: Changes the behavior of vending_over_max to allow or not setting up a vending shop when the character's zeny plus the total price of the items exceeds the maximum zeny allowed.
Also auto-closes the vending preparation window when an error is encountered.

### DIFF
--- a/src/map/vending.cpp
+++ b/src/map/vending.cpp
@@ -359,9 +359,8 @@ int8 vending_openvending( map_session_data& sd, const char* message, const uint8
 	if ( !battle_config.vending_over_max && (static_cast<int64>(sd.status.zeny) + total) > MAX_ZENY ) {
 #if PACKETVER >= 20200819
 		clif_msg_color( &sd, MSI_MERCHANTSHOP_TOTA_LOVER_ZENY_ERR, color_table[COLOR_RED] );
-#else
-		clif_skill_fail( sd, MC_VENDING );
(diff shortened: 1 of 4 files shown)


src/map/vending.cpp
```diff
@@ -344,17 +351,37 @@ int8 vending_openvending( map_session_data& sd, const char* message, const uint8
...
 		i++; // item successfully added
 	}
 
+	// check if the total value of the items plus the current zeny is over the limit
+	if ( !battle_config.vending_over_max && (static_cast<int64>(sd.status.zeny) + total) > MAX_ZENY ) {
+#if PACKETVER >= 20200819
+		clif_msg_color( &sd, MSI_MERCHANTSHOP_TOTA_LOVER_ZENY_ERR, color_table[COLOR_RED] );
+#else
+		clif_skill_fail( sd, MC_VENDING );
+#endif
```
Lemongrass3110, line 364:
```suggestion
#endif
		clif_skill_fail( sd, MC_VENDING );
```
I would keep on sending this, just like in all other cases.

Lemongrass3110 (approved):

aleos89 (approved):

PR #8479, opened 2024-07-01
Monster Losing Target Display

### LINKED ISSUE #8232: Monster moving weird when going to loot
### rAthena Hash

8cec8ce5605ac7933b40771cba51425ba64e3e43

### Client Date

20220602

### Server Mode

Renewal

### Result

https://github.com/rathena/rathena/assets/9493525/2c11637b-e805-4ec6-91e8-2332987c4804

Spawn 100 poring and kill it.

### Relevant Log Output

_No response_

### Expected Result

Monster smooth moving like before.

### How to Reproduce

Please see above video.

### Official Information

-

_No response_

Playtester:
* **Addressed Issue(s)**: #8232

* **Server Mode**: Both

- When a monster loses its target, it will now properly display movement to the next cell rather than snapping back to the previous cell
- Looters will no longer spam movement packets to the client when they go for loot
- Added some base implementation for sub-cell coordinates
- Fixes #8232

**Hints for Reviewers**:

Currently the sub-cell coordinates will always be 8 except for this one specific use case of a monster stopping to chase its target (the function unit_stop_walking_soon will reset it to 8 at the end as well). Still I decided to already add the reset to 8 in other parts of the code even though it's not yet needed, so the sub-cell coordinates can be used in other use-cases in the future without breaking the rest of the functionality.

This is more or less a quick solution to the problem and it works perfectly fine for this use-case. I'm aware a better solution would be to keep the sub-tile coordinates updated at all times, but this will not only take a lot of extra resources, it would also require a major refactoring of the code. That's why I decided it to fix it like this for now.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1566,7 +1566,7 @@ int mob_unlocktarget(struct mob_data *md, t_tick tick)
 		break;
 	default:
 		mob_stop_attack(md);
-		mob_stop_walking(md,1); //Stop chasing.
+		unit_stop_walking_soon(md->bl); //Stop chasing.
 		if (status_has_mode(&md->status,MD_ANGRY) && !md->state.aggressive)
 			md->state.aggressive = 1; //Restore angry state when switching to idle
 		md->state.skillstate = MSS_IDLE;
@@ -1908,6 +1908,8 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 			md->state.skillstate = MSS_LOOT;
 			if (!unit_walktobl(&md->bl, tbl, 0, 0))
 				mob_unlocktarget(md, tick); //Can't loot...
+			else
+				md->ud.target = tbl->id; //Remember current loot target
 			return true;
 		}
 		//Within looting range.
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1909,7 +1909,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 			if (!unit_walktobl(&md->bl, tbl, 0, 0))
 				mob_unlocktarget(md, tick); //Can't loot...
 			else
-				md->ud.target = tbl->id; //Remember current loot target
+				unit_set_target(&md->ud, tbl->id); //Remember current loot target
 			return true;
 		}
 		//Within looting range.
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -1431,12 +1431,12 @@ void unit_stop_walking_soon(struct block_list& bl)
 		return;
 
 	// Get how much percent we traversed on the timer
-	double cell_percent = 1 - ((double)DIFF_TICK(td->tick, gettick()) / (double)td->data);
+	double cell_percent = 1.0 - ((double)DIFF_TICK(td->tick, gettick()) / (double)td->data);
 
 	short ox = bl.x, oy = bl.y; // Remember original x and y coordinates
 	short path_remain = 1; // Remaining path to walk
 
-	if (cell_percent > 0 && cell_percent < 1) {
+	if (cell_percent > 0.0 && cell_percent < 1.0) {
 		// Set subcell coordinates according to timer
 		// This gives a value between 8 and 39
 		ud->sx = static_cast<decltype(ud->sx)>(24.0 + dirx[ud->walkpath.path[ud->walkpath.path_pos]] * 16.0 * cell_percent);
@@ -1454,7 +1454,7 @@ void unit_stop_walking_soon(struct block_list& bl)
 		ud->sx %= 16;
 		ud->sy %= 16;
 	}
-	else if (cell_percent >= 1) {
+	else if (cell_percent >= 1.0) {
 		// Assume exactly one cell moved
 		bl.x += dirx[ud->walkpath.path[ud->walkpath.path_pos]];
 		bl.y += diry[ud->walkpath.path[ud->walkpath.path_pos]];
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -1424,59 +1424,59 @@ void unit_stop_walking_soon(struct block_list& bl)
 
 	if (ud->walkpath.path_pos + 1 >= ud->walkpath.path_len)
 		return;
-		const struct TimerData* td = get_timer(ud->walktimer);
 
-		if (td == nullptr)
-			return;
+	const struct TimerData* td = get_timer(ud->walktimer);
 
-		// Get how much percent we traversed on the timer
-		double cell_percent = 1 - ((double)DIFF_TICK(td->tick, gettick()) / (double)td->data);
+	if (td == nullptr)
+		return;
 
-		short ox = bl.x, oy = bl.y; // Remember original x and y coordinates
-		short path_remain = 1; // Remaining path to walk
+	// Get how much percent we traversed on the timer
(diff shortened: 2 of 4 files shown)


src/map/unit.hpp
```diff
@@ -27,6 +27,7 @@ struct unit_data {
 	struct skill_unit_group_tickset skillunittick[MAX_SKILLUNITGROUPTICKSET];
 	short attacktarget_lv;
 	short to_x, to_y;
+	unsigned char sx, sy; // Subtile position (0-15, with 8 being center of cell)
```
Lemongrass3110, line 30: You want to store a number and not a character.

src/map/unit.cpp
```diff
@@ -1404,6 +1407,72 @@ int unit_warp(struct block_list *bl,short m,short x,short y,clr_type type)
...
+		}
+		else if (cell_percent >= 1) {
+			// Assume exactly one cell moved
+			bl.x = bl.x + dirx[ud->walkpath.path[ud->walkpath.path_pos]];
+			bl.y = bl.y + diry[ud->walkpath.path[ud->walkpath.path_pos]];
+			path_remain = 2;
+		}
+		// Shorten walkpath
+		if (ud->walkpath.path_pos + path_remain < ud->walkpath.path_len) {
+			ud->walkpath.path_len = ud->walkpath.path_pos + path_remain;
```
Lemongrass3110, line 1458: Here you also need to reduce the walktimer, so that it is triggered earlier.

src/map/unit.cpp
```diff
@@ -1404,6 +1407,72 @@ int unit_warp(struct block_list *bl,short m,short x,short y,clr_type type)
...
+			// Set subcell coordinates according to timer
+			// This gives a value between 8 and 39
+			ud->sx = (unsigned char)(24.0 + dirx[ud->walkpath.path[ud->walkpath.path_pos]] * 16.0 * cell_percent);
+			ud->sy = (unsigned char)(24.0 + diry[ud->walkpath.path[ud->walkpath.path_pos]] * 16.0 * cell_percent);
+			// 16-31 reflect sub position 0-15 on the current cell
+			// 8-15 reflect sub position 8-15 at -1 main coordinate
+			// 32-39 reflect sub position 0-7 at +1 main coordinate
+			if (ud->sx < 16 || ud->sy < 16 || ud->sx > 31 || ud->sy > 31) {
+				path_remain = 2;
+				if (ud->sx < 16) bl.x--;
```
Lemongrass3110, line 1442:
```suggestion
				if (ud->sx < 16) bl.x -= dirx[ud->walkpath.path[ud->walkpath.path_pos-1]];
```

src/map/unit.cpp
```diff
@@ -1404,6 +1407,72 @@ int unit_warp(struct block_list *bl,short m,short x,short y,clr_type type)
...
+				if (ud->sy < 16) bl.y--;
```
Lemongrass3110, line 1443:
```suggestion
				if (ud->sy < 16) bl.y -= diry[ud->walkpath.path[ud->walkpath.path_pos-1]];
```

src/map/unit.cpp
```diff
@@ -1404,6 +1407,72 @@ int unit_warp(struct block_list *bl,short m,short x,short y,clr_type type)
...
+				if (ud->sx > 31) bl.x++;
```
Lemongrass3110, line 1444:
```suggestion
				if (ud->sx > 31) bl.x += dirx[ud->walkpath.path[ud->walkpath.path_pos+1]];
```

src/map/unit.cpp
```diff
@@ -1404,6 +1407,72 @@ int unit_warp(struct block_list *bl,short m,short x,short y,clr_type type)
...
+				if (ud->sy > 31) bl.y++;
```
Lemongrass3110, line 1445:
```suggestion
				if (ud->sy > 31) bl.y += diry[ud->walkpath.path[ud->walkpath.path_pos+1]];
```

src/map/unit.cpp
```diff
@@ -1404,6 +1407,72 @@ int unit_warp(struct block_list *bl,short m,short x,short y,clr_type type)
...
+		short ox = bl.x, oy = bl.y; // Remember original x and y coordinates
+		short path_remain = 1; // Remaining path to walk
+
+		if (cell_percent > 0 && cell_percent < 1) {
+			// Set subcell coordinates according to timer
+			// This gives a value between 8 and 39
+			ud->sx = (unsigned char)(24.0 + dirx[ud->walkpath.path[ud->walkpath.path_pos]] * 16.0 * cell_percent);
+			ud->sy = (unsigned char)(24.0 + diry[ud->walkpath.path[ud->walkpath.path_pos]] * 16.0 * cell_percent);
+			// 16-31 reflect sub position 0-15 on the current cell
+			// 8-15 reflect sub position 8-15 at -1 main coordinate
```
Lemongrass3110, line 1438:
```suggestion

			// 8-15 reflect sub position 8-15 at -1 main coordinate
			// 16-31 reflect sub position 0-15 on the current cell
```

Lemongrass3110 (commented):
Some initial comments.
Lets discuss the rest in Discord.

src/map/mob.cpp
```diff
@@ -1908,6 +1908,8 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 			md->state.skillstate = MSS_LOOT;
 			if (!unit_walktobl(&md->bl, tbl, 0, 0))
 				mob_unlocktarget(md, tick); //Can't loot...
+			else
+				md->ud.target = tbl->id; //Remember current loot target
```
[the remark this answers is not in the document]
Playtester: I want the change to influence as little as possible, so I only put it here, as it's only needed when unit_walktobl was successful. The only time we don't want to call unit_walktobl again is when the looter is already going for the loot.

I'll check the function. I think unit_set_target is for attack targets.

src/map/unit.hpp
Playtester: Char is a number from 0 to 255.

But I mainly use this because that's what clif.cpp expects to be put into the packet.

src/map/unit.cpp
Playtester: Nothing changes with the timer. The monster just continues moving server-sided.

src/map/unit.cpp
Playtester: See above.

src/map/unit.cpp
Playtester: See above.

src/map/unit.cpp
Playtester: I put 16-31 first because the current cell is the most "normal" situation. 8-15 and 32-39 belong together as in the unit already moved to the next cell.

src/map/mob.cpp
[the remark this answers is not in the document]
Playtester: Yeah I checked, we shouldn't use unit_set_target because the loot doesn't need to remember which monsters go after it. That function's purpose is to keep track of the number of attackers currently attacking you.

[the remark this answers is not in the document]
Playtester: Okay after debugging, I can use unit_set_target.

It will do nothing because: `&& (target = map_id2bl(target_id))` just returns false (nullptr is interpreted as false). So both variants would do the same, but I guess it might be cleaner to only have one place setting ud.target directly.

src/map/unit.hpp
Lemongrass3110: Fixed that for you.

aleos89 (approved):

PR #8498, opened 2024-07-09
Assassin quest fixes and cleanup

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

-Fixes killmonsterall killing monsters from other part of the quest, causing other players who were doing that part to get stuck
-Follow-up to 979e13dd66103e3b6387e9acfa48642a57b66a26 preventing players with level under 49 to get the reward
-New header style
-Cleaned up set commands

### DIFF
--- a/npc/jobs/2-1/assassin.txt
+++ b/npc/jobs/2-1/assassin.txt
@@ -92,7 +92,6 @@ in_moc_16,19,33,1	script	Guildsman#asn	55,{
 		close2;
 		ASSIN_Q = 0;
 		ASSIN_Q2 = 0;
-		ASSIN_Q3 = 0;
 		savepoint "in_moc_16",18,14,1,1;
 		warp "in_moc_16",18,14;
 		end;
@@ -547,7 +546,8 @@ OnTouch_:
 				mes "Alright then,";
 				mes "best of luck to you!";
 				close2;
-				ASSIN_Q3 = 1;
+				if (!ASSIN_Q3)
+					ASSIN_Q3 = 1;
 				ASSIN_Q = 1;
 				changequest 8001,8002;
 				warp "in_moc_16",19,144;
@@ -570,7 +570,8 @@ OnTouch_:
 				mes "to the Test Hall.";
 				mes "Best of luck~";
 				close2;
-				ASSIN_Q3 = 2;
+				if (!ASSIN_Q3)
+					ASSIN_Q3 = 2;
 				ASSIN_Q = 1;
 				changequest 8001,8002;
 				warp "in_moc_16",19,144;
@@ -597,7 +598,6 @@ OnTouch_:
 				mes "Stop wastin' my time...";
 				close2;
 				ASSIN_Q = 0;
-				ASSIN_Q3 = 0;
 				ASSIN_Q2 = 0;
 				erasequest 8001;
 				warp "moc_fild16",206,229;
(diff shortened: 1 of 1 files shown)


npc/jobs/2-1/assassin.txt
```diff
@@ -2064,7 +2057,8 @@ OnTouch_:
 			mes "I believe you stand out amongst them...";
 			next;
 		}
-		if (ASSIN_Q3 == 1 && BaseJob == Job_Thief && ASSIN_Q > 6 && ASSIN_Q < 17) {
```
aleos89, line 2067: This is a giant OnTouch event, but is the removal of the Job check ok here? Also, the removal of the `ASSIN_Q` part doesn't seem to be anywhere below as well, is that ok?

Daegaladh: Yeah, both are already checked right at the start in line 1644
`if (ASSIN_Q == 7 && BaseJob == Job_Thief) {`
And there's no way ASSIN_Q gets a value out of those limits during all the code, even if the player relogs or times out, the value gets reseted to 7, so that check was just redundant

npc/jobs/2-1/assassin.txt
```diff
@@ -553,8 +546,9 @@ OnTouch_:
 				mes "Alright then,";
 				mes "best of luck to you!";
 				close2;
-				set ASSIN_Q3,1;
-				set ASSIN_Q,1;
+				if (!ASSIN_Q3)
+					ASSIN_Q3 = 1;
```
Atemo, line 550: For me this change is incorrect, ASSIN_Q3 should be assigned regardless of the previous value (same for other change)

npc/jobs/2-1/assassin.txt
```diff
@@ -96,9 +90,8 @@ in_moc_16,19,33,1	script	Guildsman#asn	55,{
 		mes "[Ferocious-looking guy]";
 		mes "You'll have to walk if you want to get back to town. Oh, and don't forget to save your spawn point, alright?";
 		close2;
-		set ASSIN_Q,0;
-		set ASSIN_Q2,0;
-		set ASSIN_Q3,0;
+		ASSIN_Q = 0;
+		ASSIN_Q2 = 0;
```
Atemo, line 94: For me this change is incorrect, ASSIN_Q3 should be 0 (same for other change)

npc/jobs/2-1/assassin.txt
```diff
@@ -2064,7 +2057,8 @@ OnTouch_:
 			mes "I believe you stand out amongst them...";
 			next;
 		}
-		if (ASSIN_Q3 == 1 && BaseJob == Job_Thief && ASSIN_Q > 6 && ASSIN_Q < 17) {
+		switch(ASSIN_Q3) {
+		case 1:
```
Atemo, line 2061:
```suggestion
		case 1:	// player Job Level was above 48
```

npc/jobs/2-1/assassin.txt
```diff
@@ -2098,10 +2092,10 @@ OnTouch_:
 				mes "It used to rule over the Assassin weapon market. Please take care of my gladius.";
 				getitem 1220,1; //Gladius_
 			}
-			ASSIN_Q3 = 2;
+			ASSIN_Q3 = 0;
 			next;
-		}
-		else if (ASSIN_Q3 != 2) {
+			break;
+		case 2:
```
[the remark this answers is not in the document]
Atemo:
```suggestion
		case 2:	// player Job Level was below 48
```

Atemo (approved):

PR #8501, opened 2024-07-12
Resolve Death Valley HP/SP adjustment

### LINKED ISSUE #8428: Incorrect HP & SP for WM_DEADHILLHERE
### rAthena Hash

673ef066fb261985b7be4526b444e8b735e7ed95

### Client Date

2022-04-06

### Server Mode

Renewal

### Result

- incorrect HP restoration
- incorrect SP depletion

https://github.com/rathena/rathena/assets/143317511/07350608-b66d-4116-8df4-d71246c17d01

### Relevant Log Output

_No response_

### Expected Result

A dead target is resurrected based on its SP, and the target's SP is depleted by (60 - 10 * skillLv) after being resurrected.
Lv1 50%
Lv2 40%
Lv3 30%
Lv4 20%
Lv5 10%

### How to Reproduce

1. Use #die target
2. Use WM_DEADHILLHERE Level 1 to ress a target
3. Repeat step 1 and 2 constantly

### Official Information

https://irowiki.org/wiki/Death_Valley
https://ro.gnjoy.com/guide/runemidgarts/skillview.asp?lineseq=5&skillid=2421

_No response_

aleos89:
* **Addressed Issue(s)**: Fixes #8428

* **Server Mode**: Renewal

  * The skill will now properly take the SP first before healing HP of the target player.
Thanks to @jamesandrewww!

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -11708,12 +11708,13 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			if( !status_isdead(bl) )
 				break;
 
+			status_zap(bl, 0, tstatus->sp * (60 - 10 * skill_lv) / 100);
+
 			int heal = tstatus->sp;
 
 			if( heal <= 0 )
 				heal = 1;
 			tstatus->hp = heal;
-			tstatus->sp -= tstatus->sp * ( 60 - 10 * skill_lv ) / 100;
 			clif_skill_nodamage(src,bl,skill_id,skill_lv,1);
 			pc_revive((TBL_PC*)bl,heal,0);
 			clif_resurrection( *bl );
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -11708,12 +11708,13 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			if( !status_isdead(bl) )
 				break;
 
+			status_zap(bl, 0, tstatus->sp * (60 - 10 * skill_lv) / 100);
```
Lemongrass3110, line 11711:
```suggestion
			status_zap(bl, 0, tstatus->sp * 10 * skill_lv / 100);
```
Its a flat 10% per skill level.
But I think in general you have to store this calculated amount, as it is used as value for the heal.
Now you reduce it first and therefore it would result in a smaller amount of HP recovered.
Or is this intentional?

Lemongrass3110 (approved):

PR #8502, opened 2024-07-13
Correct Death Valley behavior

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Renewal

  * Follow up to 4b32621.
  * The target player should be healed from the remaining SP before the SP reduction is applied.
  * Adjust the variable type when storing the target's SP.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -11708,7 +11708,7 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			if( !status_isdead(bl) )
 				break;
 
-			tstatus->hp += tstatus->sp;
+			tstatus->hp += max(tstatus->sp, 1);
 			tstatus->sp = tstatus->sp * 10 * skill_lv / 100;
 			clif_skill_nodamage(src,bl,skill_id,skill_lv,1);
 			pc_revive((TBL_PC*)bl,true,true);
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -11708,10 +11708,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 			if( !status_isdead(bl) )
 				break;
 
-			tstatus->hp += max(tstatus->sp, 1);
+			tstatus->hp = max(tstatus->sp, 1);
 			tstatus->sp = tstatus->sp * 10 * skill_lv / 100;
 			clif_skill_nodamage(src,bl,skill_id,skill_lv,1);
-			pc_revive((TBL_PC*)bl,true,true);
+			pc_revive(reinterpret_cast<map_session_data*>(bl),true,true);
 			clif_resurrection( *bl );
 		}
 		break;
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -11708,15 +11708,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 				break;
 
-			status_zap(bl, 0, tstatus->sp * 10 * skill_lv / 100);
-
-			int heal = tstatus->sp;
-
-			if( heal <= 0 )
-				heal = 1;
-			tstatus->hp = heal;
+			tstatus->hp += max(tstatus->sp, 1);
```
Lemongrass3110, line 11711:
```suggestion
			tstatus->hp = max(tstatus->sp, 1);
```

src/map/skill.cpp
```diff
@@ -11708,15 +11708,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+			tstatus->sp = tstatus->sp * 10 * skill_lv / 100;
 			clif_skill_nodamage(src,bl,skill_id,skill_lv,1);
-			pc_revive((TBL_PC*)bl,heal,0);
+			pc_revive((TBL_PC*)bl,true,true);
```
Lemongrass3110, line 11714:
```suggestion
			pc_revive(reinterpret_cast<map_session_data*>(bl),true,true);
```
Not sure about the "bool" for "unsigned int" usage. But I understand why you did it.

Lemongrass3110 (approved):

src/map/skill.cpp
aleos89: Yeah, I figured I could come back to this with a refactor to clean those parameters up!

src/map/skill.cpp
```diff
@@ -11708,15 +11708,10 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
 
-			status_zap(bl, 0, tstatus->sp * 10 * skill_lv / 100);
-
-			int heal = tstatus->sp;
-
-			if( heal <= 0 )
-				heal = 1;
-			tstatus->hp = heal;
+			tstatus->hp = max(tstatus->sp, 1);
+			tstatus->sp = tstatus->sp * 10 * skill_lv / 100;
```
Atemo, line 11712: I just saw that the formula is incorrect since https://github.com/rathena/rathena/commit/4b32621c61bfce397488e27b4eb962b7c04a9c6f
https://www.divine-pride.net/database/skill/2421
```
[Level 1] : 성공율 100% / 부활 대상의 SP 50% 소모
[Level 2] : 성공율 100% / 부활 대상의 SP 40% 소모
[Level 3] : 성공율 100% / 부활 대상의 SP 30% 소모
[Level 4] : 성공율 100% / 부활 대상의 SP 20% 소모
[Level 5] : 성공율 100% / 부활 대상의 SP 10% 소모 
```
It should consume 50% at level 1, 40% at level 2.. with the current formula the sp are 10% of the previous sp (90% sp consumed) at level 1 etc..

Lemongrass3110: My fault... :(

PR #8507, opened 2024-07-20
add more minimal docker compose setup

dweipert-3138720606:
The current docker compose setup talks about only being usable for development.
This is my attempt to make a minimal production ready docker compose setup where you can just run "docker compose up -d" in the root directory and be done

Issues until now:
- The database reports healthy before all sql imports are done. The other containers can't connect in the meantime and fail to start up. On subsequent runs that's not a problem though.
- the BUILDER_CONFIGURE variable needs to be set for every service (I think). ideally it just has to be set once in an .env file maybe


.editorconfig
```diff
@@ -21,6 +21,11 @@ indent_style = tab
 indent_style = space
 indent_size = 4
 
+# YAML does not support hard tabs.
+[docker-compose*.{yml,yaml}]
+indent_style = space
+indent_size = 2
```
aleos89, line 27: Might be best to just remove this addition and change the section above to 2 instead of 4 since that is the official tab value for YAML.

secretdataz (approved):

dweipert-3138720606:
Just rebased the pull request branch with current master. Don't know if that was necessary, but everything seems to be fine.

tools/docker/development/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.11
+WORKDIR /rathena
+RUN apk add --no-cache wget git cmake make gcc g++ gdb zlib-dev mariadb-dev ca-certificates linux-headers bash valgrind netcat-openbsd
+RUN wget https://raw.githubusercontent.com/eficode/wait-for/v2.2.2/wait-for -O /bin/wait-for && chmod +x /bin/wait-for
```
lighta, line 4: I'm not fan of this, better download and add the script as an asset.
Like this it look quite unsafe. (yes it's probably only meant for dev but still, why do we need to download this ?)

dweipert-3138720606: me neither. that's the part I factored out from the old docker compose setup. Everything under `docker/development` is. The things I propose to add are everything but those files.

Issue #7605, opened 2023-02-12
Confirm Homunculus skill cooldowns


Lemongrass3110:
### rAthena Hash

9d2bcc7

### Client Date

2023-01-05

### Server Mode

Pre-Renewal, Renewal

### Result

During review of https://github.com/rathena/rathena/pull/7586 it was reported that the Homunculus should use Cooldown rather than AfterCastActDelay.

### Expected Result

Unknown

### How to Reproduce

Unknown

### Official Information

I do not have any sources.
Feel free to comment some.

Daegaladh:
From my tests o officials I can confirm on Renewal they have Cooldown.
Neither the effect or the cooldown goes away on changing maps, teleport or vaporize, but the effect goes away on relog (the cooldown doesn't)

On Pre-re on the other hand, it was AfterCastActDelay, and both the effect and the delay go away on changing maps, teleport, vaporize and relog.

PR #8508, opened 2024-07-20
Fixed homunculus cooldown and aftercast delays

Daegaladh:
* **Addressed Issue(s)**: #7605

* **Server Mode**: Both

* **Description of Pull Request**: Fixes homunculus skills cooldown and aftercast delays for both pre-renewal and renewal with the exception of Homunculus S because the lack of data.
Splits the delays reset when warp and vaporize on separate configs because officially works different on each mode.
Adds missing skill fail messages for both homunculus and mercenary skills
Adds missing cooldown timer displays for both homunculus and mercenary skills

Official info:
For pre-re the aftercast delay of homunculus goes away on teleport, changing maps, vaporize and relog, along with the status effects on the homunculus.
On renewal, the cooldown doesn't go away no matter what.
On both pre-renewal and renewal the skill icon delay/cooldown clock display is shown, also both show skill fail message along with the (...) emote when you try to reuse them too soon.
All the AfterCastActDelay and Cooldown values are from official data and match what you can find in Divine Pride and Kokotewa's databases.
AfterCastWalkDelay values I calculated them testing each skill on official and using slowmo replay to get the exact timing.

TODO: Refactor the cooldown system to save homunculus and mercenary cooldowns in the skillcooldown sql database, since on officials they're saved after relog

### DIFF
--- a/db/pre-re/skill_db.yml
+++ b/db/pre-re/skill_db.yml
@@ -30729,8 +30729,8 @@ Body:
     Range: 9
     Hit: Single
     HitCount: 1
-    AfterCastWalkDelay: 1000
     AfterCastActDelay: 2000
+    AfterCastWalkDelay: 1000
     Requires:
       SpCost:
         - Level: 1
--- a/db/pre-re/skill_db.yml
+++ b/db/pre-re/skill_db.yml
@@ -30729,6 +30729,7 @@ Body:
     Range: 9
     Hit: Single
     HitCount: 1
+    AfterCastWalkDelay: 1000
     AfterCastActDelay: 2000
     Requires:
       SpCost:
@@ -30757,6 +30758,7 @@ Body:
     HitCount: 1
     SplashArea: -1
     AfterCastActDelay: 35000
+    AfterCastWalkDelay: 1000
     Duration1:
       - Level: 1
         Time: 40000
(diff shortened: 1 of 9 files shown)


db/pre-re/skill_db.yml
```diff
@@ -30729,6 +30729,7 @@ Body:
     Range: 9
     Hit: Single
     HitCount: 1
+    AfterCastWalkDelay: 1000
     AfterCastActDelay: 2000
```
Lemongrass3110, line 30733:
```suggestion
    AfterCastActDelay: 2000
    AfterCastWalkDelay: 1000
```

src/map/skill.cpp
```diff
@@ -10675,11 +10695,17 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		break;
 
 	case HAMI_CASTLE:	//[orn]
-		if (src != bl && rnd()%100 < 20 * skill_lv) {
+		if (src != bl && rnd() % 100 < 20 * skill_lv) {
```
Lemongrass3110, line 10698:
```suggestion
		if (src != bl && rnd_chance(20 * skill_lv, 100)) {
```

Lemongrass3110 (commented):
Just some general comments. Cant really say much about the content of the changes.

aleos89 (approved):

Issue #8525, opened 2024-07-29
Homunculus cooldown display not clearing


Daegaladh:
### rAthena Hash

48b105405a4d15e247947118d85d24fd0b2fd0fd

### Client Date

2018-06-21

### Server Mode

Pre-Renewal, Renewal

### Result

Homunculus cooldown timer display doesn't clear when the cooldown is removed.

On a 2022 client it gets even worse, since the client doesn't allow you to reuse the skill while the display timer is still running, even if the cooldown has finished server-side.

### Expected Result

Cooldown timer display should be cleared when the cooldown is removed on vaporize, map change and teleport.

### How to Reproduce

On pre-re is easier to reproduce, just use an homunculus skill with cooldown, and teleport, change map or vaporize/call. The timer display won't reset.

### Official Information

On official renewal servers, the server resends clif_homskillinfoblock before clearing the cooldowns, then sends clif_skill_cooldown with the remaining time. On vaporize this gets bugged if you vaporize and call the homunculus 2+ times when the cooldown is still running.

### Modifications that may affect results

None

Daegaladh:
I've tried to fix this sending `clif_homskillinfoblock(*sd->hd)` then `clif_skill_cooldown(*sd, skill, 0)` as officials do, and it fixes the problem for changing maps, but not for teleport.
For vaporize happens the same bug as officials, trying to vaporize-call the homunculus more than once somehow makes the timer display to return.

I have no idea how to fix this, since the client for some reason retains the timers after clearing on `clif_homskillinfoblock`, and shows them again with a second `clif_homskillinfoblock`.

Also, for the renewal vaporize, we need to fix #8518 first, since we need to send the remaining time, and currently is not being saved.

PR #8526, opened 2024-07-29
Fixed cooldown displays not being cleared

Daegaladh:
* **Addressed Issue(s)**: #8525

* **Server Mode**: Both

* **Description of Pull Request**: Fixes the cooldown timer displays not being cleared after the cooldown is cancelled.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -1926,6 +1926,17 @@ void clif_homskillinfoblock( homun_data& hd ){
 	}
 
 	clif_send( packet, packet->packetLength, &sd->bl, SELF );
+
+	if (battle_config.display_status_timers) {
+		// Clear cooldown display for skills that are no longer blocked
+		for (auto& it : hd.homunculus.hskill) {
+			if (it.id == 0)
+				continue;
+
+			if (util::vector_get(hd.blockskill, it.id) == hd.blockskill.end())
+				clif_skill_cooldown(*sd, it.id, 0);
+		}
+	}
 #endif
 }
 
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -23039,13 +23039,17 @@ int skill_blockpc_clear(map_session_data *sd) {
 }
 
 TIMER_FUNC(skill_blockhomun_end){
-	struct homun_data *hd = (TBL_HOM*) map_id2bl(id);
+	homun_data *hd = reinterpret_cast<homun_data *>(map_id2bl(id));
 
 	if (hd) {
-		auto skill = util::vector_get(hd->blockskill, (uint16)data);
+		auto skill = util::vector_get(hd->blockskill, static_cast<uint16>(data));
 
 		if (skill != hd->blockskill.end())
 			hd->blockskill.erase(skill);
+
+		// Make sure the cooldown display is removed
+		if (battle_config.display_status_timers)
+			clif_homskillinfoblock(*hd);
 	}
 
 	return 1;
@@ -23067,8 +23071,11 @@ int skill_blockhomun_start(struct homun_data *hd, uint16 skill_id, int tick)	//[
 
 	hd->blockskill.push_back(skill_id);
 
-	if (battle_config.display_status_timers)
+	if (battle_config.display_status_timers) {
+		// Reset the skill cooldown display first
+		clif_homskillinfoblock(*hd);
 		clif_skill_cooldown(*hd->master, skill_id, tick);
+	}
 
 	return add_timer(gettick() + tick, skill_blockhomun_end, hd->bl.id, skill_id);
 }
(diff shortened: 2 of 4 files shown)


Daegaladh:
This seems to work fine on client 20180620, but on 20220406 even if the timer display is cleared, the client retains the old data and doesn't allow you to cast the skill again.
No idea how to fix that, maybe we're missing a packet?

src/map/skill.cpp
```diff
@@ -23039,13 +23039,17 @@ int skill_blockpc_clear(map_session_data *sd) {
 }
 
 TIMER_FUNC(skill_blockhomun_end){
-	struct homun_data *hd = (TBL_HOM*) map_id2bl(id);
+	homun_data *hd = reinterpret_cast<homun_data *>(map_id2bl(id));
```
Lemongrass3110, line 23042:
```suggestion
	homun_data* hd = map_id2hd(id);
```

src/map/skill.cpp
```diff
@@ -23039,13 +23039,17 @@ int skill_blockpc_clear(map_session_data *sd) {
...
 
 	if (hd) {
-		auto skill = util::vector_get(hd->blockskill, (uint16)data);
+		auto skill = util::vector_get(hd->blockskill, static_cast<uint16>(data));
 
 		if (skill != hd->blockskill.end())
 			hd->blockskill.erase(skill);
```
Lemongrass3110, line 23048:
```suggestion
		util::vector_erase_if_exists( hd->blockskill, static_cast<uint16>( data ) );
```

src/map/clif.cpp
```diff
@@ -1926,6 +1926,17 @@ void clif_homskillinfoblock( homun_data& hd ){
...
 
 	clif_send( packet, packet->packetLength, &sd->bl, SELF );
+
+	if (battle_config.display_status_timers) {
+		// Clear cooldown display for skills that are no longer blocked
+		for (auto& it : hd.homunculus.hskill) {
+			if (it.id == 0)
+				continue;
+
+			if (util::vector_get(hd.blockskill, it.id) == hd.blockskill.end())
```
Lemongrass3110, line 1936:
```suggestion
			if (util::vector_exists(hd.blockskill, it.id))
```

AoShinRO:
> This seems to work fine on client 20180620, but on 20220406 even if the timer display is cleared, the client retains the old data and doesn't allow you to cast the skill again. No idea how to fix that, maybe we're missing a packet?

for homunculus only or every cooldown?

src/map/clif.cpp
```diff
@@ -1926,6 +1926,17 @@ void clif_homskillinfoblock( homun_data& hd ){
 	}
 
 	clif_send( packet, packet->packetLength, &sd->bl, SELF );
+
+	if (battle_config.display_status_timers) {
```
AoShinRO, line 1930: try change this to upper loop, and clif 0 tick only if skill is not in the vector
```
	for(int i = 0, count = 0; i < MAX_HOMUNSKILL; i++) {
		(...)
			if (battle_config.display_status_timers) {
				if (!util::vector_exists(hd.blockskill, id))
					clif_skill_cooldown(*sd, id, 0);
			}
		}
```

Daegaladh: That wouldn't work, sending ZC_HOSKILLINFO_LIST makes the client refresh the timers. In fact, sending it twice somehow makes the timers revert to what the client has internally.

AoShinRO:
you forget to add_func to list at  skill_init
add_timer_func_list(skill_blockhom_end, "skill_blockhom_end");
if this not  added when map server crashes with any cooldown time up this warning a missmatch tid function
image

Daegaladh:
This was just a proof of concept, but I think you're missing the point of this draft, which is tricking the client into end the timers early. And unfortunately doesn't seem to work, as the client keeps track of the timer by itself.
I haven't tested your PR yet, but doesn't seem to have anything tondo with what I'm trying to achieve here

src/map/clif.cpp
```diff
@@ -1926,6 +1926,17 @@ void clif_homskillinfoblock( homun_data& hd ){
...
+		// Clear cooldown display for skills that are no longer blocked
+		for (auto& it : hd.homunculus.hskill) {
+			if (it.id == 0)
+				continue;
+
+			if (util::vector_get(hd.blockskill, it.id) == hd.blockskill.end())
+				clif_skill_cooldown(*sd, it.id, 0);
```
AoShinRO, line 1937:
```suggestion
				int iter = skill_blockpc_get(it.id);
				if(iter>=0)
					delete_timer(hd->master->scd[iter]->timer, skill_blockpc_end);
				clif_skill_cooldown(*sd, it.id, 0);
```

maybe clear pc  timer helps

Daegaladh: But the issue is client-side

AoShinRO:
> This was just a proof of concept, but I think you're missing the point of this draft, which is tricking the client into end the timers early. And unfortunately doesn't seem to work, as the client keeps track of the timer by itself. I haven't tested your PR yet, but doesn't seem to have anything to do with what I'm trying to achieve here
> 
> Actually, I don't think this can be fixed server-side only. A client patch might be needed in order to force it to set the timer to whatever we're sending (0 in this case).

the pool request i did earlier is not related, plz forget that  haha

Issue #8529, opened 2024-07-31
Exp bar not working correctly on 2015 clients


Singe-Horizontal:
### rAthena Hash

48b105405a

### Client Date

2015-11-04

### Server Mode

Pre-Renewal

### Result

The exp bar isn't updating itself or show broken values when minimized (ex 115%)

### Expected Result

Exp bar should update consistently

### How to Reproduce

Gain experience from any source

### Official Information

-

PR #8530, opened 2024-07-31
Corrects exp packet header

Singe-Horizontal:
* **Addressed Issue(s)**: 
#8529

* **Server Mode**: 
Both

This changes the function to build a correct packet header for <2017 clients

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -3670,16 +3670,16 @@ void clif_updatestatus( map_session_data& sd, enum _sp type ){
 			break;
 #else
 		case SP_BASEEXP:
-			clif_par_change(sd, type, client_exp(sd.status.base_exp));
+			clif_longpar_change(sd, type, client_exp(sd.status.base_exp));
 			break;
 		case SP_JOBEXP:
-			clif_par_change(sd, type, client_exp(sd.status.job_exp));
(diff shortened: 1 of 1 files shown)


src/map/clif.cpp
```diff
@@ -3670,16 +3670,16 @@ void clif_updatestatus( map_session_data& sd, enum _sp type ){
 			break;
 #else
 		case SP_BASEEXP:
-			clif_par_change(sd, type, client_exp(sd.status.base_exp));
+			clif_longpar_change(sd, type, client_exp(sd.status.base_exp));
```
aleos89, line 3673: Not sure if anyone has more packet data info, but shouldn't this be adjusted to whenever the client changed from `clif_par_change` to `clif_longpar_change` and not thrown into the final else?

Daegaladh: > Not sure if anyone has more packet data info, but shouldn't this be adjusted to whenever the client changed from `clif_par_change` to `clif_longpar_change` and not thrown into the final else?

ZC_PAR_CHANGE and ZC_LONGPAR_CHANGE have been there since the alpha clients, so they're not date dependent, but depends on the stuff they send.
clif_longpar_change seems correct for exp values

Daegaladh (approved):

aleos89 (approved):

PR #8553, opened 2024-08-13
Updated/corrected EFST datas

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

Update and corrections of EFST names.


src/map/script_constants.hpp
```diff
@@ -3203,7 +3203,7 @@
 	export_constant(EFST_SPELLFIST);
 	export_constant(EFST_NETHERWORLD);
 	export_constant(EFST_SIREN);
-	export_constant(EFST_HANDICAPSTATE_DEEP_SLEEP);
```
aleos89, line 3206: Should we export these as deprecated to still have "support" but also print out the deprecation message?

Issue #8512, opened 2024-07-22
Map-server crashes from received unsupported packet a large amount session


hnomkeng:
### rAthena Hash

0a977c1

### Client Date

2021-11-16

### Server Mode

Pre-Renewal

### Result

#0  0x00000000007074f1 in connect_check_clear (tid=<optimized out>, tick=6902725, id=<optimized out>, data=<optimized out>) at socket.cpp:1622
1622                            if( (!hist->ddos && DIFF_TICK(tick,hist->tick) > ddos_interval*3) ||
Missing separate debuginfos, use: yum debuginfo-install MariaDB-shared-10.4.34-1.el8.x86_64 glibc-2.28-251.el8_10.2.x86_64 libgcc-8.5.0-22.el8_10.x86_64 libstdc++-8.5.0-22.el8_10.x86_64 openssl-libs-1.1.1k-12.el8_9.x86_64 zlib-1.2.11-26.el8.x86_64
(gdb) bt full
#0  0x00000000007074f1 in connect_check_clear (tid=<optimized out>, tick=6902725, id=<optimized out>, data=<optimized out>) at socket.cpp:1622
        i = <optimized out>
        clear = 0
        list = 0
        root = {next = 0xfe6, ip = 7376052, tick = 1295, count = 7, ddos = 0}
        prev_hist = 0x7fff4328b2f0
        hist = 0xfe6
        __FUNCTION__ = "connect_check_clear"
#1  0x000000000070b336 in do_timer (tick=<optimized out>) at timer.cpp:372
        tid = 0
        diff = -1
        __FUNCTION__ = "do_timer"
#2  0x0000000000424653 in main (argc=1, argv=0x7fff4328b498) at core.cpp:380
        next = <optimized out>

### Expected Result

(07/18/2024 01:58:24) [ Warning ] : clif_parse: Received unsupported packet (packet 0x0013, 30 bytes received), disconnecting session #213.
(07/18/2024 01:58:24) [ Warning ] : clif_parse: Received unsupported packet (packet 0x0013, 30 bytes received), disconnecting session #374.
(07/18/2024 01:58:24) [ Warning ] : clif_parse: Received unsupported packet (packet 0x0013, 30 bytes received), disconnecting session #257.
(07/18/2024 01:58:24) [ Warning ] : clif_parse: Received unsupported packet (packet 0x0013, 30 bytes received), disconnecting session #324.
[...]

PR #8558, opened 2024-08-16
hotfix/packetspammingdelay

AoShinRO:
* **Addressed Issue(s)**:  https://github.com/rathena/rathena/issues/8512

* **Server Mode**:  both

* **Description of Pull Request**:  i found an older TODO:  to prevent packetspamming, lets  work on it? on comment say click masters can send 100 packets in 10  seconds, but openkore can send 100 packets  in 1 second, it need to be fixed, i'm accepting suggestions


src/map/clif.cpp
```diff
@@ -25204,6 +25219,25 @@ static int clif_parse(int fd)
 	}
 #endif
 
+	ARR_FIND(0,packet_spammer.size(),piter,packet_spammer[piter].client_addr == session[fd]->client_addr);
```
vstumpf, line 25222: on every processed packet, we do a linear search on a list of recent packet senders? Why not just put this inside `map_session_data`?

Also, this is a c++ vector, you can just use `std::find()`.
this also doesnt work on dual clients, only the first client will get kicked

src/map/clif.cpp
```diff
@@ -25141,25 +25141,40 @@ void clif_specialpopup(map_session_data& sd, int32 id ){
...
-	int pnum;
+	int piter;
 #ifdef PACKET_OBFUSCATION
 	int cmd2;
 #endif
 
-	//TODO apply delays or disconnect based on packet throughput [FlavioJS]
-	// Note: "click masters" can do 80+ clicks in 10 seconds
-
-	for( pnum = 0; pnum < 3; ++pnum )// Limit max packets per cycle to 3 (delay packet spammers) [FlavioJS]  -- This actually aids packet spammers, but stuff like /str+ gets slow without it [Ai4rei]
```
vstumpf, line 25159: Why remove this? this is a performance decrease. If you really don't like it, make it configurable

src/map/clif.cpp
```diff
@@ -25141,25 +25141,40 @@ void clif_specialpopup(map_session_data& sd, int32 id ){
 #endif
 }
 
+static int min_tick_interval = 100;
```
vstumpf, line 25144: make this configurable, rename it to be more clear/specific

src/map/clif.cpp
```diff
@@ -25141,25 +25141,40 @@ void clif_specialpopup(map_session_data& sd, int32 id ){
...
+static int32 ban_time = min_tick_interval*10000;
```
vstumpf, line 25145: make this configurable

src/map/clif.cpp
```diff
@@ -25204,6 +25219,25 @@ static int clif_parse(int fd)
...
+		packet_spammer[piter].click.count++;
+		return 0;
+	}
+	else if(sd && piter!=packet_spammer.size() && (cmd > MAX_PACKET_DB || cmd < MIN_PACKET_DB || packet_db[cmd].len == 0)){
+		chrif_req_login_operation(sd->status.account_id, sd->status.name, CHRIF_OP_LOGIN_BAN, ban_time+(int32)gettick(), 0, 0);
+		set_eof(fd);
+		return 0;
+	} else if(piter!=packet_spammer.size()){
+		packet_spammer.erase(packet_spammer.begin()+piter);
+		packet_spammer.shrink_to_fit();
```
vstumpf, line 25235: don't shrink_to_fit

src/map/clif.cpp
```diff
@@ -25141,25 +25141,40 @@ void clif_specialpopup(map_session_data& sd, int32 id ){
...
 }
 
+static int min_tick_interval = 100;
+static int32 ban_time = min_tick_interval*10000;
+struct packet_spammer_data{
+	uint32 client_addr;
+	t_tick tick;
+	struct{
+		short count;
+		short max=10;
```
vstumpf, line 25151: Make the max packet configurable

vstumpf (changes requested):
Honestly, I'm not sure this is a good feature to have in the repo, might be better to just leave this as an addon

src/map/clif.cpp
AoShinRO: changed to store fd, it's solving that?

PR #8562, opened 2024-08-16
map_session_data container memory allocation in c++ syntax

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**:  both

* **Description of Pull Request**:  just more one c++ update


AoShinRO:
there are two initializations of map_session_data dummy which hinders implementing pc_setnewpc as a constructor method :/

PR #8564, opened 2024-08-17
Fix GD_CHARGESHOUT_BEATING check

aleos89:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Renewal

  * Add missing parentheses to ensure the player is in a guild and avoid a short circuit check.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -12834,7 +12834,7 @@ void clif_parse_skill_toid( map_session_data* sd, uint16 skill_id, uint16 skill_
 	sd->skillitem = sd->skillitemlv = 0;
 
 	if( SKILL_CHK_GUILD(skill_id) ) {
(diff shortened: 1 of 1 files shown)


src/map/clif.cpp
```diff
@@ -12834,7 +12834,7 @@ void clif_parse_skill_toid( map_session_data* sd, uint16 skill_id, uint16 skill_
 	sd->skillitem = sd->skillitemlv = 0;
 
 	if( SKILL_CHK_GUILD(skill_id) ) {
-		if (sd->guild && sd->state.gmaster_flag || skill_id == GD_CHARGESHOUT_BEATING)
+		if (sd->guild && (sd->state.gmaster_flag || skill_id == GD_CHARGESHOUT_BEATING))
```
Atemo, line 12837: (small nickpick)
```suggestion
		if (sd->guild != nullptr && (sd->state.gmaster_flag || skill_id == GD_CHARGESHOUT_BEATING))
```

secretdataz (approved):

PR #8568, opened 2024-08-20
getitem+equip scriptcmd improvment

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**:  Both

* **Description of Pull Request**:  this make  getitem2,3,4 generate an item unique id and push it back to be used  in equip  command
now you can use:
```
getitem4(item_id,1,1,.@r,0,.@c0,.@c1,.@c2,.@c3,.@grade,.@r_id,.@r_val,.@r_par);
equip_unique(.@unique_id$);
```
and  equip exactly item generated  by getitem cmd


Haydrich:
I was looking for this changes for several years. Appreciated 🙏🏻

Lemongrass3110:
For me this does not work out for the following reasons:
It would have to return an array of unique ids => then it is just as good as returning the inventory index.
On the other hand the equip function reinterprets the same parameter differently, which I do not like personally.

src/map/script.cpp
```diff
@@ -7855,6 +7856,11 @@ BUILDIN_FUNC(getitem2)
 					return SCRIPT_CMD_FAILURE;
 				}
 			}
+			int maxlen = 256;
+			char *buf = (char *)aMalloc(maxlen*sizeof(char));
```
vstumpf, line 7860: This gets leaked
 `sizeof(char)` is always 1, so you can just use maxlen
You don't need to cast the result of aMalloc() either.

src/map/script.cpp
```diff
@@ -7855,6 +7856,11 @@ BUILDIN_FUNC(getitem2)
...
+			memset(buf, 0, maxlen);
+			snprintf(buf, maxlen-1, "%llu", (unsigned long long)unique_id);
+			mapreg_setregstr(amount > 1? reference_uid(add_str("$@unique_id$"), i):add_str("$@unique_id$"), buf);
```
vstumpf, line 7863:
```suggestion
			mapreg_setregstr(reference_uid(add_str("$@unique_id$"), i), buf);
```

src/map/script.cpp
```diff
@@ -16769,21 +16775,16 @@ BUILDIN_FUNC(equip) {
...
-
-		ARR_FIND( 0, MAX_INVENTORY, i, sd->inventory.u.items_inventory[i].nameid == nameid );
-		if (i < MAX_INVENTORY) {
-			pc_equipitem(sd,i,id->equip);
-			script_pushint(st,1);
-			return SCRIPT_CMD_SUCCESS;
-		}
+	uint64 id = script_getnum64(st,2);
+	int i;
+	ARR_FIND( 0, MAX_INVENTORY, i, sd->inventory.u.items_inventory[i].nameid == (t_itemid)id || sd->inventory.u.items_inventory[i].unique_id == id);
```
vstumpf, line 16780: I wouldn't overload `equip`, instead make a new command called `equip_uniqueid`, then only search for `unique_id`s there.

vstumpf (changes requested):
If you make this change to `getitem2`, I would want it in the other get*item as well
(`getitem`, `getnameditem`, `rentitem(2)`), etc

src/map/script.cpp
Atemo: `$@` is a misuse, please use the scope variable `.@`

Atemo (commented):
Why not return the inventory index instead?

src/map/script.cpp
AoShinRO: when i look to that now it's seen dumb, i can just use std::string :smile:

AoShinRO:
> Why not return the inventory index instead?

at this point we know the inventory index without an ARR_FIND?

src/map/script.cpp
```diff
@@ -7679,13 +7679,17 @@ BUILDIN_FUNC(getitem)
...
 			e_additem_result flag = pc_additem( sd, &it, get_count, LOG_TYPE_SCRIPT );
 
 			if( flag != ADDITEM_SUCCESS ){
 				clif_additem(sd, 0, 0, flag);
 				ShowError( "buildin_getitem: Failed to add the item to player.\n" );
 				return SCRIPT_CMD_FAILURE;
 			}
+			std::string buf;
+			snprintf(buf.data(), NAME_LENGTH, "%llu", (unsigned long long)unique_id);
+			setd_sub_str(st, nullptr, ".@unique_id$", i, buf.c_str(), nullptr);
```
AoShinRO, line 7692: @Atemo is that right?

src/map/script.cpp
```diff
@@ -16788,6 +16809,26 @@ BUILDIN_FUNC(equip) {
 	return SCRIPT_CMD_FAILURE;
 }
 
+BUILDIN_FUNC(equip_unique) {
```
AoShinRO, line 16812: equip_unique or equip_uniqueid ?

AoShinRO:
> If you make this change to `getitem2`, I would want it in the other get*item as well (`getitem`, `getnameditem`, `rentitem(2)`), etc

I only added this to getitem because at the moment I personally only used it with the getitem command, but it makes sense to extend it to other command scripts, some people say this may not be useful, but those who say this have never needed to make a custom item enchanter/refiner/changer and were disappointed when trying to equip the same spawned item when the player has more than one item of the same id in their inventory and the command equipped the first id it found :smile:

PR #8570, opened 2024-08-21
Improved status_isdead and status_get_status_data

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: 

* **Description of Pull Request**:  i have some issues  refering nullptr at status_get_status_data(bl) called by status_isdead

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -1904,7 +1904,7 @@ int status_revive(struct block_list *bl, unsigned char per_hp, unsigned char per
 {
 	struct status_data *status;
 	unsigned int hp, sp, ap;
-	if (!status_isdead(bl)) return 0;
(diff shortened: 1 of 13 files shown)


src/map/status.cpp
```diff
@@ -9223,7 +9223,7 @@ std::vector<e_race2> status_get_race2(struct block_list *bl)
  */
 int status_isdead(struct block_list *bl)
```
Atemo, line 9224: Can we use ref and return a bool instead?

AoShinRO: it's that?

AoShinRO: if change status_getdata  to ref is bad idea, say and i revert it  @Atemo

AoShinRO:
@Lemongrass3110 it work's

PR #8575, opened 2024-08-26
server-side-lesseffect-utilitty

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: both

* **Description of Pull Request**:  when player disable effects with /effects, packets from unit effects stay sended to client, he don't see the skill because it's client side coded, but server still sending unneeded packets


vstumpf (changes requested):
Personally I don't like this, I'm pretty sure it breaks `forcerendereffect.lua` in the client.

PR #8600, opened 2024-08-29
Implemented ZC_WARPLIST packet

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: both

* **Description of Pull Request**: Implemented ZC_WARPLIST & ZC_WARPLIST2 new packet


src/map/clif.cpp
```diff
@@ -6154,37 +6154,37 @@ void clif_skill_poseffect(struct block_list *src,uint16 skill_id,int val,int x,i
...
-/// Presents a list of available warp destinations (ZC_WARPLIST).
-/// 011c <skill id>.W { <map name>.16B }*4
-void clif_skill_warppoint( map_session_data* sd, uint16 skill_id, uint16 skill_lv, const char* map1, const char* map2, const char* map3, const char* map4 ){
-	int fd;
-	nullpo_retv(sd);
-	fd = sd->fd;
+/// Presents a list of available warp destinations.
+/// 011c <skill id>.W { <map name>.16B }*4 (ZC_WARPLIST)
+/// 0abe <lenght>.W <skill id>.W { <map name>.16B }*4 (ZC_WARPLIST2)
+void clif_skill_warppoint( map_session_data& sd, uint16 skill_id, uint16 skill_lv, std::vector<const char*> maps){
```
aleos89, line 6160: Since you flipped this to a vector, why not use a vector of strings instead?

AoShinRO: @aleos89 i always forget of std::string existence haha

src/map/clif.cpp
```diff
@@ -6154,37 +6154,60 @@ void clif_skill_poseffect(struct block_list *src,uint16 skill_id,int val,int x,i
...
+
+	switch(skill_id){
+		case AL_TELEPORT:
+		case ALL_ODINS_RECALL:
+			maps.push_back("Random");
+			if( skill_lv > 1 || skill_id == ALL_ODINS_RECALL )
+				maps.push_back(sd.status.save_point.map);
+
+			break;
+	case AL_WARP:
```
AoShinRO, line 6172:
```suggestion
		case AL_WARP:
```

src/map/clif.cpp
```diff
@@ -6154,37 +6154,60 @@ void clif_skill_poseffect(struct block_list *src,uint16 skill_id,int val,int x,i
...
+	int len, memoCount=0;
+	for(std::string map : maps){
+		if( !map.empty() ){
+			mapindex_getmapname_ext( map.c_str(), p->maps[memoCount++].map );
+		}
 	}
-	WFIFOSET(fd,packet_len(0x11c));
+#if PACKETVER_MAIN_NUM >= 20170502 || PACKETVER_RE_NUM >= 20170419 || defined(PACKETVER_ZERO)
+	len = sizeof(PACKET_ZC_WARPLIST) + (sizeof(PACKET_ZC_WARPLIST_sub) * memoCount);
+	p->packetLength = len;
```
AoShinRO, line 6198:
```suggestion
	p->packetLength = static_cast<decltype(p->packetLength)>(len);
```

src/map/clif.cpp
```diff
@@ -6154,37 +6154,71 @@ void clif_skill_poseffect(struct block_list *src,uint16 skill_id,int val,int x,i
...
+		mapindex_getmapname_ext( map.c_str(), warp.map );
+
+		if( memoCount++ == max ){
+			break;
+		}
 	}
-	if( strcmp( "", map4 ) != 0 ){
-		mapindex_getmapname_ext( map4, WFIFOCP( fd, 52 ) );
+
+	for( ; memoCount < max; memoCount++ ){
```
AoShinRO, line 6207: @Lemongrass3110 
Is this part necessary? I'm not sure if the 0x011c packet sends empty strings to the client, in the WFIFOHEAD version the strings are only added if( strcmp( "", map1 ) != 0 )

Daegaladh:
On kRO there's a quest to expand the memo points by 2, with a total of 5 destinations, but actually it's pretty much unlimited as it's stored as a permanent status change so we might need to future-proof it a bit

src/map/clif.cpp
```diff
@@ -6154,37 +6154,71 @@ void clif_skill_poseffect(struct block_list *src,uint16 skill_id,int val,int x,i
...
-		mapindex_getmapname_ext( map3, WFIFOCP( fd, 36 ) );
+
+	clif_send( p, p->packetLength, &sd.bl, SELF );
+#else
+	PACKET_ZC_WARPLIST p = {};
+
+	p.packetType = HEADER_ZC_WARPLIST;
+	p.skillId = skill_id;
+
+	size_t memoCount = 0, max = 4;
```
Daegaladh, line 6192:
```suggestion
	size_t memoCount = 0, max = MAX_MEMOPOINTS + 1;
```

AoShinRO:
> On kRO there's a quest to expand the memo points by 2, with a total of 5 destinations, but actually it's pretty much unlimited as it's stored as a permanent status change so we might need to future-proof it a bit

I was wondering why Gravity changed the package to support more memopoints and what the limits would be without blowing up the client, now it makes sense, so it invented an expansion system, we can work with that to deliver the complete system

Daegaladh:
I think it would be better to split the PR in two parts, one for the packet and the skill changes and one for the quest itself

PR #8613, opened 2024-09-06
Implemented /resetcooltime

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: Both

 

* **Description of Pull Request**:  this packet is related to /resetcooltime GM cmd, currently you are disconnected from the server after using the command

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -14776,9 +14776,10 @@ void clif_parse_GMHide(int fd, map_session_data *sd) {
 
 /// /resetcooltime 
 /// 0a88 (CZ_CMD_RESETCOOLTIME).
-void clif_parse_GMCooldownReset(int fd, map_session_data* sd) {
-#if PACKETVER >= 20160622
-	char cmd[15];
+void clif_parse_gm_resetcooltime( int fd, map_session_data* sd ){
+#if PACKETVER_MAIN_NUM >= 20160622 || PACKETVER_RE_NUM >= 20160622 || defined(PACKETVER_ZERO)
+	const PACKET_CZ_CMD_RESETCOOLTIME* p = reinterpret_cast<const PACKET_CZ_CMD_RESETCOOLTIME*>( RFIFOP( fd, 0 ) );
+	char cmd[CHAT_SIZE_MAX];
 
 	safesnprintf(cmd,sizeof(cmd),"%cresetcooltime",atcommand_symbol);
 	is_atcommand(fd, sd, cmd, 1);
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -14757,28 +14757,6 @@ void clif_parse_GM_Item_Monster(int fd, map_session_data *sd)
 	}
 }
 
-/// /resetcooltime 
-/// 0a88 (CZ_CMD_RESETCOOLTIME).
-void clif_parse_GMCooldownReset(int fd, map_session_data* sd) {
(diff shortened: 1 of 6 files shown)


src/map/clif_packetdb.hpp
```diff
@@ -290,6 +290,7 @@
 	packet(0x01dd,47);
 	packet(0x01de,33);
 	parseable_packet(0x01df,6,clif_parse_GMReqAccountName,2);
+	parseable_packet(0x0a88,2,clif_parse_GMCooldownReset,0);
```
Lemongrass3110, line 293:
```suggestion
#if PACKETVER_MAIN_NUM >= 20160622 || PACKETVER_RE_NUM >= 20160622 || defined(PACKETVER_ZERO)
	parseable_packet(0x0a88,2,clif_parse_GMCooldownReset,0);
#endif
```
Please also move it down. Maybe replace the `2016-06-22aRagexeRE` entry that is not used anymore.

src/map/clif.cpp
```diff
@@ -14757,6 +14757,28 @@ void clif_parse_GM_Item_Monster(int fd, map_session_data *sd)
...
+/// 0a88 (CZ_CMD_RESETCOOLTIME).
+void clif_parse_GMCooldownReset(int fd, map_session_data* sd) {
+#if PACKETVER >= 20160622
+	if(sd->group_id < 99)
+		return;
+	for (int i = 0; i < MAX_SKILLCOOLDOWN; i++) {
+		if( sd->scd[i] != nullptr ) {
+			std::string buf = "Found skill '" + std::string(skill_db.find(sd->scd[i]->skill_id)->name) + "', unblocking...";
+			clif_displaymessage(sd->fd, buf.c_str());
+			buf.clear();
```
Lemongrass3110, line 14770:
```suggestion
			std::string buf = "Found skill '" + std::string(skill_db.find(sd->scd[i]->skill_id)->name) + "', unblocking...";
			clif_displaymessage(sd->fd, buf.c_str());
			buf.clear();
```
Did you add this for debugging or how do you know that this message is being triggered on AEGIS?

src/map/clif.cpp
```diff
@@ -14757,6 +14757,28 @@ void clif_parse_GM_Item_Monster(int fd, map_session_data *sd)
 	}
 }
 
+/// /resetcooltime 
+/// 0a88 (CZ_CMD_RESETCOOLTIME).
+void clif_parse_GMCooldownReset(int fd, map_session_data* sd) {
+#if PACKETVER >= 20160622
+	if(sd->group_id < 99)
```
Lemongrass3110, line 14764: It would be better to create an atcommand and forward /resetcooltime to @resetcooltime so that the permission/group check can be customized via atcommands.yml

Lemongrass3110 (changes requested):

src/map/clif.cpp
AoShinRO: > Did you add this for debugging or how do you know that this message is being triggered on AEGIS?

I was looking for how to use the 0x01df package in-game, I think it was removed, then I saw some commands in the assembly that are giving unknowpacket  and disconnect myself, one of them was this, the message i got from hercule

src/map/clif.cpp
AoShinRO: i'will

Lemongrass3110 (approved):

aleos89 (approved):

PR #8615, opened 2024-09-09
Extended Memo Slots Quest

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: Renewal

* **Description of Pull Request**: Extended Memo Slots quest

- [x] find way to reproduce same warp effect used by gvt on talking with the npcs

the quest -> https://www.youtube.com/watch?v=uF8iAE8d0fo
patch notes -> https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7393&curpage=37
image

this content depends of aproval of new warplist packet
https://github.com/rathena/rathena/pull/8600


src/map/script.cpp
```diff
@@ -27412,6 +27412,36 @@ BUILDIN_FUNC(permission_add)
...
+
+#if PACKETVER_MAIN_NUM >= 20170502 || PACKETVER_RE_NUM >= 20170419 || defined(PACKETVER_ZERO)
+	if(slot < 0 || slot >= ((MAX_MEMOPOINTS / 2) + pc_readreg2(sd,EXT_MEMO_VAR)))
+		return SCRIPT_CMD_FAILURE;
+#else
+	if(slot < 0 || slot >= MAX_MEMOPOINTS)
+		return SCRIPT_CMD_FAILURE;
+#endif
+
+	safestrncpy( sd->status.memo_point[slot].map, map_name, sizeof( sd->status.memo_point[slot].map ) );
```
AoShinRO, line 27438: i need to change it to search a free slot before alloc in slot determined

src/map/atcommand.cpp
```diff
@@ -2591,6 +2591,10 @@ ACMD_FUNC(memo)
 		clif_displaymessage(sd->fd,  msg_txt(sd,668)); // Your actual memo positions are:
 		for( i = 0; i < MAX_MEMOPOINTS; i++ )
 		{
+#if PACKETVER_MAIN_NUM >= 20170502 || PACKETVER_RE_NUM >= 20170419 || defined(PACKETVER_ZERO)
+			if (i >= ((MAX_MEMOPOINTS / 2) + pc_readreg2(sd, EXT_MEMO_VAR)))
```
Daegaladh, line 2595: MAX_MEMOPOINTS / 2 could cause problems if Gravity extends it even more in the future, consider to leave MAX_MEMOPOINTS as 3 and add a new const instead `MAX_EXT_MEMOPOINTS = MAX_MEMOPOINTS + 3`

PR #8633, opened 2024-09-12
sugestion to Bug fix #8625

### LINKED ISSUE #8625: Unable to restore stack! Double Continuation!
### rAthena Hash

1f25ebc

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

It seems that usable items with callfunc that have mes or other NPC dialog such as Kafra Card can trigger Double Continuation. Only happen on pre-renewal, tested it on fresh copy of rathena.
double continuation

### Relevant Log Output

_No response_

### Expected Result

not allow player to spam the item when an NPC dialog is active.

### How to Reproduce

1. use @item 12211 20
2. use Kafra Card and when the mes is open just use another one

### Official Information

NA

_No response_

AoShinRO:
* **Addressed Issue(s)**:  https://github.com/rathena/rathena/issues/8625

* **Server Mode**: Both

* **Description of Pull Request**: fixing #8625

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -11535,3 +11535,11 @@ Set position for NPC dialog in pixels.
 Set position for NPC dialog in screen size percent.
 
 ---------------------------------------
+
+*checknpcdialogue()
+
+Check if player is interacting with other npc
+These functions will return 1 (interacting with npc) or 0 (dialogue free)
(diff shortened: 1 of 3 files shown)


AoShinRO:
I'm not sure if this is the best approach to solve this, but it is functional, but it only solves cases where the command is entered

Issue #8608, opened 2024-09-05
Can use skill while npc dialogue is open


thanna:
### rAthena Hash

f4228abf1

### Client Date

2022-04-06

### Server Mode

Pre-Renewal, Renewal

### Result

Is it normal that I can use skill while the NPC dialog is open, though it doesnt work on offensive skills
![bug](https://github.com/user-attachments/assets/07f0e57f-dd93-4302-a7a8-526efa144c8e)

### Relevant Log Output

```Shell
No error log
```

### Expected Result

not allow to use skills while the npc dialog is open.

### How to Reproduce

1. Just click any NCP with mes or menu
2. use skill that target self

### Official Information

I tested this on iRO, which doesnt allow skills while talking to NPC.

### Modifications that may affect results

none, this is a fresh new copy of rathena.

Daegaladh:
That's the official behavior, on pre-renewal you can use skills while talking to NPCs, so not a bug.
Maybe we could add a config instead and default it to different settings depending on the mode.

thanna:
> That's the official behavior, on pre-renewal you can use skills while talking to NPCs, so not a bug. Maybe we could add a config instead and default it to different settings depending on the mode.

Thank you @Daegaladh for confirming it maybe Im not used to it.

Daegaladh:
I want to clarify because what I said is not entirely correct and it's kinda bugged on rAthena, so I'm gonna fix it and adding some config similar to what we have with item use.
On pre-renewal you can use 'self' skills while talking to NPCs, but not ground or targeted skills.
On renewal you can't use skills at all.
Also noticed on renewal homunculus and mercenaries still can use Self 'self' while talking to an NPC.

gidzdlcrz:
You can just use the following script command.
```
disable_items; // To prevent item usage when speaking to NPC
disable_command; // To prevent commands and skill usage when speaking to NPC
```
You can do it way

`-	script	Warper	-1,{
function Go; function Disp; function Pick; function Restrict;

// --------------------------------------------------
//	Main Menu:
// --------------------------------------------------

disable_items;
disable_command;

menu	"Last Warp ^777777["+lastwarp$+"]^000000",-,
    	" ~ Towns",Towns,
    	" ~ Fields",Fields,
    	" ~ Dungeons",Dungeons,
    	" ~ Guild Castles",Castles,
    	" ~ Guild Dungeons",Guild_Dungeons,`

PR #8659, opened 2024-09-19
Fix for #8608 - Can use skill while talking to NPC

thanna:
* **Addressed Issue(s)**: 
Fix for #8608 - Can use skill while talking to NPC

* **Server Mode**: 
Pre-Renewal

Can use skill while talking to NPC


Daegaladh:
Closed since this is not a bug, but intended behavior

PR #8670, opened 2024-09-21
Improved readability of distance functions  and change typeof it

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: 

* **Description of Pull Request**: cleaning up of distance functions making new macros with creator of formula names and added comments  and references

ps: this is the second time I've tried to improve the readability of path files.

image

why change distances to unsigned char?
In ragnarok, we always work with AREA_SIZE and MAX_WALKPATH to measure the maximum distances that can be reached, and both are always less than 255, this can reduce some memory allocation larger than necessary

### DIFF
--- a/src/common/database.cpp
+++ b/src/common/database.cpp
@@ -276,14 +276,6 @@ bool YamlDatabase::asBool(const ryml::NodeRef& node, const std::string &name, bo
 	}
 }
 
-bool YamlDatabase::asInt8( const ryml::NodeRef& node, const std::string& name, int8& out ){
-	return asType<int8>( node, name, out);
-}
-
-bool YamlDatabase::asUInt8( const ryml::NodeRef& node, const std::string& name, uint8& out ){
-	return asType<uint8>( node, name, out);
-}
-
 bool YamlDatabase::asInt16( const ryml::NodeRef& node, const std::string& name, int16& out ){
 	return asType<int16>( node, name, out);
 }
--- a/src/common/database.cpp
+++ b/src/common/database.cpp
@@ -276,6 +276,14 @@ bool YamlDatabase::asBool(const ryml::NodeRef& node, const std::string &name, bo
 	}
 }
 
+bool YamlDatabase::asInt8( const ryml::NodeRef& node, const std::string& name, int8& out ){
+	return asType<int8>( node, name, out);
+}
+
+bool YamlDatabase::asUInt8( const ryml::NodeRef& node, const std::string& name, uint8& out ){
+	return asType<uint8>( node, name, out);
+}
+
 bool YamlDatabase::asInt16( const ryml::NodeRef& node, const std::string& name, int16& out ){
 	return asType<int16>( node, name, out);
 }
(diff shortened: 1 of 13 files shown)


src/map/path.cpp
```diff
@@ -431,46 +431,37 @@ bool path_search(struct walkpath_data *wpd, int16 m, int16 x0, int16 y0, int16 x
...
 }
 
-
 //Distance functions, taken from http://www.flipcode.com/articles/article_fastdistance.shtml
 bool check_distance(int dx, int dy, int distance)
 {
 #ifdef CIRCULAR_AREA
-	//In this case, we just do a square comparison. Add 1 tile grace for diagonal range checks.
-	return (dx*dx + dy*dy <= distance*distance + (dx&&dy?1:0));
+	dx = std::abs(dx);
```
Playtester, line 438: This is not really needed because if you if multiply the value with itself you will always have a positive result.
We also shouldn't remove comments.

AoShinRO: i  put it here as mistake, mybad

AoShinRO:
if anyone can explain why we use CHEBYSHEV formula to calculate circular area and EUCLIDEAN formula to calculate square area i will appreciate
image

AoShinRO:
builded some macros  with comments to make  all formulas more easy to read, changed link of chebyshev_distance  reference from waybackmachine link  because link is dead

src/map/path.cpp
```diff
@@ -496,17 +502,15 @@ bool check_distance_client(int dx, int dy, int distance)
...
  */
-int distance_client(int dx, int dy)
+unsigned int distance_client(int dx, int dy)
 {
-	double temp_dist = sqrt((double)(dx*dx + dy*dy));
+	double temp_dist = euclidean_distance(dx,dy);
 
 	//Bonus factor used by client
 	//This affects even horizontal/vertical lines so they are one cell longer than expected
 	temp_dist -= 0.1;
```
AoShinRO, line 511: this -0.1 is to "std::floor" on int conversion?

src/map/path.cpp
```diff
@@ -431,46 +461,22 @@ bool path_search(struct walkpath_data *wpd, int16 m, int16 x0, int16 y0, int16 x
...
-		min = dx;
-		max = dy;
-	} else {
-		min = dy;
-		max = dx;
-	}
-   // coefficients equivalent to ( 123/128 * max ) and ( 51/128 * min )
-	return ((( max << 8 ) + ( max << 3 ) - ( max << 4 ) - ( max << 1 ) +
-		( min << 7 ) - ( min << 5 ) + ( min << 3 ) - ( min << 1 )) >> 8 );
+	return std::max(chebyshev_distance(dx, dy), (unsigned)1);
```
AoShinRO, line 477: if anyone give-me okay i'll change it to euclidean_distance because this are on #ifdef circular_area, idk why it using chebyshev bit mode

src/common/database.cpp
```diff
@@ -276,6 +276,14 @@ bool YamlDatabase::asBool(const ryml::NodeRef& node, const std::string &name, bo
 	}
 }
 
+bool YamlDatabase::asInt8( const ryml::NodeRef& node, const std::string& name, int8& out ){
+	return asType<int8>( node, name, out);
+}
+
+bool YamlDatabase::asUInt8( const ryml::NodeRef& node, const std::string& name, uint8& out ){
+	return asType<uint8>( node, name, out);
```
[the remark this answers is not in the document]
AoShinRO: removed 👍

PR #8674, opened 2024-09-22
adding missing import file

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: RE

* **Description of Pull Request**: Navi generator error->
image


aleos89:
We don't manage the import/ directory files. Those get copied from import-tmpl/. See the src/map/map-server.vcxproj file and you can see the line is missing in there!

AoShinRO:
> We don't manage the import/ directory files. Those get copied from import-tmpl/. See the src/map/map-server.vcxproj file and you can see the line is missing in there!

Now I understand, thanks for the explanation, I think I did it right this time

aleos89 (approved):

PR #8680, opened 2024-09-24
Cleanup duplicate Baby class stats

aleos89:
* **Addressed Issue(s)**: N/A

  * Join the Baby classes with their counterpart classes to reduce the duplicate values defined in job_stats.
  * Add MaxWeightBaby which will be used if a Baby class is part of the Jobs list.


db/import-tmpl/job_stats.yml
```diff
@@ -25,6 +25,7 @@
 # - Jobs:                    List of jobs associated to group.
 #     Job                    Job name.
 #   MaxWeight                Base maximum weight. (Default: 20000)
```
Atemo, line 27: Shouldn't baby be excluded from MaxWeight with this change?

Lemongrass3110 (approved):

PR #8700, opened 2024-10-02
Cleanup CLI config parsing

### LINKED ISSUE #8618: Executing char-server with "--char-config" doesn't work
### rAthena Hash

f4cd68e42

### Client Date

N/A

### Server Mode

Pre-Renewal

### Result

When passing `--char-config conf/char_athena.conf` (the same file it already reads without the parameter) to `./char-server`, it doesn't work:

image

### Relevant Log Output

_No response_

### Expected Result

Running the char server without the param, just `./char_server`, it works properly:

image

### How to Reproduce

Just run the command `./char_server --char-config conf/char_athena.conf` after having it working with `./char-server`

### Official Information

N/A

I'm on a fresh install, without any src modifications

aleos89:
* **Addressed Issue(s)**: Fixes #8618

  * When --char-config is provided, the argument value was being sent to the function that loads the inter-server config resulting in some configuration values not being properly parsed.
Thanks to @joaogardenberg!

### DIFF
--- a/src/char/char.cpp
+++ b/src/char/char.cpp
@@ -3205,7 +3205,7 @@ bool CharacterServer::initialize( int argc, char *argv[] ){
 	}
 #endif
 
-	inter_init_sql((argc > 2) ? argv[2] : SQL_CONF_NAME); // inter server configuration
+	inter_init_sql((argc > 2 && strstr(argv[2], "inter") != nullptr) ? argv[2] : SQL_CONF_NAME); // inter server configuration
 
 	char_mmo_sql_init();
 	char_read_fame_list(); //Read fame lists.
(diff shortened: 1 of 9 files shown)


Lemongrass3110 (changes requested):
Please drop cli.cpp SQL_CONF_NAME instead and use INTER_CONF_NAME instead.
This is already correctly parsed in cli.cpp

Please also drop SQL_CONF_NAME, which is redefined in sql.cpp and ignores all other settings, and replace it by using INTER_CONF_NAME.

src/char/char.cpp
```diff
@@ -3205,7 +3205,7 @@ bool CharacterServer::initialize( int argc, char *argv[] ){
 	}
 #endif
 
-	inter_init_sql((argc > 2 && strstr(argv[2], "inter") != nullptr) ? argv[2] : SQL_CONF_NAME); // inter server configuration
+	inter_init_sql((argc > 2 && strstr(argv[2], "inter") != nullptr) ? argv[2] : INTER_CONF_NAME); // inter server configuration
```
Lemongrass3110, line 3208:
```suggestion
	inter_init_sql(INTER_CONF_NAME);
```

Lemongrass3110 (approved):

PR #8711, opened 2024-10-04
Party Name Should Be Displayed

Haydrich:

* **Server Mode**: RE

* **Description of Pull Request**: Corrected the party name

### DIFF
--- a/conf/battle/party.conf
+++ b/conf/battle/party.conf
@@ -54,8 +54,8 @@ idle_no_share: no
 party_even_share_bonus: 0
 
 // Display party name regardless if player is in a guild.
-// Official servers do not display party name unless the user is in a guild. (Note 1)
-display_party_name: no
+// Official servers display party name even if the user is not in a guild. (Note 1)
+display_party_name: yes
 
 // Prevent multiple characters of the same account to join the same party. (Note 1)
 block_account_in_same_party: yes


Toshiro90:
Can confirm on kRO.
grafik

Playtester:
@Toshiro90 Can you check what happens if you do /showname.

Toshiro90:
> @Toshiro90 Can you check what happens if you do /showname.

grafik

conf/battle/party.conf
```diff
@@ -54,8 +54,8 @@ idle_no_share: no
 party_even_share_bonus: 0
 
 // Display party name regardless if player is in a guild.
-// Official servers do not display party name unless the user is in a guild. (Note 1)
-display_party_name: no
+// Official servers display party name even if the user is not in a guild. (Note 1)
+display_party_name: yes
```
Atemo, line 58: Assuming the old behaviour is pre-renewal
```suggestion
// Default on official servers: no for Pre-renewal, yes for Renewal. (Note 1)
//display_party_name: yes
```
in battle.cpp
```
#ifdef RENEWAL
	{ "display_party_name",                 &battle_config.display_party_name,              1,      0,      1,              },
#else
	{ "display_party_name",                 &battle_config.display_party_name,              0,      0,      1,              },
#endif
```

Daegaladh: According to mrj, official pre-re behavior is to always display the party name, but there was a confusion because of a client bug that won't properly display it

Atemo (approved):

aleos89 (approved):

Issue #8716, opened 2024-10-06
Greed Skill New Update


Haydrich:
### Is your request related to a missing official feature?

Yes

### Describe the solution you'd like

In the kRO Greed skill shouldn't work when the player has less than 5 free slots in inventory
https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7607

PR #8717, opened 2024-10-06
Update Greed Skill Mechanism

Haydrich:
* **Addressed Issue(s)**: #8716

* **Server Mode**: RE

* **Description of Pull Request**: Greed skill should be disabled if the character have less than 5 slots

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -9826,11 +9826,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
 		break;
 
 	case BS_GREED:
-		if(sd){
-			clif_skill_nodamage(src,*bl,skill_id,skill_lv);
-			map_foreachinallrange(skill_greed,bl,
-				skill_get_splash(skill_id, skill_lv),BL_ITEM,bl);
-		}
+		if (sd != nullptr && pc_inventoryblank(sd) >= 5) { // Greed skill should be disabled if the character have less than 5 slots [Haydrich]
+			clif_skill_nodamage(src, bl, skill_id, skill_lv, 1);
+			map_foreachinallrange(skill_greed, bl,
+				skill_get_splash(skill_id, skill_lv), BL_ITEM, bl);
+		} else
+			clif_msg_color(sd, MSI_SKILL_INVENTORY_KINDCNT_OVER, color_table[COLOR_RED]);
 		break;
 
 	case SA_ELEMENTWATER:


src/map/skill.cpp
```diff
@@ -9826,11 +9826,12 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
-			clif_skill_nodamage(src,*bl,skill_id,skill_lv);
-			map_foreachinallrange(skill_greed,bl,
-				skill_get_splash(skill_id, skill_lv),BL_ITEM,bl);
-		}
+		if (sd != nullptr && pc_inventoryblank(sd) >= 5) { // Greed skill should be disabled if the character have less than 5 slots [Haydrich]
+			clif_skill_nodamage(src, *bl, skill_id, skill_lv, 1);
+			map_foreachinallrange(skill_greed, bl,
+				skill_get_splash(skill_id, skill_lv), BL_ITEM, bl);
+		} else
+			clif_msg_color(sd, MSI_SKILL_INVENTORY_KINDCNT_OVER, color_table[COLOR_RED]);
```
Atemo, line 9834: sd can't be nullptr for clif_msg_color, missing condition

Daegaladh:
On official pre-re there's no 5 slots check, but it shows a red message "You cannot carry more items because you've exceeded your inventory limit" and the items remain on the floor.

src/map/skill.cpp
```diff
@@ -9826,10 +9826,17 @@ int skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, ui
...
+#ifdef	RENEWAL
+			if (pc_inventoryblank(sd) >= 5) { // Greed skill should be disabled if the character have less than 5 slots [Haydrich]
+				clif_skill_nodamage(src, *bl, skill_id, skill_lv, 1);
+				map_foreachinallrange(skill_greed, bl, skill_get_splash(skill_id, skill_lv), BL_ITEM, bl);
+			} else
+				clif_msg_color(sd, MSI_SKILL_INVENTORY_KINDCNT_OVER, color_table[COLOR_RED]);
+#else
+			clif_skill_nodamage(src, *bl, skill_id, skill_lv);
+			map_foreachinallrange(skill_greed, bl, skill_get_splash(skill_id, skill_lv), BL_ITEM, bl);
+#endif
```
aleos89, line 9839:
```suggestion
			if (pc_inventoryblank(sd) < 5) { // Greed skill should be disabled if the character have less than 5 slots [Haydrich]
				clif_msg_color(sd, MSI_SKILL_INVENTORY_KINDCNT_OVER, color_table[COLOR_RED]);
				break;
			}
#endif
			clif_skill_nodamage(src, *bl, skill_id, skill_lv);
			map_foreachinallrange(skill_greed, bl, skill_get_splash(skill_id, skill_lv), BL_ITEM, bl);
```
This removes the need to duplicate the item pickup code. This doesn't take in to account the comment Daegaladh left though!

Haydrich: Yes i was preparing to cleanup it but thanks, but i guess we need to add else for the MSI_CANT_GET_ITEM_BECAUSE_COUNT on pre-re

Daegaladh:
Yes, it does the motion when the inventory is full, but doesn't get any items.
When overweight it just shows the message (clif_skill_fail)

Haydrich:
@Daegaladh Can you review it?

Daegaladh:
> @Daegaladh Can you review it?

Looks good, but you missed the overweight part, I can check it on renewal too as it's probably the same

Haydrich:
> > @Daegaladh Can you review it?
> 
> Looks good, but you missed the overweight part, I can check it on renewal too as it's probably the same

Actually i tested the skill on overweight
image
What is the missing then?

Daegaladh:
> > > @Daegaladh Can you review it?
> > 
> > 
> > Looks good, but you missed the overweight part, I can check it on renewal too as it's probably the same
> 
> Actually i tested the skill on overweight image What is the missing then?

Yeah sorry, I didn't realize it was checked before the skill code, ignore it.

Haydrich:
Items that use `itemskill "BS_GREED",1;` requires a check of the checkweight, it should shows the message of MSI_SKILL_INVENTORY_KINDCNT_OVER
```
      if (checkweight(1201,5) == 0) {
         dispbottom "Please secure some room in the inventory",0xFF0000;
         end;
      }
```
But i was wondering for new script command that can send a MSI instead of using dispbottom? it would be fine to be used for those items in case of the sentence can be the same in the two cases (skill and item)

Daegaladh:
> Items that use `itemskill "BS_GREED",1;` requires a check of the checkweight, it should shows the message of MSI_SKILL_INVENTORY_KINDCNT_OVER
> 
> ```
>       if (checkweight(1201,5) == 0) {
>          dispbottom "Please secure some room in the inventory",0xFF0000;
>          end;
>       }
> ```
> 
> But i was wondering for new script command that can send a MSI instead of using dispbottom? it would be fine to be used for those items in case of the sentence can be the same in the two cases (skill and item)

Isn't that already managed by the skill execution code?

Haydrich:
> Isn't that already managed by the skill execution code?

I thought this too but no, kRO is using the packet skill fail for the skills, but for items it should be a dispbottom (Because the packet fail in kRO was pushing the skill name, but the item isn't push the skill name in case of fail)

Daegaladh:
I guess we could just export the constants then

Daegaladh (approved):

Atemo:
The result is still different
rAthena, skill name displayed when inventory > 95 items
Capture d’écran (2051)
kRO, no skill name displayed when inventory > 95 items
Capture d’écran (2052)

Haydrich:
> The result is still different rAthena, skill name displayed when inventory > 95 items Capture d’écran (2051) kRO, no skill name displayed when inventory > 95 items Capture d’écran (2052)

Actually it's not related to the skill, its related to the item itself. I've mentioned this above
```
      if (checkweight(1201,5) == 0) {
         dispbottom "Please secure some room in the inventory",0xFF0000;
         end;
      }
```
And i asked before for a change that i can call **MSI_SKILL_INVENTORY_KINDCNT_OVER** inside a dispbottom

Daegaladh:
> Actually it's not related to the skill, its related to the item itself. I've mentioned this above
> 
> ```
>       if (checkweight(1201,5) == 0) {
>          dispbottom "Please secure some room in the inventory",0xFF0000;
>          end;
>       }
> ```
> 
> And i asked before for a change that i can call **MSI_SKILL_INVENTORY_KINDCNT_OVER** inside a dispbottom

I've checked on Aegis renewal using a Greed Scroll, and it just calls the skill, is the skill that does the checks. Does it work different on kRO?

Haydrich:
> > Actually it's not related to the skill, its related to the item itself. I've mentioned this above
> > ```
> >       if (checkweight(1201,5) == 0) {
> >          dispbottom "Please secure some room in the inventory",0xFF0000;
> >          end;
> >       }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And i asked before for a change that i can call **MSI_SKILL_INVENTORY_KINDCNT_OVER** inside a dispbottom
> 
> I've checked on Aegis renewal using a Greed Scroll, and it just calls the skill, is the skill that does the checks. Does it work different on kRO?

Actually I haven't test it with a normal items, it was a rent item, so in kRO it's call the skill. but the issue was, in case you've more than 95 item in your inventory the item consumed when it's a normal item and the skill gonna failed.

Issue #8722, opened 2024-10-07
WZ_ESTIMATION shows wrong values


Singe-Horizontal:
### rAthena Hash

ad0349c0f4

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

When using WZ_ESTIMATION, showed values correspond to element level + 1

### Expected Result

WZ_ESTIMATION should shows correct elemental weakness.

### How to Reproduce

Cast WZ_ESTIMATION on a monster and compare values to element table.
For level 4 elements, it will show 100 for all of them.

### Official Information

N/A

PR #8723, opened 2024-10-07
Fix battle_attr_fix logic for ele level

Singe-Horizontal:
* **Addressed Issue(s)**: 
#8722

* **Server Mode**: 
Both

AttributeDatabase::getAttribute `level` parameter should be comprised between 1~MAX_ELE_LEVEL instead of 0 ~ MAX_ELE_LEVEL -1.
This fixes WZ_ESTIMATION display bug.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -15663,7 +15663,7 @@ AttributeDatabase elemental_attribute_db;
  * @param level Element level 1 ~ MAX_ELE_LEVEL
  */
 int16 AttributeDatabase::getAttribute(uint16 level, uint16 atk_ele, uint16 def_ele) {
-	if (!CHK_ELEMENT(atk_ele) || !CHK_ELEMENT(def_ele) || !CHK_ELEMENT_LEVEL(level+1))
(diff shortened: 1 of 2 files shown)


src/map/status.cpp
```diff
@@ -15663,7 +15663,7 @@ AttributeDatabase elemental_attribute_db;
  * @param level Element level 1 ~ MAX_ELE_LEVEL
  */
 int16 AttributeDatabase::getAttribute(uint16 level, uint16 atk_ele, uint16 def_ele) {
-	if (!CHK_ELEMENT(atk_ele) || !CHK_ELEMENT(def_ele) || !CHK_ELEMENT_LEVEL(level+1))
+	if (!CHK_ELEMENT(atk_ele) || !CHK_ELEMENT(def_ele) || !CHK_ELEMENT_LEVEL(level))
 		return 100;
 
 	return this->attr_fix_table[level][atk_ele][def_ele];
```
Atemo, line 15669:
```suggestion
	return this->attr_fix_table[level-1][atk_ele][def_ele];
```
https://github.com/rathena/rathena/blob/master/src/map/status.hpp#L138

Singe-Horizontal: Sorry about that

Issue #8643, opened 2024-09-13
guild name is not loaded on login and the guild leader is offline


AoShinRO:
### rAthena Hash

d280915

### Client Date

20221116

### Server Mode

Renewal

### Result

guild name is not loaded if you login and the guild leader is offline
anyone can confirm other clients?

https://github.com/user-attachments/assets/72e09821-f16c-4ee3-b5af-582c8599f252

HD video -> https://youtu.be/PesqTLpa9Gw

### Expected Result

guildname need to be showed

### How to Reproduce

just login with an guild member account without online guild leader

### Official Information

This probably shouldn't depend on the online leader

AoShinRO:
this did not happen in the 2023 emulator

Toshiro90:
Had this behaviour for more than 3 years on my local test server with various client versions.
I was able to reproduce this problem and it is happening in the special case when the map server did not receive (nor requested) the guild data yet and you're logging in with any guild member (excluding the leader) for the first time after server restart. Upon receiving the guild data intends to send the names (char/party/guild/rank) of any guild member (with`clif_name_area`), however only the guild leader's map_session_data is updated with the guild reference _before_ sending.

PR #8725, opened 2024-10-08
Fixed guild name not showing on login

Toshiro90:
* **Addressed Issue(s)**: #8643 

* **Server Mode**: Both

Fixes guild name not showing when a character logs in and the corresponding guild data is not yet loaded on the map server.

### DIFF
--- a/src/map/guild.cpp
+++ b/src/map/guild.cpp
@@ -862,7 +862,6 @@ int guild_recv_info(const struct mmo_guild &sg) {
 	for(i=bm=m=0;i<g->guild.max_member;i++){
 		if(g->guild.member[i].account_id>0){
 			sd = g->guild.member[i].sd = guild_sd_check(g->guild.guild_id, g->guild.member[i].account_id, g->guild.member[i].char_id);
-			if (sd) sd->guild = g;
 			m++;
 		}else
 			g->guild.member[i].sd=nullptr;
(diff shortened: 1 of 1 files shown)


src/map/guild.cpp
```diff
@@ -862,7 +862,7 @@ int guild_recv_info(const struct mmo_guild &sg) {
 	for(i=bm=m=0;i<g->guild.max_member;i++){
 		if(g->guild.member[i].account_id>0){
 			sd = g->guild.member[i].sd = guild_sd_check(g->guild.guild_id, g->guild.member[i].account_id, g->guild.member[i].char_id);
-			if (sd) clif_name_area(&sd->bl); // [LuzZza]
+			if (sd) sd->guild = g;
```
Lemongrass3110, line 865:
```suggestion
```

src/map/guild.cpp
```diff
@@ -874,7 +874,7 @@ int guild_recv_info(const struct mmo_guild &sg) {
 		sd = g->guild.member[i].sd;
 		if( sd==nullptr )
 			continue;
-		sd->guild = g;
+		clif_name_area(&sd->bl); // [LuzZza]
```
Lemongrass3110, line 877:
```suggestion
		if( sd->guild == nullptr ){
			sd->guild = g;
			clif_name_area(&sd->bl);
		}else{
			sd->guild = g;
		}
```
Should it not be like that?

Lemongrass3110 (approved):

src/map/guild.cpp
Toshiro90: In regards of fixing the mentioned bug my approach worked fine (I placed the guild assignment next to assigning the member sd, considering the first loop as "assignment" loop), however the suggested condition ensures the name update is not sent when not necessary, e.g. when guild data is received after skill level change.

PR #8738, opened 2024-10-11
improve char/int_clan

Alvisrynoc:
* **Addressed Issue(s)**: not really an issue

* **Server Mode**: both

This is a QoL pull request.
This pull request is not intended and should not change behavior observed from client side and other servers.
This pull brings the following changes and benefits:
1) modularization, extensibility
   use of namespace
   OOP-ify the module, create interface classes for map, database and clan
   extend their functions with subclass and create hook class dedicated for user-defined behaviors
2) description for every class and method
3) refactor packets into structs
4) use of mysql stored procedure to avoid hard-coding sql command to decouple the program and database
5) use of modern C++

test note:
you must execute the sql procedure defined in main.sql before test

some play test:
test build last commit id 95316736b5dec28c5f7441d1c806e8d7e996c9f8
test packet version 20021-11-03
client version 2021-11-03
1. build on ubuntu 22.04.2 LTS ok
2. server init ok
3. join clan ok
4. leave clan ok
5. view clan info ok

Hi rAthena community, please let me know if pull request like this is helpful to rAthena development.
Any feedback would be appreciated for I'd be able to understand the development direction and see what I can contribute.

### DIFF
--- a/sql-files/main.sql
+++ b/sql-files/main.sql
@@ -364,6 +364,32 @@ INSERT INTO `clan` VALUES ('2', 'Arcwand Clan', 'Devon Aire', 'geffen', '500');
 INSERT INTO `clan` VALUES ('3', 'Golden Mace Clan', 'Berman Aire', 'prontera', '500');
 INSERT INTO `clan` VALUES ('4', 'Crossbow Clan', 'Shaam Rumi', 'payon', '500');
 
+DELIMITER //
+DROP PROCEDURE IF EXISTS GetClanAndClanAlliance //
+CREATE PROCEDURE GetClanAndClanAlliance()
+BEGIN
+    SELECT 
+        c.clan_id, 
+        c.name AS clan_name, 
+        c.master, 
+        c.mapname, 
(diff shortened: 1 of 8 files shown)


Lemongrass3110:
Hello!

First of all thank your for trying to contribute!

Since you asked for general feedback:
Your PR has some good ideas in it.
However even for a team member it would be changing quite a lot and we try to prevent that as much as possible.
If you want to contribute please focus on a single feature you want to change/touch in each PR.
And with feature I am not talking about "the clan feature" like here, but more granular tasks such as:

- Converting to C++
- Adding hooks
- Changing inter-server communication to structs
- Others...

(Since you asked about development direction: all the points mentioned above are on the plan and we are slowly but steadily getting there)

Some ideas I dislike like the use of stored procedures (in the way you use them).

So sorry for having to turn down your PR this time, but please feel free to open it again - if you split it into smaller pieces so that it can seriously be reviewed and tested by a team member.

Thank you for understanding!

Issue #8748, opened 2024-10-23
Adding Systemd Unit files


astroty:
### Is your request related to a missing official feature?

No

### Describe the solution you'd like

The project offering systemd units for use on a SystemD Linux host

### Additional context

This would allow the host to fully control the processes, bypassing the need for athena-start and giving control to default system tools. This would allow it to utilize Journalctl, systemd-coredump, restart on service fail, etc while it all being taken care of by systemd, freeing maintenance burden.

PR #8749, opened 2024-10-23
Systemd unit files

astroty:
Created systemd unit files for use with Linux SystemD. Hopefully it's useful.

* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/8748

* **Server Mode**: 

Both

Unit files currently being used on my custom server.

### DIFF
--- a/tools/systemd-units/char-rathena.service
+++ b/tools/systemd-units/char-rathena.service
@@ -1,7 +1,6 @@
 [Unit]
 Description=rAthena Character Server
 After=network-online.target remote-fs.target nss-lookup.target
-Wants=network-online.target
 
 [Service]
 Type=single


tools/systemd-units/char-rathena.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Rathena Char Server
+After=network-online.target remote-fs.target nss-lookup.target
+Wants=network-online.target
+#You might wanna edit the next line if your units are running in different servers instead of all in a single server.
+Requires=map-rathena.service login-rathena.service
```
Lemongrass3110, line 6:
```suggestion
```
I think we should remove these dependencies. All of them should run independent of each other and only if installed.

tools/systemd-units/char-rathena.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Rathena Char Server
```
Lemongrass3110, line 2:
```suggestion
Description=rAthena Character Server
```

tools/systemd-units/char-rathena.service
```diff
@@ -0,0 +1,18 @@
...
+After=network-online.target remote-fs.target nss-lookup.target
```
Lemongrass3110, line 3:
```suggestion
After=network-online.target remote-fs.target nss-lookup.target mysql.service
```

tools/systemd-units/char-rathena.service
```diff
@@ -0,0 +1,18 @@
...
+Wants=network-online.target
```
Lemongrass3110, line 4:
```suggestion
```
I dont see any benefit in defining this dependency (in addition to the startup order).

tools/systemd-units/char-rathena.service
```diff
@@ -0,0 +1,18 @@
...
+#You might wanna edit the next line if your units are running in different servers instead of all in a single server.
+Requires=map-rathena.service login-rathena.service
+
+[Service]
+Type=single
+PIDFile=/run/char-serv.pid
+User=rathena
+ExecStart=${replacewithpath}/char-server
```
Lemongrass3110, line 12:
```suggestion
ExecStart=~/rathena/char-server
```
Since we specify a username as well, we can also specify the path inside his home directory or not?

Lemongrass3110 (commented):
Some initial comments and looking forward to your feedback.
If you think my input is valid, then all the comments should also be applied to login-server and map-server.

web-server is currently missing and would need to be added, before we can merge it.

tools/systemd-units/char-rathena.service
Lemongrass3110: Thats already defined by "After". Is it not?
"Wants" does not give any additional benefit, since even if starting the network fails, it will still try to start the unit.

tools/systemd-units/char-rathena.service
Lemongrass3110: Okay that is a good reason. Then we should do the same with `User=${replacewithyourrathenausername}` or something like this.

tools/systemd-units/char-rathena.service
Lemongrass3110: Sounds fine to me.

vstumpf:
Due to the amount of customizations required per user (path, user, whether or not they want to wait for their sql server), I think this would be better as a page in https://github.com/rathena/user-guides

What do you think?

Singe-Horizontal:
The logging system could also be updated to be compatible with journald.

tools/systemd-units/login-rathena.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Rathena Login Server
+After=network-online.target remote-fs.target nss-lookup.target
+Wants=network-online.target
+#You might wanna edit the next line if your units are running in different servers instead of all in a single server.
+Requires=char-rathena.service map-rathena.service
+
+[Service]  
+Type=single
```
lighta, line 9: It may be just me, but I really like to setup some quota to my process to ensure the host still responsive.
Now it may depend load but I think login server is pretty small. 

I'd add those in [Service] section, (require systemd >231 tough)
MemoryHigh=400M
MemoryMax=500M

PR #8759, opened 2024-10-27
Show party name

Cypall:
Add Document Tutorial.
How to add party name after player name ?
screenAthena000


Haydrich:
It's already showing the party name by default since #8711

PR #8778, opened 2024-11-03
Monster behavior when hiding

### LINKED ISSUE #8774: Monster AI when hiding
### rAthena Hash

5cc4a6f

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When a monster hides while its target is still in attack range, it stays in "Berserk" mode, but will be unable to use any skills.

### Relevant Log Output

_No response_

### Expected Result

When a monster hides while its target is still in attack range, it stays in "Berserk" mode and will check for skill usage every server interval (20ms). Attack motion delay is only applied when a skill is actually used.

When a monster searches for a new target and finds one and that new target is already in attack range, the monster switches to "Berserk" mode. This applies even to hidden monsters. So a way to switch an Idle hidden monster to Berserk would be to either unhide, revive or snap (while previously outside aggro range) right next to it.
This also can happen when an angry monster switches targets on "Follow".

### How to Reproduce

Can test on official servers with Martin, Obsidian and Golden Thief Bug for example (see official information).

For the follow test you can just lure an angry monster over another player. When it switches to that player while the player is in attack range it will also start using berserk skills.

### Official Information

Martin - Passive monster, keeping berserk, you can see that it's using berserk state skills while hidden and at a very fast rate.
Obsidian - Wait until it hides, snap or unhide next to it, you will see it actually start using berserk state skills, even though it was never hit at all.
Golden Thief Bug - If it hides and you stay next to it, it will not heal, but spam its berserk state skills instead. Only after you move outside its attack range, it starts healing.

_No response_

Playtester:
* **Addressed Issue(s)**: #8774 

* **Server Mode**: Both

- Monsters will now cast skills in berserk state when they are hiding and a target is in attack range
- Monsters will now switch to berserk state (losing angry mode) when they search for a target and find a new one that is already in attack range
- Fixes #8774

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1862,7 +1862,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 		map_foreachinallrange (mob_ai_sub_hard_activesearch, &md->bl, view_range, DEFAULT_ENEMY_TYPE(md), md, &tbl, mode);
 		// If a monster finds a target through search that is already in attack range it immediately switches to berserk mode
 		// This behavior overrides even angry mode and other mode-specific behavior
-		if (tbl && prev_id != md->target_id && battle_check_range(&md->bl, tbl, md->status.rhw.range)) {
+		if (tbl != nullptr && prev_id != md->target_id && battle_check_range(&md->bl, tbl, md->status.rhw.range)) {
 			md->state.aggressive = 0;
 			md->state.skillstate = MSS_BERSERK;
 		}
@@ -1982,7 +1982,8 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 				else { // Attack once and find a new random target
 					int search_size = (view_range < md->status.rhw.range) ? view_range : md->status.rhw.range;
 					unit_attack(&md->bl, tbl->id, 0);
-					if ((tbl = battle_getenemy(&md->bl, DEFAULT_ENEMY_TYPE(md), search_size))) {
+					tbl = battle_getenemy(&md->bl, DEFAULT_ENEMY_TYPE(md), search_size);
+					if (tbl != nullptr) {
 						md->target_id = tbl->id;
 						md->min_chase = md->db->range3;
 					}
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1858,7 +1858,14 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 
 	if ((mode&MD_AGGRESSIVE && (!tbl || slave_lost_target)) || md->state.skillstate == MSS_FOLLOW)
 	{
+		int prev_id = md->target_id;
 		map_foreachinallrange (mob_ai_sub_hard_activesearch, &md->bl, view_range, DEFAULT_ENEMY_TYPE(md), md, &tbl, mode);
+		// If a monster finds a target through search that is already in attack range it immediately switches to berserk mode
+		// This behavior overrides even angry mode and other mode-specific behavior
+		if (tbl && prev_id != md->target_id && battle_check_range(&md->bl, tbl, md->status.rhw.range)) {
+			md->state.aggressive = 0;
+			md->state.skillstate = MSS_BERSERK;
+		}
 	}
 	else
 	if (mode&MD_CHANGECHASE && (md->state.skillstate == MSS_RUSH || md->state.skillstate == MSS_FOLLOW))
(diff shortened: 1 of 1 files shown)


aleos89 (approved):

src/map/mob.cpp
```diff
@@ -1858,7 +1858,14 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 
 	if ((mode&MD_AGGRESSIVE && (!tbl || slave_lost_target)) || md->state.skillstate == MSS_FOLLOW)
 	{
+		int prev_id = md->target_id;
```
Lemongrass3110, line 1861:
```suggestion
		int32 prev_id = md->target_id;
```

src/map/mob.cpp
```diff
@@ -1858,7 +1858,14 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
 		map_foreachinallrange (mob_ai_sub_hard_activesearch, &md->bl, view_range, DEFAULT_ENEMY_TYPE(md), md, &tbl, mode);
+		// If a monster finds a target through search that is already in attack range it immediately switches to berserk mode
+		// This behavior overrides even angry mode and other mode-specific behavior
+		if (tbl && prev_id != md->target_id && battle_check_range(&md->bl, tbl, md->status.rhw.range)) {
```
Lemongrass3110, line 1865:
```suggestion
		if (tbl != nullptr && prev_id != md->target_id && battle_check_range(&md->bl, tbl, md->status.rhw.range)) {
```

src/map/mob.cpp
```diff
@@ -1961,24 +1968,36 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
-
-			if( !(mode&MD_RANDOMTARGET) )
-				unit_attack(&md->bl,tbl->id,1);
-			else { // Attack once and find a new random target
-				int search_size = (view_range < md->status.rhw.range) ? view_range : md->status.rhw.range;
-				unit_attack(&md->bl, tbl->id, 0);
-				if ((tbl = battle_getenemy(&md->bl, DEFAULT_ENEMY_TYPE(md), search_size))) {
-					md->target_id = tbl->id;
-					md->min_chase = md->db->range3;
+	if (battle_check_range(&md->bl, tbl, md->status.rhw.range))
```
Lemongrass3110, line 1971:
```suggestion
	// Check if the target is within range
	if (battle_check_range(&md->bl, tbl, md->status.rhw.range))
```

src/map/mob.cpp
```diff
@@ -1961,24 +1968,36 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
+	{
+		if (!(md->sc.option&OPTION_HIDE)) {
+			// Target within range and able to use normal attack, engage
```
Lemongrass3110, line 1974:
```suggestion
			// Only attack if no more attack delay left
```

src/map/mob.cpp
```diff
@@ -1961,24 +1968,36 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
+			if (md->ud.target != tbl->id || md->ud.attacktimer == INVALID_TIMER)
+			{ //Only attack if no more attack delay left
```
Lemongrass3110, line 1976:
```suggestion
			if (md->ud.target != tbl->id || md->ud.attacktimer == INVALID_TIMER){
```

src/map/mob.cpp
```diff
@@ -1961,24 +1968,36 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
+				if (tbl->type == BL_PC)
+					mob_log_damage(md, tbl, 0); //Log interaction (counts as 'attacker' for the exp bonus)
+
+				if (!(mode&MD_RANDOMTARGET))
+					unit_attack(&md->bl,tbl->id,1);
+				else { // Attack once and find a new random target
+					int search_size = (view_range < md->status.rhw.range) ? view_range : md->status.rhw.range;
```
Lemongrass3110, line 1983:
```suggestion
					int32 search_size = (view_range < md->status.rhw.range) ? view_range : md->status.rhw.range;
```

src/map/mob.cpp
```diff
@@ -1961,24 +1968,36 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
+					unit_attack(&md->bl, tbl->id, 0);
+					if ((tbl = battle_getenemy(&md->bl, DEFAULT_ENEMY_TYPE(md), search_size))) {
```
Lemongrass3110, line 1985:
```suggestion
					unit_attack(&md->bl, tbl->id, 0);
					tbl = battle_getenemy(&md->bl, DEFAULT_ENEMY_TYPE(md), search_size);
					if (tbl != nullptr) {
```

src/map/mob.cpp
```diff
@@ -1961,24 +1968,36 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
+						md->target_id = tbl->id;
+						md->min_chase = md->db->range3;
+					}
```
Lemongrass3110, line 1988: Dont we have to reset the target in the else case?

src/map/mob.cpp
```diff
@@ -1961,24 +1968,36 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
 				}
 			}
 		}
+		else if (md->state.skillstate == MSS_BERSERK && DIFF_TICK(md->ud.attackabletime, tick) <= 0) {
```
Lemongrass3110, line 1992: Here you use `DIFF_TICK(md->ud.attackabletime, tick)` for checking. Above the same comment is used for checking via `md->ud.attacktimer == INVALID_TIMER`

Lemongrass3110 (approved):

src/map/mob.cpp
Playtester: This is a special case because a hidden monster won't do normal attack and so it will never have an attack timer.

src/map/mob.cpp
Playtester: Unrelated to the current update and don't want to look into this.

src/map/mob.cpp
Playtester: md->target_id is of type "int", do you really want to add an implicit cast here?

src/map/mob.cpp
Playtester: The purpose of the comment is to explain the situation under which you get in here, not to comment on the next check. It needs to be clear that this branch is for situations where a monster can do normal attacks. This is so that if any new state is added that prevents monsters from doing normal attacks, it is easy to see it would need to be added into this branch. We shouldn't remove it.

src/map/mob.cpp
Playtester: Unrelated to current change. 

view_range is "int" and md->status.rhw.range is "unsigned short", the search_size variable is also declared in another place in this function as "int". Wouldn't it be better to rework this separately? Maybe even defining view_range differently in the first place.

src/map/mob.cpp
Playtester: Unrelated to the current change, but I don't have any concerns with this, so I'll add it.

src/map/mob.cpp
Lemongrass3110: Ah yeah. Overlooked that this was not added now and only indented.

src/map/mob.cpp
Lemongrass3110: Yes we should revisit all the code so that the same types are used.
But if you want to define is as "int" here, please use "int32" for the time being. It can still be adjusted to uint16 later on.

src/map/mob.cpp
Lemongrass3110: But you do not know, if you are able to use a normal attack (yet). That is only decided after a few checks below...
So the comment gives of a wrong impression.

Playtester: Nono, the check if the monster can do normal attacks is this:
`if (!(md->sc.option&OPTION_HIDE)) {`

Right now it only checks for Hide as we only know this situation (the check was already there I just moved it out of the main check so we can have a proper "else" added here.

Issue #8784, opened 2024-11-07
 Incorrect magic damage with SC_MAGIC_POISON debuff (comet debuff)


Atemo:
### rAthena Hash

https://github.com/rathena/rathena/commit/0a0708491b3499430d50ff1cbc53ee0f51da218f

### Client Date

2022-03-30

### Server Mode

Renewal

### Result

WL_COMET inflicts SC_MAGIC_POISON to the targets that reduces resistance to all property by 50%.
The damage on target with SC_MAGIC_POISON is currently different from the official value.

### Expected Result

Same damage on rAthena and official server.

### How to Reproduce

See below.

### Official Information

Character :

**BaseLevel** 263
**INT** 130+15
**DEX** 1+8
**LUK** 100+4
**SPL** 110+14
**CON** 31+9
**Base S.MATK** 49

**Equipments** :
+10C Glacier Shield + enchants 
	magic_addsize +3%
	magic_addele +3%
	magic_addclass +3%
	MATK + 30

**Dummy (Neutral Lv1)**	ID: [21077](https://www.divine-pride.net/database/monster/21077)
**Level** 100
**INT** 0
**Hard Mdef** = 0
Target Soft Mdef (**Mdef2**) = `(float)(status->int_ + level) / 4` = 25

Character Matk minimum :
```
status_base_matk_min : status->int_ + (status->int_ / 2) + (status->dex / 5) + (status->luk / 3) + (level / 4) + 5 * status->spl;
status_base_matk_min : 937
```

SC_RECOGNIZEDSPELL: status_base_matk_min = status_base_matk_max

AG_DESTRUCTIVE_HURRICANE `skillratio` :
```
AG_DESTRUCTIVE_HURRICANE level 5: skillratio = (600 + 2850 * skill_lv + 5 * sstatus->spl) * BaseLevel / 100;
AG_DESTRUCTIVE_HURRICANE level 5: skillratio = 40686
```

Test
-------------------------------------------------------

AG_DESTRUCTIVE_HURRICANE without comet debuff 
```
Equip                      Official server    RA
Nude                       567951             567951
Shield                     671294             671294
```

AG_DESTRUCTIVE_HURRICANE with comet debuff 
```
Equip                      Official server    RA
Nude                       851532             851926
Shield                     1007360            1006941
```

By moving the debuff in battle_calc_cardfix
-------------------------------------------------------

Nude
```
damage = status_base_matk_min = 937
[...]

PR #8785, opened 2024-11-07
Corrected magic damage with SC_MAGIC_POISON debuff (comet debuff)

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8784

* **Server Mode**: renewal

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -756,7 +756,7 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 				}
 			}
 			// Statuses that affect the target's element and should be calculated right after magic_addele, independently of it
-			if (tsc && !nk[NK_IGNOREDEFCARD] && !nk[NK_IGNOREELEMENT]) {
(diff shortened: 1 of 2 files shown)


src/map/battle.cpp
```diff
@@ -805,6 +808,11 @@ int battle_calc_cardfix(int attack_type, struct block_list *src, struct block_li
 					}
 					if (s_defele != ELE_NONE)
 						ele_fix += tsd->indexed_bonus.magic_subdefele[s_defele] + tsd->indexed_bonus.magic_subdefele[ELE_ALL];
+#ifndef RENEWAL
```
aleos89, line 811: Do we care about adding support to pre-renewal when the end goal is to remove renewal skills from the pre-re databases?

Atemo: That's from the original legacy. Skills and SCs are common by default in src. Even if there is some weird behavior like in this PR, separating pre-re and re in src will make things harder for people who mix the modes. Should we change our way and separate them like the DB?

PR #8786, opened 2024-11-07
Added script support to status changes

Lemongrass3110:

* **Server Mode**: Both

Adds the ability to script status changes for players

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -4092,11 +4092,21 @@ int status_calc_pc_sub(map_session_data* sd, uint8 opt)
 		current_equip_opt_index = -1;
 	}
 
-	if (sc->count && sc->getSCE(SC_ITEMSCRIPT)) {
-		std::shared_ptr<item_data> data = item_db.find(sc->getSCE(SC_ITEMSCRIPT)->val1);
+	if (sc->count){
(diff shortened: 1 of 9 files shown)


src/map/status.cpp
```diff
@@ -10171,6 +10181,11 @@ int status_change_start(struct block_list* src, struct block_list* bl,enum sc_ty
 	if( !sc )
 		return 0; // Unable to receive status changes
 
+	// Scripted status changes only work for players for the time being
+	if( scdb->script != nullptr && bl->type != BL_PC ){
```
Atemo, line 10185: A debug/warning would be welcome to track potential mistake

Lemongrass3110, line 10185:
```suggestion
	if( scdb->script != nullptr && bl->type != BL_PC ){
		ShowError( "status_change_start: Trying to start the scripted status change %d on a non player.\n", type );
```
Like this?

Atemo: Yes, maybe "failed" would be more explicit than "trying" tho

aleos89 (approved):
Looks good! Just missing doc/status.txt addition for the new parameter!

Atemo (approved):

aleos89 (approved):

PR #8800, opened 2024-11-09
Star Gladiator Heat Official Behavior

### LINKED ISSUE #8799: Star Gladiator Heat Official Behavior
### rAthena Hash

2fff4c5

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

Interval of Heat is 100ms (10 hits per second, 20 SP cost per second).
Heat drains 15 SP every 100ms on players (150 SP per second).
Heat works on bosses to 100% (damages exactly once per 100ms).
When the skill misses, it drains 10 SP instead of 2 SP.
Casting Heat while it's active costs 20 SP. It plays a sound when recast.
Heat has 1000ms aftercast delay in pre-renewal.
Knockback is 2 cells.

### Relevant Log Output

_No response_

### Expected Result

Interval of Heat is 20ms (50 hits per second, 100 SP cost per second).
Heat drains 10 SP every 20ms on players (500 SP per second).
Heat works on bosses to 20% (damages in irregular intervals). If heat fails, it also does not cost SP.
There is no special penalty for having the skill miss, but if it misses, there is no knockback so it triggers again after 20ms.
Casting Heat while it's active does not cost any SP. It does not play a sound when recast.
Heat has no aftercast delay.
Knockback is 2~5 cells.

### How to Reproduce

Use Heat on:
- Normal monster against wall having 100% hit chance (50 hits per second)
- Normal monster not against wall having 5% hit chance (on average it drains 40 SP per hit, knockback is 2-5 cells)
- Boss monster (irregular hits and SP cost but on average 10 hits per second)
- Guardian monster (same as normal monster)
- Battlefield monster (costs 100 SP per second but does not deal any damage)
- Player (drains 500 SP per second, no cost)

Cast Heat while it's still active (no SP cost, duration will also not be extended).

### Official Information

Above was all tested and confirmed on official servers.

As for recasting Heat while still active (HEADER_ZC_USE_SKILL2).

First cast:
{
  "Skid": 430,
  "Amount": 3,
  "Targetaid": xxxxxxx,
  "Srcaid": xxxxxxx,
  "Result": true,
  "Command": 2507
}

Second cast:
{
  "Skid": 430,
  "Amount": 3,
  "Targetaid": xxxxxxx,
  "Srcaid": xxxxxxx,
  "Result": false,
  "Command": 2507
}

_No response_

Playtester:
* **Addressed Issue(s)**: #8799 

* **Server Mode**: Both

- Heat now has an interval of 20ms instead of 100ms
- Heat no longer has an aftercast delay
- Knockback of Heat is now a random value between 2 and 5 cells
- Heat now drains 10 SP on players per interval (500 SP per second)
- Heat now has an 80% chance to not activate on bosses per interval
- Heat no longer has an SP penalty when it misses
- Heat now fails when it's already active
- If Heat fails, you will now do the "T-Pose" animation without the effect activating instead of getting a skill failed message
- Fixed order of packets sent to the client (prevents position lag)
- Fixes #8799

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -7230,6 +7230,12 @@ static struct Damage initialize_weapon_data(struct block_list *src, struct block
 			case TK_TURNKICK:
 				wd.blewcount = 0;
 				break;
+			case SG_SUN_WARM:
+			case SG_MOON_WARM:
+			case SG_STAR_WARM:
+				// Knockback is random
+				wd.blewcount += rnd_value(0, 3);
+				break;
 #ifdef RENEWAL
 			case KN_BOWLINGBASH:
 				if (sd && sd->status.weapon == W_2HSWORD) {
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -7233,7 +7233,7 @@ static struct Damage initialize_weapon_data(struct block_list *src, struct block
 			case SG_SUN_WARM:
 			case SG_MOON_WARM:
 			case SG_STAR_WARM:
-				// Knockback is random
+				// A random 0~3 knockback bonus is added to the base knockback
 				wd.blewcount += rnd_value(0, 3);
 				break;
 #ifdef RENEWAL
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -16107,14 +16107,17 @@ int skill_unit_onplace_timer(struct skill_unit *unit, struct block_list *bl, t_t
 					//If target isn't knocked back it should hit every "interval" ms [Playtester]
 					do {
 						if( bl->type == BL_PC )
-							status_zap(bl, 0, 15); // sp damage to players
-						else // mobs
-						if( status_charge(ss, 0, 2) ) { // costs 2 SP per hit
-							if( !skill_attack(BF_WEAPON,ss,&unit->bl,bl,sg->skill_id,sg->skill_lv,tick+(t_tick)count*sg->interval,0) )
-								status_charge(ss, 0, 8); //costs additional 8 SP if miss
-						} else { //should end when out of sp.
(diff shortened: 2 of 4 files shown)


src/map/battle.cpp
```diff
@@ -7230,6 +7230,12 @@ static struct Damage initialize_weapon_data(struct block_list *src, struct block
 			case TK_TURNKICK:
 				wd.blewcount = 0;
 				break;
+			case SG_SUN_WARM:
+			case SG_MOON_WARM:
+			case SG_STAR_WARM:
+				// Knockback is random
```
Daegaladh, line 7236: It would be better to clarify that there is a fixed part on the db and a random part here

src/map/skill.cpp
```diff
@@ -16107,14 +16107,17 @@ int skill_unit_onplace_timer(struct skill_unit *unit, struct block_list *bl, t_t
...
-						if( status_charge(ss, 0, 2) ) { // costs 2 SP per hit
-							if( !skill_attack(BF_WEAPON,ss,&unit->bl,bl,sg->skill_id,sg->skill_lv,tick+(t_tick)count*sg->interval,0) )
-								status_charge(ss, 0, 8); //costs additional 8 SP if miss
-						} else { //should end when out of sp.
-							sg->limit = DIFF_TICK(tick,sg->tick);
-							break;
+							status_zap(bl, 0, 10); // sp damage to players
+						else { // mobs
+							// Bosses have a 80% chance to not trigger the effect
+							if (status_get_class_(bl) == CLASS_BOSS && rnd_chance(4, 5))
```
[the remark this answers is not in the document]
Playtester: Will improve comment instead.

src/map/battle.cpp
Playtester: Comment improved.

PR #8818, opened 2024-11-12
Fixing variable "chan_banentry" is not a type name error

SapitoSucio:
* **Server Mode**: Both

Fixes the external declaration of chan_banentry by adding the missing 'struct' keyword, as chan_banentry is a structure and not a type.

Changes made:
- **From**: extern chan_banentry chan_banentry;
- **To**: extern struct chan_banentry chan_banentry;

This fixes the compiler error "variable 'chan_banentry' is not a type name", which occurred because the compiler couldn't interpret chan_banentry as a type without the struct keyword.

### DIFF
--- a/src/map/channel.hpp
+++ b/src/map/channel.hpp
@@ -59,7 +59,7 @@ struct chan_banentry {
 	uint32 char_id;
 	char char_name[NAME_LENGTH];
 };
-extern chan_banentry chan_banentry;
+extern struct chan_banentry chan_banentry;
 
 struct Channel_Config {
 	unsigned long *colors;		///< List of available colors


Lemongrass3110:
Which OS? Which OS version?
Which compiler? Which compiler version?

Can you please give us a screenshot of the error.

Thank you in advance!

Lemongrass3110:
https://github.com/rathena/rathena/actions/runs/11803806427/job/32882592506

```
g++-11 is already the newest version (11.4.0-1ubuntu1~22.04).
g++-11 set to manually installed.
gcc-11 is already the newest version (11.4.0-1ubuntu1~22.04).
gcc-11 set to manually installed.
```

Our CI contains no errors.

Are you sure that this is not just a visual glitch/recommendation of the IDE that you are using?

Daegaladh:
I've tested on ubuntu 22.04/gcc 11.4 and don't even get any warnings on compiling

Lemongrass3110:
Sorry but this is just a (invalid) visual warning of MSVS IntelliSense and not an error of the MSVC Compiler.
Everything is defined (a few lines above) as intended and allowed in C++.

PR #8834, opened 2024-11-17
Changed SC array to STL

Lemongrass3110:

* **Server Mode**: Both

Decreased RAM usage for status changes from 7.832 bytes to 120 bytes, so by 7.712 bytes and therefore by 98,47%
Decreased server start RAM usage on RENEWAL from 646.072.000 bytes to 456.664.000 bytes, so by 189.408.000 bytes and therefore by 29,32% (measurements taken on Win10 x64)
Decreased server start RAM usage on PRE-RENEWAL from 432.724.000 bytes to 319.252.000 bytes, so by 113.472.000 bytes and therefore by 26,22% (measurements taken on Win10 x64)
During server runtime this means a decreased RAM usage by 98,47% for each affected block_list

Improved speed for saving active status changes
Improved speed for executing scripts of active status changes
Removed ERI use for status changes
Removed clearSCE to not interfere with the STL containers memory management
Removed sce variable reuse in status_change_end
Removed unnecessary sce exposure for SC_CLOSECONFINE/SC_CLOSECONFINE2/SC_CURSEDCIRCLE_ATKER
Fixed a bug in SC_STOP that caused the target's status to be recalculated instead of the source's (because of variable reuse)

Thanks to @vstumpf for the preparations!


datawulf:
> I encounter a crash when script using attcomand "@some_gm_command", it will create a temporary map_session_data* sd in `int32 atcommand_sub(struct script_state* st,int32 type)` and then in `bool is_atcommand(const int32 fd, map_session_data* sd, const char* message, int32 type)` section
> 
> ```
> 	//Block NOCHAT but do not display it as a normal message
> 	if ( sd->sc.getSCE(SC_NOCHAT) && sd->sc.getSCE(SC_NOCHAT)->val1&MANNER_NOCOMMAND )
> 		return true;
> ```
> 
> this sd doesn't have correct sc, and leaks to crash.

do you have an example which leads to your crash? I'm just testing this, too. But I can't reproduce what you said.

inhyositsu:
> > I encounter a crash when script using attcomand "@some_gm_command", it will create a temporary map_session_data* sd in `int32 atcommand_sub(struct script_state* st,int32 type)` and then in `bool is_atcommand(const int32 fd, map_session_data* sd, const char* message, int32 type)` section
> > ```
> > 	//Block NOCHAT but do not display it as a normal message
> > 	if ( sd->sc.getSCE(SC_NOCHAT) && sd->sc.getSCE(SC_NOCHAT)->val1&MANNER_NOCOMMAND )
> > 		return true;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this sd doesn't have correct sc, and leaks to crash.
> 
> @inhyositsu could it be that your rAthena is not up to date? It may be that you are missing [83c093f](https://github.com/rathena/rathena/commit/83c093f1b8f41e548ea967a344b17ac9325423d1) and [0486cf1](https://github.com/rathena/rathena/commit/0486cf14c88f0fb098773ba1c680dacf8a1ebb49)

Yes, sorry!!!! I'm using rathena at [f3c177699d24f0eab94af8a475b806b1fe80f73f] version

src/map/status.cpp
```diff
@@ -1140,12 +1139,26 @@ void StatusDatabase::removeByStatusFlag(block_list *bl, std::vector<e_status_cha
 	}
 }
 
+status_change_entry::status_change_entry(){
+	this->timer = INVALID_TIMER;
```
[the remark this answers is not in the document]
Lemongrass3110: No.

src/map/status.hpp
```diff
@@ -3327,6 +3327,9 @@ struct sc_display_entry {
 struct status_change_entry {
 	int32 timer;
```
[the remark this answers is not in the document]
Lemongrass3110: No.

src/map/status.cpp
```diff
@@ -1155,51 +1168,63 @@ status_change::status_change(){
...
 bool status_change::empty(){
-	return this->count == 0;
+	return this->data.empty();
+}
+
+size_t status_change::size(){
+	return this->data.size();
+}
+
+std::unordered_map<enum sc_type, status_change_entry>::iterator status_change::begin(){
```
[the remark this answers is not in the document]
Lemongrass3110: Changed in ff168a5
Thank you!

src/map/chrif.cpp
```diff
@@ -1286,12 +1286,11 @@ int32 chrif_save_scdata(map_session_data *sd) { //parses the sc_data of the play
 	WFIFOL(char_fd,4) = sd->status.account_id;
 	WFIFOL(char_fd,8) = sd->status.char_id;
 
-	for (i = 0; i < SC_MAX; i++) {
-		auto sce = sc->getSCE(static_cast<sc_type>(i));
-		if (!sce)
-			continue;
-		if (sce->timer != INVALID_TIMER) {
-			timer = get_timer(sce->timer);
+	for( const auto& it : *sc ){
```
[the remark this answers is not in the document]
Lemongrass3110: Changed in b972b86
Thank you!

Playtester (commented):
I tested and when a monster uses NPC_SUICIDE, the server crashes. It works fine in master.

Playtester:
Assuming it really fixes the issue, it would be nice if you could mention that this fixes #7777 in the description and final commit message.

src/map/status.cpp
```diff
@@ -5461,141 +5484,289 @@ void status_calc_regen_rate(struct block_list *bl, struct regen_data *regen, sta
...
+				case SC_HIDING:
+					if( bl.type != BL_PC ){
+						restriction = true;
+					}else if( pc_checkskill( (map_session_data*)( &bl ), RG_TUNNELDRIVE ) <= 0 ){
+						restriction = true;
+					}
+					break;
+
+				case SC_DANCING:
+					if( sce.val4 != 0 && ( ( sce.val1 & 0xFFFF ) == CG_MOONLIT || ( sce.val1 & 0xFFFF ) == CG_HERMODE ) ){
```
Playtester, line 5613: There must be something going wrong with the conversion of "val" values. I noticed that Dancing no longer stops you. When debugging I can see that sce.val4 is set to 65536 instead of 0.

src/map/status.cpp
```diff
@@ -5461,141 +5484,289 @@ void status_calc_regen_rate(struct block_list *bl, struct regen_data *regen, sta
...
+}
+
 /**
  * Applies a state to a unit - See [StatusChangeStateTable]
  * @param bl: Object to change state on [PC|MOB|HOM|MER|ELEM]
  * @param sc: Object's status change data
- * @param flag: Which state to apply to bl
- * @param start: (1) start state, (0) remove state
+ * @param scdb: Database information of the status change
+ * @param start: (true) start state, (false) remove state
```
Playtester, line 5545: It would be nice to have an explanation on the return value.

src/map/status.cpp
```diff
@@ -13338,33 +13519,31 @@ int32 status_change_end(struct block_list* bl, enum sc_type type, int32 tid)
...
-			sc_start4(bl, bl, SC_REGENERATION, 100, 10,0,0,(RGN_HP|RGN_SP), skill_get_time(LK_BERSERK, sce->val1));
+			sc_start4(bl, bl, SC_REGENERATION, 100, 10,0,0,(RGN_HP|RGN_SP), skill_get_time(LK_BERSERK, val1));
 			break;
 		case SC_GOSPEL:
-			if (sce->val3) { // Clear the group.
-				std::shared_ptr<s_skill_unit_group> group = skill_id2group(sce->val3);
+			if (val3) { // Clear the group.
+				std::shared_ptr<s_skill_unit_group> group = skill_id2group(val3);
 
-				sce->val3 = 0;
```
Playtester, line 13353: Isn't this a prevention to ensure it's not deleted again? Is it okay to remove it?

Lemongrass3110: val3 is now a local value.
It contains the copy of an already [really] deleted status change entry.
This already deleted status change entry can not be deleted again.
Also except this check, I dont see any val3 checks for SC_GOSPEL.
If there is a generic val3 check that does this, we would need to look deeper into that.

src/map/status.cpp
Lemongrass3110: Fixed in c94e620
Thank you!

Playtester (approved):
No further critical issues found. I did test SC_DANCING (Ensembles), SC_STOP, Close Confine, Freeze/Stone from items, self Freeze/Stone from items and NPC_SUICIDE.

Issue #8835, opened 2024-11-18
Inaccurate calculation that ignores mdef


Atemo:
### rAthena Hash

https://github.com/rathena/rathena/commit/06358cffae406d75bdb6f0b3e1b8a229ca4b2793

### Client Date

2022-03-30

### Server Mode

Renewal

### Result

The calculation that ignores mdef is inaccurate on renewal (unknown on pre-renewal).

### Expected Result

Same damage on rAthena and official server.

### How to Reproduce

See below.

### Official Information

Character :
-----------------------------------------
**BaseLevel** 159
**INT** 130+21
**DEX** 1+17
**LUK** 1+12

**Item used** :
Almighty (ID 12883)
= Allstats + 10
= MATK + 30
Power Booster (ID 23898)
= MATK + 1%
= MATK + 30
Paradise Advanced Shadow Set (warlock)
= MATK + 25
= All property magical damage +15%
= Soul Expansion damage +15%
= Ignores magical defense rate of all race monsters (except players) by 50%

-----------------------------------------
**Character Matk (with buffs and equip)** :
```
status_base_matk : (272 + 85) + (272 + 85) * 1 / 100;
status_base_matk : 360
```

**Skill attack** :
WL_SOULEXPANSION `skillratio` :
```
WL_SOULEXPANSION level 5: skillratio = (1000 + skill_lv * 200 + sstatus->int_) * BaseLevel / 100;
WL_SOULEXPANSION level 5: skillratio = 3420
```

Target :
-----------------------------------------

**Poring**	ID: [1002](https://www.divine-pride.net/database/monster/1002)
**Level** 1
**INT** 0
**Hard Mdef** = 5
Target Soft Mdef (**Mdef2**) = `(float)(status->int_ + level) / 4` = 0

Test
-------------------------------------------------------
```
Equip                                    Official server           RA
nude                                     10964 (5482 per hit)      10964 (5482 per hit)
Equip + buff                             15994 (7997 per hit)      15852 (7926 per hit)
```

-------------------------------------------------------
The issue lies in the inaccurate calculation of ignored magic defense. With Hard mdef ignored by 3 instead of 2 :
```
damage = status_base_matk = 360
damage = damage + damage * magic_atkele / 100 = 360 + 360 * 15 / 100 = 414
damage = damage * skillratio / 100 = 414 * 3420 / 100 = 14158
[...]

PR #8836, opened 2024-11-18
Corrected calculation that ignores mdef

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8835

* **Server Mode**: renewal


Playtester:
I think we should test this in pre-renewal also, any good way to test this?

Issue #8839, opened 2024-11-20
Giant Fly Wing Bug


Miraakol:
### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

There are instances when using the Giant Fly Wing causes some party members to be teleported to different locations.

https://youtu.be/UGizBAW14Os

### Relevant Log Output

```Shell
N/A
```

### Expected Result

Party members should always be recalled to the same location where the party leader is teleported.

### How to Reproduce

- Create a party.
- Invite another player to join the party.
- Enter a dungeon together.
- Continuously spam the Giant Fly Wing, as shown in the video provided.

### Official Information

N/A

### Modifications that may affect results

- Added goldpc (from rathena branch)
- Added VIP fix char limit creation (also from rathena branch)

Daegaladh:
Not sure what might be causing it but I'll check it

Miraakol:
> Not sure what might be causing it but I'll check it Where did the other player go? Does it any error on the console?

Hello, @Daegaladh! There's no error on the console. However, both players were randomly teleported, which seems unusual.

PR #8841, opened 2024-11-20
Fixed Giant Fly Wing

Singe-Horizontal:
* **Addressed Issue(s)**: 
#8839

* **Server Mode**: 
Both

For case WARPPARTY_RANDOMALLAREA inside warpparty command, that is used by Giant Fly Wings, the coordinates searching logic is broken, almost always leading to failed attempts, thus warping party members onto the leader. However, when it does find valid coordinates, it can separate the player from the team.
This restores the wanted coordinate searching logic, warping party members in random cells around the leader.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -5892,8 +5892,8 @@ BUILDIN_FUNC(warpparty)
 			[[fallthrough]];
 		case WARPPARTY_RANDOMALLAREA:
 			if(!mapdata->getMapFlag(MF_NORETURN) && !mapdata->getMapFlag(MF_NOWARP) && pc_job_can_entermap((enum e_job)pl_sd->status.class_, m, pc_get_group_level(pl_sd))){
-				int nx = x;
-				int ny = y;
+				int32 nx = x;
+				int32 ny = y;
 				if (rx || ry) {
 					uint8 attempts = 10;
 
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -5892,24 +5892,23 @@ BUILDIN_FUNC(warpparty)
 			[[fallthrough]];
 		case WARPPARTY_RANDOMALLAREA:
 			if(!mapdata->getMapFlag(MF_NORETURN) && !mapdata->getMapFlag(MF_NOWARP) && pc_job_can_entermap((enum e_job)pl_sd->status.class_, m, pc_get_group_level(pl_sd))){
+				int nx = x;
+				int ny = y;
 				if (rx || ry) {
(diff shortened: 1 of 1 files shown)


Daegaladh:
That's not the official behavior, Giant Fly Wings uses RandomAll, not RandomAllArea, the other players must be teleported to the exact cell the leader did.

src/map/script.cpp
```diff
@@ -5892,24 +5892,23 @@ BUILDIN_FUNC(warpparty)
 			[[fallthrough]];
 		case WARPPARTY_RANDOMALLAREA:
 			if(!mapdata->getMapFlag(MF_NORETURN) && !mapdata->getMapFlag(MF_NOWARP) && pc_job_can_entermap((enum e_job)pl_sd->status.class_, m, pc_get_group_level(pl_sd))){
+				int nx = x;
+				int ny = y;
```
Daegaladh, line 5896:
```suggestion
				int32 nx = x;
				int32 ny = y;
```

Daegaladh (approved):

src/map/script.cpp
```diff
@@ -5892,24 +5892,23 @@ BUILDIN_FUNC(warpparty)
...
 					do {
-						nx = x0 + rnd_value(x0, x1);
-						ny = y0 + rnd_value(y0, y1);
+						nx = x + rnd_value(-rx, rx);
+						ny = y + rnd_value(-rx, rx);
 					} while ((--attempts) > 0 && !map_getcell(m, nx, ny, CELL_CHKPASS));
 
-					if (attempts != 0) { //Keep the original coordinates if fails to find a valid cell within the range
-						x = nx;
-						y = ny;
```
Daegaladh, line 5908: Ah so the problem was x and y were being overwritten and kept the last attempt instead of keeping the original coordinates

src/map/script.cpp
```diff
@@ -5892,24 +5892,23 @@ BUILDIN_FUNC(warpparty)
...
-					int32 x1 = x + rx, y1 = y + ry,
-						x0 = x - rx, y0 = y - ry,
-						nx, ny;
 					uint8 attempts = 10;
 
 					do {
-						nx = x0 + rnd_value(x0, x1);
-						ny = y0 + rnd_value(y0, y1);
+						nx = x + rnd_value(-rx, rx);
+						ny = y + rnd_value(-rx, rx);
```
Atemo, line 5902:
```suggestion
						ny = y + rnd_value(-ry, ry);
```

Atemo (approved):

PR #8849, opened 2024-11-24
Algorithm for opening group items in renewal

### LINKED ISSUE #8848: Opening Package Items
### rAthena Hash

97a124c

### Client Date

2021-11-03

### Server Mode

Renewal

### Result

When opening a package that uses the new renewal package system (e.g. Special_Box) we currently differentiate between "Must" and "Random" sub groups. Assuming you use the "getgroupitem" command, it will drop all items from the "Must" group and a random item from each "Random" group. The selection of the item from the "Random" group is not influenced by what items have already been received from the package.

Old package groups from pre-renewal times are defined as sub group 1.

### Relevant Log Output

_No response_

### Expected Result

There should be three sub groups "Must", "Random" and "Natural Random".
"Must" (group 0 on the client) -> You get all items from the group (already working)
"Random" (group 1-5 on the client) -> Each package has each item "rate" times in it. You get a random item that's still left in the package. Once received, the item is removed from the package. Only when all items have been taken out of the package (or the server restarts) the package is refilled.
"Natural Random" (group 6 on the client) -> A random item is received that's defined for the package, using `rate / total_rate` as chance for the item to appear. What items have already been received from the package has no influence on the chances for whoever opens it next (this is basically what we are using for all groups except 0 at the moment).

Old package groups from pre-renewal times should be defined as sub group 6.

### How to Reproduce

Get 1000 boxes of a package such as "Special_Box" that uses group 1-5 and open them all at once right after server restart (to make sure nobody else opened them yet).

You will see that you get the exact amounts specified as probability rather than it being random.

### Official Information

The packageitem.lua file on the client has this information.

We confirmed this behavior with an official test with both group 1-5 and group 6.

There's also a new group 7, which has not been tested yet.

_No response_

Playtester:
* **Addressed Issue(s)**: #8848 

* **Server Mode**: Renewal (mostly)

- Introduced a new item_group_db attribute "Algorithm" with three possible values: All, Random and SharedPool
  * The attribute now affects behavior of how an item is selected within the group instead of SubGroup
  * This affects the commands "getgroupitem", "getrandgroupitem" and "groupranditem"
  * Be aware that various commands use or default to SubGroup=1
- Implemented the "SharedPool" algorithm where rate is the amount of items in the group
  * When an item is retrieved from the group, it is removed from the group
  * Only when the group is completely empty or the server restarts, the group refills
  * This algorithm is used by many renewal item groups
- The algorithm "All" replaces the previous SubGroup=0 check
  * When using "getgroupitem" on a group with "All" as algorithm, you will receive all items from the group
  * When using "getrandgroupitem" on a group with "All" as algorithm, you will receive a random item from the group
  * When using "groupranditem" on a group with "All" as algorithm, it will return a random item from the group
- The algorithm "Random" is the previous algorithm that returns a random item using rate as chance
- Battle config "item_group_rate" now only applies to GROUP_ALGORITHM_DROP
  * This is accomplished by storing the adjusted rate in a separate variable
- Updated documentation accordingly
- Fixes #8848

### DIFF
--- a/db/pre-re/item_group_db.yml
+++ b/db/pre-re/item_group_db.yml
@@ -50,7 +50,7 @@ Header:
 Body:
   - Group: ACCESORY
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Ring
@@ -180,7 +180,7 @@ Body:
             Rate: 1
   - Group: BLEUBOX
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Milk
@@ -217,7 +217,7 @@ Body:
             Rate: 1
   - Group: BLUEBOX
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Red_Potion
@@ -3287,7 +3287,7 @@ Body:
             Rate: 1
   - Group: BOMB
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Apple_Bomb
@@ -3315,7 +3315,7 @@ Body:
             Rate: 1
   - Group: BOX
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Old_Blue_Box
@@ -3328,7 +3328,7 @@ Body:
             Rate: 1
   - Group: CANDY
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Candy
@@ -3338,7 +3338,7 @@ Body:
             Rate: 1
   - Group: CARDALBUM
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Poring_Card
@@ -4428,7 +4428,7 @@ Body:
             Rate: 1
   - Group: CASH_FOOD
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Str_Dish10_
@@ -4450,7 +4450,7 @@ Body:
             Rate: 1
   - Group: CHERISH_BOX
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Dog_Cap_
@@ -4463,7 +4463,7 @@ Body:
             Rate: 5
   - Group: CHERISH_BOX_ORI
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Ragnarok_Limited_Ed
@@ -4482,7 +4482,7 @@ Body:
             Rate: 2
   - Group: COOKIEBAG
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Candy
@@ -4534,7 +4534,7 @@ Body:
             Rate: 1
   - Group: EASTER_SCROLL
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: No_Recipient
@@ -4580,7 +4580,7 @@ Body:
             Rate: 1
   - Group: EGGBOY
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
         List:
           - Index: 0
             Item: Grape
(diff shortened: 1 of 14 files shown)


secretdataz (changes requested):
I don't like this implementation where assigning items to subgroup 6 implicitly assigns a different random algorithm (imitating aegis).  Please make an YAML database attribute that makes algorithm for each subgroup customizable.

```yaml
- Group: SOME_BOX
    SubGroups:
      - SubGroup: 1
        Algorithm: Random
        List:
             # ....
      - SubGroup: 2
        Algorithm: NaturalRandom
        List:
             # ....
```

src/tool/yamlupgrade.cpp
```diff
@@ -447,9 +446,18 @@ static bool upgrade_item_group_db( std::string file, const uint32 source_version
...
+					if( !it["SubGroup"].IsDefined() ){
+						ShowError( "Cannot upgrade automatically, SubGroup is missing." );
+						return false;
 					}
+					body << YAML::Key << "SubGroup" << YAML::Value << it["SubGroup"];
+
+					if (it["SubGroup"].as<uint16>() == 0)
+						body << YAML::Key << "Algorithm" << YAML::Value << "Must";
+					else if (it["SubGroup"].as<uint16>() == 6)
+						body << YAML::Key << "Algorithm" << YAML::Value << "NaturalRandom";
```
Playtester, line 458: This doesn't seem to match with the syntax you defined in the files.

src/map/itemdb.hpp
```diff
@@ -3009,9 +3009,11 @@ enum e_delay_consume : uint8 {
 };
 
 /// Enum for different ways to search an item group
-enum e_group_search_type : uint8 {
-	GROUP_SEARCH_BOX = 0, // Always return an item from the group, rate determines which item is more likely to be returned
-	GROUP_SEARCH_DROP = 1, // Pick one item from the group and check use rate as drop rate, on fail, do not return any item
+enum e_group_draws_type : uint8 {
+	GROUP_DRAWS_REPLACEMENT, // Draws with replacement. Always return an item from the group, rate determines which item is more likely to be returned
+	GROUP_DRAWS_DROP, // Draws with replacement. Pick one item from the group and check use rate as drop rate, on fail, do not return any item
```
Playtester, line 3014: I still suggest replacing the descriptions here.

src/map/itemdb.cpp
```diff
@@ -3029,21 +3095,20 @@ uint8 ItemGroupDatabase::pc_get_itemgroup( uint16 group_id, bool identify, map_s
...
-	}
-
-	// Get 1 'random' item from each subgroup
 	for (const auto &random : group->random) {
-		// Skip the 'must' group
-		if( random.first == 0 ){
-			continue;
+		switch( random.second->draws ) {
+			case GROUP_DRAWS_REPLACEMENT:
+				this->pc_get_itemgroup_sub( sd, identify, this->get_random_itemsubgroup( random.second, GROUP_DRAWS_REPLACEMENT ) );
```
Playtester, line 3101: Could have both cases together in one and just give random.second->draws into the function.

src/map/itemdb.cpp
```diff
@@ -2943,15 +2979,45 @@ std::shared_ptr<s_item_group_entry> ItemGroupDatabase::get_random_entry(uint16 g
...
+	if (search_type > GROUP_DRAWS_MUST) {
+		ShowError("get_random_entry: Invalid search_type %hu for group id %hu, sub group %hu.\n", search_type, group_id, sub_group);
+		return nullptr;
+	}
+
+	return this->get_random_itemsubgroup(group->random[sub_group], search_type);
+}
+
+std::shared_ptr<s_item_group_entry> ItemGroupDatabase::get_random_entry(uint16 group_id, uint8 sub_group) {
+	std::shared_ptr<s_item_group_db> group = this->find(group_id);
```
Playtester, line 2991: As mentioned before I'm again duplicating all this code. It should just be one function or one function calls the other.

But the problem here is also that you're changing how getrandgroupitem works as discussed on Discord.

db/pre-re/item_group_db.yml
```diff
@@ -45,12 +46,13 @@
...
   Type: ITEM_GROUP_DB
-  Version: 3
+  Version: 4
 
 Body:
   - Group: ACCESORY
     SubGroups:
-      - SubGroup: 1
+      - SubGroup: 6
+        Draws: NoReplacement
```
Playtester, line 55: All pre-re groups should use true random so it should be "Replacement" here, right? Maybe I'm still misunderstanding the definition of these fields, but to me calling the algorithm that always have the same chances "NoReplacement" and calling the algorithm that actually removes the items without replacing them "Replacement" seems like the opposite of what your intent was.

Atemo (approved):

secretdataz (approved):
lgtm

PR #8854, opened 2024-11-26
Updated item group DB format

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

Added "Draws" field to dehardcode the "must" subgroup associated with subgroup 0
Added "NoReplacement" for "Draws" field which will be used later
Upgraded version and tool to upgrade the file from previous version


db/pre-re/item_db_usable.yml
```diff
@@ -1061,7 +1061,7 @@ Body:
       BuyingStore: true
       Container: true
     Script: |
-      getrandgroupitem(IG_BlueBox,1);
+      getgroupitem(IG_BLUEBOX);
```
Lemongrass3110, line 1064: Doesnt that change the "identify" behavior?

getrandgroupitem:
```
item_tmp.identify = identify ? 1 : itemdb_isidentified(entry->nameid);
```

getgroupitem:
```
	bool identify = false;

	if( script_hasdata( st, 3 ) ){
		identify = script_getnum( st, 3 );
	}
```

db/import-tmpl/item_group_db.yml
```diff
@@ -25,6 +25,7 @@
 # - Group                      Group name of the item.
```
Lemongrass3110, line 25: Also needs the 2024 copyright update.

Playtester:
Shouldn't this be part of: https://github.com/rathena/rathena/pull/8849
?

src/map/itemdb.cpp
```diff
@@ -3029,21 +3029,13 @@ uint8 ItemGroupDatabase::pc_get_itemgroup( uint16 group_id, bool identify, map_s
...
-	// Get 1 'random' item from each subgroup
 	for (const auto &random : group->random) {
-		// Skip the 'must' group
-		if( random.first == 0 ){
-			continue;
+		if( random.second->algorithm != SUBGROUP_ALGO_MUST )
+			this->pc_get_itemgroup_sub( sd, identify, this->get_random_itemsubgroup( random.second ) );
+		else {
+			for (const auto &it : random.second->data)
+				this->pc_get_itemgroup_sub( sd, identify, it.second );
```
Lemongrass3110, line 3037: Maybe make it a switch already here?
So that "natural" can be added easily.

src/map/itemdb.hpp
```diff
@@ -3139,6 +3145,7 @@ struct s_item_group_entry
 struct s_item_group_random
 {
 	uint32 total_rate;
+	uint16 algorithm;
```
Lemongrass3110, line 3148:
```suggestion
	e_item_subgroup_algorithm algorithm;
```

src/map/itemdb.cpp
```diff
@@ -3315,20 +3307,38 @@ uint64 ItemGroupDatabase::parseBodyNode(const ryml::NodeRef& node) {
...
+
+				std::string sub_algo_constant = "SUBGROUP_ALGO_" + sub_algo;
+				int64 constant_algo;
+
+				if (!script_get_constant(sub_algo_constant.c_str(), &constant_algo)) {
+					this->invalidWarning(subit["Algorithm"], "Invalid Algorithm %s.\n", sub_algo.c_str());
+					continue;
+				}
+
+				random->algorithm = static_cast<uint16>(constant_algo);
```
Lemongrass3110, line 3335:
```suggestion
				random->algorithm = static_cast<e_item_subgroup_algorithm>(constant_algo);
```

src/tool/yamlupgrade.cpp
```diff
@@ -508,6 +508,86 @@ static bool upgrade_item_group_db( std::string file, const uint32 source_version
 			body << YAML::EndMap;
 		}
 
+		if( source_version == 3 ){
```
Lemongrass3110, line 511: Not sure what happens when someone with < version 3 upgrades and the upgrades above are also done at the same time. Could cause problems...

Lemongrass3110 (commented):
Looks good. Just some small remarks.

src/map/itemdb.hpp
Playtester: In my opinion, we shouldn't add a second enum for the algorithm at all but instead use the existing one `e_group_search_type`.

https://github.com/rathena/rathena/pull/8849/files

```
/// Enum for different ways to search an item group
enum e_group_search_type : uint8 {
	GROUP_SEARCH_BOX = 0, // Always return an item from the group, rate determines which item is more likely to be returned
	GROUP_SEARCH_PICK = 1, // Rate is the amount of items in the group, return a random item and remove it from the group
	GROUP_SEARCH_DROP = 2, // Pick one item from the group and check use rate as drop rate, on fail, do not return any item
};
```

db/pre-re/item_db_usable.yml
Atemo: For getrandgroupitem and getgroupitem, identify is defined to false by default. In case of getgroupitem, itemdb_isidentified is handled by `ItemGroupDatabase::pc_get_itemgroup_sub` https://github.com/rathena/rathena/blob/master/src/map/itemdb.cpp#L2964
To resume the "identify" behavior is the same in both commands

src/map/itemdb.hpp
```diff
@@ -3314,11 +3310,12 @@ class ItemGroupDatabase : public TypesafeCachedYamlDatabase<uint16, s_item_group
 	// Additional
 	bool item_exists(uint16 group_id, t_itemid nameid);
 	int16 item_exists_pc(map_session_data *sd, uint16 group_id);
-	std::shared_ptr<s_item_group_entry> get_random_entry(uint16 group_id, uint8 sub_group, e_group_search_type search_type = GROUP_SEARCH_BOX);
+	std::shared_ptr<s_item_group_entry> get_random_entry(uint16 group_id, uint8 sub_group, e_group_draws_type search_type);
```
Playtester, line 3313: Why did you duplicate the function?

src/map/itemdb.hpp
```diff
@@ -3015,9 +3009,11 @@ enum e_delay_consume : uint8 {
 };
 
 /// Enum for different ways to search an item group
-enum e_group_search_type : uint8 {
-	GROUP_SEARCH_BOX = 0, // Always return an item from the group, rate determines which item is more likely to be returned
-	GROUP_SEARCH_DROP = 1, // Pick one item from the group and check use rate as drop rate, on fail, do not return any item
+enum e_group_draws_type : uint8 {
```
Playtester, line 3012: Changing the enum now is problematic, you will never get it merged with the main PR. Please just add your changes to the main PR then.

But also I don't understand this description at all...

src/map/itemdb.cpp
```diff
@@ -2943,6 +2948,36 @@ std::shared_ptr<s_item_group_entry> ItemGroupDatabase::get_random_entry(uint16 g
...
+	if (group->random.empty()) {
+		ShowError("get_random_entry: No item entries for group id %hu.\n", group_id);
+		return nullptr;
+	}
+	if (group->random.count(sub_group) == 0) {
+		ShowError("get_random_entry: No item entries for group id %hu and sub group %hu.\n", group_id, sub_group);
+		return nullptr;
+	}
+
+	e_group_draws_type search_type = group->random[sub_group]->draws;
```
Playtester, line 2975: This just duplicates the code, why not leave it as one function and give the corresponding algorithm into the function?

src/map/itemdb.cpp
```diff
@@ -2888,33 +2888,38 @@ uint16 itemdb_searchname_array(std::map<t_itemid, std::shared_ptr<item_data>> &d
...
-			if (current_pos > pos)
+	switch( search_type ) {
+		case GROUP_DRAWS_DROP: {
+			// We pick a random item from the group and then do a drop check based on the rate. On fail, do not return any item
+			std::shared_ptr<s_item_group_entry> entry = util::umap_random(random->data);
+			if (rnd_chance_official<uint16>(entry->rate, 10000))
 				return entry;
+			break;
+		}
+		case GROUP_DRAWS_NOREPLACEMENT:	// placeholder
```
Playtester, line 2903: Another reason I recommend putting the change in the main PR instead. You'll never get this properly merged now.

src/map/itemdb.hpp
```diff
@@ -3015,9 +3009,11 @@ enum e_delay_consume : uint8 {
...
+	GROUP_DRAWS_REPLACEMENT, // Draws with replacement. Always return an item from the group, rate determines which item is more likely to be returned
+	GROUP_DRAWS_DROP, // Draws with replacement. Pick one item from the group and check use rate as drop rate, on fail, do not return any item
```
Playtester, line 3014: Drop is a bit of a special case as it shouldn't be used by Boxes, so I would place this last.

Atemo:
Merged with https://github.com/rathena/rathena/pull/8849

PR #8855, opened 2024-11-27
Corrected the order of items displayed in player inventory/storage

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

On official server, items in the player inventory/storage are displayed in the order the player placed them. (at least on renewal)

### DIFF
--- a/conf/battle/client.conf
+++ b/conf/battle/client.conf
@@ -113,8 +113,8 @@ display_status_timers: yes
 // packets for the desired number. (Note 1)
 client_reshuffle_dice: yes
 
-// Sorts the character and guild storage before it is sent to the client.
-// Official servers do not sort storage. (Note 1)
+// Sorts the cart, guild storage, inventory and storage before it is sent to the client. (Note 1)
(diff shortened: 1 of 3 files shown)


Lemongrass3110:
Should we make that a config, since it has been that way for a really long time?

Atemo:
https://github.com/rathena/rathena/blob/master/conf/battle/client.conf#L119
```
// Sorts the character and guild storage before it is sent to the client.
// Official servers do not sort storage. (Note 1)
// NOTE: Enabling this option degrades performance.
client_sort_storage: no
```
A conf is defined for storage, should I create a similar conf for inventory?

Lemongrass3110:
Yes, sounds good for me.

Maybe:
```
// Sort the inventory by item id before it is sent to the client?
// Official servers do not sort by item id. (Note 1)
client_sort_inventory: no
```

Otherwise @aleos89 maybe can gives his 5 cent.

aleos89:
Having a second config for inventory works for me!

PR #8860, opened 2024-11-30
Fixed NPC_SELFDESTRUCTION not giving items and exp

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: Fixes NPC_SELFDESTRUCTION not giving items or exp when the monster dies.
It should not show display the damage, but should give exp and items, so I used **status_zap** instead.
Follow up to dc64916ec4975b446188fc93aa4c10171c1b6bf2

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -8817,7 +8817,8 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, 
 			map_freeblock_unlock();
 			return 1;
 		}
-		status_kill(src);
+		// Won't display the damage, but drop items and give exp
+		status_zap(src, sstatus->max_hp, 0, 0);
 		break;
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -8817,7 +8817,8 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
 			map_freeblock_unlock();
 			return 1;
 		}
-		status_kill(src);
+		// Won't display the damage, but drop items and give exp
+		status_zap(src, sstatus->max_hp, 0, 0);
```
Lemongrass3110, line 8821:
```suggestion
		status_zap(src, sstatus->hp, 0, 0);
```

Lemongrass3110 (approved):

Issue #8859, opened 2024-11-30
script_command (*summon): return bl.id even without generating the md


Hyroshima:
### rAthena Hash

de4d739be78a545705ab5b6fa96a28ed89579f50

### Client Date

0000-00-00

### Server Mode

Pre-Renewal, Renewal

### Result

when trying to create a summon (slave monster), if for some reason the monster id passed is 0 or non-existent, it causes a crash because there is no verification for the monster id and even if there is a verification, if for another reason the md is not generated, script_pushint returns the bl.id null

### Expected Result

returns failure, non-existent monster and script_pushint 0
...
`if (!_class || !mobdb_checkid(_class)) {
		ShowWarning("buildin_summon: Attempted to spawn non-existing monster class %d\n", _class);
		return SCRIPT_CMD_FAILURE;
	}`
...
`script_pushint(st, (md ? md->bl.id : 0) );`

### How to Reproduce

summon "poring",0;

### Official Information

.

PR #8863, opened 2024-12-02
Fixed a crash in script command summon

Lemongrass3110:
* **Addressed Issue(s)**: #8859

* **Server Mode**: Both

Added a nullpointer check and a corresponding error message.
Also return SCRIPT_CMD_FAILURE in this case or when there is no player attached.

Thanks to @Hyroshima

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -16500,6 +16500,7 @@ BUILDIN_FUNC(summon)
 
 	if( md == nullptr ){
 		ShowError( "buildin_summon: Invalid mob ID %d.\n", _class );
+		st->state = END;
 		return SCRIPT_CMD_FAILURE;
 	}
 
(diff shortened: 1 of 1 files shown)


src/map/script.cpp
```diff
@@ -16497,20 +16496,28 @@ BUILDIN_FUNC(summon)
 		check_event(st, event);
 	}
 
-	clif_skill_poseffect(&sd->bl,AM_CALLHOMUN,1,sd->bl.x,sd->bl.y,tick);
+	mob_data* md = mob_once_spawn_sub( &sd->bl, sd->bl.m, sd->bl.x, sd->bl.y, str, _class, event, SZ_SMALL, AI_NONE );
+
+	if( md == nullptr ){
+		ShowError( "buildin_summon: Invalid mob ID %d.\n", _class );
+		return SCRIPT_CMD_FAILURE;
```
Atemo, line 16503: A value must be returned or the script should stop (`SCRIPT_CMD_FAILURE` does not stop the script)
```suggestion
		st->state = END;
		return SCRIPT_CMD_FAILURE;
```

Atemo (approved):

Issue #8858, opened 2024-11-28
atcommand stockall feature doesnt properly count the items.


gidzdlcrz:
### rAthena Hash

latest

### Client Date

2018 ~ 2024

### Server Mode

Pre-Renewal, Renewal

### Result

When you use at stockall it works but it doesnt properly count the items nor the item flags

### Relevant Log Output

```Shell
Not Applicable
```

### Expected Result

It should count the items like whats on at dropall command

### How to Reproduce

put items in your carts. then use at stockall command. You will notice that it doesnt count the items properly.

### Official Information

Not Applicable

### Modifications that may affect results

Not Applicable

Lemongrass3110:
Could you show us an example with screenshots?

Note: If you transfer stackable items, it will count each of it. So if you transfer 500 Apples and 500 Red Potions, you will see: "1000 items are transferred (0 skipped)!"

gidzdlcrz:
> Note: If you transfer stackable items, it will count each of it. So if you transfer 500 Apples and 500 Red Potions, you will see: "1000 items are transferred (0 skipped)!"

Yes. This part doesnt work. It always says "0 items are transferred (0 skipped)!"

PR #8864, opened 2024-12-02
Fixed stockall counter

Lemongrass3110:
* **Addressed Issue(s)**: #8858

* **Server Mode**: Both

`@stockall` deletes the items from the cart and after deletion it tries to access the item amount from the same cart index again, which of course will always return 0. Fixed this behavior by saving the amount before deleting the item.
Similar to what was done in #4331 for `@dropall`

Thanks to @gidzdlcrz

### DIFF
--- a/src/map/atcommand.cpp
+++ b/src/map/atcommand.cpp
@@ -6043,7 +6043,8 @@ ACMD_FUNC(stockall)
 				return -1;
 		}
 	}
-	uint16 count = 0, count2 = 0;
+
+	int32 count = 0, count2 = 0;
 	for ( uint16 i = 0; i < MAX_CART; i++ ) {
 		if ( sd->cart.u.items_cart[i].amount > 0 ) {
 			std::shared_ptr<item_data> id = item_db.find(sd->cart.u.items_cart[i].nameid);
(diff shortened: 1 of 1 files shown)


gidzdlcrz:
Thanks! This fixes the issue. <3

gidzdlcrz:
Hi. As i checked. It doesnt count the skipped items.
image

This should count the skipped items as well. In this case its the 2 items.

Lemongrass3110:
> Hi. As i checked. It doesnt count the skipped items. image
> 
> This should count the skipped items as well. In this case its the 2 items.

@gidzdlcrz What command did you use? The "skipped" only counts if the items would be of the correct type and cannot be moved. So if you use `@stockall 0` to stock healing(=0) items the weapons (=5) will not count as skipped.

gidzdlcrz:
> > Hi. As i checked. It doesnt count the skipped items. image
> > This should count the skipped items as well. In this case its the 2 items.
> 
> @gidzdlcrz What command did you use? The "skipped" only counts if the items would be of the correct type and cannot be moved. So if you use `@stockall 0` to stock healing(=0) items the weapons (=5) will not count as skipped.

Yes i have used @stockall 0 to get those 100 berries. I thought those weapons are included in skip if you do @stockall <flag>. Nevermind then. Thank you! <3

PR #8870, opened 2024-12-07
Implemented bonus bMatk2

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

Implemented `bonus bMatk2`, a bonus not visible in the player status window.
`bonus bMatk2,n;   Magical attack power + n (not visible in status window)`

This bonus is made especially for the status change used in `Tyr's Blessing` (id: 14601), `Herb of Incantation` (id: 683)..


src/map/status.cpp
```diff
@@ -5989,6 +5994,12 @@ void status_calc_bl_main(struct block_list& bl, std::bitset<SCB_MAX> flag)
 			matk_max += sd->bonus.ematk;
 		}
 
+		// Flat MATK not visible in status window
+		if (sd != nullptr && sd->bonus.ematk_hidden > 0) {
```
vstumpf, line 5998: we should just add ematk_hidden even if it's 0, because right now it'll skip if it's negative

Atemo: Correct! ematk also need a fix

Issue #8873, opened 2024-12-08
Allow to have an option to end VIP immediately


gidzdlcrz:
### Is your request related to a missing official feature?

No

### Describe the solution you'd like

When your VIP expires, you will have 4 to 5 minutes before your actual VIP to end.

### Describe alternatives you've considered

if its not too much. Put battle conf to use the original settings or the modded type.

### Additional context

1. To Reproduce: Have your other character to become VIP by using `@vip 30s <PLAYERNAME>`. That is 30 seconds VIP.
2. Create NPC that checks if you are VIP.
```
prontera,150,150,5	script	VIP Checker	909,{
	if(vip_status(VIP_STATUS_ACTIVE)) {
		mes "You are VIP.";
	} else {
		mes "You are not VIP.";
	}
	close;
}
```
3. Talk to NPC to confirm if you are VIP. Wait for 30 secs (this should end your VIP)
4. Talk to the NPC again after 30 secs to check if you still have VIP.
5. See Result.

PR #8877, opened 2024-12-15
Fixed VIP not ending immediately

Lemongrass3110:
* **Addressed Issue(s)**: #8873

* **Server Mode**: Both

Added a monitoring to the login-server to notify the map-server, when the VIP status runs out.
Removed VIP status checking from periodic player saving of the map-server.
Added seconds to the output of the `@vip` command.
Added the player name to the output of the `@vip` command.
Fixed a bug where a player would not immediately get the defined gemstone bonus of the VIP group.

Thanks to @gidzdlcrz


gidzdlcrz:
Thank you so much @Lemongrass3110! <3

PR #8894, opened 2024-12-23
Removed player damage log

Lemongrass3110:

* **Server Mode**: Both

It has been confirmed that players that log out stay in the damage log of the monsters, that they have attacked. Once the monster dies their online status is being checked and players that are offline are ignored at that point. This offline check is already implemented in rAthena additionally, since the player damage log seem to only have been used for MVPs.


Playtester:
Actually, let me test if logging out really doesn't remove you as first attacker before finishing this.

Playtester:
Okay I can confirm:

Player A deals 49% damage, Player B deals 51% damage -> Player A gets loot priority
Player A deals 49% damage, logs out, Player B deals 51% damage -> Player B gets loot priority
Player A deals 49% damage, logs out, logs in, Player B deals 51% damage -> Player A gets loot priority

So it only matters if the player is online at the time of the kill.

Issue #8847, opened 2024-11-23
Loot Priority (first attacker) and MVP bonus (received damage)


Playtester:
### rAthena Hash

4de6c26

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

Loot Priority is currently determined by the amount of damage dealt. There is no bonus for the first attacker. Also, slave damage isn't added to player damage.

MVP bonus does not consider damage tanked.

When a player logs out, he is removed from the damage log.

### Expected Result

For loot priority, the first attacker gets a 30% bonus of the total damage dealt to the monster. Damage received is not considered. Slave damage (homunculus + mercenary) is added to player damage.

MVP bonus is damage_dealt + damage_received (normal attacks only). No first attacker bonus.

When a player logs out, he remains in the damage log, but when the monster dies while he is logged out, he will not get loot priority.

### How to Reproduce

Elder Willow 693 HP
Player 1 - Deals 300 damage to Elder Willow
Player 2 - Deals 500 damage to Elder Willow

Result: Player 1 gets loot priority.
Math: Player 1: `300 + 800 * 0.3 = 540` Player 2: `500`

Any MVP
Player 1 - Deals first hit.
Player 2 - Tanks more damage than MVP has HP.
Player 1 - Kills MVP (without overkill damage).

Result: Player 2 gets MVP bonus.

Ant Egg 420 HP
Player 1 - Deals 91 damage to Ant Egg
Player 2 - Deals 164 damage to Ant Egg
Player 2's Homunculus - Deals 165 damage to Ant Egg

Result: Player 2 gets loot priority.
Math: Player 1: `91 + 420 * 0.3 = 217` Player 2: `164 + 165 = 329`

### Official Information

Verified on Elder Willow, MVP and Ant Egg (see tests above).

PR #8895, opened 2024-12-23
Monster Damage Log Improvements

Playtester:
* **Addressed Issue(s)**: #8847 (this is just a preparation PR for the actual fix)

* **Server Mode**: Both

- If a monster does a normal attack we now log the damage it dealt (will later be used for MVP reward)
- Official DAMAGELOG_SIZE of 20
- When there are more than 20 entries in the damage log, the oldest entry will be removed (this also updates the "first player")
- Related to #8847

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2400,14 +2400,16 @@ TIMER_FUNC(mob_respawn){
 	return 1;
 }
 
-void mob_log_damage(mob_data* md, block_list* src, int32 damage, int32 attdamage)
+void mob_log_damage(mob_data* md, block_list* src, int32 damage, int32 damage_tanked)
 {
 	uint32 char_id = 0;
 	int32 flag = MDLF_NORMAL;
 
 	if( damage < 0 )
 		return; //Do nothing for absorbed damage.
-	if( !damage && !attdamage && !(src->type&DEFAULT_ENEMY_TYPE(md)) )
+	if (damage_tanked < 0)
+		return; //Just to make sure we don't subtract it (should not happen)
+	if( !damage && !damage_tanked && !(src->type&DEFAULT_ENEMY_TYPE(md)) )
 		return; //Do not log non-damaging effects from non-enemies.
 
 	switch( src->type )
@@ -2502,7 +2504,7 @@ void mob_log_damage(mob_data* md, block_list* src, int32 damage, int32 attdamage
 				md->dmglog[i].flag= flag;
 				// Damage is added outside the loop, we reset it here to be safe
 				md->dmglog[i].dmg = 0;
-				md->dmglog[i].attdmg = 0;
+				md->dmglog[i].dmg_tanked = 0;
 
 				if( md->get_bosstype() == BOSSTYPE_MVP )
 					pc_damage_log_add(map_charid2sd(char_id),md->bl.id);
@@ -2512,7 +2514,7 @@ void mob_log_damage(mob_data* md, block_list* src, int32 damage, int32 attdamage
 		// Character or empty slot was found, just add damage to it
 		if (i < DAMAGELOG_SIZE) {
 			md->dmglog[i].dmg += damage;
-			md->dmglog[i].attdmg += attdamage;
+			md->dmglog[i].dmg_tanked += damage_tanked;
 		}
 		else {
 			// Damage log is full, remove oldest entry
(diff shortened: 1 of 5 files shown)


src/map/mob.cpp
```diff
@@ -2403,14 +2400,14 @@ TIMER_FUNC(mob_respawn){
...
-void mob_log_damage(struct mob_data *md, struct block_list *src, int32 damage)
+void mob_log_damage(struct mob_data *md, struct block_list *src, int32 damage, int32 attdamage)
 {
 	uint32 char_id = 0;
 	int32 flag = MDLF_NORMAL;
 
 	if( damage < 0 )
 		return; //Do nothing for absorbed damage.
-	if( !damage && !(src->type&DEFAULT_ENEMY_TYPE(md)) )
+	if( !damage && !attdamage && !(src->type&DEFAULT_ENEMY_TYPE(md)) )
```
Lemongrass3110, line 2410:
```suggestion
	if( ( damage != 0 || attdamage != 0 ) && !(src->type&DEFAULT_ENEMY_TYPE(md)) )
```
Wouldnt that be the more correct logic now?

src/map/mob.hpp
```diff
@@ -350,6 +350,7 @@ struct mob_data {
 	struct s_dmglog {
 		int32 id; //char id
 		uint32 dmg;
+		uint32 attdmg;
 		uint32 flag : 2; //0: Normal. 1: Homunc exp. 2: Pet exp
```
Lemongrass3110, line 354:
```suggestion
		uint32 flag : 2; //0: Normal. 1: Homunc exp. 2: Pet exp. 3: Self.
```

src/map/status.cpp
```diff
@@ -1600,6 +1601,11 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp,
 		break;
 	}
 
+	// Normal attack damage is logged in the monster's dmglog as attack damage
+	// This counts as exp tap and is used for determining the MVP
+	if (src && src->type == BL_MOB && skill_id == 0)
+		mob_log_damage(reinterpret_cast<mob_data*>(src), target, 0, (int32)dhp);
```
Lemongrass3110, line 1607:
```suggestion
		mob_log_damage(reinterpret_cast<mob_data*>(src), target, 0, static_cast<int32>(dhp));
```

src/map/mob.hpp
```diff
@@ -515,7 +516,7 @@ int32 mob_spawn(struct mob_data *md);
 TIMER_FUNC(mob_delayspawn);
 int32 mob_setdelayspawn(struct mob_data *md);
 int32 mob_parse_dataset(struct spawn_data *data);
-void mob_log_damage(struct mob_data *md, struct block_list *src, int32 damage);
+void mob_log_damage(struct mob_data *md, struct block_list *src, int32 damage, int32 attdamage = 0);
```
Lemongrass3110, line 519:
```suggestion
void mob_log_damage(mob_data* md, block_list* src, int32 damage, int32 attdamage = 0);
```

src/map/mob.cpp
```diff
@@ -2403,14 +2400,14 @@ TIMER_FUNC(mob_respawn){
 	return 1;
 }
 
-void mob_log_damage(struct mob_data *md, struct block_list *src, int32 damage)
+void mob_log_damage(struct mob_data *md, struct block_list *src, int32 damage, int32 attdamage)
```
Lemongrass3110, line 2403:
```suggestion
void mob_log_damage(mob_data* md, block_list* src, int32 damage, int32 attdamage)
```

src/map/mob.cpp
```diff
@@ -2403,14 +2400,14 @@ TIMER_FUNC(mob_respawn){
...
 {
 	uint32 char_id = 0;
 	int32 flag = MDLF_NORMAL;
 
 	if( damage < 0 )
```
Lemongrass3110, line 2408: Do we need a similar check for attdamage?

src/map/mob.cpp
```diff
@@ -2493,32 +2490,44 @@ void mob_log_damage(struct mob_data *md, struct block_list *src, int32 damage)
 
 	if( char_id )
 	{ //Log damage...
-		int32 i,minpos;
-		uint32 mindmg;
-		for(i=0,minpos=DAMAGELOG_SIZE-1,mindmg=UINT_MAX;i<DAMAGELOG_SIZE;i++){
+		int32 i;
```
Lemongrass3110, line 2493:
```suggestion
		size_t i;
```

Lemongrass3110 (approved):
Looks fine to me, although some players/server owners may be used to the previous calculation (by damage and not by time of hitting). So not 100% sure, if we should provide a config for it or not.

I would still recommend to rename "attdamage" to something that is easier to understand.

src/map/mob.cpp
Playtester: Only if both are 0, you don't want to record it.

src/map/mob.cpp
Playtester: attdamage cannot be negative

Playtester: I'll add the check anyway, but just for safety, so nothing results in a substraction.

Issue #8125, opened 2024-02-07
Tuna party take HP from target


Melk3000:
### rAthena Hash

lastest

### Client Date

20180620

### Server Mode

Pre-Renewal

### Result

Casting TUNA PARTY on an enemy, the shield takes the HP of the Enemy, when it should take the HP of the caster

### Expected Result

The size of the shield should be the size of the caster

### How to Reproduce

Cast Tuna Party on an enemy, and the defense is equal to the enemy's HP

### Official Information

https://www.divine-pride.net/database/skill/5039

PR #8904, opened 2024-12-28
Corrected SU_TUNAPARTY

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8125

* **Server Mode**: renewal

The shield from `SU_TUNAPARTY` is now according to the Caster's MaxHP%

Thanks to @Melk3000

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -12145,7 +12145,7 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
 			break;
 		case SC_TUNAPARTY:
 			val2 = status_get_max_hp(src) * (val1 * 10) / 100; // Shield according to the Caster's MaxHP%
-			if (pc_checkskill(reinterpret_cast<map_session_data*>(src), SU_SPIRITOFSEA) > 0)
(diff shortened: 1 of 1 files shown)


src/map/status.cpp
```diff
@@ -12144,9 +12144,9 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
 			}
 			break;
 		case SC_TUNAPARTY:
-			val2 = (status->max_hp * (val1 * 10) / 100); // Max HP% to absorb
-			if (sd && pc_checkskill(sd, SU_SPIRITOFSEA))
-				val2 *= 2; // Double the shield life
+			val2 = status_get_max_hp(src) * (val1 * 10) / 100; // Shield according to the Caster's MaxHP%
+			if (pc_checkskill(reinterpret_cast<map_session_data*>(src), SU_SPIRITOFSEA) > 0)
```
Lemongrass3110, line 12148:
```suggestion
			if (map_session_data* ssd = BL_CAST(BL_PC, src); pc_checkskill(ssd, SU_SPIRITOFSEA) > 0)
```
Here you cannot use reinterpret_cast, since you cannot be sure that src is always a player (unless I missed something above).

Note: pc_checkskill contains a nullptr check.

Lemongrass3110 (approved):

PR #8905, opened 2024-12-28
Corrected SC_FRESHSHRIMP

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

Heal value is now calculated based on the caster status.
Heal value is now displayed on the player.

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -12107,7 +12107,7 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
 			break;
 		case SC_FRESHSHRIMP: {
 				int32 min = 0, max = 0;
-				map_session_data* ssd = reinterpret_cast<map_session_data*>(src);
+				map_session_data* ssd = BL_CAST( BL_PC, src );
 
 #ifdef RENEWAL
 				status_data* sstatus = status_get_status_data(*src);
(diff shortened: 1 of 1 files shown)


src/map/status.cpp
```diff
@@ -12100,43 +12100,45 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
...
 			if (sd && (sd->class_&MAPID_BASEMASK) == MAPID_SUMMONER)
-				val4 = 10; // Ranged ATK increase
+				val4 = 10; // Ranged ATK increase if the target is a Doram
 			break;
 		case SC_SHRIMP:
 			val2 = 10; // BATK%, MATK%
 			break;
 		case SC_FRESHSHRIMP: {
 				int32 min = 0, max = 0;
+				map_session_data* ssd = reinterpret_cast<map_session_data*>(src);
```
Lemongrass3110, line 12110:
```suggestion
				map_session_data* ssd = BL_CAST( BL_PC, src );
```
Here you cannot use reinterpret_cast, since you cannot be sure that src is always a player (unless I missed something above).

Lemongrass3110 (approved):

src/map/status.cpp
Atemo: Correct !

PR #8919, opened 2025-01-06
Converted dmglog to std::deque

Lemongrass3110:

* **Server Mode**: Both

Converted dmglog to std::deque

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1204,7 +1204,6 @@ int32 mob_spawn (struct mob_data *md)
 		md->spotted_log[i] = 0;
 
 	md->dmglog.clear();
-	md->tdmg = 0;
 
 	if (md->lootitems)
 		memset(md->lootitems, 0, sizeof(*md->lootitems));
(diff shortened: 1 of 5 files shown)


Atemo (approved):
I didn't test, it seems good to me

aleos89 (approved):

src/map/mob.cpp
```diff
@@ -2700,40 +2699,60 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
...
-			continue; // skip homunc's share if inactive
-		if (md->dmglog[i].flag == MDLF_PET && (!tsd->status.pet_id || !tsd->pd))
-			continue; // skip pet's share if inactive
 
-		if(md->dmglog[i].dmg > mvp_damage) {
+		switch( entry.flag ){
+			case MDLF_NORMAL:
+				dmgbltypes |= BL_PC;
+				break;
+
```
Playtester, line 2728: The "break;" already serves pretty much as a clear divider, no need for extra empty line here. We should try to keep the "height" of code as small as possible so you can overview more on a single screen. (But I will rewrite this soon anyway, so you can leave it like this and I'll improve it later.)

Playtester (approved):
Good to go now from my viewpoint!

PR #8923, opened 2025-01-07
Changed mob drop array to STL

Lemongrass3110:

* **Server Mode**: Both

Changed mob drop array to STL
Harmonized output of `@monsterinfo` between drops and MVP drops
Fixed a problem with addmonsterdrop, which would report an error if you would try to edit the first item (index = 0).
delmonsterdrop now deletes all items with the given item id and not only the first found entry.
Reduced memory use of all mobs, especially slave monsters and non MVPs.


Playtester (approved):
Tested: 
- @mi with drop rates 
- Max loot drop
- Drop order
- Drop position
- MVP drops
- Jewel Sword
- Ore Discovery

PR #8926, opened 2025-01-08
Corrected SU_CN_METEOR formula to match the skill description

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

Corrected SU_CN_METEOR formula to match the skill description : "If the caster's base level is 100 or higher, damage is additionally increased according to the caster's base level and INT."
Tests on official server confirm that INT ratio is 5.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -8393,8 +8393,8 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 						skillratio += -100 + 200 + 100 * skill_lv;
 						if (status_get_lv(src) > 99) {
 							skillratio += sstatus->int_ * 5;
-							RE_LVL_DMOD(100);
 						}
+						RE_LVL_DMOD(100);
 						break;
 					case NPC_VENOMFOG:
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -8390,8 +8390,11 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
 						break;
 					case SU_CN_METEOR:
 					case SU_CN_METEOR2:
-						skillratio += 100 + 100 * skill_lv + sstatus->int_ * 5; // !TODO: Confirm INT bonus
-						RE_LVL_DMOD(100);
+						skillratio += -100 + 200 + 100 * skill_lv;
+						if (status_get_lv(src) > 99) {
+							skillratio += sstatus->int_ * 5;
+							RE_LVL_DMOD(100);
+						}
```
Lemongrass3110, line 8397:
```suggestion
						}
						RE_LVL_DMOD(100);
```
I understand the intention behind it, but lets leave it like that.

Lemongrass3110 (approved):

aleos89 (approved):

PR #8936, opened 2025-01-18
Monster Ranges, Looters and Blind

### LINKED ISSUE #8914: Monster Ranges, Looters and Blind
### rAthena Hash

f4dfc59

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

Aggro Range is Range2 (10) and 3 if blind. This is the range at which an aggressive monster starts chasing its target.
Chase Range is Range3 (12). This is the range at which a chasing monster gives up chasing its target.
Loot Range is Range2 (10) and 3 if blind. This is the range at which a looter monster searching for loot to pick up.

Monsters always start chasing when they find a target or their target leaves their attack range.

Monsters check for their target while chasing at the end of their walkpath. A ranged monster's chase path is only up to the point from which it can attack.

Monsters stop attacking when their target moves outside their attack range.

### Relevant Log Output

_No response_

### Expected Result

Aggro Range is Range2 (10) and 1 if blind (only starts attacking if you are right next to it).
Chase Range is Range3 (12) and 1 if blind.
Loot Range is 12 (fixed hardcoded value). Not affected by blindness.

Monsters with aMotion > aDelay and monsters not in Berserk mode start chasing in [Chase Range].
Monsters with aDelay > aMotion that are in Berserk mode always start chasing (only limited by max walkpath).

Monsters check for their target in [Chase Range] one cell before the end of their walkpath. A ranged monster's chase path leads all the way next to the target, but the monster immediately stops when the target gets into attack range.

Monsters stop attacking when at the end of their attack delay, there is no target in [Chase Range]. A monster without attack delay can always initiate an attack without checking for this.

**Range At Which A Monster Switches Target**
Angry Mode - [Range3] (12) range and 1 if blind
Berserk Mode - [aRan+1] range

### How to Reproduce

Use Ankle Snare or Ice Wall trick to get a monster stuck, then position yourself (or an item) at the test distance and see how the monster reacts.

Use a Monk. Blind a monster (e.g. 4x Familiar Card), then use Snap to get far away once it's blinded.

Ambernite - The monster will chase you all the way to where you snapped, but then drops target even if you are right next to it.

Zombie / Orc Skeleton - The monster will chase you all the way to where you snapped, then drops target, but because it still scans for targets in 1 range, it finds you as a target again. (Note: These are Berserk mode due to being blinded.)

Antique Firelock / Khalitzburg / Dancing Dragon - The monster drops target (no chase). In very rare cases it might starting moving (timer inaccuracy issue).

Archer Skeleton:
From Range 1 -> Snap 8 Cells - Monster stops attacking
From Range 1 -> Snap 9 Cells - Moves one cell closer, stops attacking
From Range 1 -> Snap 14 Cells right when attack starts - Moves 5 cells closer, stops attacking
From Range 1 -> Snap 14 Cells when AttackDelay is about to end - Moves 5 cells closer, then attacks once, then stops attacking

---------------------------

Use a Monk - Use Snap to make the monster chase you, then use Snap again to gain exactly the distance from the monster's target walk tile that you want to test with. Below tests assume you start next to the mob and snap away from it.

Angry Monster (e.g. Zombie / Orc Skeleton / Civil Servant / Hyegun):
Snap 2x10 tiles -> Monster starts chasing for 10 tiles, then continues chasing for another 10 tiles
Snap 2x11 tiles -> Monster starts chasing for 11 tiles, then gives up
Snap 2x12 tiles -> Monster doesn't start chasing at all
Snap 2x13 tiles -> Monster doesn't start chasing at all

aDelay > aMotion Monster (e.g. Ambernite; **After Hit**: Zombie / Orc Skeleton):
Snap 2x10 tiles -> Monster starts chasing for 10 tiles, then continues chasing for another 10 tiles
Snap 2x11 tiles -> Monster starts chasing for 11 tiles, then gives up
Snap 2x12 tiles -> Monster starts chasing for 12 tiles, then gives up
Snap 2x13 tiles -> Monster starts chasing for 13 tiles, then gives up

aMotion > aDelay Monster (e.g. Khalitzburg / Hyegun)
Snap 2x10 tiles -> Monster starts chasing for 10 tiles, then continues chasing for another 10 tiles
Snap 2x11 tiles -> Monster starts chasing for 11 tiles, then gives up
Snap 2x12 tiles -> Monster doesn't start chasing at all
Snap 2x13 tiles -> Monster doesn't start chasing at all

Tank a monster in Angry Mode (don't hit it) with Range3=12. Then do the first hit with an archer from x range:
Range 11 - Monster completely switches to Archer as target.
Range 12 - Monster completely switches to Archer as target.
Range 13 - Monster switches to Archer as target, but then switches back to the Tank.

Blind Elder Test (need to modify stats so it's not immune to blind):
When moving out of attack range of Elder, it starts off with Frost Diver / Thunderbolt, but then uses Firebolt, which implies it switched to idle mode (it doesn't move, so it's still outside attack range).

--------------------------

Sniper with Flash Arrow Tests:

Fight Archer Skeleton and **blind** it (aDelay > aMotion):
Range 1 - Monster keeps attacking
Range 2 - Monster stops attacking
Attack at Range 2 - Monster attacks once, then stops attacking
Attack at Range 14 - Monster walks to Range 9, then attacks once, then stops attacking
Attack at Range 14 then move two cells away - Monster walks 5+2 cells to Range 9, then attacks once, then stops attacking

Fight Antique Firelock and **blind** it (aMotion > aDelay):
Range 1 - Monster keeps attacking
Range 2 - Monster stops attacking
Attack at Range 2 - Monster attacks once, then stops attacking
Attack at Range 14 - Monster walks to Range 10, then attacks once, then stops attacking
Attack at Range 14 then move two cells away - Monster walks 4+2 cells to Range 10, then attacks once, then stops attacking

Fight Hyegun and **blind** it (aMotion > aDelay):
Range 1 - Monster keeps attacking
Range 2 - Monster stops attacking
Attack at Range 2 - Monster walks to Range 1, then continues attacking
Attack at Range 14 - Monster walks to Range 1, then continues attacking
Attack at Range 14 then move one cell away - Monster walks 13 cells to to Range 2, then stops

Fight Shecil (Range1=14, Range2=10, Range3=12); move into her aggro range, once she attacks, move away:
Range 10 - Shecil starts attacking
Range 13 - Shecil stops attacking

-----------------------------------

Two players: One Monk with Snap & Smokie Card
Monk snaps to the other side of Player B and then immediately hides.

Goblin 5: Walks past Player B until one cell before he end of the walkpath, then turns around and goes for Player B. 
(Note: Cannot be fully fixed as part of this PR as even though Goblin will now properly drop target one cell earlier we currently didn't implement sub cells to allow changing taking between two cells.)
Khalitzburg: Immediately attacks Player B when moving past it even if Monk is hiding.

### Official Information

Verified through tests.

Please note that apart from these ranges, there's also a complex path finding algorithm going on that limits pathfinding to 60 walkable nodes, which can be a lot shorter if there are a lot of walkable tiles nearby. Implementing this algorithm correctly is not part of this bug report. Above only applies for simple pathfinding.

_No response_

Playtester:
* **Addressed Issue(s)**: #8914

* **Server Mode**: Both

- Looters now have a loot range of 12 (configurable) that is no longer reduced by Blind
- Chase range is no longer increased when hitting a monster from outside its chase range 
  * Removed the "min_chase" custom solution
- Blind now reduces both aggro range and chase range to 1
- Chase range at the beginning of a chase is now only checked in Angry state (blind doesn't matter)
- Chase range (including blind) is now checked whenever the attack timer ends
  * The attack timer can no longer be stopped by pre-planning a chase
  * This will make monsters stop attacking when outside chase range at the end of the timer
- Monsters will now think about updating their chase already one cell before they reach the end of their path
  * This is not only official but also makes their movement and behavior smoother
  * If at this time no enemy is in chase range, the monster will drop its target but still finish moving
- A monster's chase path will now always go all the way next to the target, even if the monster has ranged attacks
  * Monsters will now check for each cell moved if target is in attack range and then immediately attack
  * If they still have attack delay they will stop and attack when ready (if in chase range at this point)
- Fixed a timer to delay movement during hit lock
  * Previously the timer didn't work at all and the whole mob AI had to be processed again
  * Now the monster will start moving immediately once the hit lock is gone
- Fixed an endless loop that occurred when we call mob_randomwalk while the mob still has a target
  * This happens because we pre-plan chases with timers but the state remains Idle until the chase starts
- Monsters now become active once a player is within 18 cells (AREA_SIZE+4)
- Slaves now teleport back to their master if they are more than 15 cells away (AREA_SIZE+1) 
  * This only applies to AI_ABR and AI_BIONIC unless "slave_stick_with_master" is enabled
- Improved code structure in unit.cpp: Bundling repeating code segments into new functions
- Fixes #8914

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -1699,8 +1699,7 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp, 
 		unit_remove_map(target,CLR_DEAD);
 	else { // Some death states that would normally be handled by unit_remove_map
 		unit_stop_attack(target);
-		unit_stop_walking(target,1);
-		unit_stop_chase(*target);
+		unit_stop_walking(target,USW_FIXPOS|USW_RELEASE_TARGET);
 		unit_skillcastcancel(target,0);
 		clif_clearunit_area( *target, CLR_DEAD );
 		skill_unit_move(target,gettick(),4);
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -2243,7 +2243,7 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
  * @return src can see (1) or target is invisible (0)
  * @author [Skotlex]
  */
-int32 status_check_visibility(struct block_list *src, struct block_list *target, bool checkblind)
+int32 status_check_visibility(block_list* src, block_list* target, bool checkblind)
 {
 	int32 view_range;
 	status_change* tsc = status_get_sc(target);
@@ -2260,7 +2260,7 @@ int32 status_check_visibility(struct block_list *src, struct block_list *target,
 
 	if (checkblind) {
 		status_change* sc = status_get_sc(src);
-		if (sc && sc->getSCE(SC_BLIND))
+		if (sc != nullptr && sc->getSCE(SC_BLIND) != nullptr)
 			view_range = 1;
 	}
 
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -2240,10 +2240,10 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
  * @param src:	Object using skill on target [PC|MOB|PET|HOM|MER|ELEM]
  * @param target: Object being targeted by src [PC|MOB|HOM|MER|ELEM]
  * @param checkblind: Whether blind condition should be considered (sets view range to 1)
- * @return src can see (1) or target is invisible (0)
+ * @return src can see (true) or target is invisible (false)
  * @author [Skotlex]
  */
-int32 status_check_visibility(block_list* src, block_list* target, bool checkblind)
+bool status_check_visibility(block_list* src, block_list* target, bool checkblind)
 {
 	int32 view_range;
 	status_change* tsc = status_get_sc(target);
@@ -2265,10 +2265,10 @@ int32 status_check_visibility(block_list* src, block_list* target, bool checkbli
 	}
 
 	if (src->m != target->m || !check_distance_bl(src, target, view_range))
-		return 0;
+		return false;
 
 	if ( src->type == BL_NPC) // NPCs don't care for the rest
-		return 1;
+		return true;
 
 	if (tsc) {
 		bool is_boss = (status_get_class_(src) == CLASS_BOSS);
@@ -2279,24 +2279,24 @@ int32 status_check_visibility(block_list* src, block_list* target, bool checkbli
 					map_session_data *tsd = (TBL_PC*)target;
 
 					if (((tsc->option&(OPTION_HIDE|OPTION_CLOAK|OPTION_CHASEWALK)) || tsc->getSCE(SC_CAMOUFLAGE) || tsc->getSCE(SC_STEALTHFIELD) || tsc->getSCE(SC_SUHIDE)) && !is_boss && (tsd->special_state.perfect_hiding || !is_detector))
-						return 0;
+						return false;
 					if ((tsc->getSCE(SC_CLOAKINGEXCEED) || tsc->getSCE(SC_NEWMOON)) && !is_boss && ((tsd && tsd->special_state.perfect_hiding) || is_detector))
-						return 0;
+						return false;
 					if (tsc->getSCE(SC__FEINTBOMB) && !is_boss && !is_detector)
-						return 0;
+						return false;
 				}
 				break;
 			case BL_ELEM:
 				if (tsc->getSCE(SC_ELEMENTAL_VEIL) && !is_boss && !is_detector)
-					return 0;
+					return false;
 				break;
 			default:
 				if (((tsc->option&(OPTION_HIDE|OPTION_CLOAK|OPTION_CHASEWALK)) || tsc->getSCE(SC_CAMOUFLAGE) || tsc->getSCE(SC_STEALTHFIELD) || tsc->getSCE(SC_SUHIDE)) && !is_boss && !is_detector)
-					return 0;
+					return false;
 		}
 	}
 
-	return 1;
+	return true;
 }
 
 /**
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -1700,6 +1700,7 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp, 
 	else { // Some death states that would normally be handled by unit_remove_map
 		unit_stop_attack(target);
 		unit_stop_walking(target,1);
+		unit_stop_chase(*target);
 		unit_skillcastcancel(target,0);
 		clif_clearunit_area( *target, CLR_DEAD );
 		skill_unit_move(target,gettick(),4);
@@ -2248,7 +2249,10 @@ int32 status_check_visibility(struct block_list *src, struct block_list *target)
 	status_change* tsc = status_get_sc(target);
 	switch (src->type) {
 		case BL_MOB:
-			view_range = ((TBL_MOB*)src)->min_chase;
+			if (((TBL_MOB*)src)->sc.getSCE(SC_BLIND))
+				view_range = 1;
+			else
+				view_range = ((TBL_MOB*)src)->db->range3;
 			break;
 		case BL_PET:
 			view_range = ((TBL_PET*)src)->db->range2;
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -2240,25 +2240,29 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
  * Checks whether the src can see the target
  * @param src:	Object using skill on target [PC|MOB|PET|HOM|MER|ELEM]
  * @param target: Object being targeted by src [PC|MOB|HOM|MER|ELEM]
+ * @param checkblind: Whether blind condition should be considered (sets view range to 1)
  * @return src can see (1) or target is invisible (0)
  * @author [Skotlex]
  */
-int32 status_check_visibility(struct block_list *src, struct block_list *target)
+int32 status_check_visibility(struct block_list *src, struct block_list *target, bool checkblind)
 {
 	int32 view_range;
 	status_change* tsc = status_get_sc(target);
 	switch (src->type) {
 		case BL_MOB:
-			if (((TBL_MOB*)src)->sc.getSCE(SC_BLIND))
-				view_range = 1;
-			else
-				view_range = ((TBL_MOB*)src)->db->range3;
+			view_range = ((TBL_MOB*)src)->db->range3;
 			break;
 		case BL_PET:
 			view_range = ((TBL_PET*)src)->db->range2;
 			break;
 		default:
 			view_range = AREA_SIZE;
+	}
+
+	if (checkblind) {
+		status_change* sc = status_get_sc(src);
+		if (sc && sc->getSCE(SC_BLIND))
+			view_range = 1;
 	}
 
 	if (src->m != target->m || !check_distance_bl(src, target, view_range))
--- a/src/map/status.hpp
+++ b/src/map/status.hpp
@@ -3622,7 +3622,7 @@ void status_calc_state(struct block_list *bl, status_change *sc, std::bitset<SCS
 void status_calc_slave_mode(mob_data& md);
 
 bool status_check_skilluse(struct block_list *src, struct block_list *target, uint16 skill_id, int32 flag);
-int32 status_check_visibility(block_list* src, block_list* target, bool checkblind);
+bool status_check_visibility(block_list* src, block_list* target, bool checkblind);
 
 int32 status_change_spread(block_list *src, block_list *bl);
 
--- a/src/map/status.hpp
+++ b/src/map/status.hpp
@@ -3622,7 +3622,7 @@ void status_calc_state(struct block_list *bl, status_change *sc, std::bitset<SCS
 void status_calc_slave_mode(mob_data& md);
 
 bool status_check_skilluse(struct block_list *src, struct block_list *target, uint16 skill_id, int32 flag);
-int32 status_check_visibility(struct block_list *src, struct block_list *target, bool checkblind);
+int32 status_check_visibility(block_list* src, block_list* target, bool checkblind);
 
 int32 status_change_spread(block_list *src, block_list *bl);
 
--- a/src/map/status.hpp
+++ b/src/map/status.hpp
@@ -3622,7 +3622,7 @@ void status_calc_state(struct block_list *bl, status_change *sc, std::bitset<SCS
 void status_calc_slave_mode(mob_data& md);
 
 bool status_check_skilluse(struct block_list *src, struct block_list *target, uint16 skill_id, int32 flag);
-int32 status_check_visibility(struct block_list *src, struct block_list *target);
+int32 status_check_visibility(struct block_list *src, struct block_list *target, bool checkblind);
 
 int32 status_change_spread(block_list *src, block_list *bl);
 
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -89,18 +89,17 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
  * @return Whether the chase path was updated (true) or current movement can continue (false)
  */
 bool unit_update_chase(block_list& bl, t_tick tick, bool fullcheck) {
-
 	unit_data* ud = unit_bl2ud(&bl);
 
 	if (ud == nullptr)
 		return true;
 
-	struct block_list* tbl = nullptr;
-	if (ud->target_to)
+	block_list* tbl = nullptr;
+	if (ud->target_to != 0)
 		tbl = map_id2bl(ud->target_to);
 
 	// Reached destination, start attacking
-	if (tbl && tbl->m == bl.m && check_distance_bl(&bl, tbl, ud->chaserange) && status_check_visibility(&bl, tbl, false)) {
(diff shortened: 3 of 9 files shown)


src/map/unit.cpp
```diff
@@ -840,7 +895,7 @@ static TIMER_FUNC(unit_walktobl_sub){
 	struct block_list *bl = map_id2bl(id);
 	struct unit_data *ud = bl?unit_bl2ud(bl):nullptr;
 
-	if (ud && ud->walktimer == INVALID_TIMER && ud->target && ud->target == data) {
+	if (ud && ud->walktimer == INVALID_TIMER && ud->target_to && ud->target_to == data) {
```
Lemongrass3110, line 898:
```suggestion
	if (ud != nullptr && ud->walktimer == INVALID_TIMER && ud->target_to != 0 && ud->target_to == data) {
```

src/map/unit.cpp
```diff
@@ -1702,8 +1755,8 @@ int32 unit_set_walkdelay(struct block_list *bl, t_tick tick, t_tick delay, int32
 			else {
 				unit_stop_walking(bl,4);
 
-				if(ud->target)
-					add_timer(ud->canmove_tick+1, unit_walktobl_sub, bl->id, ud->target);
+				if(ud->target_to)
```
Lemongrass3110, line 1758:
```suggestion
				if(ud->target_to != 0)
```

src/map/unit.cpp
```diff
@@ -2972,8 +3023,25 @@ static TIMER_FUNC(unit_attack_timer){
 
 	bl = map_id2bl(id);
 
-	if(bl && unit_attack_timer_sub(bl, tid, tick) == 0)
-		unit_unattackable(bl);
+	if (bl) {
```
Lemongrass3110, line 3026:
```suggestion
	if (bl != nullptr) {
```

src/map/unit.cpp
```diff
@@ -614,53 +708,14 @@ static TIMER_FUNC(unit_walktoxy_timer)
...
-		// For some reason sometimes the walk timer is not empty here
-		// TODO: Need to check why (e.g. when the monster spams NPC_RUN)
-		if (ud->walktimer != INVALID_TIMER) {
-			delete_timer(ud->walktimer, unit_walktoxy_timer);
-			ud->walktimer = INVALID_TIMER;
-		}
-		ud->walktimer = add_timer(tick+speed,unit_walktoxy_timer,id,speed);
-		if( md && DIFF_TICK(tick,md->dmgtick) < 3000 ) // Not required not damaged recently
-			clif_move( *ud );
+	if(unit_walktoxy_nextcell(*bl, (md && DIFF_TICK(tick, md->dmgtick) < 3000), tick)) {
```
Lemongrass3110, line 711:
```suggestion
	if(unit_walktoxy_nextcell(*bl, (md != nullptr && DIFF_TICK(tick, md->dmgtick) < 3000), tick)) {
```

src/map/unit.cpp
```diff
@@ -320,7 +416,7 @@ TIMER_FUNC(unit_step_timer){
 	} else {
 		//If a player has target_id set and target is in range, attempt attack
 		struct block_list *tbl = map_id2bl(target_id);
-		if (!tbl || !status_check_visibility(bl, tbl)) {
+		if (!tbl || !status_check_visibility(bl, tbl, false)) {
```
Lemongrass3110, line 419:
```suggestion
		if (tbl == nullptr || !status_check_visibility(bl, tbl, false)) {
```

src/map/unit.cpp
```diff
@@ -108,15 +217,16 @@ int32 unit_walktoxy_sub(struct block_list *bl)
 
 	int32 i;
 
-	if (ud->target_to && ud->chaserange>1) {
+	// Monsters always target an adjacent tile even if ranged, no need to shorten the path
+	if (ud->target_to && ud->chaserange>1 && bl->type != BL_MOB) {
```
Lemongrass3110, line 221:
```suggestion
	if (ud->target_to != 0 && ud->chaserange>1 && bl->type != BL_MOB) {
```

src/map/unit.cpp
```diff
@@ -78,6 +78,115 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
...
+
+/**
+ * Handles everything that happens when movement to the next cell is initiated
+ * @param bl: Moving bl
+ * @param sendMove: Whether move packet should be sent or not
+ * @param tick: Current tick
+ * @return Whether movement was initialized (true) or not (false)
+ */
+bool unit_walktoxy_nextcell(block_list& bl, bool sendMove, t_tick tick) {
+
```
Lemongrass3110, line 148:
```suggestion
```

src/map/unit.cpp
```diff
@@ -78,6 +78,115 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
...
+	unit_data* ud = unit_bl2ud(&bl);
+
+	if (ud == nullptr)
+		return true;
+
+	int32 speed;
+
+	if (ud->walkpath.path_pos >= ud->walkpath.path_len)
+		speed = -1;
+	else if (direction_diagonal(ud->walkpath.path[ud->walkpath.path_pos]))
```
Lemongrass3110, line 158:
```suggestion
	if (ud->walkpath.path_pos >= ud->walkpath.path_len) {
		speed = -1;
		return false;
	} else if (direction_diagonal(ud->walkpath.path[ud->walkpath.path_pos]))
```

src/map/unit.cpp
```diff
@@ -78,6 +78,115 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
...
+		// Make sure there is no active walktimer
+		if (ud->walktimer != INVALID_TIMER) {
+			delete_timer(ud->walktimer, unit_walktoxy_timer);
+			ud->walktimer = INVALID_TIMER;
+		}
+		ud->walktimer = add_timer(tick + speed, unit_walktoxy_timer, bl.id, speed);
+		if (sendMove)
+			clif_move(*ud);
+		return true;
+	}
```
Lemongrass3110, line 185: Looks worse than it is.
Just removed 1 tab (see change above that makes the negative speed check obsolete) and changed short to int16
```suggestion
	// Monsters check if their target is in range each cell
	if (bl.type == BL_MOB && ud->target_to) {
		int16 tx = ud->to_x;
		int16 ty = ud->to_y;
		// Monsters update their chase path one cell before reaching their final destination
		if (unit_update_chase(bl, tick, (ud->walkpath.path_pos == ud->walkpath.path_len-1)))
			return true;
		// Continue moving, restore to_x and to_y
		ud->to_x = tx;
		ud->to_y = ty;
	}

	// Make sure there is no active walktimer
	if (ud->walktimer != INVALID_TIMER) {
		delete_timer(ud->walktimer, unit_walktoxy_timer);
		ud->walktimer = INVALID_TIMER;
	}
	ud->walktimer = add_timer(tick + speed, unit_walktoxy_timer, bl.id, speed);
	if (sendMove)
		clif_move(*ud);
	return true;
```

src/map/unit.cpp
```diff
@@ -78,6 +78,115 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
...
+ * If target in attack range -> attack
+ * If target out of sight -> drop target
+ * Otherwise update chase path
+ * @param bl: Moving bl
+ * @param tick: Current tick
+ * @param fullcheck: If false, only check for attack, don't drop target or update chase path
+ * @return Whether the chase path was updated (true) or current movement can continue (false)
+ */
+bool unit_update_chase(block_list& bl, t_tick tick, bool fullcheck) {
+
```
Lemongrass3110, line 92:
```suggestion
```

src/map/unit.cpp
```diff
@@ -78,6 +78,115 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
...
+	unit_data* ud = unit_bl2ud(&bl);
+
+	if (ud == nullptr)
+		return true;
+
+	struct block_list* tbl = nullptr;
```
Lemongrass3110, line 98:
```suggestion
	block_list* tbl = nullptr;
```

src/map/unit.cpp
```diff
@@ -78,6 +78,115 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
...
+	if (ud->target_to)
```
Lemongrass3110, line 99:
```suggestion
	if (ud->target_to != 0)
```

src/map/unit.cpp
```diff
@@ -78,6 +78,115 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
...
+		tbl = map_id2bl(ud->target_to);
+
+	// Reached destination, start attacking
+	if (tbl && tbl->m == bl.m && check_distance_bl(&bl, tbl, ud->chaserange) && status_check_visibility(&bl, tbl, false)) {
```
Lemongrass3110, line 103:
```suggestion
	if (tbl != nullptr && tbl->m == bl.m && check_distance_bl(&bl, tbl, ud->chaserange) && status_check_visibility(&bl, tbl, false)) {
```

src/map/status.hpp
```diff
@@ -3622,7 +3622,7 @@ void status_calc_state(struct block_list *bl, status_change *sc, std::bitset<SCS
 void status_calc_slave_mode(mob_data& md);
 
 bool status_check_skilluse(struct block_list *src, struct block_list *target, uint16 skill_id, int32 flag);
-int32 status_check_visibility(struct block_list *src, struct block_list *target);
+int32 status_check_visibility(struct block_list *src, struct block_list *target, bool checkblind);
```
Lemongrass3110, line 3625:
```suggestion
int32 status_check_visibility(block_list* src, block_list* target, bool checkblind);
```

src/map/status.cpp
```diff
@@ -2257,6 +2258,12 @@ int32 status_check_visibility(struct block_list *src, struct block_list *target)
 			view_range = AREA_SIZE;
 	}
 
+	if (checkblind) {
+		status_change* sc = status_get_sc(src);
+		if (sc && sc->getSCE(SC_BLIND))
```
Lemongrass3110, line 2263:
```suggestion
		if (sc != nullptr && sc->getSCE(SC_BLIND) != nullptr)
```

src/map/status.cpp
```diff
@@ -2239,16 +2239,17 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
  * Checks whether the src can see the target
  * @param src:	Object using skill on target [PC|MOB|PET|HOM|MER|ELEM]
  * @param target: Object being targeted by src [PC|MOB|HOM|MER|ELEM]
+ * @param checkblind: Whether blind condition should be considered (sets view range to 1)
  * @return src can see (1) or target is invisible (0)
  * @author [Skotlex]
  */
-int32 status_check_visibility(struct block_list *src, struct block_list *target)
+int32 status_check_visibility(struct block_list *src, struct block_list *target, bool checkblind)
```
Lemongrass3110, line 2246:
```suggestion
int32 status_check_visibility(block_list* src, block_list* target, bool checkblind)
```

src/map/status.cpp
```diff
@@ -2239,16 +2239,17 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
  * Checks whether the src can see the target
  * @param src:	Object using skill on target [PC|MOB|PET|HOM|MER|ELEM]
  * @param target: Object being targeted by src [PC|MOB|HOM|MER|ELEM]
+ * @param checkblind: Whether blind condition should be considered (sets view range to 1)
  * @return src can see (1) or target is invisible (0)
```
Lemongrass3110, line 2243: If we already touch the function signature, it would be good to change the return type from int32 to bool

Lemongrass3110 (commented):
Just a style review. I will try to understand the logic changes this weekend and give you feedback

src/map/unit.cpp
Playtester: No need to set speed here then, but good point.

src/map/unit.cpp
Playtester: Will apply this part myself (without setting speed to -1).

src/map/status.cpp
Playtester: Done

src/map/unit.cpp
Playtester: Fixed, please see https://github.com/rathena/rathena/pull/8936/commits/abd715baa71756bb29ff30940816227bfa4c5bd6.

src/map/unit.cpp
Playtester: Fixed, please see https://github.com/rathena/rathena/pull/8936/commits/abd715baa71756bb29ff30940816227bfa4c5bd6.

vstumpf (approved):

src/map/status.cpp
```diff
@@ -2239,16 +2239,17 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
...
  */
-int32 status_check_visibility(struct block_list *src, struct block_list *target)
+bool status_check_visibility(block_list* src, block_list* target, bool checkblind)
 {
 	int32 view_range;
 	status_change* tsc = status_get_sc(target);
 	switch (src->type) {
 		case BL_MOB:
-			view_range = ((TBL_MOB*)src)->min_chase;
+			view_range = ((TBL_MOB*)src)->db->range3;
```
Lemongrass3110, line 2252:
```suggestion
			view_range = reinterpret_cast<mob_data*>(src)->db->range3;
```

PR #8942, opened 2025-01-19
Corrected SO_CLOUD_KILL

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

- Corrected skill duration that should scale with skill level
- Added missing flag for SC_CLOUD_POISON (missing visual effect on target)
- Moved extra damage from SC_CLOUD_POISON to battle_calc_cardfix to match damage on official server
- Added missing INT ratio in the skill formula
- SO_CLOUD_KILL now inflicts SC_CLOUD_POISON (100% on first hit, status not refreshed)
- Corrected UNT_CLOUD_KILL skill_damage flag : clif_skill_damage flag should be DMG_SPLASH and skill level -1


src/map/skill.cpp
```diff
@@ -1965,6 +1965,11 @@ int32 skill_additional_effect( struct block_list* src, struct block_list *bl, ui
 	case SO_POISON_BUSTER:
 		sc_start(src,bl, SC_POISON, 5 * skill_lv, skill_lv, skill_get_time(skill_id, skill_lv));
 		break;
+	case SO_CLOUD_KILL: {
+		sc_type sct = skill_get_sc(skill_id);
+		sc_start(src, bl, sct, 100, skill_lv, skill_get_time2(skill_id, skill_lv));
+ 		break;
+	}
```
aleos89, line 1972:
```suggestion
	case SO_CLOUD_KILL:
		sc_start(src, bl, skill_get_sc(skill_id), 100, skill_lv, skill_get_time2(skill_id, skill_lv));
 		break;
```
Was there meant to be a safety check for `sct`?

Atemo: I forgot to update this part after updating the status in status_db.yml

aleos89 (approved):

PR #8959, opened 2025-01-23
Changed all unsigned short to uint16

Lemongrass3110:

* **Server Mode**: Both

Replaced all platform dependent basetype usages of unsigned short by platform safe uint16 usages.

### DIFF
--- a/src/char/char_clif.cpp
+++ b/src/char/char_clif.cpp
@@ -54,7 +54,7 @@ bool pincode_allowed( char* pincode );
 // reason
 // 0: success
 // 1: failed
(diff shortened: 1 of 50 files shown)


aleos89:
Should we take this moment to try to capture mob ID into a `t_mobid` like we do for items or have that in another PR?

Lemongrass3110:
> Should we take this moment to try to capture mob ID into a `t_mobid` like we do for items or have that in another PR?

Guess that will take more than just a "replace all" and mob id is mixed with too many types at the moment. But we can clear it up in the near future.

aleos89 (approved):

Issue #8544, opened 2024-08-07
New Skills 4th Jobs


Pokye:
### Is your request related to a missing official feature?

Yes

### Describe the solution you'd like

In the 2024-07-08 kRO update, new skills were implemented.

**Shiranui/Shinkiro**
https://www.divine-pride.net/database/skill/5499/
**Night Watch**
https://www.divine-pride.net/database/skill/5500/
https://www.divine-pride.net/database/skill/5501/
**Imperial Guard**
https://www.divine-pride.net/database/skill/6503/
https://www.divine-pride.net/database/skill/6504/
**Meister**
https://www.divine-pride.net/database/skill/6506/
https://www.divine-pride.net/database/skill/6507/
https://www.divine-pride.net/database/skill/6508/
**Biolo**
https://www.divine-pride.net/database/skill/6509/
https://www.divine-pride.net/database/skill/6510/
**Shadow Cross**
https://www.divine-pride.net/database/skill/6511/
**Abyss Chaser**
https://www.divine-pride.net/database/skill/6512/
https://www.divine-pride.net/database/skill/6513/
https://www.divine-pride.net/database/skill/6514/
Troubadour/Trouvere
https://www.divine-pride.net/database/skill/6521/

https://www.divine-pride.net/forum/index.php?/topic/7803-kro-main-changelog-august-07-2024/

iberryRO:
#8661 
#8662 
#8663 
#8664 
#8665
#8666 
#8671 
#8673

iberryRO:
New 14 skill
	DK_DRAGONIC_PIERCE = 6502,
	IG_IMPERIAL_PRESSURE = 6505,
	ABC_ABYSS_FLAME = 6515,
	ABC_ABYSS_FLAME_ATK = 6522,
	AG_ENERGY_CONVERSION = 6516,
	EM_PSYCHIC_STREAM = 6517,
	CD_DIVINUS_FLOS = 6518,
	IQ_BLAZING_FLAME_BLAST = 6519,
	WH_WILD_WALK = 6520,
	SKE_SKY_SUN = 5502,
	SKE_SKY_MOON = 5503,
	SKE_STAR_LIGHT_KICK = 5504,
	HN_OVERCOMING_CRISIS = 5505,
	SH_CHUL_HO_BATTERING = 5506,
	SH_HYUN_ROK_SPIRIT_POWER = 5507

[...]

PR #8983, opened 2025-01-25
Implemented CD_DIVINUS_FLOS

Haydrich:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8544

* **Server Mode**: RE

* **Description of Pull Request**: Implementation of the new skill of the Cardinal


db/re/skill_db.yml
```diff
@@ -45018,6 +45018,36 @@ Body:
...
+    Range: 9
+    Hit: Multi_Hit
+    HitCount: 5
+    Element: Holy
+    SplashArea: 3
+    GiveAp: 2
+    CastCancel: true
+    FixedCastTime: 1500
+    AfterCastActDelay: 500
+    Cooldown: 2500
```
Atemo:
```suggestion
    CastTime: 2000
    AfterCastActDelay: 850
    Cooldown: 700
    FixedCastTime: 1500
```

db/re/skill_db.yml
```diff
@@ -45018,6 +45018,37 @@ Body:
...
+    Name: CD_DIVINUS_FLOS
+    Description: Divinus Flos
+    MaxLevel: 5
+    Type: Magic
+    TargetType: Attack
+    DamageFlags:
+      Splash: true
+    Range: 9
+    Hit: Multi_Hit
+    HitCount: 5
```
Atemo: I guess Hitcount should be negative? 5 hits with a base skillratio of 20000 seems unbalanced. Please confirm (if you know)
```suggestion
    HitCount: -5
```

src/map/battle.cpp
```diff
@@ -3595,8 +3595,9 @@ int32 battle_get_magic_element(struct block_list* src, struct block_list* target
 			if(sd && sd->spiritcharm_type != CHARM_TYPE_NONE && sd->spiritcharm > 0)
 				element = sd->spiritcharm_type;
 			break;
+		case CD_DIVINUS_FLOS:
 		case AB_ADORAMUS:
-			if (sc && sc->getSCE(SC_ANCILLA))
+			if (sc != nullptr && sc->getSCE(SC_ANCILLA) != nullptr)
```
Atemo:
```suggestion
			if (sc != nullptr && sc->hasSCE(SC_ANCILLA))
```

src/map/battle.cpp
```diff
@@ -8747,6 +8748,12 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
 							skillratio += 50 * skill_lv;
 						RE_LVL_DMOD(100);
 						break;
+					case CD_DIVINUS_FLOS:
+						skillratio += -100 + 4000 * skill_lv;
+						skillratio += 70 * pc_checkskill(sd,CD_FIDUS_ANIMUS);
+						skillratio += 5 * sstatus->spl;
```
Atemo: Can you confirm the ratio? otherwise :
```suggestion
						skillratio += 5 * sstatus->spl;	// !TODO: unknown ratio
```

Atemo (approved):

db/re/skill_db.yml
Haydrich: I'll keep it here until i test

src/map/battle.cpp
```diff
@@ -8747,6 +8748,12 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
 							skillratio += 50 * skill_lv;
 						RE_LVL_DMOD(100);
 						break;
+					case CD_DIVINUS_FLOS:
+						skillratio += -100 + 4000 * skill_lv;
+						skillratio += 70 * pc_checkskill(sd,CD_FIDUS_ANIMUS);
+						skillratio += 5 * sstatus->spl;	// !TODO: unknown ratio
```
Haydrich:
```suggestion
						skillratio += 10 * sstatus->spl;
```

db/re/skill_db.yml
Haydrich: Confirmed. Divinus Flos is -5 but also it's able to damage Plants.

db/re/skill_db.yml
```diff
@@ -45018,6 +45018,37 @@ Body:
...
     Status: WeaponBlock_On
+  - Id: 6518
+    Name: CD_DIVINUS_FLOS
+    Description: Divinus Flos
+    MaxLevel: 5
+    Type: Magic
+    TargetType: Attack
+    DamageFlags:
+      Splash: true
+    Range: 9
```
Haydrich, line 45810:
```suggestion
    #Flags:
      #MultiHitOnPlants: true
    Range: 9
```
Currently negative hit count is not damaging plants, only normal values

Lemongrass3110: Can you add this commented (and with a TODO mark) for the time being?
This should be enabled with #9491 afterwards

db/re/skill_db.yml
Lemongrass3110: So this change could be accepted?

src/map/battle.cpp
Lemongrass3110: Should this change be pushed?

Lemongrass3110 (approved):

db/re/skill_db.yml
Atemo: Yes but it seems that it requires https://github.com/rathena/rathena/pull/9491

Haydrich: So i can add it with commented flag MultiHitOnPlants then on #9491 i'll uncomment it.

PR #8994, opened 2025-01-26
Converted CZ_MOVE_ITEM_FROM_CART_TO_STORE to struct

Lemongrass3110:

* **Server Mode**: Both

Converted CZ_MOVE_ITEM_FROM_CART_TO_STORE to struct


src/map/clif.cpp
```diff
@@ -13567,19 +13566,32 @@ void clif_parse_MoveToKafraFromCart(int32 fd, map_session_data *sd){
...
+	}
+
+	if( p->amount > MAX_AMOUNT ){
 		return;
+	}
+
 	if( sd->inventory.u.items_inventory[idx].equipSwitch ){
 		clif_msg( sd, MSI_SWAP_EQUIPITEM_UNREGISTER_FIRST );
 		return;
 	}
```
Lemongrass3110, line 13587:
```suggestion
```
Cannot be on equip switch, when it is inside cart.

src/map/clif.cpp
```diff
@@ -13567,19 +13566,32 @@ void clif_parse_MoveToKafraFromCart(int32 fd, map_session_data *sd){
...
+	}
+
+	if( p->amount <= 0 ){
+		return;
+	}
+
+	if( p->amount > MAX_AMOUNT ){
 		return;
+	}
+
```
Lemongrass3110, line 13583:
```suggestion
```
Checks are done in storage_storageaddfromcart

aleos89 (approved):

PR #9013, opened 2025-01-27
Monster Activity on Teleport, Assist/Slave Behavior, NPC_EMOTION

### LINKED ISSUE #8933: NPC_EMOTION can cause server crash
### rAthena Hash

3a9240a783

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

Monsters casting NPC_EMOTION can result in a server crash in rare occurences, when the following conditions are met:

1. NPC_EMOTION is reached in skill_castend_nodamage_id and mob_unlocktarget is called line 10071
https://github.com/rathena/rathena/blob/3a9240a783547dbfd52546cc279badebd665ef86/src/map/skill.cpp#L10055-L10073

2. `DIFF_TICK(tick,` md->last_skillcheck) >= `MOB_SKILL_INTERVAL` yields true and mobskill_use is called line 1564
https://github.com/rathena/rathena/blob/3a9240a783547dbfd52546cc279badebd665ef86/src/map/mob.cpp#L1552-L1565

3. [`md->skill_idx = -1;`](https://github.com/rathena/rathena/blob/3a9240a783547dbfd52546cc279badebd665ef86/src/map/mob.cpp#L4130) is reached inside `mobskill_use`

4. Back to 1., `md->db->skill[md->skill_idx]` is reached, tries to access the array to -1 thus causing the crash.

### Relevant Log Output

_No response_

### Expected Result

N/A

### How to Reproduce

Any action that could cause `md->last_skillcheck` to be higher than MOB_SKILL_INTERVAL.
The reasons it happens in normal play are still unclear for me, but spawning hundreds of mobs with frequent idle usage of NPC_EMOTION such as Banshees can do it.
The guaranteed way is to put a break point after having spawned the mobs, thus having enough tick difference.

### Official Information

N/A

_No response_

### LINKED ISSUE #8964: Any kind of warp (teleport, recall slave) should reset monster to inactive
### rAthena Hash

ea359b4

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When a monster teleports, is recalled or gets warped out of any other reason, its AI stays active.

### Expected Result

When a monster teleports, is recalled or gets warped out of any other reason, its AI goes inactive (it goes active again if a player is within 18 cells).

### How to Reproduce

Use Convex Mirror and hunt Mistress / Lady Tanee. You can see they only teleport when you're nearby, then stay in place.

Another good test is to trap a monster with slaves and recall in icewalls and move to 19 distance. After a while you can see that the master is still moving but without the slaves (as they got inactive from the recall).

### Official Information

Tested on official servers. See above.

_No response_

Playtester:
* **Addressed Issue(s)**: #8964, #8933 

* **Server Mode**: Both

- When monsters teleport or are otherwise removed from the map they now become inactive immediately (fixes #8964)
  * This for example means that when Mistress and Lady Tanee teleport, they won't teleport again
  * They become active again when a player is within 18 cells of them
  * If you want them to stay active, please set mob_active_time and boss_active_time to the desired values
- When slaves are recalled, they now also become inactive, even if their master is still active
  * Added a new option slave_active_with_master if you still want them to be active with their master
  * All slave-related options are now in one place, to keep a better overview
- When you enable slave_stick_with_master, slaves now drop their target if outside the visible area
  * This fixes them looping from their master back out of allowed range
- Fixed an issue that caused change-chase monsters to keep target_to on their old target when attacking
  * This also caused slaves to attack the wrong target
- Assist range is now 11
  * This refers to the range between two assist monsters to trigger assist aggro
  * The value is now configurable in monster.conf
- Assist no longer has a range limit between the assist mob and the target
  * But it might drop target again immediately if walkpath fails due to being too long
- Reduced MIN_MOBLINKTIME to 300ms by default
  * This causes assist and slave monsters to be more reactive
  * If you want fully official behavior both MIN_MOBTHINKTIME and MIN_MOBLINKTIME need to be set to 20
  * If this takes too many resources on your server, just increase this back to 1000ms
- Fixed NPC_EMOTION recursively calling mobskill_use, potentially causing a crash (fixes #8933)

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -2984,7 +2984,7 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
 			if (status_has_mode(sstatus,MD_ASSIST) && DIFF_TICK(tick, md->last_linktime) >= MIN_MOBLINKTIME) { 
 				// Link monsters nearby [Skotlex]
 				md->last_linktime = tick;
-				map_foreachinrange(mob_linksearch, src, md->db->range2, BL_MOB, md->mob_id, target, tick);
+				map_foreachinshootrange(mob_linksearch, src, battle_config.assist_range, BL_MOB, md->mob_id, target->id, tick);
 			}
 		}
 
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -1440,10 +1440,26 @@ int32 unit_warp(struct block_list *bl,int16 m,int16 x,int16 y,clr_type type)
 	bl->y = ud->to_y = y;
 	bl->m = m;
 
-	if (bl->type == BL_NPC) {
-		TBL_NPC *nd = (TBL_NPC*)bl;
-		map_addnpc(m, nd);
-		npc_setcells(nd);
+	switch (bl->type) {
+		case BL_NPC:
+		{
+			TBL_NPC* nd = reinterpret_cast<npc_data*>(bl);
+			map_addnpc(m, nd);
+			npc_setcells(nd);
+			break;
+		}
+		case BL_MOB:
+		{
+			TBL_MOB* md = reinterpret_cast<mob_data*>(bl);
+			// If slaves are set to stick with their master they should drop target if recalled at range
+			if (battle_config.slave_stick_with_master && md->target_id != 0) {
+				block_list* tbl = map_id2bl(md->target_id);
+				if (tbl == nullptr || !check_distance_bl(bl, tbl, AREA_SIZE)) {
+					md->target_id = 0;
+				}
+			}
+			break;
+		}
 	}
 
 	if(map_addblock(bl))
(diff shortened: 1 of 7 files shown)


src/map/unit.cpp
```diff
@@ -3412,6 +3428,10 @@ int32 unit_remove_map_(struct block_list *bl, clr_type clrtype, const char* file
 			if (!battle_config.mob_slave_keep_target)
 				md->target_id=0;
 
+			// When a monster is removed from map, its spotted log is cleared
+			for (int32 i = 0; i < DAMAGELOG_SIZE; i++)
+				md->spotted_log[i] = 0;
```
aleos89, line 3433: On mobile right now, but I think we have that ARRAYCLEAR() macro which could be used here until this gets flipped to a dynamic array type.

PR #9031, opened 2025-02-01
Converted CZ_CHANGE_GROUPEXPOPTION to struct

Lemongrass3110:

* **Server Mode**: Both

Converted CZ_CHANGE_GROUPEXPOPTION to struct

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -13937,31 +13937,41 @@ void clif_parse_RemovePartyMember( int32 fd, map_session_data* sd ){
 /// 07d7 <exp share rule>.L <item pickup rule>.B <item share rule>.B (CZ_GROUPINFO_CHANGE_V2)
 void clif_parse_PartyChangeOption(int32 fd, map_session_data *sd)
 {
-	struct party_data *p;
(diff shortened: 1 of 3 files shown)


src/map/clif.cpp
```diff
@@ -13937,31 +13937,41 @@ void clif_parse_RemovePartyMember( int32 fd, map_session_data* sd ){
...
-	if( p == nullptr )
+	if( party == nullptr ){
 		return;
-	ARR_FIND( 0, MAX_PARTY, i, p->data[i].sd == sd );
-	if( i == MAX_PARTY )
-		return; //Shouldn't happen
-	if( !p->party.member[i].leader )
+	}
+
+	if( !party_isleader( sd ) ){
```
Atemo, line 13954: Some parts could possibly be removed
- `if( sd->status.party_id == 0 ){` is also in `party_search`
- `party_search` is also used in `party_isleader`

Also should we accept negative value for party_id? (`int32 party_id` in `struct mmo_charstatus`) shouldn't it be `if( sd->status.party_id <= 0 ){` ?

Issue #1136, opened 2016-04-04
Eska


Playtester:
Currently on rAthena Eska sets HardDEF to a random value between 0 and 100. It is also coded that this value changes every second, but that's not working anyway.
Also Eska's effect currently doesn't disappear when the skill ends.

Testing Eska on Renewal:
- Physical damage decreases by a fixed 80 no matter against which monster
- Does NOT increase damage from Occult Impact

Testing Eska on Pre-Renewal:
- Physical damage decreases by 0~99 no matter against which monster (measured 17~88)
- The damage varies for every single hit no matter how fast you attack
- Damage from Occult Impact is increased and varies strongly (with STR 113, +0 Ice Pick it varies from 6~269)

Conclusion: Eska actually increases the randomness part of the soft DEF formula by 100. In renewal the randomness part was removed and instead is multiplied by 0.8x.
There also shouldn't be an interval on the status change.

Also on Aegis the effect disappears when the status change ends.

Playtester:
Don't really want to reopen this issue, but my assumption that the increase goes up to 99 was wrong, it's actually 0-89 (so it increases the random factor by 90).

PR #9032, opened 2025-02-01
ASPD cap, Slow Grace, Eska

### LINKED ISSUE #8937: Too low cap for aMotion / Slow Grace
### rAthena Hash

fb637d9

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

We cap aMotion to 10-2000 and aDelay to 20-4000.

Some monsters like Dryad that have official values above 2000 are attacking faster than officially because of that.

Players can't get below 0 ASPD.

Slow Grace reduces ASPD of monsters and mercenaries.

### Expected Result

aMotion has natural values up to 4032 (Red Lichtern has highest official value).

Dryad attacks every 2520ms.

Officially, this value can be increased to up to 32767 after which it would cause an integer overflow and become -32768 which is then capped to the minimum value of 100 (195 ASPD).

To prevent exploits, I propose to cap the values as following for monsters:
aMotion 100-8000
aDelay 100-8000
(-200 to 195 ASPD)

Players can get negative ASPD and attack less often than every 4000ms.

To prevent exploits, I propose to cap the values as following for PC:
aMotion 100-4000
aDelay 200-8000
(-200 to 190 ASPD)

Slow Grace's ASPD penalty only works on players. In pre-renewal there is an exploit that it gives a -1702296% ASPD penalty to monsters but it's not directly recalculated, you need to trigger the calculation of ASPD through other buffs or debuffs. This bug was "fixed" in renewal by making Slow Grace only usable in PVP.

### How to Reproduce

Compare attack speed of Dryad with official servers. Then put Dryad in Slow Grace and compare attack speed again.

Priest with Waghnak and 4 Howard Eisen Cards has -38 ASPD without no Agi/Dex bonus in pre-renewal. Also attacks only every 4.76 seconds then.

### Official Information

Tested on official servers.

Issue was created by this commit: https://github.com/rathena/rathena/commit/453af93a07c1a9f83002f7abe263c60f6cce45c7

Playtester:
* **Addressed Issue(s)**: #8937, #1136

* **Server Mode**: Both

- ASPD of monsters now allows values between -200 ASPD and 195 ASPD (100-8000ms)
  * Maximum ASPD value is configurable
  * Base ASPD value is no longer limited, but will be capped after adjustments
- ASPD of players can now go as low as -200 ASPD (8000ms)
- Slow Grace no longer adjusts ASPD of non-players
- Fixes #8937
- Eska no longer adjusts ASPD and movement speed
- Eska now sets the soft def variance to 90 instead of increasing it by 100
- Follow-up to 2369f39 (see #1136)

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -4571,8 +4571,8 @@ s_mob_db::s_mob_db()
 	status.luk = 1;
 	status.ele_lv = 1;
 	status.speed = DEFAULT_WALK_SPEED;
-	status.adelay = cap_value(0, battle_config.monster_max_aspd * 2, 4000);
-	status.amotion = cap_value(0, battle_config.monster_max_aspd, 2000);
+	status.adelay = battle_config.monster_max_aspd;
+	status.amotion = battle_config.monster_max_aspd;
 	status.clientamotion = cap_value(status.amotion, 1, USHRT_MAX);
 	status.mode = static_cast<e_mode>(MONSTER_TYPE_06);
 
@@ -4985,7 +4985,8 @@ uint64 MobDatabase::parseBodyNode(const ryml::NodeRef& node) {
 		if (!this->asUInt16(node, "AttackDelay", speed))
 			return 0;
 
-		mob->status.adelay = cap_value(speed, battle_config.monster_max_aspd * 2, 4000);
+		// The base value can be below monster_max_aspd
+		mob->status.adelay = cap_value(speed, 1, 8000);
 	}
 	
 	if (this->nodeExists(node, "AttackMotion")) {
@@ -4994,7 +4995,8 @@ uint64 MobDatabase::parseBodyNode(const ryml::NodeRef& node) {
 		if (!this->asUInt16(node, "AttackMotion", speed))
 			return 0;
 
-		mob->status.amotion = cap_value(speed, battle_config.monster_max_aspd, 2000);
+		// The base value can be below monster_max_aspd
+		mob->status.amotion = cap_value(speed, 1, 8000);
 	}
 
 	if (this->nodeExists(node, "ClientAttackMotion")) {
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -4566,7 +4566,7 @@ int32 status_calc_pc_sub(map_session_data* sd, uint8 opt)
 
 	// Basic ASPD value
 	i = status_base_amotion_pc(sd,base_status);
-	base_status->amotion = cap_value(i,pc_maxaspd(sd),2000);
+	base_status->amotion = cap_value(i,pc_maxaspd(sd),4000);
 
 	// Relative modifiers from passive skills
 	// Renewal modifiers are handled in status_base_amotion_pc
@@ -6166,7 +6166,7 @@ void status_calc_bl_main(struct block_list& bl, std::bitset<SCB_MAX> flag)
 			amotion += sd->bonus.aspd_add;
 #endif
 			amotion = status_calc_fix_aspd(&bl, sc, amotion);
-			status->amotion = cap_value(amotion,pc_maxaspd(sd),2000);
+			status->amotion = cap_value(amotion,pc_maxaspd(sd),4000);
 
 			status->adelay = 2 * status->amotion;
 		} else { // Mercenary and mobs
@@ -6175,10 +6175,10 @@ void status_calc_bl_main(struct block_list& bl, std::bitset<SCB_MAX> flag)
 			amotion = amotion*status->aspd_rate/1000;
 
 			amotion = status_calc_fix_aspd(&bl, sc, amotion);
-			status->amotion = cap_value(amotion, battle_config.monster_max_aspd, 2000);
+			status->amotion = cap_value(amotion, battle_config.monster_max_aspd, 8000);
 
 			temp = b_status->adelay*status->aspd_rate/1000;
-			status->adelay = cap_value(temp, battle_config.monster_max_aspd*2, 4000);
+			status->adelay = cap_value(temp, battle_config.monster_max_aspd, 8000);
 		}
 	}
 
@@ -7865,8 +7865,6 @@ static uint16 status_calc_speed(struct block_list *bl, status_change *sc, int32 
 				val = max( val, sc->getSCE(SC_SUITON)->val3 );
 			if( sc->getSCE(SC_SWOO) )
 				val = max( val, 300 );
-			if( sc->getSCE(SC_SKA) )
-				val = max( val, 25 );
 			if( sc->getSCE(SC_FREEZING) )
 				val = max( val, 30 );
 			if( sc->getSCE(SC_MARSHOFABYSS) )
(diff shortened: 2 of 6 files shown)


src/map/status.hpp
```diff
@@ -36,6 +36,25 @@ class status_change;
...
+#ifndef MIN_ASPD
+	#define MIN_ASPD 8000
+#endif
+
+// DO NOT MODIFY THESE!
+// The amotion divider defines the relation between amotion and adelay
+// For players amotion is half of adelay
+#define AMOTION_DIVIDER_PC 2
+// For monsters, homunculus, mercenaries and elementals, amotion and adelay are equal or independent
+#define AMOTION_DIVIDER_NOPC 1
```
vstumpf, line 52: If this is always 1, why do we need it?

src/map/status.hpp
```diff
@@ -36,6 +36,25 @@ class status_change;
...
+// Delay between attacks that represents 0 ASPD
+#define ZERO_ASPD 4000
```
vstumpf, line 54: Is this supposed to be exactly 4000, or half of MIN_ASPD?

src/map/status.cpp
```diff
@@ -8139,7 +8144,7 @@ static int16 status_calc_aspd(struct block_list *bl, status_change *sc, bool fix
 static int16 status_calc_fix_aspd(struct block_list *bl, status_change *sc, int32 aspd)
 {
 	if (sc == nullptr || sc->empty())
-		return cap_value(aspd, 0, 2000);
+		return cap_value(aspd, 1, 8000);
```
vstumpf, line 8147: Use MIN_ASPD?

src/map/status.hpp
Playtester: Lemongrass requested it so we don't have fixed number in the source code.

src/map/status.hpp
Playtester: It's always 4000, because 0 ASPD means 4000ms delay. You can freely change MIN_ASPD, but what 0 ASPD means will never change.

src/map/status.cpp
Playtester: True, I missed this one here, will fix it.

src/map/status.cpp
```diff
@@ -2351,8 +2350,13 @@ int32 status_base_amotion_pc(map_session_data* sd, struct status_data* status)
...
+	aspd = ((int32)(temp_aspd + ((float)(status_calc_aspd(&sd->bl, &sd->sc, true) + val) * status->agi / 200)) - min(aspd, 200));
+	return aspd;
 #else
+	if (job == nullptr)
+		return ZERO_ASPD/AMOTION_DIVIDER_PC;
+
+	int32 amotion;
 	// Angra Manyu disregards aspd_base and similar
 	if (pc_checkequip2(sd, ITEMID_ANGRA_MANYU, EQI_ACC_L, EQI_MAX))
 		return 0;
```
Lemongrass3110, line 2362: Shouldnt this return MAX_ASPD for the specific job?

Note: If we can dehardcode this, that would also be good.

src/map/status.cpp
```diff
@@ -2351,8 +2350,13 @@ int32 status_base_amotion_pc(map_session_data* sd, struct status_data* status)
...
 	else if (pc_isridingdragon(sd))
 		val -= 25 - 5 * pc_checkskill(sd, RK_DRAGONTRAINING);
-	amotion = ((int32)(temp_aspd + ((float)(status_calc_aspd(&sd->bl, &sd->sc, true) + val) * status->agi / 200)) - min(amotion, 200));
+	aspd = ((int32)(temp_aspd + ((float)(status_calc_aspd(&sd->bl, &sd->sc, true) + val) * status->agi / 200)) - min(aspd, 200));
+	return aspd;
 #else
+	if (job == nullptr)
+		return ZERO_ASPD/AMOTION_DIVIDER_PC;
+
+	int32 amotion;
```
Lemongrass3110, line 2359: Not needed here, please define it where it is first used (3 lines below).

Lemongrass3110 (approved):
As far as I do understand that PR, it looks good. But I must admit I dont fully understand everything that is being changed here.

But: You are the expert in this field and I have confidence that you know what you are doing!

src/map/mercenary.cpp
```diff
@@ -864,7 +864,7 @@ uint64 MercenaryDatabase::parseBodyNode(const ryml::NodeRef& node) {
 		if (!this->asUInt16(node, "AttackDelay", speed))
 			return 0;
 
-		mercenary->status.adelay = cap_value(speed, 0, 4000);
+		mercenary->status.adelay = cap_value(speed, 1, MIN_ASPD);
 	} else {
 		if (!exists)
 			mercenary->status.adelay = 4000;
```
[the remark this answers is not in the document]
Playtester: I was thinking that these are default values instead of min/max values. Like how for elementals someone put 504 aDelay and 1020 aMotion (not sure where that came from). But I don't really have any means to verify this.

aMotion=2000 would display as 0 ASPD for mercenary and aDelay=4000 would be the player-like represenation of that value.

src/map/mob.cpp
```diff
@@ -4985,7 +4985,8 @@ uint64 MobDatabase::parseBodyNode(const ryml::NodeRef& node) {
 		if (!this->asUInt16(node, "AttackDelay", speed))
 			return 0;
 
-		mob->status.adelay = cap_value(speed, battle_config.monster_max_aspd * 2, 4000);
+		// The base value can be below MAX_ASPD_NOPC
```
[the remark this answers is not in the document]
Playtester: We tested this with Mummy:
```
    AttackDelay: 1772
    AttackMotion: 72
```
As long as you don't buff or debuff a Mummy's ASPD it will still send 72 as aMotion to the client. Only when the ASPD value is modified AttackMotion gets capped to 100.

Also it affects how debuffs work. If we cap it to 100 here, then a debuff that halves you attack speed would set your aMotion to 200 instead of 144.

Let me run some more checks on this though.

src/map/status.cpp
```diff
@@ -2304,26 +2304,25 @@ bool status_check_visibility(block_list* src, block_list* target, bool checkblin
...
  *	  base amotion after single/dual weapon and stats adjustments [PRE-RENEWAL]
  */
 int32 status_base_amotion_pc(map_session_data* sd, struct status_data* status)
 {
 	std::shared_ptr<s_job_info> job = job_db.find(sd->status.class_);
 
+#ifdef RENEWAL_ASPD
 	if (job == nullptr)
-		return 2000;
+		return 0;
```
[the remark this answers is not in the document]
Playtester: In renewal, this function returns the ASPD value and not the amotion value, see also my updated function comment.

src/map/status.cpp
Playtester: This is just some custom feature. You can accomplish the same thing by giving the item a high ASPD bonus. We could just remove it? Maybe a separate PR. Just fixing the two GM items for both pre-re and re.

src/map/status.cpp
Playtester: Improved as suggested.

src/map/status.cpp
Lemongrass3110: Ok fine for me!

src/map/mob.cpp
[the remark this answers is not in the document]
Playtester: Finished investigation and added some improvements.

The delay between attacks can never be faster than 100ms, however, the amotion value that is send to the client is below 100 until there is a buff/debuff effect, which then caps it to 100ms if it is below.

I added capping for adelay to 100ms on Monsters, Mercenaries and Elementals as that should result in the same effect for now.

amotion needs to stay uncapped when loading the values from the db and only calculated on buff/debuff. I made that clear with a comment I added for all three bl types.

PR #9046, opened 2025-02-08
Cleanup for pc_setreg2 and pc_readreg2

Lemongrass3110:

* **Server Mode**: Both

Cleanup for pc_setreg2 and pc_readreg2
Use existing functions, rather than hardcoding conditions again.

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -11476,19 +11476,19 @@ bool pc_setreg2(map_session_data *sd, const char *reg, int64 val) {
 
 	nullpo_retr(false, sd);
 
-	if (reg[strlen(reg)-1] == '$') {
-		ShowError("pc_setreg2: Invalid variable scope '%s'\n", reg);
+	if( is_string_variable( reg ) ){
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -11476,28 +11476,26 @@ bool pc_setreg2(map_session_data *sd, const char *reg, int64 val) {
...
-	if (reg[strlen(reg)-1] == '$') {
-		ShowError("pc_setreg2: Invalid variable scope '%s'\n", reg);
+	if( is_string_variable( reg ) ){
+		ShowError( "pc_setreg2: Invalid variable '%s'. String type variables are not supported.\n", reg );
 		return false;
 	}
 
+	if( !not_server_variable( prefix ) ){
+		ShowError( "pc_setreg2: Invalid variable scope '%s'.\n", reg );
+		return 0;
```
Lemongrass3110, line 11486:
```suggestion
		return false;
```

aleos89 (approved):

PR #9069, opened 2025-02-12
Improved Alchemist Summon Setting

Playtester:
* **Addressed Issue(s)**: (Feature)

* **Server Mode**: Both (though the default setting does not impact pre-renewal)

- Replaced "summon_flora_setting" with "alchemist_summon_setting"
- Allows you to now also set if Marine Spheres can be attacked or not
- Allows you now to set whether Marine Spheres can hit your other summons outside PVP
- As behavior is different between pre-re and re, the config is commented out by default now
  * As long as it's commented, it automatically uses the official behavior
  * In pre-re all options are enabled by default
  * In renewal it's no longer possible to attack summons yourself at all by default

### DIFF
--- a/conf/battle/skill.conf
+++ b/conf/battle/skill.conf
@@ -125,12 +125,15 @@ gvg_traps_target_all: 1
 // Default on official servers: 0 for Pre-renewal, 2 for Renewal
 //traps_setting: 0
 
-// Restrictions applied to the Alchemist's Summon Flora skill (add as necessary)
+// Restrictions applied to the Alchemist's Summon skills (add as necessary)
 // 1: Enable players to damage the floras outside of versus grounds.
-// 2: Disable having different types out at the same time
+// 2: Enable players to damage marine spheres outside of versus grounds.
+// 4: Disable having different types out at the same time
 //    (eg: forbid summoning anything except hydras when there's already 
 //     one hydra out)
-summon_flora_setting: 3
+// 8: Enable Marine Spheres to damage own Homunculus and summons outside PVP
+// Default on official servers: 15 for Pre-renewal, 12 for Renewal
+//alchemist_summon_setting: 15
 
 // Whether placed down skills will check walls (Note 1)
 // (ex. Storm Gust cast against a wall will not hit the other side.) 
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -10993,8 +10993,8 @@ int32 battle_check_target( struct block_list *src, struct block_list *target,int
 
 			if (ud && ud->immune_attack)
 				return 0;
-			if(((md->special_state.ai == AI_SPHERE || //Marine Spheres
-				(md->special_state.ai == AI_FLORA && battle_config.summon_flora&1)) && s_bl->type == BL_PC && src->type != BL_MOB) || //Floras
+			if((((md->special_state.ai == AI_SPHERE && battle_config.alchemist_summon_setting&2) || //Marine Spheres
+				 (md->special_state.ai == AI_FLORA && battle_config.alchemist_summon_setting&1)) && s_bl->type == BL_PC && src->type != BL_MOB) || //Floras
 				(md->special_state.ai == AI_ZANZOU && t_bl->id != s_bl->id) || //Zanzou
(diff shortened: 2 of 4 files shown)


conf/battle/skill.conf
```diff
@@ -125,12 +125,15 @@ gvg_traps_target_all: 1
 // Default on official servers: 0 for Pre-renewal, 2 for Renewal
 //traps_setting: 0
 
-// Restrictions applied to the Alchemist's Summon Flora skill (add as necessary)
+// Restrictions applied to the Alchemist's Summon skills (add as necessary)
 // 1: Enable players to damage the floras outside of versus grounds.
-// 2: Disable having different types out at the same time
+// 2: Enable players to damage marine spheres outside of versus grounds.
+// 4: Disable having different types out at the same time
```
Lemongrass3110, line 131: This is really bad.
Please switch 2 and 4.
Otherwise this affects existing configurations.

src/map/skill.cpp
```diff
@@ -9038,8 +9038,14 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
 	case NPC_SELFDESTRUCTION:
 		//Self Destruction hits everyone in range (allies+enemies)
 		//Except for Summoned Marine spheres on non-versus maps, where it's just enemies and your own slaves.
-		i = ((!md || md->special_state.ai == AI_SPHERE) && !map_flag_vs(src->m))?
-			BCT_ENEMY|BCT_SLAVE:BCT_ALL;
+		if ((!md || md->special_state.ai == AI_SPHERE) && !map_flag_vs(src->m)) {
```
Lemongrass3110, line 9041:
```suggestion
		if ((md == nullptr || md->special_state.ai == AI_SPHERE) && !map_flag_vs(src->m)) {
```

Lemongrass3110 (changes requested):

conf/battle/skill.conf
Playtester: It affects existing configurations anyway since the setting has a new name. This order of settings is more logical.
And it should affect existing configurations since it's currently defaulting incorrectly for renewal.

src/map/battle.cpp
```diff
@@ -10993,8 +10993,8 @@ int32 battle_check_target( struct block_list *src, struct block_list *target,int
 
 			if (ud && ud->immune_attack)
 				return 0;
-			if(((md->special_state.ai == AI_SPHERE || //Marine Spheres
-				(md->special_state.ai == AI_FLORA && battle_config.summon_flora&1)) && s_bl->type == BL_PC && src->type != BL_MOB) || //Floras
+			if((((md->special_state.ai == AI_SPHERE && battle_config.alchemist_summon_setting&2) || //Marine Spheres
+				 (md->special_state.ai == AI_FLORA && battle_config.alchemist_summon_setting&1)) && s_bl->type == BL_PC && src->type != BL_MOB) || //Floras
```
Lemongrass3110, line 10997:
```suggestion
				(md->special_state.ai == AI_FLORA && battle_config.alchemist_summon_setting&1)) && s_bl->type == BL_PC && src->type != BL_MOB) || //Floras
```

src/map/skill.cpp
```diff
@@ -9038,8 +9038,14 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
 	case NPC_SELFDESTRUCTION:
 		//Self Destruction hits everyone in range (allies+enemies)
 		//Except for Summoned Marine spheres on non-versus maps, where it's just enemies and your own slaves.
-		i = ((!md || md->special_state.ai == AI_SPHERE) && !map_flag_vs(src->m))?
-			BCT_ENEMY|BCT_SLAVE:BCT_ALL;
+		if ((md == nullptr || md->special_state.ai == AI_SPHERE) && !map_flag_vs(src->m)) {
+			if (battle_config.alchemist_summon_setting&8)
```
Lemongrass3110, line 9042:
```suggestion
			// Enable Marine Spheres to damage own Homunculus and summons outside PVP
			if (battle_config.alchemist_summon_setting&8)
```

Lemongrass3110 (approved):

src/map/battle.cpp
Playtester: I intentionally added a space here, so you can see that the two lines belong to an inner () because I had missed that before.

PR #9082, opened 2025-02-16
Chase Skill Fixes and Code Optimization

### LINKED ISSUE #8989: Stopping on Chase Skills
### rAthena Hash

ec70190

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When a monster wants to start chasing and it uses NPC_EMOTION (and potentially other skills), it will not start to chase until the delay ends.

### Expected Result

When a monster wants to start chasing and it uses NPC_EMOTION (and potentially other skills), it will start chasing at the same time.

### How to Reproduce

GOBLIN_5 often uses NPC_EMOTION and has a high attack delay. Just keep snapping away from it.

### Official Information

Tested on official servers.

_No response_

Playtester:
* **Addressed Issue(s)**: #8989 

* **Server Mode**: Both

- Unified the code for using chase skills and put it at the ideal location
  * Removed the check before starting a chase
  * Removed the check at the end of having moved a cell (but not the last cell)
  * Instead we now check it at the beginning of the movement to the next cell (unit_walktoxy_nextcell)
  * This fixes an issue that monsters did not start moving when using NPC_EMOTION at the start of a chase
- Unified the location where a monsters switches to chase state
  * This needs to happen before we check for chase skills
- Fixed an issue that caused position lag when a monster recalculated its chase while having walk delay
- Monsters will now also use walk skills when nobody is on the map if they have been spotted
  * For example relevant for Aliza who casts Heal in walk state
- Fixes #8989

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -171,17 +171,17 @@ bool unit_walktoxy_nextcell(block_list& bl, bool sendMove, t_tick tick) {
 
 	// Monsters first check for a chase skill and if they didn't use one if their target is in range each cell after checking for a chase skill
 	if (bl.type == BL_MOB) {
-		mob_data* md = reinterpret_cast<mob_data*>(&bl);
+		mob_data& md = reinterpret_cast<mob_data&>(bl);
 		// Walk skills are triggered regardless of target due to the idle-walk mob state.
 		// But avoid triggering when already reached the end of the walkpath.
 		// Monsters use walk/chase skills every second, but we only get here every "speed" ms
 		// To make sure we check one skill per second on average, we substract half the speed as ms
 		if (!ud->state.force_walk &&
-			DIFF_TICK(tick, md->last_skillcheck) > MOB_SKILL_INTERVAL - md->status.speed / 2 &&
-			mobskill_use(md, tick, -1)) {
-			if ((ud->skill_id != NPC_SPEEDUP || md->trickcasting == 0) //Stop only when trickcasting expired
+			DIFF_TICK(tick, md.last_skillcheck) > MOB_SKILL_INTERVAL - md.status.speed / 2 &&
+			mobskill_use(&md, tick, -1)) {
+			if ((ud->skill_id != NPC_SPEEDUP || md.trickcasting == 0) //Stop only when trickcasting expired
 				&& ud->skill_id != NPC_EMOTION && ud->skill_id != NPC_EMOTION_ON //NPC_EMOTION doesn't make the monster stop
-				&& md->state.skillstate != MSS_WALK) //Walk skills are supposed to be used while walking
+				&& md.state.skillstate != MSS_WALK) //Walk skills are supposed to be used while walking
 			{
 				// Skill used, abort walking
 				// Fix position as walk has been cancelled.
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -79,6 +79,20 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
 }
 
 /**
+ * Sets a mob's CHASE/FOLLOW state
+ * This should not be done if there's no path to reach
+ * @param bl: Mob to set state on
+ * @param flag: Whether to set state or not
+ */
+static inline void set_mobstate(struct block_list* bl, int32 flag)
+{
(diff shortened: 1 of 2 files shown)


src/map/unit.cpp
```diff
@@ -246,6 +286,11 @@ int32 unit_walktoxy_sub(struct block_list *bl)
 		unit_refresh( bl, true );
 	}
 #endif
+
+	// Set mobstate here already as chase skills can be used on the first frame of movement
+	// If we don't set it now the monster will always move a full cell before checking
+	set_mobstate(bl, ud->state.attack_continue);
```
Lemongrass3110, line 292: Since we more or less have a half switch for the different bl_types above already.
Could you change it to a switch (bl->type) please and merge the code into it?

src/map/unit.cpp
```diff
@@ -78,6 +78,20 @@ struct unit_data* unit_bl2ud(struct block_list *bl)
 	}
 }
 
+/**
+ * Sets a mob's CHASE/FOLLOW state
+ * This should not be done if there's no path to reach
+ * @param bl: Mob to set state on
+ * @param flag: Whether to set state or not
+ */
+static inline void set_mobstate(struct block_list* bl, int32 flag)
```
Lemongrass3110, line 87: Make it a mob_data& please.
Probably should be renamed to mob_setstate and moved to mob.cpp/.hpp then too.
Since this is not really a unit function.

src/map/unit.cpp
```diff
@@ -149,29 +165,53 @@ bool unit_walktoxy_nextcell(block_list& bl, bool sendMove, t_tick tick) {
...
+			if (unit_update_chase(bl, tick, (ud->walkpath.path_pos == ud->walkpath.path_len - 1)))
+				return true;
+			// Continue moving, restore to_x and to_y
+			ud->to_x = tx;
+			ud->to_y = ty;
+		}
+	}
+
+	// Get current speed
+	int32 speed;
```
Lemongrass3110, line 209: I know it is not scope of this PR, but since you moved it down here:
Could be changed to speed = status_get_speed(&bl);
And inside the if just make it speed *=

src/map/unit.cpp
```diff
@@ -149,29 +165,53 @@ bool unit_walktoxy_nextcell(block_list& bl, bool sendMove, t_tick tick) {
...
 
-	int32 speed;
-
 	// Reached end of walkpath
 	if (ud->walkpath.path_pos >= ud->walkpath.path_len)
 		return false;
 
+	// Monsters first check for a chase skill and if they didn't use one if their target is in range each cell after checking for a chase skill
+	if (bl.type == BL_MOB) {
+		mob_data* md = reinterpret_cast<mob_data*>(&bl);
```
Lemongrass3110, line 174:
```suggestion
		mob_data& md = reinterpret_cast<mob_data&>(bl);
```
And change the rest below accordingly.

Lemongrass3110 (approved):

src/map/unit.cpp
Playtester: I think you might have missed the purpose of this function. The main purpose is to be able to call it independent of type, so that you don't have to worry about it and make the calling code only require a single line. As such it's also only needed in unit.cpp because in mob.cpp you know the type anyway.

The proposed changes would destroy half of this function's purpose.

Also please note that I did not touch this function (except for adding the "!= nullptr"), the only reason it's marked as changed is because I had to move it to the top of unit.cpp because it's a "static inline" function and thus needs to be declared before it is called.

I can still think of possible improvements for example calling it `unit_setstateifmob` so the purpose is more clear. We could also write a mob_setstate function as you proposed, but this is then called by `unit_setstateifmob`. But I think this should be in a separate PR.

Lemongrass3110: It is a mob specific function, setting mob specific states.
This is not a unit function that could be used on any other unit.
That is also documented in the function header, so i dont know how I am missing the purpose of the function.
Unit functions should not be an excuse for being too lazy to check the bl type.

Playtester: Created enhancement request for it so we don't forget about it: https://github.com/rathena/rathena/issues/9088

PR #9084, opened 2025-02-16
Converted ZC_DISAPPEAR_ENTRY to struct

Lemongrass3110:

* **Server Mode**: Both

Converted ZC_DISAPPEAR_ENTRY to struct


src/map/vending.cpp
```diff
@@ -67,7 +67,7 @@ void vending_closevending(map_session_data* sd)
 
 		sd->state.vending = false;
 		sd->vender_id = 0;
-		clif_closevendingboard(&sd->bl, 0);
+		clif_closevendingboard( sd->bl );
```
Atemo, line 70: Default values seem useless for 1 occurrence
```suggestion
		clif_closevendingboard( sd->bl, AREA_WOS, nullptr );
```

PR #9093, opened 2025-02-21
Massive Monster Delay/AI Refactor

### LINKED ISSUE #9075: Settings Monster Skill Delays
### rAthena Hash

435b2b6

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

The monster skill DB delays are set when a monster picks a skill.

### Expected Result

The monster skill DB delays are set when the monsters starts casting the skill or in the moment the skill is cast-cancelled.

Same also goes for setting the "AttackDelay".

### How to Reproduce

Hit an Elder once at range 1-2 and wait.

You can see that Elder casts Firebolt, then Stone Curse and then attacks two times before using Firebolt again. On rAthena it currently only attacks once before casting again.

If you hit it while it's casting Firebolt, it will start using Stone Curse or Attack after "Attack Delay" going from the point the skill was cancelled.

### Official Information

Verified by testing.

Setting the delays for monsters works quite different from players. Regardless of how the skill finishes (castend_id, castend_pos or castcancel) the monster will always have its AI inactive for "AttackMotion", cannot attack for "AttackDelay" and cannot use the skill for its skill delay on the mob_skill_db, from that time onward.

_No response_

### LINKED ISSUE #9088: Rework set_mobstate

No

Currently we have a function called "set_mobstate" in unit.cpp that serves the purpose to be able to be called independent of type but then only sets the mob state if the type if BL_MOB.
It also considers the angry state logic, but only supports RUSH/FOLLOW and not BERSERK/ANGRY.

Also the function does not follow function naming conventions!

It would be cooler to have a function in mob.cpp called "mob_set_skillstate" that handles both RUSH/FOLLOW and BERSERK/ANGRY. We could then even enforce going through this function to set the skillstate.

Then we could for example rename set_mobstate in unit.cpp to "unit_set_skillstate" which only handles the conversion from bl to md and then calls mob_set_skillstate.

We could also just remove the function and code the conversion everywhere we need it, but I don't like that solution as much.

### Additional context

_No response_

Playtester:
* **Addressed Issue(s)**: #9075, #9088

* **Server Mode**: Both

- Refactored the monster Delay/AI so that the setting/execution order is much closer to official
  * This makes it much easier to have a stable implementation without having to write hundreds of exception
  * It will now be much easier to do further improvements without breaking other things
  * For detailed information, see points below
- A monster will now always check for skills at the beginning of its hard AI and cancel out if it used a skill
  * For BERSERK state it will check for attack delay instead of the skill interval
  * Removed all the special "skill use" checks that now work automatically due to having the right code structure
  * Cleaned up functions by moving unrelated AI logic into the main AI function (e.g. randomly walking)
  * Not refactored yet are skills-while-moving, event-triggered skills and dead state skills
- Setting of monster delays at the end of attacks and skills now works as on official servers
  * A monster's AI will be inactive for its attack motion, this applies even to bosses now
  * A monster cannot attack for its attack delay (this applies even if the skill was cast-canceled)
  * A monster will no longer get skill-based walk delays unless the skill used has one defined
  * Skill delays now work like cooldowns and are set at cast end and cast-cancel rather than at cast begin
  * Skill delays are set based on the status the monster is in at cast end
  * The custom feature to set delays per monster skill entry still works as previously
  * Removed all other special delay settings all over the code as they are not needed anymore
  * Disabling the AI when not needed will also save a lot of resources!
- Setting of various "tick" variables is now a lot more accurate
  * Spawned monsters will now act immediately, but cannot cast skills immediately
  * Summoned/recalled monsters will now act immediately without requiring a low MIN_MOBLINKTIME (now back at 1000ms)
  * Attack-linking in general will now be more reactive while requiring less resources in total
  * A good part of the code allows for frame-perfect behavior now (especially if MIN_MOBTHINKTIME is set to 20)
- Implemented a "mob_setstate" function which is now consistently used to set a monster's state
  * This fixes a lot of issues where the aggressive bit was not set correctly
  * Just makes setting of the state much more comfortable
  * Makes it easier to add actions in the future that should always happen when setting a certain state
  * Removed set_mobstate function from unit.cpp
- The attacktimer event for monsters now loops through the monster's AI instead of executing an attack directly
  * This allows for consistent and correct implementation in just a single location
  * The monster AI will no longer block itself from executing within MIN_MOBTHINKTIME to allow earlier reaction to events
  * This also allows for frame-perfect attack behavior even if the MIN_MOBTHINKTIME is higher than 20
  * Moved the "end of attack timer" check from unit.cpp into mob.cpp to have less AI logic in unit.cpp
- Monsters will no longer start walking when still able to cast skills
- A blind monster (e.g. Elder) will now drop target exactly at the moment it runs out of spells to cast
- Fixed an issue that a monster would sometimes switch into chase mode while randomly walking
- A monster that finds an enemy in attack range will now start off with a normal attack and then switch to Berserk
  * This still overwrites angry mode as before
- A monster that reaches an enemy will no longer start with a normal attack if it still has attack delay
- The refactor probably fixes a lot more issues we don't even know about due to them being fixed automatically
- Fixes #9075
- Fixes #9088 (enhancement)

**Note for Reviewers**: This is quite complex, but it was needed! If you need a call for me to explain everything, please let me know.

**Note to Others**: I would be glad if you could help us test this! If you find any issues, please let me know.

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1121,7 +1121,7 @@ int32 mob_spawn (struct mob_data *md)
 	int32 i=0;
 	t_tick tick = gettick();
 
-	md->last_thinktime = tick;
+	md->next_thinktime = tick;
 	if (md->bl.prev != nullptr)
 		unit_remove_map(&md->bl,CLR_RESPAWN);
 	else
@@ -1187,21 +1187,18 @@ int32 mob_spawn (struct mob_data *md)
 //	md->master_id = 0;
 	md->master_dist = 0;
 
-	md->state.aggressive = status_has_mode(&md->status,MD_ANGRY)?1:0;
-	md->state.skillstate = MSS_IDLE;
+	mob_setstate(*md, MSS_IDLE);
 	md->ud.state.blockedmove = false;
 	md->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME;
-	md->last_linktime = tick;
+	md->last_linktime = 0;
 	md->dmgtick = tick - 5000;
 	md->last_pcneartime = 0;
 	md->last_canmove = tick;
-	md->last_skillcheck = 0;
+	md->last_skillcheck = tick;
 	md->trickcasting = 0;
 
-	t_tick c = tick - MOB_MAX_DELAY;
-
 	for (i = 0; i < MAX_MOBSKILL; i++)
-		md->skilldelay[i] = c;
+		md->skilldelay[i] = 0;
 	for (i = 0; i < DAMAGELOG_SIZE; i++)
 		md->spotted_log[i] = 0;
 
@@ -1509,7 +1506,6 @@ static int32 mob_ai_sub_hard_slavemob(struct mob_data *md,t_tick tick)
 	if (DIFF_TICK(tick, md->last_linktime) >= MIN_MOBLINKTIME && !md->target_id)
   	{
 		struct unit_data *ud = unit_bl2ud(bl);
-		md->last_linktime = tick;
 
 		if (ud) {
 			struct block_list *tbl=nullptr;
@@ -1523,9 +1519,18 @@ static int32 mob_ai_sub_hard_slavemob(struct mob_data *md,t_tick tick)
 				if (tbl && battle_check_target(&md->bl, tbl, BCT_ENEMY) <= 0)
 					tbl = nullptr;
 			}
-			if (tbl && status_check_skilluse(&md->bl, tbl, 0, 0)) {
-				md->target_id=tbl->id;
-				return 1;
+			else if (bl->type == BL_MOB) {
+				// If master is a monster, it might still have a target after using a skill
+				mob_data* mmd = reinterpret_cast<mob_data*>(bl);
+				if (mmd->target_id > 0)
+					tbl = map_id2bl(mmd->target_id);
+			}
+			if (tbl != nullptr) {
+				md->last_linktime = tick;
+				if (status_check_skilluse(&md->bl, tbl, 0, 0)) {
+					md->target_id = tbl->id;
+					return 1;
+				}
 			}
 		}
 	}
@@ -1550,36 +1555,16 @@ int32 mob_unlocktarget(struct mob_data *md, t_tick tick)
 		if (md->ud.walktimer != INVALID_TIMER)
 			break;
 		//Because it is not unset when the mob finishes walking.
-		md->state.skillstate = MSS_IDLE;
+		mob_setstate(*md, MSS_IDLE);
 		[[fallthrough]];
 	case MSS_IDLE:
-		// When walking we want to trigger the idle skills through the walk routine so we can prevent stopping
-		// This situation happens when an immobile monster uses a skill to move
-		if (md->ud.walktimer != INVALID_TIMER)
-			break;
-		if (md->idle_event[0] && npc_event_do_id(md->idle_event, md->bl.id) > 0) {
+		if (md->ud.walktimer == INVALID_TIMER && md->idle_event[0] && npc_event_do_id(md->idle_event, md->bl.id) > 0)
 			md->idle_event[0] = 0;
-			break;
-		}
-		// Idle skill.
-		if (DIFF_TICK(tick, md->last_skillcheck) >= MOB_SKILL_INTERVAL && mobskill_use(md, tick, -1))
-			break;
-		//Random walk.
-		if (!md->master_id &&
-			DIFF_TICK(md->next_walktime, tick) <= 0 &&
-			!mob_randomwalk(md,tick))
-			//Delay next random walk when this one failed.
-			if (md->next_walktime < md->ud.canmove_tick)
-				md->next_walktime = md->ud.canmove_tick;
-			else
-				md->next_walktime = tick+rnd()%1000;
 		break;
 	default:
 		unit_stop_attack( &md->bl );
 		unit_stop_walking_soon(md->bl); //Stop chasing.
-		if (status_has_mode(&md->status,MD_ANGRY) && !md->state.aggressive)
-			md->state.aggressive = 1; //Restore angry state when switching to idle
-		md->state.skillstate = MSS_IDLE;
+		mob_setstate(*md, MSS_IDLE);
 		if(battle_config.mob_ai&0x8) //Walk instantly after dropping target
 			md->next_walktime = tick+rnd()%1000;
 		else
@@ -1623,6 +1608,7 @@ int32 mob_randomwalk(struct mob_data *md,t_tick tick)
 		return 0;
 
 	// Make sure the monster has no target anymore, otherwise the walkpath will check for it
+	md->ud.state.attack_continue = 0;
 	md->ud.target_to = 0;
 
 	r=rnd();
@@ -1693,7 +1679,7 @@ int32 mob_randomwalk(struct mob_data *md,t_tick tick)
 		}
 		return 0;
 	}
-	md->state.skillstate=MSS_WALK;
+	mob_setstate(*md, MSS_WALK);
 	md->move_fail_count=0;
 	md->next_walktime = tick+rnd()%1000+MIN_RANDOMWALKTIME + unit_get_walkpath_time(md->bl);
 	return 1;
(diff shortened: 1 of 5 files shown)


src/map/mob.cpp
```diff
@@ -2168,7 +2217,7 @@ static int32 mob_ai_sub_lazy(struct mob_data *md, va_list args)
 
 	if (md->ud.walktimer == INVALID_TIMER) {
 		// Because it is not unset when the mob finishes walking.
-		md->state.skillstate = MSS_IDLE;
+		mob_setstate(*md, MSS_IDLE);
 		if (md->idle_event[0] && npc_event_do_id( md->idle_event, md->bl.id ) > 0) {
```
Lemongrass3110, line 2221: Could be removed, if moved to mob_setstate

src/map/mob.cpp
```diff
@@ -1898,8 +1944,20 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
+
+		// Random walk
+		if (!md->master_id &&
+			DIFF_TICK(md->next_walktime, tick) <= 0 &&
+			!mob_randomwalk(md, tick)) {
+			// Delay next random walk when this one failed
+			if (md->next_walktime < md->ud.canmove_tick)
+				md->next_walktime = md->ud.canmove_tick;
+			else
+				md->next_walktime = tick + rnd()%1000;
```
Lemongrass3110, line 1958: Shouldnt that also be some kind of defined tick?
MIN_RANDOMWALKTIME? But that seems to high for this.

src/map/mob.cpp
```diff
@@ -1871,12 +1919,10 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 	{
 		int32 prev_id = md->target_id;
 		map_foreachinallrange (mob_ai_sub_hard_activesearch, &md->bl, view_range, DEFAULT_ENEMY_TYPE(md), md, &tbl, mode);
-		// If a monster finds a target through search that is already in attack range it immediately switches to berserk mode
+		// If a monster finds a target through search that is already in attack range it switches to berserk mode
```
Lemongrass3110, line 1922:
```suggestion
		// If a monster finds a target through search, that is already in attack range, it switches to berserk mode
```

src/map/mob.cpp
```diff
@@ -1871,12 +1919,10 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
 		// This behavior overrides even angry mode and other mode-specific behavior
-		if (tbl != nullptr && prev_id != md->target_id && battle_check_range(&md->bl, tbl, md->status.rhw.range)) {
+		if (tbl != nullptr && prev_id != md->target_id && battle_check_range(&md->bl, tbl, md->status.rhw.range))
 			md->state.aggressive = 0;
-			md->state.skillstate = MSS_BERSERK;
```
Lemongrass3110, line 1878: Comment still says it switches mode. Did you intentionally remove this or should we call mob_setstate here?

src/map/mob.cpp
```diff
@@ -1738,10 +1747,11 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 	if (md->ud.state.force_walk)
 		return false;
 
-	if (DIFF_TICK(tick, md->last_thinktime) < MIN_MOBTHINKTIME)
+	if (DIFF_TICK(tick, md->next_thinktime) < 0)
```
Lemongrass3110, line 1750: Do you really want to trigger this with 0 difference?

src/map/mob.cpp
```diff
@@ -1550,36 +1555,16 @@ int32 mob_unlocktarget(struct mob_data *md, t_tick tick)
 		if (md->ud.walktimer != INVALID_TIMER)
 			break;
 		//Because it is not unset when the mob finishes walking.
-		md->state.skillstate = MSS_IDLE;
+		mob_setstate(*md, MSS_IDLE);
 		[[fallthrough]];
```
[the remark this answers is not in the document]
Playtester: This would make monsters not execute their Idle event when they switch from Walk to Idle.

src/map/mob.cpp
```diff
@@ -1550,36 +1555,16 @@ int32 mob_unlocktarget(struct mob_data *md, t_tick tick)
...
 	case MSS_IDLE:
-		// When walking we want to trigger the idle skills through the walk routine so we can prevent stopping
-		// This situation happens when an immobile monster uses a skill to move
-		if (md->ud.walktimer != INVALID_TIMER)
-			break;
-		if (md->idle_event[0] && npc_event_do_id(md->idle_event, md->bl.id) > 0) {
+		if (md->ud.walktimer == INVALID_TIMER && md->idle_event[0] && npc_event_do_id(md->idle_event, md->bl.id) > 0)
```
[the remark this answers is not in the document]
Playtester: If someone can confirm to me that this event should always be executed when switching to idle, we could do that.

If we need to test all situations where a monster switches to Idle and see if the event triggers there that would be a larger test that could result in many more changes and I can't effectively test in renewal and I don't think pre-renewal uses such events at all. So I would prefer if someone who actively tests on renewal would do that change.

In regards to the change here, it would result that moving monsters that are already in IDLE and are set to IDLE again will trigger this. From looking at the original implementation from Atemo it looks to be like it should only happen when the monster is not moving, but I asked them.

src/map/mob.cpp
Playtester: Yes, if the "end of attack" timer event happens in the same millisecond as they hard AI was executed, we need to run through the AI again because now the attacktimer was cleared which can lead to a different result (usually the mob starting the attack).

I made sure that if we come from lazy -> hard AI that it won't trigger again on the same millisecond in the lazy AI code.

src/map/mob.cpp
Playtester: I don't think that's correct English punctuation, especially not in front of a "that": 
https://editorsmanual.com/articles/commas-with-that-which/

You can optionally set a comma in front of the second part of an if-sentence in English, but that's also optional:
https://grammarhow.com/comma-before-if/

src/map/mob.cpp
Playtester: I removed the word "immediately" because it doesn't set the mode here but rather later in the code (after the attack).

It's related to this from PR description:

- A monster that finds an enemy in attack range will now start off with a normal attack and then switch to Berserk
  * This still overwrites angry mode as before

The confusing part is a bit because we sometimes call setting "aggressive = 0" as "leaving angry mode / entering berserk mode". Maybe I need to elaborate it in the comment a little.

src/map/mob.cpp
Playtester: I wanted to fix random walking later on. This refactor is actually the precondition to fix it properly.

Since I tried my best to keep this PR small, I only included things that depend on each other and left the things that can be fixed in standalone PRs for later.

How it actually works on Aegis is that is only remembers when it switched to IDLE mode last and then it does actually check if "4000 + rand()%5000" has passed since then. Since it checks this every 20ms however and rolls the random value each time, it will never end up being longer than a second.

Basically after 9 seconds being in IDLE, it will just try again every 20ms until it works.

src/map/unit.cpp
```diff
@@ -3032,9 +3024,14 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
...
 
-		// You can't move if you can't attack neither.
-		if (src->type&battle_config.attack_walk_delay)
-			unit_set_walkdelay(src, tick, sstatus->amotion, 1);
+		// You can't move during your attack motion
+		if (src->type&battle_config.attack_walk_delay) {
+			// Monsters set their AI inactive instead of having a walkdelay
+			if (md != nullptr)
+				md->next_thinktime = tick + status_get_amotion(src);
+			else
```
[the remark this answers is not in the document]
Playtester: It still causes monsters to not walk for this time, so I think it's the still the same functionally.

From what I learned yesterday, there is actually no walkdelay at all in Aegis, so even for the other types, we need to eventually refactor that and then it would be in the same code again.

src/map/mob.cpp
Playtester: Comment improved.

src/map/unit.cpp
```diff
@@ -2896,19 +2893,26 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
...
 
+	// Make sure attacktimer is removed before doing anything else
+	ud->attacktimer = INVALID_TIMER;
+
+	// Note: Officially there is no such thing as an attack timer. All actions are driven by the client or the AI.
+	// We use the continuous attack timers to have accurate attack timings that don't depend on the AI interval.
+	// However, for a clean implementation we still should channel through the whole AI code so the same rules
+	// apply as usual and we don't need to code extra rules. Currently we resolved this only for monsters.
+	// We don't want this to trigger on direct calls of the timer function as that should just execute the attack.
+	if (src->type == BL_MOB && tid != INVALID_TIMER) {
```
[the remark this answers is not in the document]
Playtester: Fixed

PR #9095, opened 2025-02-22
Endless Loop without Official Walkpath Fix

### LINKED ISSUE #9092: Range=2 Monsters Crash Map Server
### rAthena Hash

3b9d3e9

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When disabling OFFICIAL_WALKPATH and have a monster with attack range of 2 get into attack range, but there is a wall in the way of the attack, the map server crashes.

### Expected Result

Officially, a monster in this situation just stops and becomes idle.

### How to Reproduce

Summon an Apocalips and aggro it and then get into the situation as depicted below.

Image

### Official Information

Tested on official server and confirmed that monsters just stop in that situation and do not try to continue the chase.

However, disabling OFFICIAL_WALKPATH is not an official feature in the first place, and I don't know what its purpose is here.

Need to disable OFFICIAL_WALKPATH to reproduce.

Playtester:
* **Addressed Issue(s)**: #9092 

* **Server Mode**: Both

- Fixed an issue that could cause an endless loop when OFFICIAL_WALKPATH is disabled (fixes #9092)
- Improved comments so it's easier to understand what disabling OFFICIAL_WALKPATH actually means

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -2970,9 +2970,8 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
 		if (ud->state.attack_continue && ud->chaserange > 1) {
 			ud->chaserange = std::max(1, ud->chaserange - 2);
 
-			// Walk around the obstacle and start attacking once you are in range
-			unit_walktobl(src,target,ud->chaserange,ud->state.walk_easy|2);
-			return 1;
+			// Walk closer / around the obstacle and start attacking once you are in range
+			return unit_walktobl(src,target,ud->chaserange,ud->state.walk_easy|2);
 		}
 		// Can't attack even though next to the target? Giving up here.
 		return 0;
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -2969,6 +2969,8 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
 		// This code can usually only be reached if OFFICIAL_WALKPATH is disabled
 		if (ud->state.attack_continue && ud->chaserange > 1) {
 			ud->chaserange = std::max(1, ud->chaserange - 2);
+
+			// Walk around the obstacle and start attacking once you are in range
 			unit_walktobl(src,target,ud->chaserange,ud->state.walk_easy|2);
 			return 1;
 		}
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -2965,15 +2965,15 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
 	}
 
 	if( !battle_check_range(src,target,range) ) {
-	  	// Within range, but no direct line of attack
-		if( ud->state.attack_continue ) {
-			if(ud->chaserange > 2)
-				ud->chaserange-=2;
-
+		// Within range, but no direct line of attack
+		// This code can usually only be reached if OFFICIAL_WALKPATH is disabled
+		if (ud->state.attack_continue && ud->chaserange > 1) {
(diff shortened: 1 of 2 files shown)


src/map/unit.cpp
```diff
@@ -2965,15 +2965,15 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
...
 	if( !battle_check_range(src,target,range) ) {
-	  	// Within range, but no direct line of attack
-		if( ud->state.attack_continue ) {
-			if(ud->chaserange > 2)
-				ud->chaserange-=2;
-
+		// Within range, but no direct line of attack
+		// This code can usually only be reached if OFFICIAL_WALKPATH is disabled
+		if (ud->state.attack_continue && ud->chaserange > 1) {
+			ud->chaserange = std::max(1, ud->chaserange - 2);
```
Lemongrass3110, line 2971:
```suggestion
			ud->chaserange = std::max(1, ud->chaserange - 2);

			// Walk around the obstacle and start attacking once you are in range
```

src/map/unit.cpp
```diff
@@ -2965,15 +2965,15 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
...
 			unit_walktobl(src,target,ud->chaserange,ud->state.walk_easy|2);
+			return 1;
```
Lemongrass3110, line 2973: Based on your explanations in discord:
Should we not return the result of unit_walktobl here?
Because if you fail to start walking, then you actually should give up or not?
Just like you changed it below.

src/map/unit.cpp
```diff
@@ -2965,15 +2965,15 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
...
-	  	// Within range, but no direct line of attack
-		if( ud->state.attack_continue ) {
-			if(ud->chaserange > 2)
-				ud->chaserange-=2;
-
+		// Within range, but no direct line of attack
+		// This code can usually only be reached if OFFICIAL_WALKPATH is disabled
+		if (ud->state.attack_continue && ud->chaserange > 1) {
+			ud->chaserange = std::max(1, ud->chaserange - 2);
 			unit_walktobl(src,target,ud->chaserange,ud->state.walk_easy|2);
```
Lemongrass3110, line 2972: I am not sure, if we should respect ud->state.walk_easy here.
Obviously we are standing in front of an obstacle.
Using easy path search will not walk around the obstacle, which sounds like an infinite loop to me again.
So maybe we only should start walking, if walk_easy == 0 and otherwise give up?
Maybe it should be added to the conditions of the if around this block?

Playtester: I think it would be weird if a unit that's supposed to only walk in a straight line would suddenly start using complex pathfinding.

But we should do as you commented in the previous comment - return 0 when walkpath fails.

There can't be an endless loop anymore because if the walkpath fails it will never trigger the "reached end of walkpath" event again.

Lemongrass3110: Why I am thinking of an infinite loop here:
You are already in attack range (variable name "range" here).
You try to walk chaserange - 2 cells close to the target.
But I think in quite a few scenarios, you will already stand inside chaserange - 2 cells distance to the target.

Do you not think that this could loop as well?

Playtester: We give ud->chaserange back into unit_walktobl which defines the range at which it stops next.

So let's say you are at distance 3 and ud->chaserange is 5. Then we set ud->chaserange to 3 and start chasing again -> it immediately finds out it's already in attack range, stops and comes back here.
Now it's at distance 3 and ud->chaserange is 3. We set ud->chaserange to 1 and start chasing again.
And if it still fails at ud->chaserange=1, we quit.

Issue #9109, opened 2025-03-02
Pneuma stopped working


krad309:
### rAthena Hash

435b2b6

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

After updating to 435b2b6 and beyond, pneuma, when used against range monsters, has stopped working. Players are unaffected.

### Relevant Log Output

```Shell

```

### Expected Result

Monsters will miss when standing on or near pneuma.

### How to Reproduce

1. Summon Raydric Archer
2. Cast Pneuma on self

### Official Information

Pneuma effect hasn't been touched for a long time.

PR #9110, opened 2025-03-02
Fixed Pneuma against monsters

krad309:
Pneuma against monsters has stopped working after 435b2b6

* **Addressed Issue(s)**: #9109

* **Description of Pull Request**: Fixes an issue with Pneuma when used against range monsters.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -2868,12 +2868,12 @@ static bool is_skill_using_arrow(struct block_list *src, int32 skill_id)
 	if (sd == nullptr && skill_id != 0) {
 		if (skill_get_ammotype(skill_id) != 0)
 			return true;
-
-		status_data* sstatus = status_get_status_data(*src);
-
-		if (sstatus->rhw.range > 3)
-			return true;
 	}
+
+	status_data* sstatus = status_get_status_data(*src);
+
+	if (sstatus->rhw.range > 3)
+		return true;
 
 	switch( skill_id ) {
 		case HT_PHANTASMIC:


Playtester:
I think more things got broken with it. For example the exception that Freezing Trap counts as ranged attack (at least in pre-renewal) is missing.

Please see 8cec8ce

src/map/battle.cpp
```diff
@@ -2868,12 +2868,12 @@ static bool is_skill_using_arrow(struct block_list *src, int32 skill_id)
...
 			return true;
+	}
 
-		status_data* sstatus = status_get_status_data(*src);
+	status_data* sstatus = status_get_status_data(*src);
 
-		if (sstatus->rhw.range > 3)
-			return true;
-	}
+	if (sstatus->rhw.range > 3)
```
Playtester, line 2875: If I just look at the previous code. I think this should still be within the `sd == nullptr` area.

Breaking down the old formula:

1.
(sd && sd->state.arrow_atk) || 

2.
(!sd && (
	(skill_id && skill_get_ammotype(skill_id)) ||   //Skill ID is only checked for the skill_get_ammotype call
	sstatus->rhw.range>3
)
	
3.
|| skill_id == HT_FREEZINGTRAP || (skill_id == HT_PHANTASMIC) || (skill_id == GS_GROUNDDRIFT));

Lemongrass3110:
Thank you very much for your contribution.
Sadly you used your master branch and therefore we cannot push any changes (as suggested by @Playtester) to your repo, before merging the pull request.

Therefore I will have to close it and fix it in a separate PR.
Thank you for understanding!

PR #9114, opened 2025-03-02
Cleaned up pc_setinventorydata

Lemongrass3110:

* **Server Mode**: Both

Fixed a problem when increasing MAX_INVENTORY over 255

Thanks to @LadyNanuia


src/map/pc.cpp
```diff
@@ -1429,14 +1429,15 @@ int32 pc_equippoint(map_session_data *sd,int32 n){
  * @param sd : player session
  * @return 0 sucess, 1:invalid sd
```
Lemongrass3110, line 1430:
```suggestion
```

PR #9126, opened 2025-03-04
Speed up StatusDatabase::hasSCF with cache

inhyositsu:
Speed up StatusDatabase::hasSCF with cache to prevent iterate whole status_db for every mob_ai_sub_hard on each mob.

* **Server Mode**: Both

* **Description of Pull Request**: On my server, I notice there's a unusal high cpu usage in StatusDatabase::hasSCF, which iterates every status and create a shared_ptr and destroy it every time when mob_ai_sub_hard runs. I speed up with a cache code.


Lemongrass3110:
Can you try #8834 instead maybe?

inhyositsu:
> Can you try #8834 instead maybe?

Yes, good improvement on that!

inhyositsu:
> Can you try #8834 instead maybe?

I see some crashes on my server after this update.
I don't recommend this PR at this time til I figure out.
I'll restore to the original one I pushed.

Lemongrass3110:
> > Can you try #8834 instead maybe?
> 
> I see some crashes on my server after this update. I don't recommend this PR at this time til I figure out. I'll restore to the original one I pushed.

Hey! It was not meant that you have to include it into this PR.
But if you encounter a crash on #8834, could you please give me the stack trace, so that I can look into it and fix it?
Thank you in advance!

PR #9127, opened 2025-03-04
Improved Monster Warpchase Feature

Playtester:

* **Server Mode**: Both

- Monsters will now reliably chase you through warp portals
  * This includes Priest portals if the corresponding setting is enabled in monster.conf
  * It will work regardless of other settings
- This also fixes a small issue with monsters not unlocking their target when nobody is on the map

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1413,14 +1413,11 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
 }
 
 static int32 mob_warpchase_sub(struct block_list *bl,va_list ap) {
-	block_list *target;
-	block_list **target_warp;
-	int32 *min_distance;
 	int32 cur_distance;
 
-	target= va_arg(ap, struct block_list*);
-	target_warp= va_arg(ap, struct block_list**);
-	min_distance= va_arg(ap, int32*);
+	block_list* target= va_arg(ap, struct block_list*);
+	block_list** target_warp= va_arg(ap, struct block_list**);
+	int32* min_distance= va_arg(ap, int32*);
 
 	switch (bl->type) {
 		case BL_NPC:
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1729,36 +1729,6 @@ int32 mob_randomwalk(struct mob_data *md,t_tick tick)
 }
 
 /**
- * Makes a monster inactive and immune to damage during the warp invincibility of its target
- * @param md: Mob that should be become inactive and invincible
- */
-void mob_warpchase_invincible(mob_data& md) {
-	if (!(battle_config.mob_warp&16))
-		return;
-
-	if (md.target_id == 0)
-		return;
-
-	map_session_data *sd = map_id2sd(md.target_id);
-	if (sd == nullptr)
-		return;
-
-	const TimerData* timer = get_timer(sd->invincible_timer);
-	if (timer == nullptr)
-		return;
-
-	t_tick duration = DIFF_TICK(timer->tick, gettick());
-	if (duration <= 0)
-		return;
-
-	// To prevent the monster losing the target, we make the monster AI inactive for the pc invincibility duration
-	md.next_thinktime = timer->tick + TIMER_MIN_INTERVAL;
-
-	// To prevent exploits the monster is immune to damage during that time
-	sc_start(&md.bl, &md.bl, SC_BARRIER, 100, 1, duration);
-}
-
-/**
  * Makes a monster move to the closest warp if corresponding configs are set
  * @param md: Mob that should move to the warp
  * @param target: Target the mob should follow
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1422,55 +1422,61 @@ static int32 mob_warpchase_sub(struct block_list *bl,va_list ap) {
 	target_warp= va_arg(ap, struct block_list**);
 	min_distance= va_arg(ap, int32*);
 
-	// NPC Warp
-	if (bl->type == BL_NPC) {
-		npc_data *nd = reinterpret_cast<npc_data*>(bl);
+	switch (bl->type) {
+		case BL_NPC:
+		{
+			// NPC Warp
+			npc_data* nd = reinterpret_cast<npc_data*>(bl);
 
-		// Not a warp
-		if (nd->subtype != NPCTYPE_WARP)
-			return 0;
-
-		// Does not lead to the same map
-		if (nd->u.warp.mapindex != map_getmapdata(target->m)->index)
-			return 0;
-
-		// Get distance from warp exit to target
-		cur_distance = distance_blxy(target, nd->u.warp.x, nd->u.warp.y);
-
-		//Pick warp that leads closest to target
-		if (cur_distance < *min_distance) {
-			*target_warp = &nd->bl;
-			*min_distance = cur_distance;
-			return 1;
-		}
-	}
-
-	// Skill Warp
-	if (bl->type == BL_SKILL) {
-		skill_unit *su = reinterpret_cast<skill_unit*>(bl);
-
-		if (su->group == nullptr)
-			return 0;
-
-		switch (su->group->unit_id) {
-			case UNT_WARP_WAITING:
-				// Does not lead to the same map.
-				if (su->group->val3 != map_getmapdata(target->m)->index)
-					return 0;
-				// Get distance from warp exit to target
-				cur_distance = distance_blxy(target, su->group->val2>>16, su->group->val2&0xffff);
-				break;
-			default:
-				// Skill cannot warp
+			// Not a warp
+			if (nd->subtype != NPCTYPE_WARP)
 				return 0;
-		}
 
-		//Pick warp that leads closest to target.
-		if (cur_distance < *min_distance) {
-			*target_warp = &su->bl;
-			*min_distance = cur_distance;
-			return 1;
+			// Does not lead to the same map
+			if (nd->u.warp.mapindex != map_getmapdata(target->m)->index)
+				return 0;
+
+			// Get distance from warp exit to target
+			cur_distance = distance_blxy(target, nd->u.warp.x, nd->u.warp.y);
+
+			//Pick warp that leads closest to target
+			if (cur_distance < *min_distance) {
+				*target_warp = &nd->bl;
+				*min_distance = cur_distance;
+				return 1;
+			}
 		}
+		break;
+
+		case BL_SKILL:
+		{
+			// Skill Warp
+			skill_unit* su = reinterpret_cast<skill_unit*>(bl);
+
+			if (su->group == nullptr)
+				return 0;
+
+			switch (su->group->unit_id) {
+				case UNT_WARP_WAITING:
+					// Does not lead to the same map.
+					if (su->group->val3 != map_getmapdata(target->m)->index)
+						return 0;
+					// Get distance from warp exit to target
+					cur_distance = distance_blxy(target, su->group->val2 >> 16, su->group->val2 & 0xffff);
+					break;
+				default:
+					// Skill cannot warp
+					return 0;
+			}
+
+			//Pick warp that leads closest to target.
+			if (cur_distance < *min_distance) {
+				*target_warp = &su->bl;
+				*min_distance = cur_distance;
+				return 1;
+			}
+		}
+		break;
 	}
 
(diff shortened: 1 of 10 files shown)


src/map/mob.cpp
```diff
@@ -1413,31 +1413,66 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
 	int32 cur_distance;
 
 	target= va_arg(ap, struct block_list*);
-	target_nd= va_arg(ap, struct npc_data**);
+	target_warp= va_arg(ap, struct block_list**);
 	min_distance= va_arg(ap, int32*);
 
-	nd = (TBL_NPC*) bl;
+	// NPC Warp
+	if (bl->type == BL_NPC) {
```
Atemo, line 1426: Maybe consider using switch instead of if?

Playtester: Done.

src/map/mob.cpp
```diff
@@ -1695,9 +1736,16 @@ int32 mob_randomwalk(struct mob_data *md,t_tick tick)
  */
 int32 mob_warpchase(struct mob_data *md, struct block_list *target)
 {
-	struct npc_data *warp = nullptr;
+	block_list *warp = nullptr;
```
Lemongrass3110, line 1739:
```suggestion
	if( battle_config.mob_warp == 0 ){
		return 0;
	}
	
	block_list *warp = nullptr;
```

src/map/mob.cpp
```diff
@@ -1695,9 +1736,16 @@ int32 mob_randomwalk(struct mob_data *md,t_tick tick)
...
 	int32 distance = AREA_SIZE;
-	if (!(target && battle_config.mob_ai&0x40 && battle_config.mob_warp&1))
+
+	int32 type = BL_NUL;
+	if (battle_config.mob_warp&1)
+		type |= BL_NPC;
+	if (battle_config.mob_warp&2)
+		type |= BL_SKILL;
+
+	if (!(target && battle_config.mob_ai&0x40 && type != BL_NUL))
```
Lemongrass3110, line 1748: 0x40 needs an update in conf/battle/monster.conf it only talks about NPC warps.

src/map/mob.cpp
```diff
@@ -2217,6 +2269,13 @@ static int32 mob_ai_sub_lazy(struct mob_data *md, va_list args)
 	if (DIFF_TICK(tick, md->last_skillcheck) < MOB_SKILL_INTERVAL)
 		return 0;
 
+	// Check for warp chase if there's still a target
```
Lemongrass3110, line 2272: Did I miss something above, or why are we assuming that the target is not on the same map (or not even in range) here anymore?

src/map/mob.cpp
```diff
@@ -2231,14 +2290,9 @@ static int32 mob_ai_sub_lazy(struct mob_data *md, va_list args)
...
-		// Because it is not unset when the mob finishes walking.
-		mob_setstate(*md, MSS_IDLE);
-		if (md->idle_event[0] && npc_event_do_id( md->idle_event, md->bl.id ) > 0) {
-			md->idle_event[0] = 0;
-			return 0;
-		}
-	}
+	// Remove target and set to idle state when movement was finished
+	// This also triggers the idle NPC event
+	mob_unlocktarget(md, tick);
```
Lemongrass3110, line 2295: Unrelated change.

src/map/mob.cpp
```diff
@@ -1413,31 +1413,72 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
+		{
+			// Skill Warp
+			skill_unit* su = reinterpret_cast<skill_unit*>(bl);
+
+			if (su->group == nullptr)
+				return 0;
+
+			switch (su->group->unit_id) {
+				case UNT_WARP_WAITING:
+					// Does not lead to the same map.
```
Lemongrass3110, line 1461: Misses check for battle_conf.mob_warp&2

src/map/mob.cpp
```diff
@@ -1413,31 +1413,72 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
+					if (su->group->val3 != map_getmapdata(target->m)->index)
+						return 0;
+					// Get distance from warp exit to target
+					cur_distance = distance_blxy(target, su->group->val2 >> 16, su->group->val2 & 0xffff);
+					break;
```
Lemongrass3110, line 1466: Misses case for UNT_DIMENSIONDOOR

Lemongrass3110 (changes requested):

src/map/mob.cpp
Playtester: I put the check in the calling function by only searching for the affected bl type. But I guess we can check here again?

src/map/mob.cpp
Playtester: Dimension Door doesn't make sense here since it leads to a random location and not to the chase target.

src/map/mob.cpp
Playtester: It's related because we clear the target here to prevent it from attempting warp chase again.

src/map/mob.cpp
Playtester: I added a case for UNT_DIMENSIONDOOR which explains why it is not considered for warp chase.

src/map/mob.cpp
Playtester: Done!

src/map/mob.cpp
Playtester: Added additional check here. You're right, it's safer in case we add more warp types later.

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
 }
 
 static int32 mob_warpchase_sub(struct block_list *bl,va_list ap) {
-	struct block_list *target;
-	struct npc_data **target_nd;
-	struct npc_data *nd;
+	block_list *target;
+	block_list **target_warp;
```
Lemongrass3110, line 1417:
```suggestion
```
If you already touch it.

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
 	int32 *min_distance;
 	int32 cur_distance;
 
 	target= va_arg(ap, struct block_list*);
```
Lemongrass3110, line 1421:
```suggestion
	block_list* target= va_arg(ap, struct block_list*);
```

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
-	target_nd= va_arg(ap, struct npc_data**);
+	target_warp= va_arg(ap, struct block_list**);
```
Lemongrass3110, line 1422:
```suggestion
	block_list** target_warp= va_arg(ap, struct block_list**);
```

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
 }
 
 static int32 mob_warpchase_sub(struct block_list *bl,va_list ap) {
-	struct block_list *target;
-	struct npc_data **target_nd;
-	struct npc_data *nd;
+	block_list *target;
+	block_list **target_warp;
 	int32 *min_distance;
```
Lemongrass3110, line 1418:
```suggestion
```

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
 	int32 cur_distance;
 
 	target= va_arg(ap, struct block_list*);
-	target_nd= va_arg(ap, struct npc_data**);
+	target_warp= va_arg(ap, struct block_list**);
 	min_distance= va_arg(ap, int32*);
```
Lemongrass3110, line 1423:
```suggestion
	int32* min_distance= va_arg(ap, int32*);
```

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
-	if(nd->subtype != NPCTYPE_WARP)
-		return 0; //Not a warp
+			// Not a warp
+			if (nd->subtype != NPCTYPE_WARP)
+				return 0;
 
-	if(nd->u.warp.mapindex != map_getmapdata(target->m)->index)
-		return 0; //Does not lead to the same map.
+			// Does not lead to the same map as target
+			if (nd->u.warp.mapindex != map_getmapdata(target->m)->index)
```
Lemongrass3110, line 1436:
```suggestion
			if (m < 0) {
				return 0;
			}

			if (nd->u.warp.mapindex != map_getmapdata(target->m)->index)
```

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
+				return 0;
 
-	cur_distance = distance_blxy(target, nd->u.warp.x, nd->u.warp.y);
-	if (cur_distance < *min_distance)
-	{	//Pick warp that leads closest to target.
-		*target_nd = nd;
-		*min_distance = cur_distance;
-		return 1;
+			// Leads to a different map that is set to not be accessible through warp chase
+			int16 warp_m = map_mapindex2mapid(nd->u.warp.mapindex);
```
Lemongrass3110, line 1440: Check for the battle_config first and then do this lookup inside an if, only if necessary.

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
+
+		case BL_SKILL:
+		{
+			// Skill Warp
+			skill_unit* su = reinterpret_cast<skill_unit*>(bl);
+
+			if (su->group == nullptr)
+				return 0;
+
+			int16 warp_m = map_mapindex2mapid(su->group->val3);
```
Lemongrass3110, line 1464: This one depends on the skill unit type and should be moved into the respective case of the switch statement below.

src/map/mob.cpp
```diff
@@ -1413,31 +1413,91 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
...
+
+			switch (su->group->unit_id) {
+				case UNT_WARP_WAITING:
+					// Monsters cannot use priest warps
+					if (!(battle_config.mob_warp&2))
+						return 0;
+
+					// Does not lead to the same map as target
+					if (su->group->val3 != map_getmapdata(target->m)->index)
```
Lemongrass3110, line 1473: Note: You already calculated warp_m above, so you could directly compare target->m with warp_m here.

src/map/mob.cpp
Lemongrass3110: It does so pretty late and with additional distance checking.
Do we really want it this way?

Lemongrass3110 (approved):

src/map/mob.cpp
Playtester: What is m?

Playtester: target->m is verified before even calling this function now (for better performance)

src/map/mob.cpp
Playtester: Decided to go for the approach you suggested for the NPC warps to only calculatated warp_m when mob_warp&4 is set. Already agreed on this on Discord.

src/map/mob.cpp
Playtester: Fixed

src/map/mob.cpp
Playtester: Done

src/map/mob.cpp
Playtester: I sorted the checks in mob_warpchase so that they are in the optimal order (at least as to the best of my knowledge).

The distance checking happens at the last moment and only if the map is equal and all other requirements for warp chase are met.

None of these checks are very cost-intenstive and we only call mob_warpchase once and then unlock target, so as long as we map sure not to reach map_foreachinallrange unless we already have checked everything else, we should be fine.

PR #9134, opened 2025-03-08
Alternative Solution for Exact Position

Playtester:
* **Addressed Issue(s)**: Feature

* **Server Mode**: Both

- This is an alternative solution to fetch the exact position of a unit (see 3cb7c9f)
- This makes use of a private struct inside unit_data that remembers the last calculated exact coordinates
- Recalculation only happens when 20 or more milliseconds (1 frame) have passed since the last calculation
- Access to the coordinates is only possible through helper functions that make sure the data is up-to-date

Please let me know if you like this solution more, otherwise we can also just cancel this PR.

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -44,6 +44,12 @@ using namespace rathena;
 	#define MAX_SHADOW_SCAR 100 /// Max Shadow Scars
 #endif
 
+// How many milliseconds need to pass before we calculated the exact position of a unit
+// Calculation will only happen on demand and when at least the time defined here has passed
+#ifndef MIN_POS_INTERVAL
+	#define MIN_POS_INTERVAL 20
+#endif
+
 // Directions values
 // 1 0 7
 // 2 . 6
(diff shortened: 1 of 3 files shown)


src/map/unit.cpp
```diff
@@ -1488,101 +1494,108 @@ int32 unit_warp(struct block_list *bl,int16 m,int16 x,int16 y,clr_type type)
...
+	// Remember time at which we did the last calculation
+	this->pos.tick = tick;
 
-	const TimerData* td = get_timer(ud->walktimer);
+	if (this->walkpath.path_pos >= this->walkpath.path_len)
+		return;
 
-	if (td == nullptr)
-		return false;
+	if (this->walktimer == INVALID_TIMER)
```
Lemongrass3110, line 1523: Shoudnt we move this very important low-cost check almost to the top?

src/map/unit.hpp
```diff
@@ -64,6 +64,20 @@ struct unit_data {
 	int32 group_id;
 
 	std::vector<int32> shadow_scar_timer;
+
+	// Functions and struct to calculate and store exact position at a certain tick
+	int16 getx(t_tick tick = gettick());
+	int16 gety(t_tick tick = gettick());
+	void getpos(int16 &x, int16 &y, uint8 &sx, uint8 &sy, t_tick tick = gettick());
```
Lemongrass3110, line 71: Why is it necessary to hand over the tick from the external caller?

src/map/unit.cpp
```diff
@@ -44,6 +44,12 @@ using namespace rathena;
 	#define MAX_SHADOW_SCAR 100 /// Max Shadow Scars
 #endif
 
+// How many milliseconds need to pass before we calculated the exact position of a unit
+// Calculation will only happen on demand and when at least the time defined here has passed
+#ifndef MIN_POS_INTERVAL
+	#define MIN_POS_INTERVAL 20
```
Lemongrass3110, line 50: Why does this need to happen 5x more often than MIN_MOBTHINKTIME?
I know you say it should be per frame, but do we have a define already for that?
Should this be unit type specific?

src/map/unit.hpp
```diff
@@ -64,6 +64,20 @@ struct unit_data {
...
+private:
+	void update_pos(t_tick tick);
+	struct s_udPos {
```
Lemongrass3110, line 74:
```suggestion
	struct {
```
If you want to declare it locally/private you do not have to give it a name.
If you do want to give it a name, then please follow the naming standards for example: s_ud_pos

Lemongrass3110 (commented):
Looks good. Like this much better than the old version, good work!

src/map/unit.cpp
Playtester: It only happens on demand. This is just to determine the exact position of a moving unit between two cells, if it is much higher than a frame, there won't be much point to it (Mimic for example would just move a full cell in 100ms, so the whole feature would be pointless if it used MIN_MOBTHINKTIME).

I think it will mostly be used in the future for changing the walkpath with having to finish moving to the next cell. As such "per frame" would still make it look smooth, whereas any valeu higher than that would make it look choppy (usually the client will display a changed walk speed to re-synchronize with the reported server-sided position).

We probably should also use it for time-sensitive unit-touch effects such as firewall, where the frame-accuracy is also fairly important (even if we only check for the touch itself every 100ms at the moment, we still would want the frame-accurate position when it matters).

In my opinion it's completely unrelated to unit type (except it doesn't apply for units that cannot move).

The only place where we already use this interval is:
const t_tick TIMER_MIN_INTERVAL = 20;

But I didn't want to make it dependent on that, that's why I decided to add a new define.

src/map/unit.cpp
Playtester: If the walktimer is invalid, we still need to set the position correctly. The unit is just at the cell center currently. That's why it needs to come after the setting of the coordinates (and pos.tick).

Lemongrass3110: You mean because this->bl X/Y may be async to the new cached this->pos X/Y?

src/map/unit.hpp
Lemongrass3110: Then we should not use default values - that go against this principle - actually.

Playtester: Some functions don't have a tick, so I wanted to make the call shorter for those cases.

I was wondering if we could use macros for making it shorter instead though.

Maybe I'm overthinking, my logic was that `bl->x` is just 5 characters and I didn't want to turn getting coordinates this way into like 20 characters.

src/map/unit.cpp
Playtester: Yes, we need to update the exact position also if the unit is not moving. Otherwise it might contain data from minutes ago. As we only calculate on demand.

src/map/unit.cpp
```diff
@@ -1608,7 +1621,7 @@ void unit_stop_walking_soon(struct block_list& bl, t_tick tick)
 	}
 	else {
 		// Set coordinates to exact coordinates
-		unit_pos(bl, tick, bl.x, bl.y, ud->sx, ud->sy);
+		ud->getpos(bl.x, bl.y, ud->sx, ud->sy, tick);
```
Lemongrass3110, line 1624: Why do we still need ud->sx and ud->sy here?
I think with your new functions we could even remove the ox/oy caching in this function.

Lemongrass3110 (approved):

PR #9143, opened 2025-03-10
Fixed timer SCs killing units

Lemongrass3110:
* **Addressed Issue(s)**: See description below.

* **Server Mode**: Both

The block list pointer bl, the status change pointer sc and the status change entry pointer sce may point to already freed memory, in cases where the status change has killed the unit.


Lemongrass3110:
Has to be addressed in a broader scope.

PR #9144, opened 2025-03-10
Added elemental constructor/destructor

Lemongrass3110:

* **Server Mode**: Both

Added the previously unused constructor and destructor call for elementals.
Follow up to 8515dee


Playtester (approved):
Only verified that the code looks similar to Homunculus and Mercenary types. Did not test ingame.

PR #9145, opened 2025-03-10
Fixed Monsters Not Getting Skill Delays if Target Invalid on Castend

### LINKED ISSUE #9142: Monster Skill Delays are not set when skill fails at castend
### rAthena Hash

36033f8

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When a monster casts a skill and it fails at cast end because the target is no longer valid, no delays are set.

### Expected Result

When a monster casts a skill and it fails at cast end because the target is no longer valid, attackabletime and skill delays are set. The AI is inactive for AttackMotion. The emotion of the skill entry will still be shown.

### How to Reproduce

Hit an Elder at melee so it starts casting skill Fire Bolt. Hide before castend to make the skill fail. After castend, immediately unhide so that Elder doesn't switch to Idle mode. You can see that Elder will do 3 normal attacks before casting Fire Bolt again.

### Official Information

Verified through tests on official servers (see how to reproduce).

I also verified that this is only the case for monster skills. Player aftercast delays are not set if the target is invalid at castend.

Full Elder test results:
Cast Cancel - Sets Attack Delay and Monster Skill Delay, but does not deactivate AI
Cast End Invalid Target - Sets Attack Delay, Monster Skill Delay and does deactivate AI
Cast End Caster Dead - Does not set any delays

Seems the only way to avoid setting delays is when the caster is dead at castend.

_No response_

Playtester:
* **Addressed Issue(s)**: #9142

* **Server Mode**: Both

- Fixed monsters not getting skill delays if the target is invalid on castend
- Fixes #9142

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -13957,7 +13957,7 @@ TIMER_FUNC(skill_castend_id){
 			break;
 
 		// These actions happen even if the skill fails except when the caster is already dead
-		if (md) {
+		if (md != nullptr) {
 			// When a monster uses a skill, its AI will be inactive for its attack motion
 			// This is also the reason why it doesn't move during this time
 			md->next_thinktime = tick + status_get_amotion(src);
@@ -14335,7 +14335,7 @@ TIMER_FUNC(skill_castend_pos){
 			break;
 
 		// These actions happen even if the skill fails except when the caster is already dead
-		if (md) {
+		if (md != nullptr) {
 			// When a monster uses a skill, its AI will be inactive for its attack motion
 			// This is also the reason why it doesn't move during this time
 			md->next_thinktime = tick + status_get_amotion(src);
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -13953,10 +13953,26 @@ TIMER_FUNC(skill_castend_id){
 		bool fail = false;
 		int8 res = USESKILL_FAIL_LEVEL;
 
+		if (status_isdead(*src))
+			break;
+
+		// These actions happen even if the skill fails except when the caster is already dead
+		if (md) {
+			// When a monster uses a skill, its AI will be inactive for its attack motion
+			// This is also the reason why it doesn't move during this time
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -14330,6 +14334,19 @@ TIMER_FUNC(skill_castend_pos){
 		if( status_isdead(*src) )
 			break;
 
+		// These actions happen even if the skill fails except when the caster is already dead
+		if (md != nullptr) {
+			// When a monster uses a skill, its AI will be inactive for its attack motion
+			// This is also the reason why it doesn't move during this time
+			md->next_thinktime = tick + status_get_amotion(src);
+
+			if (md->skill_idx >= 0 && md->db->skill[md->skill_idx]->emotion >= 0)
```
Lemongrass3110, line 14343:
```suggestion
			if (md->skill_idx >= 0 && md->db->skill[md->skill_idx]->emotion >= ET_SURPRISE && md->db->skill[md->skill_idx]->emotion <= ET_MAX)
```

src/map/skill.cpp
```diff
@@ -13953,10 +13953,26 @@ TIMER_FUNC(skill_castend_id){
...
+		if (status_isdead(*src))
+			break;
+
+		// These actions happen even if the skill fails except when the caster is already dead
+		if (md != nullptr) {
+			// When a monster uses a skill, its AI will be inactive for its attack motion
+			// This is also the reason why it doesn't move during this time
+			md->next_thinktime = tick + status_get_amotion(src);
+
+			if (md->skill_idx >= 0 && md->db->skill[md->skill_idx]->emotion >= 0)
```
Lemongrass3110, line 13965:
```suggestion
			if (md->skill_idx >= 0 && md->db->skill[md->skill_idx]->emotion >= ET_SURPRISE && md->db->skill[md->skill_idx]->emotion <= ET_MAX)
```

Lemongrass3110 (approved):

Issue #9146, opened 2025-03-10
Autoloottype input not working as intended


Stingor:
https://github.com/rathena/rathena/blob/b18d4e7e6a476c9b1995c925084ca21388f699bf/src/map/atcommand.cpp#L7053-L7076

Hi, as this (atoi(message) == 0) is tested first, any input like +wea +car +amm
aren't working because atoi(message) will result 0 in all cases

Suggested fix is to place if ((strncmp(message, "healing", 3) == 0) || (atoi(message) == 0)) in last

^^

Lemongrass3110:
#1609

Issue #1609, opened 2016-10-11
Replace atoi usages with strtol


Lemongrass3110:
Currently our code uses atoi way to often, which is considered unsafe and might result in unexpected behaviours.

We have a total of 865 usages in 47 files:
- [ ] src\char\char.c (51 hits)
- [ ] src\char\char_clif.c (1 hit)
- [ ] src\char\char_logif.c (3 hits)
- [ ] src\char\char_mapif.c (12 hits)
- [ ] src\char\inter.c (16 hits)
- [ ] src\char\int_auction.c (15 hits)
- [ ] src\char\int_elemental.c (15 hits)
- [ ] src\char\int_guild.c (38 hits)
- [ ] src\char\int_homun.c (22 hits)
- [ ] src\char\int_mail.c (31 hits)
- [ ] src\char\int_mercenary.c (12 hits)
- [ ] src\char\int_party.c (12 hits)
- [ ] src\char\int_pet.c (10 hits)
- [ ] src\char\int_storage.c (25 hits)
- [ ] src\common\core.c (2 hits)
- [ ] src\common\msg_conf.c (1 hit)
- [ ] src\common\socket.c (4 hits)
- [ ] src\common\sql.c (2 hits)
- [ ] src\common\timer.c (1 hit)
- [ ] src\login\account.c (8 hits)
- [ ] src\login\ipban.c (4 hits)
- [ ] src\login\login.c (14 hits)
- [ ] src\map\atcommand.c (67 hits)
- [ ] src\map\battle.c (1 hit)
- [ ] src\map\buyingstore.c (11 hits)
- [ ] src\map\cashshop.c (4 hits)
- [ ] src\map\channel.c (1 hit)
- [ ] src\map\clif.c (6 hits)
- [ ] src\map\elemental.c (29 hits)
- [ ] src\map\guild.c (5 hits)
- [ ] src\map\homunculus.c (58 hits)
- [ ] src\map\instance.c (5 hits)
- [ ] src\map\itemdb.c (66 hits)
- [ ] src\map\log.c (4 hits)
- [ ] src\map\map.c (11 hits)
- [ ] src\map\mapreg.c (1 hit)
- [ ] src\map\mercenary.c (27 hits)
- [ ] src\map\mob.c (68 hits)
- [ ] src\map\npc.c (9 hits)
- [ ] src\map\pc.c (47 hits)
- [ ] src\map\pc.h (1 hit)
- [ ] src\map\pet.c (18 hits)
[...]

PR #9147, opened 2025-03-10
Fixed atoi usage in atcmd autoloottype

Lemongrass3110:
* **Addressed Issue(s)**: #9146

* **Server Mode**: Both

Replaced atoi usage with strol (see #1609).
Ordered the remaining text checks alphabetically (to clearly show, why it sometimes compares 3 or 4 characters).

Thanks to @Stingor

### DIFF
--- a/src/map/atcommand.cpp
+++ b/src/map/atcommand.cpp
@@ -7051,25 +7051,62 @@ ACMD_FUNC(autoloottype)
 	}
 
 	if (action < 3) { // add or remove
-		if ((strncmp(message, "healing", 3) == 0) || (atoi(message) == 0))
-			type = IT_HEALING;
-		else if ((strncmp(message, "usable", 3) == 0) || (atoi(message) == 2))
-			type = IT_USABLE;
-		else if ((strncmp(message, "etc", 3) == 0) || (atoi(message) == 3))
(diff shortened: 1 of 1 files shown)


src/map/atcommand.cpp
```diff
@@ -7051,25 +7051,62 @@ ACMD_FUNC(autoloottype)
...
-		else if ((strncmp(message, "usable", 3) == 0) || (atoi(message) == 2))
-			type = IT_USABLE;
-		else if ((strncmp(message, "etc", 3) == 0) || (atoi(message) == 3))
-			type = IT_ETC;
-		else if ((strncmp(message, "armor", 3) == 0) || (atoi(message) == 4))
+		char* ptr = nullptr;
+		int32 val = strtol( message, &ptr, 10 );
+
+		// If there was only a number in the input and no other characters
+		if( ptr != nullptr && *ptr == '\0' ){
```
Playtester, line 7058: I found this part a bit hard to understand, because the variable "ptr" doesn't really make it clear that it's the last character from the conversion and it's not easy to understand why the value is a number when the last character is '\0'.

Also wouldn't this also be true if the value is empty string?

The c++ reference says that on fail: "the value of str is stored in the object pointed to by str_end."

PR #9152, opened 2025-03-11
Quick Draw Shot Code Cleanup

Playtester:

* **Server Mode**: Renewal

- Removed useless code for Quick Draw Shot
- Should improve performance
- No functional changes (further fixes required)

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -4528,16 +4528,6 @@ static int32 skill_check_condition_mercenary(struct block_list *bl, uint16 skill
  *------------------------------------------*/
 int32 skill_area_sub_count (struct block_list *src, struct block_list *target, uint16 skill_id, uint16 skill_lv, t_tick tick, int32 flag)
 {
-	switch (skill_id) {
-		case RL_QD_SHOT:
-			{
-				if (src->type == BL_PC && BL_CAST(BL_PC,src)) {
-					struct unit_data *ud = unit_bl2ud(src);
-					if (ud && ud->target == target->id)
-						return 1;
-				}
-			}
-	}
 	return 1;
 }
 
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -12959,22 +12949,16 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
 		}
 		break;
 	case RL_QD_SHOT:
```
Lemongrass3110, line 12951: Why is this even in skill_castend_nodamage_id and not directly in skill_castend_damage_id?

src/map/skill.cpp
```diff
@@ -12959,22 +12949,16 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
...
-			skill_attack(skill_get_type(skill_id), src, src, bl, skill_id, skill_lv, tick, flag|BCT_ENEMY|SD_LEVEL);
-		} else {
-			clif_skill_nodamage(src, *src, skill_id, skill_lv);
-			map_foreachinrange(skill_area_sub, src, skill_get_splash(skill_id, skill_lv), BL_CHAR, src, skill_id, skill_lv, tick, flag|BCT_ENEMY|SD_SPLASH|1, skill_castend_damage_id);
-		}
-		status_change_end(src, SC_QD_SHOT_READY); // End here to prevent spamming of the skill onto the target.
-		skill_area_temp[0] = 0;
+		// Main target always receives damage
+		clif_skill_nodamage(src, *src, skill_id, skill_lv);
+		skill_attack(skill_get_type(skill_id), src, src, bl, skill_id, skill_lv, tick, flag|BCT_ENEMY|SD_LEVEL);
```
Lemongrass3110, line 12958: Should be moved over to skill_castend_damage_id.
There we check: `if (skill_area_temp[1] == bl->id)` already and it could be turned into a simple if/else logic there.

Playtester: I was thinking that too, but there was a deviation in the code. One is using SD_LEVEL and the other is using SD_SPLASH and I wanted to avoid functional changes.

Trying to get a packet capture for this at the moment, but didn't plan to include it in this PR. 
(Usually I would just fix the whole skill, but for the sake of small PRs I only wanted to do a cleanup here.)

src/map/skill.cpp
Playtester: Good question, maybe to prevent recursive calling of the same function?

src/map/skill.cpp
Playtester: Upon further investigation it looks like it sends the same damage packet to all targets including the current one, so I guess you're right and I can optimize it further.

Playtester: How about now?

src/map/skill.cpp
```diff
@@ -7304,10 +7294,9 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 		break;
 
 	case RL_QD_SHOT:
-		if (skill_area_temp[1] == bl->id)
-			break;
-		if (flag&1 && tsc && tsc->getSCE(SC_C_MARKER))
-			skill_attack(skill_get_type(skill_id), src, src, bl, skill_id, skill_lv, tick, flag|SD_ANIMATION);
+		// Except for main target, only units marked with crimson marker are valid targets
+		if (skill_area_temp[1] == bl->id || (tsc && tsc->getSCE(SC_C_MARKER)))
```
Lemongrass3110, line 7298:
```suggestion
		if (skill_area_temp[1] == bl->id || (tsc != nullptr && tsc->getSCE(SC_C_MARKER) != nullptr))
```

Lemongrass3110 (approved):

PR #9156, opened 2025-03-11
Prevent Skill Usage after Normal Attacks for Players

### LINKED ISSUE #9154: Normal attacks should delay skills
### rAthena Hash

b9021da

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When I do a normal attack, I can immediately cast a skill afterwards, even at low ASPD.

### Expected Result

When I do a normal attack, I cannot use a skill for AttackMotion+100 (70 in renewal).

This uses the same logic as the delays given by parry.

Above only applies to players.

### How to Reproduce

Get really low ASPD for example by using a Claw with a Priest or using ASPD debuffs.

Start a normal attack and then spam a skill with no cast time until it works.

### Official Information

Tried with a 70 ASPD Wizard:
After normal attack ~1400ms until I can cast Icewall.

Also further information available.

_No response_

Playtester:
* **Addressed Issue(s)**: #9154 

* **Server Mode**: Both

- Normal attacks now prevent players from using skills for AttackMotion+[200-MaxASPD]*10
  * This is the same delay that is already applied when parrying
- Improved structure of unit_set_attackdelay
  * Now uses an event enum so we know from where it was called
  * Added commented code for the skill use delays that still need to be integrated in the future
- Fixes #9154

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -1815,10 +1815,6 @@ TIMER_FUNC(unit_resume_running){
 
 /**
  * Sets the delays that prevent attacks and skill usage considering the bl type
- * Officially for players it just remembers the last attack time here and applies the delays during the comparison
- * But we pre-calculate the delays instead and store them in attackabletime and canact_tick
- * For ground skills this also applies to mercenaries
- * For non-PCs this function also handles setting of attack delay (which we also store in attackabletime)
  * TODO: Currently this function is only called for normal attacks and parry events and is a work-in-progress
  * @param bl Object to apply attack delay to
  * @param tick Current tick
@@ -1835,9 +1831,10 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
 		case BL_PC:
 			switch (event) {
 				case DELAY_EVENT_ATTACK:
-				//case DELAY_EVENT_CASTBEGIN_ID:
-				//case DELAY_EVENT_CASTBEGIN_POS:
 				case DELAY_EVENT_PARRY:
+					// TODO: This should also happen on cast begin
+					// Officially for players it just remembers the last attack time here and applies the delays during the comparison
+					// But we pre-calculate the delays instead and store them in attackabletime and canact_tick
 					ud->attackabletime = tick + status_get_adelay(&bl);
 					// A fixed delay is added here which is equal to the minimum attack motion you can get
 					// This ensures that at max ASPD attackabletime and canact_tick are equal
@@ -1845,20 +1842,10 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
 					break;
 			}
 			break;
-		case BL_MER:
-			switch (event) {
-				case DELAY_EVENT_ATTACK:
-					ud->attackabletime = tick + status_get_adelay(&bl);
-					break;
-				//case DELAY_EVENT_CASTBEGIN_POS:
-				//	ud->attackabletime = tick + status_get_adelay(&bl);
-				//	ud->canact_tick = tick + status_get_amotion(&bl) + MAX_ASPD_NOPC;
-				//	break;
-			}
-			break;
 		default:
 			switch (event) {
 				case DELAY_EVENT_ATTACK:
+					// This represents setting of attack delay (recharge time) that happens for non-PCs
 					ud->attackabletime = tick + status_get_adelay(&bl);
 					break;
 			}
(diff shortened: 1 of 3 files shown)


src/map/unit.cpp
```diff
@@ -1815,14 +1815,16 @@ TIMER_FUNC(unit_resume_running){
 
 /**
  * Sets the delays that prevent attacks and skill usage considering the bl type
- * Officially it just remembers the last attack time here and applies the delays during the comparison
+ * Officially for players it just remembers the last attack time here and applies the delays during the comparison
```
Lemongrass3110, line 1818: We should not expose the full logic into the documentation of the function.
Only caller relevant information should be in there.
For the caller the first line is enough.
The rest should be incorporated into the logic of the function itself (now we have a [partially] undocumented switch again).

src/map/unit.cpp
```diff
@@ -1815,14 +1815,16 @@ TIMER_FUNC(unit_resume_running){
...
  * But we pre-calculate the delays instead and store them in attackabletime and canact_tick
- * Currently these delays are only applied to PCs
- * TODO: This function should also be called on attacks and cast begin
+ * For ground skills this also applies to mercenaries
+ * For non-PCs this function also handles setting of attack delay (which we also store in attackabletime)
+ * TODO: Currently this function is only called for normal attacks and parry events and is a work-in-progress
  * @param bl Object to apply attack delay to
  * @param tick Current tick
```
Lemongrass3110, line 1824: If it is always current tick, why do we even make the caller send it?

src/map/unit.cpp
```diff
@@ -1831,16 +1833,35 @@ void unit_set_attackdelay(block_list& bl, t_tick tick)
...
 	switch (bl.type) {
 		case BL_PC:
-			ud->attackabletime = tick + status_get_adelay(&bl);
-			// A fixed delay is added here which is equal to the minimum attack motion you can get
-			// This ensures that at max ASPD attackabletime and canact_tick are equal
-			ud->canact_tick = tick + status_get_amotion(&bl) + (pc_maxaspd(reinterpret_cast<map_session_data*>(&bl)) / AMOTION_DIVIDER_PC);
+			switch (event) {
+				case DELAY_EVENT_ATTACK:
+				//case DELAY_EVENT_CASTBEGIN_ID:
+				//case DELAY_EVENT_CASTBEGIN_POS:
```
Lemongrass3110, line 1839:
```suggestion
```
Please add them, when it is time.

src/map/unit.cpp
```diff
@@ -1831,16 +1833,35 @@ void unit_set_attackdelay(block_list& bl, t_tick tick)
...
+				case DELAY_EVENT_PARRY:
+					ud->attackabletime = tick + status_get_adelay(&bl);
+					// A fixed delay is added here which is equal to the minimum attack motion you can get
+					// This ensures that at max ASPD attackabletime and canact_tick are equal
+					ud->canact_tick = tick + status_get_amotion(&bl) + (pc_maxaspd(reinterpret_cast<map_session_data*>(&bl)) / AMOTION_DIVIDER_PC);
+					break;
+			}
 			break;
 		case BL_MER:
-			// TODO: Should set this, but only for ground skills
```
Lemongrass3110, line 1840: Could be restored?

src/map/unit.cpp
```diff
@@ -1831,16 +1833,35 @@ void unit_set_attackdelay(block_list& bl, t_tick tick)
...
+			switch (event) {
+				case DELAY_EVENT_ATTACK:
+					ud->attackabletime = tick + status_get_adelay(&bl);
+					break;
+				//case DELAY_EVENT_CASTBEGIN_POS:
+				//	ud->attackabletime = tick + status_get_adelay(&bl);
+				//	ud->canact_tick = tick + status_get_amotion(&bl) + MAX_ASPD_NOPC;
+				//	break;
```
Lemongrass3110, line 1856:
```suggestion
```
Should be added, when it is time.
Which leaves BL_MER = default case for the time being...

src/map/unit.cpp
Playtester: Using tick from the timer passed through the functions is common to everything we do in rAthena.

src/map/unit.cpp
Playtester: Yeah I was also wondering where to put all the information.

src/map/unit.cpp
Playtester: I moved the code solution to the corresponding issue on github instead.

Lemongrass3110 (approved):

PR #9158, opened 2025-03-12
Monster RaceGroups corrections

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

 - Added missing columns in doc/yaml/sql/mob_db*.sql files, files used by the yaml to sql convertion tool
 - Removed extra combo bonus (mistake after rebase). Thanks to @Pokye !


src/map/map.hpp
```diff
@@ -334,7 +334,7 @@ enum e_race : int8{
 };
 
 enum e_race2 : uint8{
-	RC2_NONE = 0,
+	RC2_NONE = 0,	// Allows to remove RaceGroups in import
```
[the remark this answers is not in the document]
Atemo: No I didn't test, but there is misinterpretation of mine after opening the PR, PR in progress

Lemongrass3110, line 337:
```suggestion
	RC2_NONE = 0,
```
This does not allow anything

src/map/map.hpp
```diff
@@ -1253,7 +1253,7 @@ bool map_setmapflag_sub(int16 m, enum e_mapflag mapflag, bool status, union u_ma
 #define CHK_ELEMENT(ele) ((ele) > ELE_NONE && (ele) < ELE_MAX) /// Check valid Element
 #define CHK_ELEMENT_LEVEL(lv) ((lv) >= 1 && (lv) <= MAX_ELE_LEVEL) /// Check valid element level
 #define CHK_RACE(race) ((race) > RC_NONE_ && (race) < RC_MAX) /// Check valid Race
-#define CHK_RACE2(race2) ((race2) >= RC2_NONE && (race2) < RC2_MAX) /// Check valid Race2
+#define CHK_RACE2(race2) ((race2) > RC2_NONE && (race2) < RC2_MAX) /// Check valid Race2
```
Lemongrass3110, line 1256:
```suggestion
```
Please remove the function(s) completely and move the logic into PC_BONUS_CHK_RACE2

doc/yaml/sql/mob_db.sql
```diff
@@ -63,6 +63,12 @@ CREATE TABLE `mob_db` (
   `racegroup_illusion_turtle` tinyint(1) unsigned DEFAULT NULL,
   `racegroup_rachel_sanctuary` tinyint(1) unsigned DEFAULT NULL,
   `racegroup_illusion_luanda` tinyint(1) unsigned DEFAULT NULL,
+  `racegroup_illusion_frozen` tinyint(1) unsigned DEFAULT NULL,
+  `racegroup_illusion_moonlight` tinyint(1) unsigned DEFAULT NULL,
+  `racegroup_ep16_def` tinyint(1) unsigned DEFAULT NULL,
+  `racegroup_edda_arunafeltz` tinyint(1) unsigned DEFAULT NULL,
+  `racegroup_lasagna` tinyint(1) unsigned DEFAULT NULL,
+  `racegroup_glast_heim_abyss` tinyint(1) unsigned DEFAULT NULL,
```
Lemongrass3110, line 71: Could we detect these issues with CI somehow?

Lemongrass3110 (changes requested):

src/map/mob.cpp
```diff
@@ -5096,7 +5096,7 @@ uint64 MobDatabase::parseBodyNode(const ryml::NodeRef& node) {
 				continue;
 			}
 
-			if (!CHK_RACE2(constant)) {
+			if (constant <= RC2_NONE || constant >= RC2_MAX) {
```
Lemongrass3110, line 5099: I thought we wanted RC2_NONE support for import?
If you want to make a difference between import and normal check it with the "exists" variable.

src/map/script_constants.hpp
```diff
@@ -3464,7 +3464,6 @@
 	export_constant(RC_PLAYER_DORAM);
 	export_constant(RC_ALL);
 
-	export_constant(RC2_NONE);
```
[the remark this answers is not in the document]
Atemo: Do we need it? I cant find any entry anywhere. The format in mob_db.yml is actually using `Groupname: true/false` like
```
    RaceGroups:
      Clocktower: true
```
false to remove it in import

src/map/script_constants.hpp
```diff
@@ -3503,7 +3502,6 @@
 	export_constant(RC2_EDDA_ARUNAFELTZ);
 	export_constant(RC2_LASAGNA);
 	export_constant(RC2_GLAST_HEIM_ABYSS);
-	export_constant(RC2_MAX);
```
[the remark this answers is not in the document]
Atemo: Do you have an example of usage? I cna't find any entry of RC2_MAX in our db, npc folders

src/map/mob.cpp
Atemo: I misinterpreted as said above mb
```
			bool active;

			if (!this->asBool(raceNode, raceName, active))
				return 0;

			if (active)
				mob->race2.push_back(static_cast<e_race2>(constant));
			else
				util::vector_erase_if_exists(mob->race2, static_cast<e_race2>(constant));
		}
```

doc/yaml/sql/mob_db.sql
vstumpf: these should probably be in a new table, just like drops
`monsterid|racegroup|enabled` (enabled is true by default, only set to false if the import removes it)

Lemongrass3110: How is that related to my comment?

But, yes, they should be. But the whole SQL support is only there for backwards compatibility with flux and other CPs.

src/map/mob.cpp
Lemongrass3110: Ah, fine then!

src/map/script_constants.hpp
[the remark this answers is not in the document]
Lemongrass3110: Yes we dont use it currently. But i guess we should deprecate it instead.

src/map/script_constants.hpp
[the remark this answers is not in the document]
Lemongrass3110: Yes we dont use it currently. But i guess we should deprecate it instead.

doc/yaml/sql/mob_db.sql
Atemo: hmm anyway I skip that, the check/revamp should not be done in this PR

Lemongrass3110 (approved):

PR #9160, opened 2025-03-13
Attack Delays on Cast Begin

### LINKED ISSUE #9155: Skills should delay normal attacks
### rAthena Hash

b9021da

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When I use a skill, I can immediately do a normal attack afterwards, even at low ASPD.

### Expected Result

When I use a skill, I cannot do a normal attack for AttackMotion*2 (AttackDelay).

Above only applies to players. The delay is given at castbegin, not at castend.

For mercenaries, these delays only apply when using ground skills.

### How to Reproduce

Get really low ASPD for example by using a Claw with a Priest or using ASPD debuffs.

Use a skill without cast time, then try to do a normal attack afterwards.

Use a skill with cast time, then try to do a normal attack afterwards.

### Official Information

Tried with a 70 ASPD Wizard:

After Icewall ~2600ms until I can do a normal attack.
After Cold Bolt with cast time of 1000ms, ~1600ms until I can do a normal attack.
After Cold Bolt with cast time of 2000ms, ~600ms until I can do a normal attack.

Also further information available.

Playtester:
* **Addressed Issue(s)**: #9155 

* **Server Mode**: Both

- Players can no longer immediately do a normal attack after using a skill
  * Same goes for mercenaries after using a ground skill
- Homunculus and mercenaries now have a fixed server-sided delay of 200ms between skill casts
  * Client-sided it's still slower as it uses attack motion
  * Added a config to go back to the previous, more-secure behavior that uses attack motion server-sided as well
- Code optimization
- Fixes #9155

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -50,6 +50,12 @@ using namespace rathena;
 	#define MIN_POS_INTERVAL 20
 #endif
 
+// Minimum delay after which new client-sided commands for slaves are accepted
+// Applies to mercenaries and homunculus
+#ifndef MIN_DELAY_SLAVE
+	#define MIN_DELAY_SLAVE MAX_ASPD_NOPC * 2
+#endif
+
 // Directions values
 // 1 0 7
 // 2 . 6
@@ -1828,7 +1834,7 @@ TIMER_FUNC(unit_resume_running){
 
 /**
  * Sets the delays that prevent attacks and skill usage considering the bl type
- * TODO: Currently this function is only called for normal attacks and parry events and is a work-in-progress
+ * Currently this function only handles normal attacks, cast begin and parry events
  * @param bl Object to apply attack delay to
  * @param tick Current tick
  * @param event The event that resulted in calling this function
@@ -1844,8 +1850,9 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
 		case BL_PC:
 			switch (event) {
 				case DELAY_EVENT_ATTACK:
+				case DELAY_EVENT_CASTBEGIN_ID:
+				case DELAY_EVENT_CASTBEGIN_POS:
 				case DELAY_EVENT_PARRY:
-					// TODO: This should also happen on cast begin
 					// Officially for players it just remembers the last attack time here and applies the delays during the comparison
 					// But we pre-calculate the delays instead and store them in attackabletime and canact_tick
 					ud->attackabletime = tick + status_get_adelay(&bl);
@@ -1855,7 +1862,7 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
 					break;
 			}
 			break;
-		default:
+		case BL_MOB:
 			switch (event) {
 				case DELAY_EVENT_ATTACK:
 					// This represents setting of attack delay (recharge time) that happens for non-PCs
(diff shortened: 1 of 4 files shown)


src/map/unit.cpp
```diff
@@ -1840,30 +1847,109 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
...
 				case DELAY_EVENT_ATTACK:
 					// This represents setting of attack delay (recharge time) that happens for non-PCs
-					ud->attackabletime = tick + status_get_adelay(&bl);
+					attack_delay = status_get_adelay(&bl);
+					break;
+				case DELAY_EVENT_CASTBEGIN_ID:
+					// For non-PCs that can be controlled from the client, there is a security delay of 200ms
+					// However to prevent tricks to use skills faster, we have a config to use amotion instead
+					if (battle_config.amotion_min_skill_delay == 1 && status_get_amotion(&bl) > MIN_DELAY_SLAVE)
+						act_delay = status_get_amotion(&bl);
```
Lemongrass3110, line 1904:
```suggestion
					if (battle_config.amotion_min_skill_delay == 1)
						act_delay = status_get_amotion(&bl) + MAX_ASPD_NOPC;
```

src/map/unit.cpp
```diff
@@ -1840,30 +1847,109 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
...
+				case DELAY_EVENT_ATTACK:
+					// This represents setting of attack delay (recharge time) that happens for non-PCs
+					attack_delay = status_get_adelay(&bl);
+					break;
+				case DELAY_EVENT_CASTBEGIN_ID:
+				case DELAY_EVENT_CASTBEGIN_POS:
+					// For non-PCs that can be controlled from the client, there is a security delay of 200ms
+					// However to prevent tricks to use skills faster, we have a config to use amotion instead
+					if (battle_config.amotion_min_skill_delay == 1 && status_get_amotion(&bl) > MIN_DELAY_SLAVE)
+						act_delay = status_get_amotion(&bl);
```
Lemongrass3110, line 1888:
```suggestion
					if (battle_config.amotion_min_skill_delay == 1)
						act_delay = status_get_amotion(&bl) + MAX_ASPD_NOPC;
```

src/map/unit.cpp
```diff
@@ -1840,30 +1847,109 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
 	if (ud == nullptr)
 		return;
 
+	uint16 attack_delay = 0;
+	uint16 act_delay = 0;
```
Lemongrass3110, line 1851:
```suggestion
	t_tick attack_delay = 0;
	t_tick act_delay = 0;
```

src/map/unit.cpp
```diff
@@ -1840,30 +1847,109 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
...
 					break;
 			}
 			break;
-		default:
+		case BL_MOB:
+			switch (event) {
+				case DELAY_EVENT_ATTACK:
+					// This represents setting of attack delay (recharge time) that happens for non-PCs
+					attack_delay = status_get_adelay(&bl);
+					break;
```
Lemongrass3110, line 1874: case DELAY_EVENT_CASTBEGIN_ID:
				case DELAY_EVENT_CASTBEGIN_POS:
Please add documentation, why we dont do anything here.

src/map/unit.cpp
```diff
@@ -1840,30 +1847,109 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
...
+		ud->canact_tick = i64max(tick + act_delay, ud->canact_tick);
+}
+
+/**
+ * Updates skill delays according to cast time and minimum delay, and applies security casttime
+ * @param bl Object to apply update delay for
+ * @param tick Current tick
+ * @param event The event that resulted in calling this function
+ */
+void unit_set_castdelay(unit_data& ud, t_tick tick, int32 casttime) {
```
Lemongrass3110, line 1942: I am kind of confused, why this is not part of the other function?

src/map/unit.cpp
```diff
@@ -2378,8 +2462,13 @@ int32 unit_skilluse_id2(struct block_list *src, int32 target_id, uint16 skill_id
 	if( casttime <= 0 )
 		ud->state.skillcastcancel = 0;
 
-	if (!sd || sd->skillitem != skill_id || skill_get_cast(skill_id, skill_lv))
-		ud->canact_tick = tick + i64max(casttime, max(status_get_amotion(src), battle_config.min_skill_delay_limit)) + SECURITY_CASTTIME;
+	// Skills used from items don't seem to give any attack or act delay
+	if (sd == nullptr || sd->skillitem != skill_id) {
```
Lemongrass3110, line 2466: I think this logic should be moved into unit_set_attackdelay.
The function call should be moved after setting ud->skill_id, then you access it in unit_set_attackdelay

src/map/unit.cpp
```diff
@@ -2561,8 +2650,14 @@ int32 unit_skilluse_pos2( struct block_list *src, int16 skill_x, int16 skill_y,
 #endif
 
 	ud->state.skillcastcancel = castcancel&&casttime>0?1:0;
-	if (!sd || sd->skillitem != skill_id || skill_get_cast(skill_id, skill_lv))
-		ud->canact_tick = tick + i64max(casttime, max(status_get_amotion(src), battle_config.min_skill_delay_limit)) + SECURITY_CASTTIME;
+
+	// Skills used from items don't seem to give any attack or act delay
+	if (sd == nullptr || sd->skillitem != skill_id) {
```
Lemongrass3110, line 2655: I think this logic should be moved into unit_set_attackdelay.
The function call should be moved after setting ud->skill_id, then you access it in unit_set_attackdelay

src/map/unit.cpp
Playtester: The return type of status_get_adelay and status_get_amotion is uint16, that's why I'm using it here. Why reserve more memory when not needed?

src/map/unit.cpp
Playtester: Do we really want to increase the delay by MAX_ASPD_NOPC? Also here you are making the assumption that MIN_DELAY_SLAVE is always MAX_ASPD_NOPC*2, which only works until someone customizes that value.

src/map/unit.cpp
Playtester: Same question as above.

src/map/unit.cpp
Playtester: This is only needed at castbegin and independent of bl type.

It's based on your suggestion to unify similar code from _id and _pos functions.

src/map/unit.cpp
Playtester: Mainly did it like this based on the principle that functions should be kept small. Ideally it is said a function should be fully visible on 1-2 screens.

Sure, I could make unit_set_attackdelay a massive 10 page function that handles everything delay-related, but that doesn't seem to follow good coding standards.

And making it depend on certain variables being set beforehand seems also dangerous.

src/map/unit.cpp
Playtester: See above.

src/map/unit.cpp
Playtester: Also please consider that unit_set_attackdelay is designed to be bl specific. Things that should happen independent of bl don't really fit into it (you'd need to make a separate switch/if outside of the bl->type switch to add logic like this there).

src/map/unit.cpp
Playtester: Done.

src/map/unit.cpp
Lemongrass3110: When that setting is set, I would harmonize the behavior.
Like you did it for BL_MER in DELAY_EVENT_CASTBEGIN_POS

Lemongrass3110 (approved):

Issue #9157, opened 2025-03-12
Damage beyond 2.1B not showed up


Goendoelz:
### rAthena Hash

5e6d6bf9b728738467d041ba46e99f89303bd61e

### Client Date

2024-10-30

### Server Mode

Renewal

### Result

If the damage dealt higher than 2.1b (assuming the cap for 32 bit), the damage number doesnt show up

![Image](https://github.com/user-attachments/assets/358b8de0-dc94-4d32-95e0-81754f30596b)

![Image](https://github.com/user-attachments/assets/92e71987-2708-44d8-9715-56c6aa5dc26b)

### Relevant Log Output

```Shell

```

### Expected Result

It should be shown even got capped numbers.

### How to Reproduce

Deals any damage beyond the capped number.

### Official Information

It should be shown as capped number (2.147m)

Pokye:
Invalid. 

![Image](https://github.com/user-attachments/assets/562171b6-de3e-49f1-8890-190bb219d872)

skstrife:
This issue occurs since **commit** ea359b4 made by @Atemo and @Lemongrass3110

On my server I used a temporary solution because when I found the problem I was told to open a PR, but honestly, I don't identify much with GitHub, so I didn't.

skstrife:
I hope I'm not mistaken (because one of those commits also broke the pet windows), but what I did to fix this problem was the following:

First, I commented out:
`//auto damage = std::min(static_cast<decltype(packet.damage)>(sdamage), std::numeric_limits<decltype(packet.damage)>::max());` proposed in the commit I mentioned above.

Second, I wrote it like this:
```
auto max_damage = std::numeric_limits<decltype(packet.damage)>::max();
auto damage = (sdamage > max_damage) ? max_damage : static_cast<decltype(packet.damage)>(sdamage);
```

This was my solution, as I mentioned above the ideal is to wait for the official solution from @Lemongrass3110 and @Atemo. I hope they can provide an official solution soon and I hope this temporary solution helps you. @Goendoelz

dweipert-3138720606:
@skstrife you could also just post the output of `git diff` as a comment, then you can be 100% certain that you communicated all the lines that you changed and someone else can make the PR on github.

PR #9161, opened 2025-03-13
Corrected "min" usage

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9157

* **Server Mode**: both

 - [ ] Further testing

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -6051,7 +6051,7 @@ int32 clif_skill_damage( block_list& src, block_list& dst, t_tick tick, int32 sd
 	packet.attackMT = sdelay;
 	packet.attackedMT = ddelay;
 
-	auto damage = std::min( static_cast<decltype(packet.damage)>( sdamage ), std::numeric_limits<decltype(packet.damage)>::max() );
+	auto damage = static_cast<decltype(packet.damage)>( std::min<decltype( sdamage )>( sdamage, std::numeric_limits<decltype(packet.damage)>::max() ) );
 
 	if (battle_config.hide_woe_damage && map_flag_gvg(src.m)) {
 		packet.damage = static_cast<decltype(packet.damage)>(damage ? div : 0);
@@ -6157,7 +6157,7 @@ bool clif_skill_nodamage( block_list* src, block_list& dst, uint16 skill_id, int
 
 	p.PacketType = HEADER_ZC_USE_SKILL;
 	p.SKID = skill_id;
-	p.level = std::min( static_cast<decltype(p.level)>( heal ), std::numeric_limits<decltype(p.level)>::max() );
+	p.level = static_cast<decltype(p.level)>( std::min<decltype( heal )>( heal, std::numeric_limits<decltype(p.level)>::max() ) );
 	p.targetAID = dst.id;
 	p.result = success;
 	if(src != nullptr){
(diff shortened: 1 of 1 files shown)


vstumpf:
Can we make this a common function? something like: 
```c++
template <typename T, typename V>
using larger_type = std::conditional_t<sizeof(V) < sizeof(T), T, V>;

template<class T, class V>
constexpr const std::enable_if_t<
    (std::is_signed<T>::value && std::is_signed<V>::value) ||
    (!std::is_signed<T>::value && !std::is_signed<V>::value), T>
 clampToSize(const V& value) {
    return std::clamp<larger_type<T, V>>(value, numeric_limits<T>::min(), numeric_limits<T>::max());
}

template<class T, class V>
void clampToSizeWrite(T& out, const V& value) {
    out = clampToSize<T>(value);
}


// you can use this like
    int32_t pdamage;
    clampToSizeWrite(pdamage, 0x7fff'ffff'ffff'ffffll);

packet.damage = clampToSize<decltype(packet.damage)>(sdamage);
```

https://godbolt.org/z/jePM9aGfa

vstumpf:
If we want to be able to use different signs, then here's an implementation:
```

// cases
// T = signed <=> V signed
// just use intmax_t

template<class T, class V>
constexpr const std::enable_if_t<
    (std::is_signed<T>::value && std::is_signed<V>::value), T>
  clampToSize(const V& value) {
    return std::clamp<intmax_t>(value, numeric_limits<T>::min(), numeric_limits<T>::max());
}

// T = unsigned <=> V unsigned
// just use uintmax_t

template<class T, class V>
constexpr const std::enable_if_t<
    (!std::is_signed<T>::value && !std::is_signed<V>::value), T>
  clampToSize(const V& value) {
    return std::clamp<uintmax_t>(value, numeric_limits<T>::min(), numeric_limits<T>::max());
}

// T = signed > V = unsigned
// the value of V will fit in T, because sizeof(T) > sizeof(V)
template<class T, class V>
constexpr const std::enable_if_t<
    (std::is_signed<T>::value && !std::is_signed<V>::value && sizeof(T) > sizeof(V)), T>
  clampToSize(const V& value) {
    return std::clamp<intmax_t>(value, numeric_limits<T>::min(), numeric_limits<T>::max());
  }
// T = signed <= V = unsigned
// we need to just minimize the value to the T::max()
template<class T, class V>
constexpr const std::enable_if_t<(std::is_signed<T>::value && !std::is_signed<V>::value && sizeof(T) <= sizeof(V)), T>
clampToSize(const V& value) {
    return std::min<V>(value, numeric_limits<T>::max());
}

// T = unsigned > V = signed
// if V < 0, then V => 0. If V > 0, stays the same
template<class T, class V>
constexpr const std::enable_if_t<
    (!std::is_signed<T>::value && std::is_signed<V>::value && sizeof(T) > sizeof(V)), T>
  clampToSize(const V& value) {
    if (value < 0)
        return 0;
    return value;
  }

// T = unsigned <= V = signed
// if V < 0, then V => 0. If V > 0, then minimize value to T::max()
template<class T, class V>
constexpr const std::enable_if_t<
    (!std::is_signed<T>::value && std::is_signed<V>::value && sizeof(T) <= sizeof(V)), T>
  clampToSize(const V& value) {
    if (value < 0)
        return 0;
    return std::min<V>(value, numeric_limits<T>::max());
  }
```
https://godbolt.org/z/Wc8nqTGMa

lighta:
lot of auto magic but seem ok.
since this only impact client it have reduced bad impact.
PR is ok for me.

PR #9162, opened 2025-03-13
Removed JName column from mob DB

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

This is a suggestion to remove the jname column from the mob DB. This extra column was added in [2006](https://github.com/eathena/eathena/commit/b96b459dfa780dc7f96e75cc8a72043b97c06a4c). The column has never been updated and, unless I'm mistaken, is no longer used. If needed, import DB feature covers most of its functionality.


neico:
How about turning the name field into a keyvalue/map instead?
That way it can be mapped to things like `@langtype`, setting a generalized access-point for multilanguage support.
And if it's not a map, meaning a singular value, then it acts like the current name field, the english name.
(otherwise, non-english entries override the english name on a per-account basis via `@langtype`/`#langtype`)

Hm, looking at the changes though, it seems that removing an entry doesn't require increasing the db version?
Because I assume my suggestion would certainly end up requiring an increase, which is why I brought it up, that when it's changed already then that'd be the best place to put that in.

conf/battle/monster.conf
```diff
@@ -117,7 +117,6 @@ monster_active_enable: yes
 // Should the mob_db names override the mob names specified in the spawn files?
```
Lemongrass3110, line 117:
```suggestion
// Should the mob_db names override the mob names specified in the spawn files? (Note 1)
```

conf/battle/monster.conf
```diff
@@ -117,7 +117,6 @@ monster_active_enable: yes
...
 // 0: No
```
Lemongrass3110, line 118:
```suggestion
// no: use the name specified at the spawn (default)
```

conf/battle/monster.conf
```diff
@@ -117,7 +117,6 @@ monster_active_enable: yes
...
 // 1: always use the mob_db Name column (english mob name)
```
Lemongrass3110, line 119:
```suggestion
// yes: always use the mob_db Name column
```

conf/battle/monster.conf
```diff
@@ -117,7 +117,6 @@ monster_active_enable: yes
...
-// 2: always use the mob_db JName column (original Kro mob name)
 override_mob_names: 0
```
Lemongrass3110, line 120:
```suggestion
override_mob_names: no
```

src/map/atcommand.cpp
```diff
@@ -8029,9 +8028,9 @@ ACMD_FUNC(mobinfo)
 #endif
 		// stats
 		if( mob->get_bosstype() == BOSSTYPE_MVP )
-			sprintf(atcmd_output, msg_txt(sd,1240), mob->name.c_str(), mob->jname.c_str(), mob->sprite.c_str(), mob->id); // MVP Monster: '%s'/'%s'/'%s' (%d)
+			sprintf(atcmd_output, msg_txt(sd,1240), mob->name.c_str(), mob->name.c_str(), mob->sprite.c_str(), mob->id); // MVP Monster: '%s'/'%s'/'%s' (%d)
```
Lemongrass3110, line 8031: Think this needs to be removed instead (also %s from msg 1240)

src/map/atcommand.cpp
```diff
@@ -8029,9 +8028,9 @@ ACMD_FUNC(mobinfo)
...
 		else
-			sprintf(atcmd_output, msg_txt(sd,1241), mob->name.c_str(), mob->jname.c_str(), mob->sprite.c_str(), mob->id); // Monster: '%s'/'%s'/'%s' (%d)
+			sprintf(atcmd_output, msg_txt(sd,1241), mob->name.c_str(), mob->name.c_str(), mob->sprite.c_str(), mob->id); // Monster: '%s'/'%s'/'%s' (%d)
```
Lemongrass3110, line 8033: Think this needs to be removed instead (also %s from msg 1241)

src/map/script.cpp
```diff
@@ -14577,7 +14577,7 @@ BUILDIN_FUNC(strmobinfo)
 
 	switch (num) {
 	case 1: script_pushstrcopy(st,mob->name.c_str()); break;
-	case 2: script_pushstrcopy(st, mob->jname.c_str()); break;
+	case 2: script_pushstrcopy(st, mob->name.c_str()); break;	// !TODO
```
Lemongrass3110, line 14580: Maybe add a warning here?

PR #9167, opened 2025-03-14
Fixed Looters on monster_loot_search_type=0

Playtester:
* **Addressed Issue(s)**: #8232

* **Server Mode**: Both

- Looters will now fully benefit from the previous fixes even if monster_loot_search_type is set to 0
- Follow-up to 78a2bab
- Related to #8232

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1406,7 +1406,7 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
 	}
 	else if( !battle_config.monster_loot_search_type ){
 		// Stop walking immediately if item is no longer on the ground.
-		unit_stop_walking( &md->bl, USW_FIXPOS );
+		unit_stop_walking_soon(md->bl, USW_FIXPOS);
 	}
 
 	return 0;


src/map/mob.cpp
```diff
@@ -1406,7 +1406,7 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
 	}
 	else if( !battle_config.monster_loot_search_type ){
 		// Stop walking immediately if item is no longer on the ground.
```
vstumpf, line 1408: change the comment?

Playtester: Fixed!

vstumpf (approved):

src/map/mob.cpp
```diff
@@ -1405,8 +1405,8 @@ static int32 mob_ai_sub_hard_lootsearch(struct block_list *bl,va_list ap)
 		md->target_id = bl->id;
 	}
 	else if( !battle_config.monster_loot_search_type ){
-		// Stop walking immediately if item is no longer on the ground.
-		unit_stop_walking( &md->bl, USW_FIXPOS );
+		// Stop walking after 0.5-1.5 cells if item is no longer on the ground.
```
Lemongrass3110, line 1408:
```suggestion
		// Stop walking as soon as possible, if the target item is no longer on the ground.
```

Lemongrass3110 (approved):

PR #9168, opened 2025-03-14
Loot Closest Item Feature Improvement

Playtester:

* **Server Mode**: Both

- When setting monster_loot_search_type to 0, monsters will now reliably switch to the closest item target
  * This works now even when a new item appears that is closer while the monster is already chasing an item
  * Does not apply when the monster already has a target that is not an item
- Removed previous code that shortened the walkpath but only worked in a few edge cases
  * Now uses the normal process of stopping when target is gone

Changes walkpath from:
grafik

To:
grafik

(Assuming new item was dropped when monster is at "6" and monster_loot_search_type is set to 0.)

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1994,7 +1994,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 	if (can_move && mode&MD_LOOTER && md->lootitems && DIFF_TICK(tick, md->ud.canact_tick) > 0 &&
 		(md->lootitem_count < LOOTITEM_SIZE || battle_config.monster_loot_type != 1))
 	{
-		if (!tbl) {
+		if (tbl == nullptr) {
 			// Search for items in loot range
 			map_foreachinshootrange(mob_ai_sub_hard_lootsearch, &md->bl, battle_config.loot_range, BL_ITEM, md, &tbl);
 		}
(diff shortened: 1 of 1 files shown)


src/map/mob.cpp
```diff
@@ -1993,10 +1989,21 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
...
 	// Scan area for targets
-	if (!tbl && can_move && mode&MD_LOOTER && md->lootitems && DIFF_TICK(tick, md->ud.canact_tick) > 0 &&
+
+	// Scan area for items to loot, avoid trying to loot if the mob is full and can't consume the items.
+	if (can_move && mode&MD_LOOTER && md->lootitems && DIFF_TICK(tick, md->ud.canact_tick) > 0 &&
 		(md->lootitem_count < LOOTITEM_SIZE || battle_config.monster_loot_type != 1))
-	{	// Scan area for items to loot, avoid trying to loot if the mob is full and can't consume the items.
-		map_foreachinshootrange (mob_ai_sub_hard_lootsearch, &md->bl, battle_config.loot_range, BL_ITEM, md, &tbl);
+	{
+		if (!tbl) {
```
Lemongrass3110, line 1997:
```suggestion
		if (tbl == nullptr) {
```

Lemongrass3110 (approved):

PR #9169, opened 2025-03-14
Speedrate Bonus Cleanup

crazy-arashi:

* **Server Mode**: Renewal

Implemented a new bonus that increases speed and display the necessary icon client side when an item is equipped.
Correct bSpeedAddRate behavior to match its documented behavior.

I only added the new bonus on items that I was able to test in kRO.

To Do:

- [ ] Confirm if the icon exists in pre-renewal
- [ ] Test other items that display the icon when given the movement speed bonus


Playtester:
If you're already doing this, please check also how this bonus stacks with usable items that give the same status change. Some examples would be Inspector_Certificate, Box_Of_Thunder and Moon_Cake.

PR #9173, opened 2025-03-17
Restored INF in clif_skillinfo

Lemongrass3110:

* **Server Mode**: Both

In fb066e9 the possibility to manually force set INF was accidentally removed.
This causes combos to not work correctly.

Thanks to @Playtester


src/map/clif.cpp
```diff
@@ -5868,7 +5868,11 @@ void clif_skillinfo( map_session_data& sd, uint16 skill_id ){
 	p.level = sd.status.skill[idx].lv;
 	p.sp = static_cast<decltype(p.sp)>( skill_get_sp( skill_id,sd.status.skill[idx].lv ) );
 	p.range2 = static_cast<decltype(p.range2)>( skill_get_range2( &sd.bl,skill_id,sd.status.skill[idx].lv,false ) );
-	p.inf = skill_get_inf( skill_id );
+	if( inf == INF_PASSIVE_SKILL ){
```
Playtester, line 5871: Should we maybe introduce a separate constant for this? Just in case someone wants to implement a custom skill that can switch between passive and active.

Playtester (approved):
Verified: 
Combo Finish -> Asura Strike Combo
Tumbling -> Jump Kick Combo
Both working fine again now!

PR #9177, opened 2025-03-18
Overweight checks cleanup and fixes

Daegaladh:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* Replaced specific weight checks pc_is50overweight, pc_is70overweight and pc_is90overweight with pc_getpercentweight() function and configurable battle conf parameters.
* Unified natural_heal_weight_rate battle conf
* Fixed open box weight check, it's actually 70% by default on pre-re and  90% on renewal
* Cleanup of pc_updateweightstatus() function
* Replaced some weight checks with status SC_WEIGHT90 / SC_WEIGHT50 status checks
* Used new pc_getpercentweight() function where possible.
* Changed overweight regen state to bool and updated some checks that previously checked for SC_WEIGHT50 and SC_WEIGHT90 instead
* Added EndOnStart values on status.yml for SC_WEIGHT50 and SC_WEIGHT90 to cancel out each other

### DIFF
--- a/conf/battle/player.conf
+++ b/conf/battle/player.conf
@@ -60,15 +60,16 @@ natural_healsp_interval: 8000
 // Automatic healing skill's time interval. (in milliseconds)
 natural_heal_skill_interval: 10000
 
-// The maximum weight for a character to carry before the it stops healing naturally. (in %)
-// For renewal: Requires client 20171025 or newer to display properly
+// The maximum weight for a character to carry before it stops healing naturally. (Note 2)
+// Depending on this configuration you may need to change the corresponding TGA image file
+// in your data in order to show the proper percentage in the status icon.
 // Default on official servers: 50 for Pre-renewal, 70 for Renewal
 //natural_heal_weight_rate: 70
 
-// The maximum weight for a character to get an item from item boxes. (in %)
+// The maximum weight for a character to get an item from item boxes. (Note 2)
 open_box_weight_rate: 70
 
-// The maximum weight for a character to carry before entering the major overweight state. (in %)
+// The maximum weight for a character to carry before entering the major overweight state. (Note 2)
 // In this state the character cannot heal naturally, attack or use skills.
 major_overweight_rate: 90
 
--- a/conf/battle/player.conf
+++ b/conf/battle/player.conf
@@ -60,10 +60,17 @@ natural_healsp_interval: 8000
 // Automatic healing skill's time interval. (in milliseconds)
 natural_heal_skill_interval: 10000
 
-// The maximum weight for a character to carry when the character stops healing naturally. (in %)
+// The maximum weight for a character to carry before the it stops healing naturally. (in %)
 // For renewal: Requires client 20171025 or newer to display properly
-natural_heal_weight_rate: 50
-natural_heal_weight_rate_renewal: 70
+// Default on official servers: 50 for Pre-renewal, 70 for Renewal
+//natural_heal_weight_rate: 70
+
+// The maximum weight for a character to get an item from item boxes. (in %)
+open_box_weight_rate: 70
+
+// The maximum weight for a character to carry before entering the major overweight state. (in %)
+// In this state the character cannot heal naturally, attack or use skills.
+major_overweight_rate: 90
 
 // Maximum atk speed. (Default 190, Highest allowed 199)
 max_aspd: 190
--- a/conf/battle/player.conf
+++ b/conf/battle/player.conf
@@ -67,7 +67,8 @@ natural_heal_skill_interval: 10000
 //natural_heal_weight_rate: 70
 
 // The maximum weight for a character to get an item from item boxes. (Note 2)
-open_box_weight_rate: 70
+// Default on official servers: 70 for Pre-renewal, 90 for Renewal
+//open_box_weight_rate: 90
 
 // The maximum weight for a character to carry before entering the major overweight state. (Note 2)
 // In this state the character cannot heal naturally, attack or use skills.
--- a/src/map/buyingstore.cpp
+++ b/src/map/buyingstore.cpp
@@ -215,7 +215,7 @@ int8 buyingstore_create( map_session_data* sd, int32 zenylimit, unsigned char re
 		return 5;
 	}
 
-	if( (sd->max_weight*90)/100 < weight )
+	if( pc_getpercentweight(*sd) >= battle_config.major_overweight_rate )
 	{// not able to carry all wanted items without getting overweight (90%)
 		sd->buyingstore.slots = 0;
 		clif_buyingstore_open_failed(sd, BUYINGSTORE_CREATE_OVERWEIGHT, weight);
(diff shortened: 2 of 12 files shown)


src/map/pc.cpp
```diff
@@ -2942,44 +2942,51 @@ uint64 pc_calc_skilltree_normalize_job( map_session_data *sd ){
 	return c;
 }
 
+uint8 pc_getpercentweight(map_session_data &sd)
```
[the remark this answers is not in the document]
Daegaladh: Yeah you're right, I'll change it to uint16. If some other function such as min and max requires a specific type we can just cast

conf/battle/player.conf
```diff
@@ -60,10 +60,17 @@ natural_healsp_interval: 8000
 // Automatic healing skill's time interval. (in milliseconds)
 natural_heal_skill_interval: 10000
 
-// The maximum weight for a character to carry when the character stops healing naturally. (in %)
+// The maximum weight for a character to carry before the it stops healing naturally. (in %)
```
Playtester, line 63: Grammar: "the it"

conf/battle/player.conf
```diff
@@ -60,10 +60,17 @@ natural_healsp_interval: 8000
...
 // For renewal: Requires client 20171025 or newer to display properly
```
Playtester, line 64: Probably should instead say that a client of 20171025 or newer displays "70%" and an older client displays "50% on the icon regardless of this setting. (At least I'm assuming this is what this hint here means.)

conf/battle/player.conf
```diff
@@ -60,10 +60,17 @@ natural_healsp_interval: 8000
...
-natural_heal_weight_rate: 50
-natural_heal_weight_rate_renewal: 70
+// Default on official servers: 50 for Pre-renewal, 70 for Renewal
+//natural_heal_weight_rate: 70
+
+// The maximum weight for a character to get an item from item boxes. (in %)
```
Playtester, line 68: (Note 2)

conf/battle/player.conf
```diff
@@ -60,10 +60,17 @@ natural_healsp_interval: 8000
...
+open_box_weight_rate: 70
+
+// The maximum weight for a character to carry before entering the major overweight state. (in %)
```
Playtester, line 71: (Note 2)

src/map/buyingstore.cpp
```diff
@@ -215,7 +215,7 @@ int8 buyingstore_create( map_session_data* sd, int32 zenylimit, unsigned char re
 		return 5;
 	}
 
-	if( (sd->max_weight*90)/100 < weight )
```
Playtester, line 218: Are the adjustments done to the "weight" variable above no longer relevant?

conf/battle/player.conf
Daegaladh: Actually it depends on the ¹«°Ô50ÀÌ»ó.tga icon. Gravity changed the number in the icon, but not the name file.
I'll try to explain it there.

Playtester (approved):
Did not test ingame, but looks good now.

src/map/skill.cpp
```diff
@@ -19191,11 +19191,7 @@ bool skill_check_condition_castbegin( map_session_data& sd, uint16 skill_id, uin
 			}
 			break;
 		case ST_RECOVER_WEIGHT_RATE:
-#ifdef RENEWAL
-			if(pc_is70overweight(&sd)) {
-#else
-			if(pc_is50overweight(&sd)) {
-#endif
+			if( pc_getpercentweight(sd) >= battle_config.natural_heal_weight_rate ) {
```
Lemongrass3110, line 19194: Should use SC_WEIGHT50 check?

src/map/skill.cpp
```diff
@@ -19191,11 +19191,7 @@ bool skill_check_condition_castbegin( map_session_data& sd, uint16 skill_id, uin
...
 				clif_skill_fail( sd, skill_id );
```
Lemongrass3110, line 19195: Did we confirm that this is the correct error message for TF_PICKSTONE and AC_MAKINGARROW in this case?

src/map/pc.cpp
```diff
@@ -2942,44 +2942,58 @@ uint64 pc_calc_skilltree_normalize_job( map_session_data *sd ){
...
-	// start new status change
-	if( new_overweight == 1 )
-		sc_start(&sd->bl,&sd->bl, SC_WEIGHT50, 100, 0, 0);
-	else if( new_overweight == 2 )
-		sc_start(&sd->bl,&sd->bl, SC_WEIGHT90, 100, 0, 0);
 
-	// update overweight status
-	sd->regen.state.overweight = new_overweight;
+	// Update overweight status
+	sd.regen.state.overweight = new_overweight;
```
Lemongrass3110, line 2996:
```suggestion
	sd.regen.state.overweight = new_overweight != 0;
```
The code in status.cpp does not really care, if it is 1 or 2.

src/map/skill.cpp
Daegaladh: Yeah I've checked and once you have the SC_WEIGHT50 status all skills that get you items, show the default skill fail message

Daegaladh: Yeah I've checked and once you have the SC_WEIGHT50 status all skills that get you items, show the default skill fail message

src/map/status.cpp
```diff
@@ -13418,7 +13418,7 @@ int32 status_change_end(struct block_list* bl, enum sc_type type, int32 tid)
 			}
 			break;
 		case SC_TENSIONRELAX:
-			if (sc && (sc->getSCE(SC_WEIGHT50) || sc->getSCE(SC_WEIGHT90)))
+			if ( sc != nullptr && ( sc->getSCE(SC_WEIGHT50) != nullptr || sc->getSCE(SC_WEIGHT90) != nullptr ) )
```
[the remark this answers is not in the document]
Daegaladh: Maybe we should check sd.regen.state.overweight instead?

src/map/skill.cpp
Daegaladh: I've changed it to check SC_WEIGHT50 and SC_WEIGHT90, but maybe we should check sd.regen.state.overweight instead?

src/map/pc.cpp
Daegaladh: Should I change sd.regen.state.overweight to bool?

src/map/status.cpp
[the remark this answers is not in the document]
Lemongrass3110: Would make sense!

src/map/skill.cpp
Lemongrass3110: Would make sense!

src/map/pc.cpp
Lemongrass3110: Yes please!

Lemongrass3110 (approved):

PR #9184, opened 2025-03-19
Walk Delay on Attack/Skill Cleanup/Fix

Playtester:
* **Addressed Issue(s)**: Refactor/Cleanup

* **Server Mode**: Both

- Replaced the special "walkdelay by inactive AI on attack" solution with a separate monster_ai setting
- Cast end and cast cancel delay events will now also go through unit_set_attackdelay
- Monster-specific delays are now handled by a new mob_set_delay function called from unit_set_attackdelay
  * It handles setting the AI inactive on attack and cast end
  * And setting monster skill delays on cast end and cast cancel
- Attack delays at cast end for monsters is now handled by unit_set_attackdelay instead of mobskill_end
  * Renamed mobskill_end to mobskill_delay as this function now only handles setting monster skill delays
- Disabled attack_walk_delay by default and improved the documentation of this setting
  * This has no visible change as the client will still send movement requests delayed by amotion
  * Feel free to re-enable for "PC" if you want to prevent client edits allowing to move earlier


conf/battle/battle.conf
```diff
@@ -25,11 +25,12 @@ mob_critical_rate: 100
 critical_rate: 100
 
 // Should normal attacks give you a walk delay? (Note 3)
```
Lemongrass3110, line 27:
```suggestion
// Which unit types should get a walk delay from normal attacks? (Note 3)
```

src/map/unit.cpp
```diff
@@ -3260,13 +3273,8 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
...
 		// You can't move during your attack motion
-		if (src->type&battle_config.attack_walk_delay) {
-			// Monsters set their AI inactive instead of having a walkdelay
-			if (md != nullptr)
-				md->next_thinktime = tick + status_get_amotion(src);
-			else
-				unit_set_walkdelay(src, tick, sstatus->amotion, 1);
-		}
+		if (src->type&battle_config.attack_walk_delay)
+			unit_set_walkdelay(src, tick, sstatus->amotion, 1);
```
Lemongrass3110, line 3277: Why dont we do that in unit_set_attackdelay?

src/map/unit.cpp
```diff
@@ -1938,8 +1942,17 @@ void unit_set_attackdelay(block_list& bl, t_tick tick, e_delay_event event)
...
+	if (act_delay > 0) {
+		if (bl.type == BL_MOB && !(battle_config.mob_ai&0x2000)) {
+			// Monsters don't process their AI for this duration instead of having a canact_tick
+			// This is also the reason why they don't move during this time
+			mob_data& md = reinterpret_cast<mob_data&>(bl);
+			md.next_thinktime = i64max(tick + act_delay, md.next_thinktime);
+		}
+		else if(bl.type != BL_MOB) {
+			ud->canact_tick = i64max(tick + act_delay, ud->canact_tick);
+		}
```
Lemongrass3110, line 1954: Please move this into the switch above and set act_delay to 0 if needed.

src/map/unit.cpp
Playtester: Originally I was thinking about integrating unit_set_walkdelay into unit_set_attackdelay later on, but since I know now that this is just a custom feature, I kinda want to leave it separate.

src/map/unit.cpp
Playtester: Why? Here, it's the perfect place so we have the logic (and the check for the config) all in one place. Otherwise we run the risk that a developer will set canact_tick for monsters.

I also asked you if you're okay with this change before I started working on it.

src/map/unit.cpp
Lemongrass3110: battle_config.mob_ai&0x2000 is also a custom feature and was also moved into the function

src/map/unit.cpp
Playtester: Reworked it completely not to have a clean split between attack/act delays and monster-specific delays.

Lemongrass3110 (approved):

Issue #9174, opened 2025-03-18
Updated some script of Utan kid


StrawberriesX1:
### rAthena Hash

3f1cee7a

### Client Date

20220406

### Server Mode

Renewal

### Result

1. talk to npc Utan kid umbala,59,243
2. select option 1 (give him meat)
3. now Utan kid gives all 3 items, jellopy fluff and clover, and gives the same amount of 2 for all.

### Relevant Log Output

```Shell

```

### Expected Result

Utan kid should only randomly pick 1 item at a time and give 3 ea.

### How to Reproduce

-

### Official Information

In the official server of thRO and iRO. Utan kid only gives 1 item and the amount of items is 3.
References
https://irowiki.org/wiki/Umbala
https://www.youtube.com/watch?v=z0-Hom9-w-I
https://www.youtube.com/watch?v=naUVBRbb7Zw
https://www.youtube.com/watch?v=eLcuO3EDKlM

Playtester:
Reopening this due to two issues with the current fix:
1. This is not true for pre-renewal. In pre-renewal you get 2 of each item.
2. The chances in renewal are not correct. There is a 50% chance to get 3 Clover, a 25% chance to get 3 Jellopy and a 25% chance to get 3 Fluff.

PR #9185, opened 2025-03-19
Follow up to 6d5b26f

crazy-arashi:
* **Addressed Issue(s)**: #9174

* **Server Mode**: Both

Returned the pre-renewal script for umbala exchange quest that was changed on 6d5b26f and updated the renewal script to match the official script.


Playtester (approved):
Looks good. Also verified ingame that it's working as intended.

You might want to give the PR/commit a better description.

Issue #9181, opened 2025-03-19
Robe Costume Sprites are no longer displayed


krad309:
### rAthena Hash

ab78935

### Client Date

2020-07-15

### Server Mode

Pre-Renewal

### Result

Upon login, robe costumes worn by others are not visible. At the same time, your character's robe also won't be displayed to the others, however it will be visible to yourself. For it to be visible to others, you have to re-equip the item. The same must be done by other players to make their robes visible.

### Relevant Log Output

```Shell

```

### Expected Result

Robe costumes will be visible both ways.

### How to Reproduce

1. Do a dual login. We will have char A and char B
2. Both chars should wear a robe costume (20509 - renewal item), at this time both robe costumes will be visible to both.
3. Char A goes to char selection screen and logs back in.
4. Check char A through char B, this results to missing robe sprites of char A. Char A won't also won't be able see the robe costume worn by char B.
5. Go back to char A, unequip, and equip back the robe costume
6. Switch to char B and check char A, char A's costume is now visible to char B. However, char B's robe won't be visible to char A.
7. Unequip and equip the robe of char B. It will become visible to char A

Each time a character logs off or goes to char selection screen, robe costume is only visible to self.

### Official Information

N/A

PR #9193, opened 2025-03-24
Fixed robe display

Lemongrass3110:
* **Addressed Issue(s)**: #9181

* **Server Mode**: Both

Robes are not being copied correctly from the data saved by the char-server into the map_session_data.

Thanks to @krad309


Playtester (approved):
Don't know this feature and didn't test but the copying looks like how we copy other visual data, so I guess it's fine.

PR #9201, opened 2025-03-26
Fixed potential division by 0

Daegaladh:
* **Addressed Issue(s)**: Follow up to 1a85e1c925af94452adf6a4478aab914df0eadd7

* **Server Mode**: Both

* **Description of Pull Request**: Fixes potential division by 0 and matches official behavior

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -2951,7 +2951,7 @@ uint16 pc_getpercentweight(map_session_data& sd, uint32 weight)
 {
 	if (weight == 0)
 		weight = sd.weight;
-	return static_cast<uint16>(weight * 100 / std::max(sd.max_weight, static_cast<uint32>(1)));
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -2951,7 +2951,7 @@ uint16 pc_getpercentweight(map_session_data& sd, uint32 weight)
 {
 	if (weight == 0)
 		weight = sd.weight;
-	return static_cast<uint16>(weight * 100 / sd.max_weight);
+	return static_cast<uint16>(weight * 100 / std::max(sd.max_weight, static_cast<uint32>(1)));
```
Playtester, line 2954: You could put the variable type you want to base the max calculation on behind std::max I think.
`std::max<uint32>(sd.max_weight, 1);`
But didn't test.

Playtester (approved):
Did not test, but looks good now!

PR #9202, opened 2025-03-26
Custom Bonus - Skill Ratio

Playtester:
* **Addressed Issue(s)**: It's a feature!

* **Server Mode**: Both

- Added a custom bonus bSkillRatio that adds a fixed value directly to the skill ratio of attacks/skills
  * Normal attacks are affected (a bonus of 5 makes normal attacks have a skillratio of 105%)
  * Skills are only affected if they use skill ratio and don't deal fixed damage instead
  * Both physical and magical skills are affected
  * Bonus benefits weaker skills more than stronger skills (bonus=5: 200% -> 205%, 800% -> 805%)
  * Bonus is applied to each hit of skills that multiply damage by number of hits (bonus=5: 100%*10 -> 105%*10)
  * This bonus works the same way as "atkpercent" in renewal (e.g. Provoke) and stacks linearly with it


doc/item_bonus.txt
```diff
@@ -198,6 +198,7 @@ bonus2 bSkillUseSPrate,sk,n;		Decreases SP consumption of skill sk by n%
 Atk/Def
 -------
 bonus2 bSkillAtk,sk,n;    		Increases damage of skill sk by n%
+bonus bSkillRatio,n;			Adds n to the skillratio of all attacks/skills that use it
```
Atemo, line 201: I don't know what to add here, but I had to read the whole commit and the pull request description to get all the info. I'm missing some infos in this doc

Playtester: I guess I could put the whole PR info in there but it doesn't really fit the "explained in a single line" style of this file.

Maybe need to have a separate section for it. /hm

doc/item_bonus.txt
```diff
@@ -198,7 +198,14 @@ bonus2 bSkillUseSPrate,sk,n;		Decreases SP consumption of skill sk by n%
...
 -------
 bonus2 bSkillAtk,sk,n;    		Increases damage of skill sk by n%
-bonus bSkillRatio,n;			Adds n to the skillratio of all attacks/skills that use it
+bonus bSkillRatio,n;			Adds n to the skillratio of all attacks/skills
+					Detailed Explanation (all examples assume n=5):
+					- Normal attacks will have their skillratio increased (100% -> 105%)
+					- Only skills that have a skill ratio are affected, skills that deal fixed damage are not affected
+					- Both physical and magical skills are affected
+					- Bonus benefits skills with low skill ratio more (200% -> 205%, 800% -> 805%)
+					- Bonus is applied to each hit of skills that multiply damage by number of hits (100%*8 -> 105%*8)
```
Atemo, line 207: These 2 are more like example, they could be removed

Playtester: Isn't it necessary to understand it?

Atemo: I don't think so, but let's see what others think

Lemongrass3110: I think the explanation is too long.
All others only have half a line, without any examples.

doc/item_bonus.txt
```diff
@@ -198,6 +198,14 @@ bonus2 bSkillUseSPrate,sk,n;		Decreases SP consumption of skill sk by n%
 Atk/Def
 -------
 bonus2 bSkillAtk,sk,n;    		Increases damage of skill sk by n%
+bonus bSkillRatio,n;			Adds n to the skillratio of all attacks/skills
+					Detailed Explanation (all examples assume n=5):
+					- Normal attacks will have their skillratio increased (100% -> 105%)
+					- Only skills that have a skill ratio are affected, skills that deal fixed damage are not affected
+					- Both physical and magical skills are affected
+					- Bonus benefits skills with low skill ratio more (200% -> 205%, 800% -> 805%)
+					- Bonus is applied to each hit of skills that multiply damage by number of hits (100%*8 -> 105%*8)
```
Lemongrass3110, line 207:
```suggestion
bonus bSkillRatio,n;			Adds n to the skillratio of all attacks/skills that use skillratio
```

doc/item_bonus.txt
```diff
@@ -198,6 +198,14 @@ bonus2 bSkillUseSPrate,sk,n;		Decreases SP consumption of skill sk by n%
...
+					- This bonus works the same way as status changes like Provoke in renewal and stacks linearly with them
```
Lemongrass3110, line 208:
```suggestion
bonus bSkillRatio,n;			Adds n to the skillratio of all attacks/skills
```

src/map/battle.cpp
```diff
@@ -8160,6 +8162,9 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
 
 	if (!flag.infdef) { //No need to do the math for plants
 		uint32 skillratio = 100; //Skill dmg modifiers.
+		if (sd != nullptr)
+			skillratio += sd->bonus.skill_ratio;
```
Lemongrass3110, line 8166: Is it really okay to apply this here for renewal?
See has_skillratio below.

Playtester: Yes, that should be fine. Skills that don't make use of skillratio are simply not affected then. Or should we force them to be affected by it then?

doc/item_bonus.txt
Playtester: I also prefer single line explanation (see my original version), but Atemo requested more information to be added.

src/map/battle.cpp
Lemongrass3110: Ok perfect. Just didnt know if you saw that.

doc/item_bonus.txt
Lemongrass3110: I think we should keep it the way you originally had it.
If we want to add more information to the specific bonus, then we have to rethink/redesign the whole file.
We can discuss that, if we think this would be useful.

@Atemo is that okay for you?

Atemo: Yeah it's okay

Lemongrass3110 (approved):

PR #9207, opened 2025-03-27
Remove Exp Tap Bonus from No Damage Skills

### LINKED ISSUE #9206: Exp Tap should only happen when damage is applied
### rAthena Hash

11d373a

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When using skills on a monster that deal no damage, it will give 25% more exp.

### Expected Result

When using skills on a monster that deal no damage, it does not increase the monster's exp.

Note that this is a popular feature and should be configurable.

### How to Reproduce

Cast Lex Aeterna on a monster, then kill it with another player. Check the Exp received. You will see that it was not increased by 25%.

### Official Information

Verified by testing.

_No response_

Playtester:
* **Addressed Issue(s)**: #9206 

* **Server Mode**: Both

- By default skills that don't deal damage no longer count towards the exp tap bonus
- Added an option to re-enable this feature
- Fixes #9206

### DIFF
--- a/conf/battle/exp.conf
+++ b/conf/battle/exp.conf
@@ -45,6 +45,11 @@ exp_bonus_attacker: 25
 // (eg: if set at 5, the max bonus is 4*bonus-per-char regardless of attackers)
 exp_bonus_max_attacker: 12
 
+// Should casting skill that deal no damage still make you count as an attacker? (Note 1)
+// Setting this to yes makes skills like Lex Aeterna increase the exp a monster gives.
+// Officially you need to deal damage for it to count. 
+exp_bonus_nodamage_attacker: no
+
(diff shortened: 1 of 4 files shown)


conf/battle/exp.conf
```diff
@@ -45,6 +45,11 @@ exp_bonus_attacker: 25
 // (eg: if set at 5, the max bonus is 4*bonus-per-char regardless of attackers)
 exp_bonus_max_attacker: 12
 
+// Should casting skill that deal no damage still make you count as an attacker? (Note 1)
```
Lemongrass3110, line 48:
```suggestion
// Should casting skills that deal no damage still make you count as an attacker? (Note 1)
```

PR #9208, opened 2025-03-28
Harmonized minimum CMake version to 3.13

Lemongrass3110:

* **Server Mode**: Both

c4core requires 3.13 and since this is a core dependency it does not make sense to set others to different values.
Should fix CI compilation.


Playtester (approved):
Approved under the assumption that the automatic builds are passing.

PR #9211, opened 2025-03-29
Correct Determination of Endure Effect

### LINKED ISSUE #9068: Endure Effect on dmotion=0 units
### rAthena Hash

17f0447

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

Currently we set dmotion to 0 in various situation. Then say we show the "Endure" effect when it is 0.
Even monsters that naturally have dmotion of 0, will show the Endure effect.

### Expected Result

SC_ENDURE but also other situations create a timed "No Damage Motion Possible" effect, which causes dmotion to be 0, prevent stopping and also shows the Endure effect. Monsters that naturally have dmotion of 0 don't have this effect unless granted through other means.

SC_ENDURE (all effects that grant it) is disabled on GVG2/Battleground maps, but other effects granting "No Damage Motion Possible" are not. This is already working but just written here for reference.

Backsliding grants "No Damage Motion Possible" for 200ms.

Firewall hitting an undead creature (including Evil Druid Armor) grant "No Damage Motion Possible" for 2000ms.

NPC_BARRIER grants this effect for 15000ms.
NPC_KEEPING grants this effect for 30000ms.

SC_RUN and similar effects don't actually grant this effect, but still prevent stopping (should probably use "can_escape" similar to NPC_RANDOMMOVE and NPC_RUN).

### How to Reproduce

Use skill that grant endure and test damage against monsters with dmotion 0 (e.g. test dummies in tra_fild).

### Official Information

Atemo confirmed that the dummies in tra_fild don't use the Endure effect.

I confirmed through testing that our logic for the SC_ENDURE-granting effects was already correct.

Other information is available but untested.

_No response_

Playtester:
* **Addressed Issue(s)**: #9068 

* **Server Mode**: Both

- Damage motion (dmotion) is no longer set to 0 if the unit has the endure effect
  * This is because even on Endure the dmotion value sent to the client should remain unchanged
- Instead, added a function status_isendure that determines if a unit has the endure effect
  * The function can also return if the unit can be stopped
- Added a new variable endure_tick that contains the time until which a unit cannot be stopped
  * It is not used for everything yet, SC_ENDURE and no_walk_delay are still separate from it
  * When a fire or undead unit is hit by firewall, it now has the endure effect for 2000ms
  * When a unit uses backslide, it now has the endure effect for 200ms if it didn't have it before
  * NPC_BARRIER and NPC_KEEPING also grant endure for their duration (only affects display)
- Damage packet sent to client is now accurate in regards to damage type and dmotion
  * Bosses no longer have infinite endure (but still cannot be stopped in renewal)
  * SC_RUN and SC_WUGDASH also don't grant endure anymore, just prevent stopping
- Moved hardcoded Endure in GVG check to status_disabled.txt
  * Infinite endure from items will no longer be started on the corresponding maps
- Fixes #9068

### DIFF
--- a/db/status_disabled.txt
+++ b/db/status_disabled.txt
@@ -26,7 +26,7 @@
 // SC_ENDURE,4   // Endure status will be removed when the player enters GvG and WoE Castle maps; Also cannot be inflicted again.
 
 //----------------------------------------------------------------------------
-// Disabled/Removed on all WoE maps and battlegrounds
+// Disabled/Removed on all GVG and battleground maps and in WoE:TE Castles
 //----------------------------------------------------------------------------
 SC_ENDURE,28
 
--- a/db/status_disabled.txt
+++ b/db/status_disabled.txt
@@ -26,9 +26,13 @@
 // SC_ENDURE,4   // Endure status will be removed when the player enters GvG and WoE Castle maps; Also cannot be inflicted again.
 
 //----------------------------------------------------------------------------
+// Disabled/Removed on all WoE maps and battlegrounds
+//----------------------------------------------------------------------------
+SC_ENDURE,28
+
+//----------------------------------------------------------------------------
 // Disabled/Removed statuses in WoE:TE Castles (16)
 //----------------------------------------------------------------------------
-SC_ENDURE,16
 SC_BERSERK,16
 SC_ASSUMPTIO,16
 
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -8153,9 +8153,8 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 		case MG_FIREWALL:
 			if (tstatus->def_ele == ELE_FIRE || battle_check_undead(tstatus->race, tstatus->def_ele)) {
 				ad.blewcount = 0; //No knockback
-				unit_data* ud = unit_bl2ud(target);
 				// Fire and undead units hit by firewall cannot be stopped for 2 seconds
-				if (ud != nullptr)
+				if (unit_data* ud = unit_bl2ud(target); ud != nullptr)
 					ud->endure_tick = gettick() + 2000;
 				break;
 			}
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -8151,12 +8151,19 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list 
 
 	switch(skill_id) {
 		case MG_FIREWALL:
-			if ( tstatus->def_ele == ELE_FIRE || battle_check_undead(tstatus->race, tstatus->def_ele) )
+			if (tstatus->def_ele == ELE_FIRE || battle_check_undead(tstatus->race, tstatus->def_ele)) {
 				ad.blewcount = 0; //No knockback
+				unit_data* ud = unit_bl2ud(target);
+				// Fire and undead units hit by firewall cannot be stopped for 2 seconds
+				if (ud != nullptr)
+					ud->endure_tick = gettick() + 2000;
+				break;
+			}
 			[[fallthrough]];
 		case NJ_KAENSIN:
 		case PR_SANCTUARY:
-			ad.dmotion = 1; //No flinch animation.
+			// TODO: This code is a temporary workaround because right now we stop monsters for their dmotion which is not official
+			ad.dmotion = 1;
 			break;
 	}
 
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -5119,12 +5119,12 @@ void clif_getareachar_unit( map_session_data* sd,struct block_list *bl ){
 
 //Modifies the type of damage according to target status changes [Skotlex]
 //Aegis data specifies that: 4 endure against single hit sources, 9 against multi-hit.
-static enum e_damage_type clif_calc_delay(block_list& bl, e_damage_type type, int32 div, int64 damage, int32 delay) {
+static enum e_damage_type clif_calc_delay(block_list& bl, e_damage_type type, int32 div, int64 damage, int32 delay, t_tick tick) {
 	if (damage < 1)
 		return type;
 
 	// Check if unit has endure
-	if (!status_isendure(bl, gettick()))
+	if (!status_isendure(bl, tick, true))
 		return type;
 
 	// General change of type based on div against target with endure effect
@@ -5156,26 +5156,12 @@ static enum e_damage_type clif_calc_delay(block_list& bl, e_damage_type type, in
 /*==========================================
  * Estimates walk delay based on the damage criteria. [Skotlex]
  *------------------------------------------*/
-static int32 clif_calc_walkdelay( block_list &bl, int32 delay, e_damage_type type, int64 damage, int32 div_ ) {
+static int32 clif_calc_walkdelay(block_list &bl, int32 delay, e_damage_type type, int64 damage, int32 div_, t_tick tick) {
 	if (damage < 1)
 		return 0;
 
-	switch( type ) {
-		case DMG_ENDURE:
-		case DMG_MULTI_HIT_ENDURE:
-		case DMG_SPLASH_ENDURE:
-			return 0;
-	}
-
-	// TODO: This logic is only temporary until we refactor this function, we should check this when the unit is stopped instead!
-	// Below are situations that don't send "Endure" to the client, but still prevent being stopped
-	// This includes bosses in renewal (episode 20+) and running / dashing units
-#ifdef RENEWAL
-	if (bl.type == BL_MOB && status_get_class_(&bl) == CLASS_BOSS)
-		return 0;
-#endif
-	status_change* sc = status_get_sc(&bl);
-	if (sc != nullptr && !sc->empty() && (sc->getSCE(SC_RUN) || sc->getSCE(SC_WUGDASH)))
+	// Check if unit can be stopped by damage
+	if (status_isendure(bl, tick, false))
 		return 0;
 
 	if (bl.type == BL_PC) {
@@ -5253,8 +5239,7 @@ int32 clif_damage(block_list& src, block_list& dst, t_tick tick, int32 sdelay, i
 	int32 damage = (int32)cap_value(sdamage,INT_MIN,INT_MAX);
 	int32 damage2 = (int32)cap_value(sdamage2,INT_MIN,INT_MAX);
 
-	if (type != DMG_MULTI_HIT_CRITICAL)
-		type = clif_calc_delay( dst, type, div, damage+damage2, ddelay );
+	type = clif_calc_delay(dst, type, div, damage+damage2, ddelay, tick);
 
 	damage = static_cast<decltype(damage)>(clif_hallucination_damage(dst, damage));
 	damage2 = static_cast<decltype(damage2)>(clif_hallucination_damage(dst, damage2));
@@ -5325,13 +5310,8 @@ int32 clif_damage(block_list& src, block_list& dst, t_tick tick, int32 sdelay, i
 	if(&src == &dst) 
 		unit_setdir(&src, unit_getdir(&src));
 
-	// In case this assignment is bypassed by DMG_MULTI_HIT_CRITICAL
-	if( type == DMG_MULTI_HIT_CRITICAL ){
-		type = clif_calc_delay( dst, type, div, damage+damage2, ddelay );
-	}
-
 	//Return adjusted can't walk delay for further processing.
-	return clif_calc_walkdelay(dst, ddelay, type, damage+damage2, div);
+	return clif_calc_walkdelay(dst, ddelay, type, damage+damage2, div, tick);
 }
 
 /*==========================================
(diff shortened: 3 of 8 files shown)


src/map/clif.cpp
```diff
@@ -5169,6 +5167,17 @@ static int32 clif_calc_walkdelay( block_list &bl, int32 delay, e_damage_type typ
...
+	// Below are situations that don't send "Endure" to the client, but still prevent being stopped
+	// This includes bosses in renewal (episode 20+) and running / dashing units
+#ifdef RENEWAL
+	if (bl.type == BL_MOB && status_get_class_(&bl) == CLASS_BOSS)
+		return 0;
+#endif
+	status_change* sc = status_get_sc(&bl);
+	if (sc != nullptr && !sc->empty() && (sc->getSCE(SC_RUN) || sc->getSCE(SC_WUGDASH)))
+		return 0;
+
```
Lemongrass3110, line 5180: Why is this not part of status_isendure?
And yes I have read the TODO note above.

For me that sounds like status_isendure needs a parameter "visible" true/false.

src/map/skill.cpp
```diff
@@ -10079,6 +10079,11 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
 
 	case TF_BACKSLIDING: //This is the correct implementation as per packet logging information. [Skotlex]
 		{
+			// Backsliding makes you immune to being stopped for 200ms, but only if you weren't immune already
+			unit_data* ud = unit_bl2ud(bl);
+			if (ud != nullptr && DIFF_TICK(ud->endure_tick, tick) < 0)
+				ud->endure_tick = tick + 200;
```
Lemongrass3110, line 10085: Not really a friend of that DIFF_TICK syntax, would prefer simply using std::max or i64max

src/map/status.cpp
```diff
@@ -9266,6 +9232,36 @@ int32 status_isimmune(struct block_list *bl)
...
+ * @param bl: Object to check [PC|MOB|HOM|MER|ELEM]
+ * @param tick: Current tick
+ * @return Whether object can be stopped (false) or not (true)
+ */
+bool status_isendure(block_list& bl, t_tick tick)
+{
+	// Officially the bonus "no_walk_delay" is actually just SC_ENDURE with unlimited duration
+	// Endure is forbidden on some maps, but we don't apply this on this bonus
+	// That's why we need to check it here
+	if (bl.type == BL_PC && !status_change_isDisabledOnMap(SC_ENDURE, map_getmapdata(bl.m))) {
```
Lemongrass3110, line 9247:
```suggestion
	if (bl.m >= 0 && bl.type == BL_PC && !status_change_isDisabledOnMap(SC_ENDURE, map_getmapdata(bl.m))) {
```

src/map/status.cpp
```diff
@@ -13331,7 +13327,7 @@ int32 status_change_end( struct block_list* bl, enum sc_type type, int32 tid ){
 			unit_data *ud = unit_bl2ud(bl);
 
 			if (ud)
```
Lemongrass3110, line 13329:
```suggestion
			if (ud != nullptr) {
```

src/map/status.cpp
```diff
@@ -13331,7 +13327,7 @@ int32 status_change_end( struct block_list* bl, enum sc_type type, int32 tid ){
...
-				ud->attackabletime = ud->canact_tick = ud->canmove_tick = gettick();
+				ud->attackabletime = ud->canact_tick = ud->canmove_tick = ud->endure_tick = gettick();
```
Lemongrass3110, line 13330:
```suggestion
				ud->attackabletime = ud->canact_tick = ud->canmove_tick = gettick();
				ud->endure_tick = gettick();
			}
```

src/map/unit.cpp
```diff
@@ -3421,7 +3421,8 @@ void unit_dataset(struct block_list *bl)
 	ud->steptimer      = INVALID_TIMER;
 	ud->attackabletime =
 	ud->canact_tick    =
-	ud->canmove_tick   = gettick();
+	ud->canmove_tick   =
```
Lemongrass3110, line 3424:
```suggestion
	ud->canmove_tick   = gettick();
```

src/map/status.hpp
```diff
@@ -3619,6 +3619,7 @@ status_change *status_get_sc(struct block_list *bl);
 
 bool status_isdead(block_list &bl);
 int32 status_isimmune(struct block_list *bl);
+bool status_isendure(block_list &bl, t_tick tick);
```
Lemongrass3110, line 3622:
```suggestion
bool status_isendure(block_list& bl, t_tick tick);
```

src/map/status.cpp
```diff
@@ -10659,7 +10655,7 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
 			unit_data *ud = unit_bl2ud(bl);
 
 			if (ud)
-				ud->attackabletime = ud->canact_tick = ud->canmove_tick = gettick() + tick;
+				ud->attackabletime = ud->canact_tick = ud->canmove_tick = ud->endure_tick = gettick() + tick;
```
Lemongrass3110, line 10658:
```suggestion
				ud->attackabletime = ud->canact_tick = ud->canmove_tick = gettick() + tick;
				ud->endure_tick = gettick() + tick;
```

src/map/battle.cpp
```diff
@@ -8151,12 +8151,19 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
...
+				// Fire and undead units hit by firewall cannot be stopped for 2 seconds
+				if (ud != nullptr)
+					ud->endure_tick = gettick() + 2000;
+				break;
+			}
 			[[fallthrough]];
 		case NJ_KAENSIN:
 		case PR_SANCTUARY:
-			ad.dmotion = 1; //No flinch animation.
+			// TODO: This code is a temporary workaround because right now we stop monsters for their dmotion which is not official
```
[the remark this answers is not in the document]
Playtester: It's a general workaround, I guess. But it's mainly to get monsters "use up" the firewall faster. This will be removed once I'm done with the walk delay rework.

src/map/skill.cpp
Playtester: I think we should consistently use DIFF_TICK mainly because then we can easily do adjustments that should affect all t_tick comparisons.

src/map/status.cpp
Playtester: Why?

db/status_disabled.txt
```diff
@@ -25,10 +25,14 @@
 // Example:
 // SC_ENDURE,4   // Endure status will be removed when the player enters GvG and WoE Castle maps; Also cannot be inflicted again.
 
+//----------------------------------------------------------------------------
+// Disabled/Removed on all WoE maps and battlegrounds
```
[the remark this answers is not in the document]
Playtester: We used the term "WoE:TE Castles" below so I wanted to be consistent. GVG might mislead into thinking WoE TE maps are not included.

src/map/unit.cpp
Playtester: If you go this far I would also remove the extra spaces.

src/map/status.hpp
Playtester: Saw that the other status_is function use it like that and thought it looks better. You prefer it next to the type rather than the object name?

db/status_disabled.txt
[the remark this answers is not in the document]
Playtester: No they don't, the "16" within the 28 stands for WoE:TE castles.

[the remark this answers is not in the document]
Daegaladh: Ah right, makes sense, then maybe specify (GVG + WoE:TE) ?

src/map/clif.cpp
Playtester: Updated, please check.

src/map/status.cpp
```diff
@@ -9266,6 +9232,49 @@ int32 status_isimmune(struct block_list *bl)
...
+ * @param tick: Current tick
+ * @param visible: If true, function returns if the endure effect should be shown on the client
+ * @return Whether object can be stopped (false) or not (true)
+ */
+bool status_isendure(block_list& bl, t_tick tick, bool visible)
+{
+	// Officially the bonus "no_walk_delay" is actually just SC_ENDURE with unlimited duration
+	// Endure is forbidden on some maps, but we don't apply this on this bonus
+	// That's why we need to check it here
+	if (bl.m >= 0 && bl.type == BL_PC && !status_change_isDisabledOnMap(SC_ENDURE, map_getmapdata(bl.m))) {
```
Lemongrass3110, line 9248: Is it okay to ignore status_change_isDisabledOnMap for other BL types?
In status_change_start we check it independent of BL type.

src/map/status.cpp
```diff
@@ -9266,6 +9232,49 @@ int32 status_isimmune(struct block_list *bl)
...
+		if (reinterpret_cast<map_session_data&>(bl).special_state.no_walk_delay)
+			return true;
+	}
+
+	unit_data* ud = unit_bl2ud(&bl);
+	if (ud != nullptr && DIFF_TICK(ud->endure_tick, tick) > 0)
+		return true;
+
+	status_change* sc = status_get_sc(&bl);
+	if (sc != nullptr && !sc->empty()) {
```
Lemongrass3110, line 9258:
```suggestion
	if (status_change* sc = status_get_sc(&bl); sc != nullptr && !sc->empty()) {
```

src/map/status.cpp
```diff
@@ -9266,6 +9232,49 @@ int32 status_isimmune(struct block_list *bl)
...
+	// Officially the bonus "no_walk_delay" is actually just SC_ENDURE with unlimited duration
+	// Endure is forbidden on some maps, but we don't apply this on this bonus
+	// That's why we need to check it here
+	if (bl.m >= 0 && bl.type == BL_PC && !status_change_isDisabledOnMap(SC_ENDURE, map_getmapdata(bl.m))) {
+		if (reinterpret_cast<map_session_data&>(bl).special_state.no_walk_delay)
+			return true;
+	}
+
+	unit_data* ud = unit_bl2ud(&bl);
+	if (ud != nullptr && DIFF_TICK(ud->endure_tick, tick) > 0)
```
Lemongrass3110, line 9254:
```suggestion
	if (unit_data* ud = unit_bl2ud(&bl); ud != nullptr && DIFF_TICK(ud->endure_tick, tick) > 0)
```

src/map/status.cpp
```diff
@@ -9266,6 +9232,49 @@ int32 status_isimmune(struct block_list *bl)
...
+		if (sc->getSCE(SC_ENDURE))
+			return true;
+
+		// These status changes don't send "Endure" to the client, but still prevent being stopped
+		if (!visible && (sc->getSCE(SC_RUN) || sc->getSCE(SC_WUGDASH)))
+			return true;
+	}
+
+	// Bosses in renewal (episode 20+) cannot be stopped by damage, but still show damage normally
+#ifdef RENEWAL
```
Lemongrass3110, line 9270:
```suggestion
#ifdef RENEWAL
	// Bosses in renewal (episode 20+) cannot be stopped by damage, but still show damage normally
```

src/map/status.cpp
Lemongrass3110: Track changes / git blame.
If you want something like this, please define a variable above and asign it to all.

src/map/status.cpp
Lemongrass3110: Here you touched the line.

src/map/unit.cpp
Lemongrass3110: Track changes / git blame.
If you want something like this, please define a variable above and asign it to all.

Feel free to remove the useless extra spacing.

src/map/status.cpp
Lemongrass3110: Track changes / git blame.
If you want something like this, please define a variable above and asign it to all.

src/map/skill.cpp
Lemongrass3110: I think i dont really understand your explanation.
But if i understand you correctly, then we should introduce a utility function that does this.

src/map/status.cpp
Playtester: special_state.no_walk_delay only applies to BL_PC

The restriction itself applies to all BL types of course, but it's not relevant here.

src/map/status.cpp
Playtester: I didn't. It's white.

Lemongrass3110: Ah yeah I suggested it here, because I had to suggest a { for the changes below.

Atemo (approved):
It looks good to me

src/map/status.cpp
```diff
@@ -9266,6 +9232,47 @@ int32 status_isimmune(struct block_list *bl)
...
+			return true;
+	}
+
+	if (unit_data* ud = unit_bl2ud(&bl); ud != nullptr && DIFF_TICK(ud->endure_tick, tick) > 0)
+		return true;
+
+	if (status_change* sc = status_get_sc(&bl); sc != nullptr && !sc->empty()) {
+		// Officially endure also sets endure_tick
+		// However, we have a lot of extra logic for infinite endure, so we use the status change for now
+		if (sc->getSCE(SC_ENDURE))
```
Lemongrass3110, line 9259:
```suggestion
		if (sc->getSCE(SC_ENDURE) != nullptr)
```

src/map/status.cpp
```diff
@@ -9266,6 +9232,47 @@ int32 status_isimmune(struct block_list *bl)
...
+			return true;
+
+		// These status changes don't send "Endure" to the client, but still prevent being stopped
+		if (!visible && (sc->getSCE(SC_RUN) || sc->getSCE(SC_WUGDASH)))
```
Lemongrass3110, line 9263:
```suggestion
		if (!visible && (sc->getSCE(SC_RUN) != nullptr || sc->getSCE(SC_WUGDASH) != nullptr))
```

src/map/status.cpp
```diff
@@ -9266,6 +9232,47 @@ int32 status_isimmune(struct block_list *bl)
...
+ * @param bl: Object to check [PC|MOB|HOM|MER|ELEM]
+ * @param tick: Current tick
+ * @param visible: If true, function returns if the endure effect should be shown on the client
+ * @return Whether object can be stopped (false) or not (true)
+ */
+bool status_isendure(block_list& bl, t_tick tick, bool visible)
+{
+	// Officially the bonus "no_walk_delay" is actually just SC_ENDURE with unlimited duration
+	// Endure is forbidden on some maps, but we don't apply this on this bonus
+	// That's why we need to check it here
```
Lemongrass3110, line 9247:
```suggestion
	// Officially the bonus "no_walk_delay" is actually just SC_ENDURE with unlimited duration.
	// Endure is forbidden on some maps, but we still allow the bonus on such maps.
	// That's why we need to check it here and ignore the bonus on such maps to correctly mimic official behavior.
```

Lemongrass3110 (approved):

Issue #9212, opened 2025-03-30
[Warning]: clif_parse: Received unsupported packet (packet 0x0bf3, 36 bytes received), disconnecting session #3.


brynner23:
### rAthena Hash

0811bfef323d1dd03bbf77371aa0b02fb5d9d2c9

### Client Date

2024-09-19

### Server Mode

Renewal

### Result

Adventure Guide you will disconnected from the server
[Warning]: clif_parse: Received unsupported packet (packet 0x0bf3, 36 bytes received), disconnecting session #3.

![Image](https://github.com/user-attachments/assets/9b0a9abc-178d-490b-bc86-d1a27c7f1f57)
![Image](https://github.com/user-attachments/assets/0789a6d3-3050-4f9a-8816-28e3ff8c081a)

### Relevant Log Output

```Shell

```

### Expected Result

[Warning]: clif_parse: Received unsupported packet (packet 0x0bf3, 36 bytes received), disconnecting session #3.

### How to Reproduce

click the adventure guide

### Official Information

.

monbazko:
#8730 
Unsupported client. so most likely it will be treated as custom feature.
Make a dummy packet ignore for now as commented on the linked issue

brynner23:
oh, I see. it's being reported. wel,l this will be a duplicate report. I'm going to close this now

PR #9215, opened 2025-03-31
Added dummy parsing for 0xbf3

Lemongrass3110:
* **Addressed Issue(s)**: #9212

* **Server Mode**: Both

Added dummy parsing for 0xbf3

Thanks to @brynner23

### DIFF
--- a/src/map/clif_packetdb.hpp
+++ b/src/map/clif_packetdb.hpp
@@ -2041,6 +2041,10 @@
 	parseable_packet( HEADER_CZ_REQ_EMOTION_EXPANSION, sizeof( struct PACKET_CZ_REQ_EMOTION_EXPANSION ), clif_parse_dull, 0 );
 #endif
 
+#if PACKETVER_MAIN_NUM >= 20230802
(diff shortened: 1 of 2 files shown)


src/map/clif_packetdb.hpp
```diff
@@ -2041,6 +2041,10 @@
 	parseable_packet( HEADER_CZ_REQ_EMOTION_EXPANSION, sizeof( struct PACKET_CZ_REQ_EMOTION_EXPANSION ), clif_parse_dull, 0 );
 #endif
 
+#if PACKETVER_MAIN_NUM >= 20230802
+	parseable_packet( HEADER_CZ_QUEST_STATUS_REQ, -1, clif_parse_dull, 0 );
```
Playtester, line 2045: So length of the packet is dynamic?

Lemongrass3110: Yes, the size in the client is defined with minimum 4 and variable maximum.

Playtester (approved):
Looks similar to the others I have already approved before.

Issue #9216, opened 2025-03-31
[Error]: map_destroyblock: unknown type 0


brynner23:
### rAthena Hash

d4993ebd2d6976bc6200cbb27d719f1b367c3bd0

### Client Date

2024-09-19

### Server Mode

Renewal

### Result

i just got this error on my map server
`[Error]: map_destroyblock: unknown type 0
[Error]: map_destroyblock: unknown type 0
[Error]: map_destroyblock: unknown type 0
[Error]: map_destroyblock: unknown type 0`

I don't know how to produce this error
it randomly appears on the map server

![Image](https://github.com/user-attachments/assets/e96ba55d-e5ea-44b0-9fad-b76889f5ccef)
![Image](https://github.com/user-attachments/assets/b4b8a043-50f4-4c41-a3c4-618a818d4b07)
![Image](https://github.com/user-attachments/assets/0b6eb131-04e0-48c2-bbbc-7ef3bd8e74c0)

### Relevant Log Output

```Shell

```

### Expected Result

i just got this error on my map server
`[Error]: map_destroyblock: unknown type 0
[Error]: map_destroyblock: unknown type 0
[Error]: map_destroyblock: unknown type 0
[Error]: map_destroyblock: unknown type 0`

### How to Reproduce

I don't know how to produce this error
it randomly appears on the map server

### Official Information

.

```
### Modifications that may affect results

/// Uncomment to enable VIP system.
#define VIP_ENABLE
```

```
#ifndef PACKETVER
	/// Do NOT edit this line! To set your client version, please do this instead:
	/// In Windows: Add this line in your src\custom\defines_pre.hpp file: #define PACKETVER YYYYMMDD
	/// In Linux: The same as above or run the following command: ./configure --enable-packetver=YYYYMMDD
	#define PACKETVER 20240919
#endif
```

```
//#define RENEWAL_CAST
//#define RENEWAL_DROP
//#define RENEWAL_EXP
//#define RENEWAL_LVDMG
//#define RENEWAL_ASPD
//#define RENEWAL_STAT
```

skstrife:
This is an issue that started when 75c8875 was merged.

Playtester:
@Lemongrass3110 is investigating this issue.

PR #9217, opened 2025-04-01
Fixed unknown type in map_destroyblock

Lemongrass3110:
* **Addressed Issue(s)**: #9216

* **Server Mode**: Both

Fixed double call to elemental destructor
Removed old comment regarding map_session_data, that was copied to map_destroyblock
Added BL_NUL case

Thanks to @brynner23

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -234,6 +234,10 @@ void map_destroyblock( block_list* bl ){
 	}
 
 	switch( bl->type ){
+		case BL_NUL:
+			// Dummy type, has no destructor
+			break;
+
 		case BL_PC:
 			// Do not call the destructor here, it will be done in chrif_auth_delete
 			//reinterpret_cast<map_session_data*>( bl )->~map_session_data();
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -3979,9 +3979,6 @@ int32 unit_free(struct block_list *bl, clr_type clrtype)
 
 			skill_clear_unitgroup(bl);
 			status_change_clear(bl,1);
-
-			// Do not call the destructor here, it will be done in chrif_auth_delete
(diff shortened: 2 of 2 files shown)


src/map/unit.cpp
```diff
@@ -4151,8 +4148,6 @@ int32 unit_free(struct block_list *bl, clr_type clrtype)
 
 			skill_clear_unitgroup(bl);
 			status_change_clear(bl,1);
-
-			ed->~s_elemental_data();
```
Playtester, line 4155: So the previous commit was wrong? Maybe should link to it: https://github.com/rathena/rathena/commit/273d1034251ce8a8d25f3ce0d1a24f50b1b8e955

Lemongrass3110: It was not wrong.
All the destructor calls should have been moved in c198bb3
But I missed this one.
The call to the destructor is still there and now happens in map_destroyblock

Playtester (approved):
Not quite sure if this solves the issue that a BL_NUL type is created in the first place. But it will certainly resolve the issue with the error message.

Destructor call removal seems good when looking at the previous PRs.

src/map/map.cpp
```diff
@@ -234,6 +234,10 @@ void map_destroyblock( block_list* bl ){
 	}
 
 	switch( bl->type ){
+		case BL_NUL:
+			// Dummy type, has no destructor
```
Playtester, line 238: After seeing the crash reports, I think the whole cause of this is that the destructor sets the BL type to BL_NUL, so if you call the destructor a second time you get the warning.

I'm wondering if we should really suppress this warning then as it could give us necessary information that a destructor was called somewhere were it's not supposed to.

Maybe we could have a message in debug mode here that tells us to check for bl destructor calls?

PR #9219, opened 2025-04-02
Allow to define conditional SCS only

Lemongrass3110:

* **Server Mode**: Both

Currently the conditional status change state (SCS) check is only triggered, when the unconditional SCS is also defined.
This can lead to mistakes and is not user friendly.

### DIFF
--- a/db/re/status.yml
+++ b/db/re/status.yml
@@ -1173,7 +1173,6 @@ Body:
       bonus2 bSubEle,Ele_Wind, getstatus(SC_ARMOR_ELEMENT_WATER, 4);
   - Status: Nochat
     States:
-      NoPickItem: true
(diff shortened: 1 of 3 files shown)


Playtester (approved):
Did not test ingame, but reviewed all changes and seems good.

db/re/status.yml
```diff
@@ -3991,9 +3981,7 @@ Body:
     Icon: EFST_COLD
     DurationLookup: SO_DIAMONDDUST
     States:
-      NoMove: true
       NoMoveCond: true
-      NoCast: true
       NoCastCond: true
```
Playtester, line 3985: Crystalize is missing code for "NoCastCond".

PR #9227, opened 2025-04-03
Fixed SC_CRYSTALIZE

Lemongrass3110:
* **Addressed Issue(s)**:
```
[Error]: status_calc_state_sub: status "SC_CRYSTALIZE" is defined with "SCS_NOCASTCOND", but the condition is not implemented.
```

* **Server Mode**: Both

Crystalize had the NoCastCond flag in status.yml, but no source side condition was implemented. Actually this was wrong and it is unconditional.
Also it has been confirmed that the movement restriction is also applying for monsters, so also the movement restriction is unconditional.

Additionally synced pre-renewal and renewal flags in status.yml

Thanks to @Playtester

### DIFF
--- a/db/pre-re/status.yml
+++ b/db/pre-re/status.yml
@@ -3827,8 +3827,8 @@ Body:
     Icon: EFST_COLD
     DurationLookup: SO_DIAMONDDUST
     States:
-      NoMoveCond: true
-      NoCastCond: true
+      NoMove: true
+      NoCast: true
       NoConsumeItem: true
       NoAttack: true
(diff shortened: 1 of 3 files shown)


db/pre-re/status.yml
```diff
@@ -3839,11 +3839,15 @@ Body:
       RemoveOnLuxAnima: true
       BossResist: true
       StopWalking: true
+      StopWalking: true
```
rabbit-lunatic, line 3842:
```suggestion
```
It's already above.

PR #9228, opened 2025-04-03
Improved solution to sync position after endure hit

Playtester:

* **Server Mode**: Both

- Replaced the monster-specific solution to fix position lag after an endure hit with a common one
  * It remembers the time of the last damage and updates the position every cell for a certain time span
  * This is needed because sometimes an Endure hit still stops a unit on the client (e.g. firewall on undead)
  * Works for all BL types now and can even sync position on the current client of a player
  * Reduced the duration from 3000ms to 1000ms (max move time per cell)
  * Code optimizations such as using a CONSTANT for the duration


Lemongrass3110 (approved):
Looks good. But the MAX_WALKSPEED still looks strange

PR #9233, opened 2025-04-08
Moved Walkdelay Calculation to Correct Place

Playtester:

* **Server Mode**: Both

- Replaced clif_calc_walkdelay with battle_calc_walkdelay
- Removed the individual calculations of walkdelay through clif_damage and clif_skill_damage
- When calling battle_damage, battle_delay_damage and battle_fix_damage, we now send the number of hits instead of walkdelay
- Walkdelay is now calculated at one central place and at the time the damage is applied
  * This matches the place where it is calculated officially

Note: Certain situations that had an individual walkdelay calculations work differently now. They didn't work correctly in the first place as there is no such thing as a walkdelay officially, just a delay until when a unit reacts to damage. But we need to find out one-by-one why they were coded the way they were and then maybe integrate them into the normal system. I'd prefer if we could finish this PR first and then fix the different skills and status changes one-by-one.

Support in testing this PR would be appreciated.

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -281,11 +281,10 @@ static t_tick battle_calc_walkdelay(block_list& bl, int64 damage, int16 div_, t_
 		delay += battle_config.multihit_delay * (div_ - 1);
 	}
 
-	if (bl.type == BL_PC) {
-		delay = apply_rate( delay, battle_config.pc_walk_delay_rate );
-	}
+	if (bl.type == BL_PC)
+		delay = apply_rate(delay, battle_config.pc_walk_delay_rate);
 	else
-		delay = apply_rate( delay, battle_config.walk_delay_rate );
+		delay = apply_rate(delay, battle_config.walk_delay_rate);
 
 	return (delay > 0) ? delay : 1; //Return 1 to specify there should be no noticeable delay, but you should stop walking.
 }
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -7273,7 +7273,7 @@ static void battle_calc_attack_gvg_bg(struct Damage* wd, struct block_list *src,
 				int64 damage = wd->damage + wd->damage2, rdamage = 0;
 				map_session_data *tsd = BL_CAST(BL_PC, target);
 				status_data* sstatus = status_get_status_data(*src);
-				t_tick tick = gettick(), rdelay = 0;
+				t_tick tick = gettick();
 
 				rdamage = battle_calc_return_damage(target, src, &damage, wd->flag, skill_id, false);
 				if( rdamage > 0 ) { //Item reflect gets calculated before any mapflag reducing is applicated
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -256,6 +256,42 @@ struct block_list* battle_getenemyarea(struct block_list *src, int32 x, int32 y,
 	return bl_list[rnd()%c];
 }
 
+/*==========================================
+ * Returns the walkdelay a unit receives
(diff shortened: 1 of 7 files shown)


src/map/battle.cpp
```diff
@@ -7617,11 +7659,11 @@ void battle_do_reflect(int32 attack_type, struct Damage *wd, struct block_list*
 			if( attack_type == BF_WEAPON && tsc->getSCE(SC_REFLECTDAMAGE) ) // Don't reflect your own damage (Grand Cross)
 				map_foreachinshootrange(battle_damage_area,target,skill_get_splash(LG_REFLECTDAMAGE,1),BL_CHAR,tick,target,wd->amotion,sstatus->dmotion,rdamage,wd->flag);
 			else if( attack_type == BF_WEAPON || attack_type == BF_MISC) {
-				rdelay = clif_damage(*src, (d_bl == nullptr) ? *src : *d_bl, tick, wd->amotion, sstatus->dmotion, rdamage, 1, DMG_ENDURE, 0, false);
+				clif_damage(*src, (d_bl == nullptr) ? *src : *d_bl, tick, wd->amotion, sstatus->dmotion, rdamage, 1, DMG_ENDURE, 0, false);
 				if( tsd )
 					battle_drain(tsd, src, rdamage, rdamage, sstatus->race, sstatus->class_);
 				// It appears that official servers give skill reflect damage a longer delay
```
Lemongrass3110, line 7665:
```suggestion
```

src/map/battle.cpp
```diff
@@ -10626,7 +10668,7 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
 		}
 	}
 
-	wd.dmotion = clif_damage(*src, *target, tick, wd.amotion, wd.dmotion, wd.damage, wd.div_ , wd.type, wd.damage2, wd.isspdamage);
+	clif_damage(*src, *target, tick, wd.amotion, wd.dmotion, wd.damage, wd.div_ , wd.type, wd.damage2, wd.isspdamage);
```
Lemongrass3110, line 10671:
```suggestion
	clif_damage(*src, *target, tick, wd.amotion, wd.dmotion, wd.damage, wd.div_, wd.type, wd.damage2, wd.isspdamage);
```

src/map/battle.cpp
```diff
@@ -7312,7 +7353,8 @@ static void battle_calc_weapon_final_atk_modifiers(struct Damage* wd, struct blo
 	{
 		ATK_RATER(wd->damage, 50)
 		clif_skill_nodamage(target, *target,ST_REJECTSWORD, tsc->getSCE(SC_REJECTSWORD)->val1);
-		battle_fix_damage(target,src,wd->damage,clif_damage(*target,*src,gettick(),0,0,wd->damage,0,DMG_NORMAL,0,false),ST_REJECTSWORD);
+		clif_damage(*target, *src, gettick(), 0, 0, wd->damage, 0, DMG_NORMAL, 0, false);
+		battle_fix_damage(target,src,wd->damage,wd->div_,ST_REJECTSWORD);
```
Lemongrass3110, line 7357:
```suggestion
		battle_fix_damage(target, src, wd->damage, wd->div_, ST_REJECTSWORD);
```

Lemongrass3110 (approved):
I did not test it, but looks okay.

src/map/battle.cpp
Playtester: I think this comment doesn't refer to the "rdelay" variable, but rather the delay until which the damage is applied, but I guess might as well remove it since nobody knows what this means anyway.

PR #9236, opened 2025-04-11
Infinite Endure Config

### LINKED ISSUE #9235: Permanent Endure for Players & Monsters Doesnt work
### rAthena Hash

420ebb2

### Client Date

2024/10/30

### Server Mode

Renewal

### Result

Since the last update, using Permanent Endure effect in Conf settings for monsters and players is no longer working, players and monsters are both staggering and getting stuck from skills and attacks.

### Relevant Log Output

```Shell
None
```

### Expected Result

Setting Permanent Endure in conf settings for players and monsters should prevent this from happening

### How to Reproduce

latest rA, set player and monster endure permanently on and then get hit and you'll see.

### Official Information

Eddga card on official makes walking smooth when you take damage, should be the same thing here, monsters should also be able to get this effect.

None

Playtester:
* **Addressed Issue(s)**: #9235 

* **Server Mode**: Both

- Added config that allows to make certain bl types always have the endure effect
- Updated comment on damage_delay_rate configs that no longer grant endure since 420ebb2
- Fixes #9235

### DIFF
--- a/conf/battle/battle.conf
+++ b/conf/battle/battle.conf
@@ -46,8 +46,13 @@ damage_walk_delay_rate: 100
 multihit_delay: 80
 
 // Damaged delay rate for players (Note 2)
-// (Setting to no/0  will be like always endure)
+// This affects the damage delay that is sent to the client and is the base value for the walk delay.
 player_damage_delay_rate: 100
+
+// Always have endure? (Note 3)
+// The unit types defined here cannot be stopped by damage.
+// Please note that this makes it impossible to display critical hits on these unit types.
+infinite_endure: 0
 
 // Should race or element be used to consider someone undead?
 // 0 = element undead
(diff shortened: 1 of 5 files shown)


conf/battle/battle.conf
```diff
@@ -46,9 +46,14 @@ damage_walk_delay_rate: 100
 multihit_delay: 80
 
 // Damaged delay rate for players (Note 2)
-// (Setting to no/0  will be like always endure)
+// This affects the damage delay that is sent to the client and is the base value for the walk delay.
 player_damage_delay_rate: 100
 
+// Always have endure? (Note 3)
+// The unit types defined here cannot be stopped by damage.
+// Please note that this makes it impossible to display critical hits on these unit types.
```
Daegaladh, line 54: Maybe briefly explain why it's impossible so people don't start asking about it

Daegaladh (approved):

conf/battle/battle.conf
Playtester: Improved description.

PR #9237, opened 2025-04-12
Fixed Monster Cooldowns on "Any" and "Anytarget" states

Playtester:
* **Addressed Issue(s)**: Follow-up to bdfa5fe

* **Server Mode**: Both

- Entries in mob_skill_db with state "Any" and "Anytarget" now properly set the cooldown again
- Follow-up to bdfa5fe

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -4073,10 +4073,19 @@ void mobskill_delay(mob_data& md, t_tick tick)
 		// If the skill cannot be found anymore because the monster's state has changed no delay will be applied
 		int32 delay = 0;
 		for (int32 i = 0; i < ms.size(); i++) {
-			if (ms[i]->state == md.state.skillstate && ms[i]->skill_id == ms[md.skill_idx]->skill_id) {
+			if (ms[i]->skill_id == ms[md.skill_idx]->skill_id) {
+				bool match = false;
+				if (ms[i]->state == md.state.skillstate)
+					match = true;
+				else if (ms[i]->state == MSS_ANY)
+					match = true;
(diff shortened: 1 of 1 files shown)


Lemongrass3110 (approved):

src/map/mob.cpp
```diff
@@ -4073,10 +4073,19 @@ void mobskill_delay(mob_data& md, t_tick tick)
 		// If the skill cannot be found anymore because the monster's state has changed no delay will be applied
 		int32 delay = 0;
 		for (int32 i = 0; i < ms.size(); i++) {
-			if (ms[i]->state == md.state.skillstate && ms[i]->skill_id == ms[md.skill_idx]->skill_id) {
+			if (ms[i]->skill_id == ms[md.skill_idx]->skill_id) {
```
Atemo, line 4076: I wonder, why don't we use foreach here?

Atemo (approved):

src/map/mob.cpp
Lemongrass3110: Yeah I thought the same.
Also it would require less indentation if we would invert the logic and skip entries.
But I did not want to suggest to change this in this PR.

Lemongrass3110: It would be purely stylistic and no functional change.

Playtester: Could still break something!

Playtester: Decided to finish this one without refactor. Mainly because I can't think of a good way to invert it, because all conditions would need to fail to continue. Feel free to make a follow-up PR!

PR #9238, opened 2025-04-12
Correct Determination of Loot Priority and MVP Reward

### LINKED ISSUE #8847: Loot Priority (first attacker) and MVP bonus (received damage)

Playtester:
* **Addressed Issue(s)**: #8847 

* **Server Mode**: Both

- The MVP Reward is now granted to the player that has the highest [dmg + dmg_tanked] value
  * It is not influenced by the first attacker bonus
- Fixed loot priority not properly considering first attacker and slaves
  * There is now a 30% of total damage bonus for the first person in the damage log
  * Damage dealt by players and their slaves is now added together
- Exp rewards are unaffected by this and work as previously
- Fixes #8847

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2809,6 +2809,41 @@ int32 mob_getdroprate(struct block_list *src, std::shared_ptr<s_mob_db> mob, int
 	return drop_rate;
 }
 
+/**
+ * Returns the MVP player based on the monster's damage log
+ * This player has the highest value when damage dealt and damage tanked are added together
+ * @return The MVP player
+ */
+map_session_data* mob_data::get_mvp() {
+	// There cannot be an MVP player if the monster is not an MVP
+	if (this->get_bosstype() != BOSSTYPE_MVP)
+		return nullptr;
+
+	int64 mvp_damage = 0;
+	map_session_data* mvp_sd = nullptr;
+
+	for (int32 i = 0; i < this->dmglog.size(); i++) {
+		const s_dmglog& entry = this->dmglog[i];
+
+		map_session_data* tsd = map_charid2sd(entry.id);
+
+		if (tsd == nullptr)
+			continue; // skip players that are offline
+		if (tsd->bl.m != this->bl.m)
+			continue; // skip players not on this map
+		if (pc_isdead(tsd))
+			continue; // skip dead players
+
+		if (mvp_damage >= entry.dmg + entry.dmg_tanked)
+			continue;
+
+		mvp_damage = entry.dmg + entry.dmg_tanked;
+		mvp_sd = tsd;
+	}
+
+	return mvp_sd;
+}
+
 /*==========================================
  * Signals death of mob.
  * type&1 -> no drops, type&2 -> no exp
@@ -2817,7 +2852,7 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 {
 	struct status_data *status;
 	map_session_data *sd = nullptr, *tmpsd[DAMAGELOG_SIZE];
-	map_session_data *mvp_sd = nullptr, *second_sd = nullptr, *third_sd = nullptr;
+	map_session_data *top_sd = nullptr, *second_sd = nullptr, *third_sd = nullptr;
 
 	struct {
 		struct party_data *p;
@@ -2834,7 +2869,7 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 
 	if( src && src->type == BL_PC ) {
 		sd = (map_session_data *)src;
-		mvp_sd = sd;
+		top_sd = sd;
 	}
 
 	if( md->guardian_data && md->guardian_data->number >= 0 && md->guardian_data->number < MAX_GUARDIANS )
@@ -2857,8 +2892,15 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 	// filter out entries not eligible for exp distribution
 	memset(tmpsd,0,sizeof(tmpsd));
 
+	// Struct that only contains entries eligible for loot distribution
+	// Also contains combined damage of the players and their slaves
+	struct s_dmg_entry {
+		map_session_data* sd;
+		int64 damage;
+	};
+	std::vector<s_dmg_entry> lootdmg;
+
 	count = 0;
-	int64 mvp_damage = 0;
 	int64 total_damage = 0;
 	for( i = 0; i < md->dmglog.size(); i++ ){
 		const s_dmglog& entry = md->dmglog[i];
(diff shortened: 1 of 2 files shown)


src/map/mob.cpp
```diff
@@ -2901,21 +2943,51 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
...
+
+		// Gather data to determine loot priority
+		// Check if player already has an entry
+		auto it = lootdmg.begin();
+		for (it; it != lootdmg.end(); ++it)
+			if (it->sd->bl.id == tsd->bl.id)
+				break;
+
+		if (it == lootdmg.end())
+		{
```
Lemongrass3110, line 2956:
```suggestion
		if (it == lootdmg.end()) {
```

src/map/mob.cpp
```diff
@@ -2901,21 +2943,51 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
...
+			// No matching player found, create new entry for player
+			s_dmg_entry dmg_entry;
+			dmg_entry.sd = tsd;
+			dmg_entry.damage = entry.dmg;
+			lootdmg.push_back(dmg_entry);
+		}
+		else
+		{
```
Lemongrass3110, line 2964:
```suggestion
		} else {
```

src/map/mob.cpp
```diff
@@ -2901,21 +2943,51 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
...
-			third_sd = second_sd;
-			second_sd = mvp_sd;
-			mvp_sd = tsd;
-			mvp_damage = entry.dmg;
+		tmpsd[i] = tsd; // record as valid damage-log entry
+
+		// Gather data to determine loot priority
+		// Check if player already has an entry
+		auto it = lootdmg.begin();
+		for (it; it != lootdmg.end(); ++it)
```
Lemongrass3110, line 2951:
```suggestion
		for (; it != lootdmg.end(); ++it)
```

src/map/mob.cpp
```diff
@@ -2901,21 +2943,51 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
...
 		}
+	}
 
-		tmpsd[i] = tsd; // record as valid damage-log entry
+	if (!lootdmg.empty()) {
+		// First player in the damage log gets 30% of total damage as bonus for loot priority
+		lootdmg[0].damage += (total_damage * 30) / 100;
+
+		// Sort list by damage now and determine top 3 damage dealers
+		std::sort(lootdmg.begin(), lootdmg.end(), [](s_dmg_entry& a, s_dmg_entry& b) { return a.damage > b.damage; });
```
Lemongrass3110, line 2975:
```suggestion
		std::sort(lootdmg.begin(), lootdmg.end(), [](s_dmg_entry& a, s_dmg_entry& b) {
			return a.damage > b.damage;
		});
```

src/map/mob.hpp
```diff
@@ -398,6 +398,7 @@ struct mob_data {
 	uint16 damagetaken;
 
 	e_mob_bosstype get_bosstype();
+	map_session_data* get_mvp();
```
Lemongrass3110, line 401:
```suggestion
	map_session_data* get_mvp_player();
```
Just to make it more obvious? Otherwise it may be misinterpreted as getting the MVP flag from the DB or something like this?

src/map/mob.cpp
```diff
@@ -3264,6 +3336,7 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 		add_timer(tick + (!battle_config.delay_battle_damage ? 500 : 0), mob_delay_item_drop, md->bl.id, 0);
 	}
 
+	// MVP Reward
 	if( mvp_sd && md->get_bosstype() == BOSSTYPE_MVP ){
```
Lemongrass3110, line 3340:
```suggestion
	if( mvp_sd != nullptr ){
```
Bosstype check is included in getmvp already.

src/map/mob.cpp
```diff
@@ -2809,6 +2809,41 @@ int32 mob_getdroprate(struct block_list *src, std::shared_ptr<s_mob_db> mob, int
...
+	if (this->get_bosstype() != BOSSTYPE_MVP)
+		return nullptr;
+
+	int64 mvp_damage = 0;
+	map_session_data* mvp_sd = nullptr;
+
+	for (int32 i = 0; i < this->dmglog.size(); i++) {
+		const s_dmglog& entry = this->dmglog[i];
+
+		map_session_data* tsd = map_charid2sd(entry.id);
```
Lemongrass3110, line 2828:
```suggestion
		map_session_data* sd = map_charid2sd(entry.id);
```
Here it is not a "target" sd.

Lemongrass3110 (approved):
Please add explicit nullptr checks in the modified lines.

src/map/mob.cpp
Playtester: The usage also needs to adjusted then. (I think it stands for temp sd, but you're right that we don't need to do that in this isolated function.)

src/map/mob.cpp
```diff
@@ -2817,7 +2852,7 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 {
 	struct status_data *status;
 	map_session_data *sd = nullptr, *tmpsd[DAMAGELOG_SIZE];
-	map_session_data *mvp_sd = nullptr, *second_sd = nullptr, *third_sd = nullptr;
+	map_session_data *top_sd = nullptr, *second_sd = nullptr, *third_sd = nullptr;
```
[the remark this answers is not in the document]
Playtester: I had this at first but thought people might mistake it for the first attacker, that's why I decided to rename it to the top_sd.

src/map/mob.cpp
```diff
@@ -2809,6 +2809,41 @@ int32 mob_getdroprate(struct block_list *src, std::shared_ptr<s_mob_db> mob, int
...
+			continue; // skip players that are offline
+		if (tsd->bl.m != this->bl.m)
+			continue; // skip players not on this map
+		if (pc_isdead(tsd))
+			continue; // skip dead players
+
+		if (mvp_damage >= entry.dmg + entry.dmg_tanked)
+			continue;
+
+		mvp_damage = entry.dmg + entry.dmg_tanked;
```
Atemo, line 2840: Shouldn't we use `util::safe_addition_cap` ?

src/map/mob.cpp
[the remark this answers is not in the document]
Lemongrass3110: They could also mistake second_sd as the second attacker.
I dont think that this is a valid argument for breaking the pattern.

src/map/mob.cpp
Lemongrass3110: True

Playtester: Unlikely that these will exceed int64. Also Aegis would also just loop into the negatives if that happened, unless they changed the code when they moved from int32 to another type.

Lemongrass3110: Does not matter. Please use it.

src/map/mob.cpp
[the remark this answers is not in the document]
Playtester: Renamed to first_sd based on majority vote.

PR #9243, opened 2025-04-13
Follow up to aa676e6

Lemongrass3110:

* **Server Mode**: Both

Cleaned up the logic a bit and added comments

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -4071,27 +4071,57 @@ void mobskill_delay(mob_data& md, t_tick tick)
 		// Even though we already know which skill was picked as we stored that in skill_idx
 		// Officially it actually tries to find the skill again using the monster's current state
 		// If the skill cannot be found anymore because the monster's state has changed no delay will be applied
-		int32 delay = 0;
-		for (int32 i = 0; i < ms.size(); i++) {
-			if (ms[i]->skill_id == ms[md.skill_idx]->skill_id) {
-				bool match = false;
-				if (ms[i]->state == md.state.skillstate)
-					match = true;
-				else if (ms[i]->state == MSS_ANY)
-					match = true;
-				else if (ms[i]->state == MSS_ANYTARGET && md.target_id != 0 && md.state.skillstate != MSS_LOOT)
-					match = true;
-				// State and skill match, use the first delay found
-				if (match) {
-					delay = ms[i]->delay;
-					break;
-				}
+		t_tick delay = 0;
+		for (const auto& it : ms) {
+			// Check if the skill id matches
+			if (it->skill_id != ms[md.skill_idx]->skill_id) {
+				// Skip other skills
+				continue;
+			}
+
+			// Check if the skill state matches the current monster skill state exactly
+			if (it->state == md.state.skillstate) {
+				// Skill matches and state matches exactly too
+				delay = it->delay;
+				// Use the first delay found
+				break;
+			}
+
+			// Check if the skill state applies to any monster skill state
+			if (it->state == MSS_ANY) {
+				// Skill matches and any state applies
+				delay = it->delay;
+				// Use the first delay found
+				break;
+			}
+
+			// Check if the skill state applies to any target skill states
+			// However do not apply it, if the monster already has a target or when the monster is currently looting
+			if (it->state == MSS_ANYTARGET && md.target_id != 0 && md.state.skillstate != MSS_LOOT) {
+				// Skill matches and any target skill state applies
+				delay = it->delay;
+				// Use the first delay found
+				break;
 			}
 		}
-		// Apply delay found to all entries of the skill
-		for (int32 i = 0; i < ms.size(); i++)
-			if (ms[i]->skill_id == ms[md.skill_idx]->skill_id)
-				md.skilldelay[i] = tick + delay;
+
+		// Only calculate the delay, if there is any
+		if (delay > 0) {
+			// Calculate the delay once
+			delay = tick + delay;
+
+			// Apply delay found to all entries of the skill
+			for (size_t i = 0; i < ms.size(); i++) {
+				// Check if the skill id matches
+				if (ms[i]->skill_id != ms[md.skill_idx]->skill_id) {
+					// Skip other skills
+					continue;
+				}
+
+				// Apply the delay
+				md.skilldelay[i] = delay;
+			}
+		}
 	}
 	else
 		md.skilldelay[md.skill_idx] = tick + ms[md.skill_idx]->delay;


src/map/mob.cpp
```diff
@@ -4071,27 +4071,57 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+			if (it->state == MSS_ANYTARGET && md.target_id != 0 && md.state.skillstate != MSS_LOOT) {
+				// Skill matches and any target skill state applies
+				delay = it->delay;
+				// Use the first delay found
+				break;
+			}
+		}
+
+		// Only calculate the delay, if there is any
+		if (delay > 0) {
```
Playtester, line 4109: Officially, it always sets the delay time, even if the delay is not found (it sets it to current timestamp then), but it doesn't seem to have any negative impact as far as I can see.

src/map/mob.cpp
```diff
@@ -4071,27 +4071,57 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+			// Calculate the delay once
+			delay = tick + delay;
+
+			// Apply delay found to all entries of the skill
+			for (size_t i = 0; i < ms.size(); i++) {
```
Playtester, line 4114: Any reason why we don't use for each here as well?

src/map/mob.cpp
```diff
@@ -4071,27 +4071,57 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+				// Check if the skill id matches
+				if (ms[i]->skill_id != ms[md.skill_idx]->skill_id) {
+					// Skip other skills
+					continue;
```
Playtester, line 4118: Using the "continue" style here where there is only a single condition seems a waste, I liked the previous version more.

src/map/mob.cpp
Lemongrass3110: Well... If only there was a comment in the code, that would explain this...

src/map/mob.cpp
Lemongrass3110: Because it is two different containers, which seem to share the index.

src/map/mob.cpp
Lemongrass3110: It is better this way, if you add another condition later on.

Atemo (approved):
Look good!
But imo there is too many comment now

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
-					delay = ms[i]->delay;
-					break;
-				}
+		t_tick delay = 0;
+		for (const auto& it : ms) {
+			// Check if the skill id matches
+			if (it->skill_id != ms[md.skill_idx]->skill_id) {
+				// Skip other skills
+				continue;
+			}
```
vstumpf, line 4080: like this

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+
+			// Check if the skill state matches the current monster skill state exactly
+			if (it->state == md.state.skillstate) {
+				// Skill matches and state matches exactly too
+				delay = it->delay;
+				// Use the first delay found
+				break;
```
vstumpf, line 4087: This reads like
```
// check if a is the same as b
if (a == b) {
   // a and b match
   doSomething()
   // we should do something
}
```

just remove the comments

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+			// However do not apply it, if the monster already has a target or when the monster is currently looting
+			if (it->state == MSS_ANYTARGET && md.target_id != 0 && md.state.skillstate != MSS_LOOT) {
+				// Skill matches and any target skill state applies
+				delay = it->delay;
+				// Use the first delay found
+				break;
 			}
 		}
+
+		// Calculate the delay once
```
vstumpf, line 4108: remove this

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+		// Note: Officially the delay gets set, even if it is 0.
+		delay = tick + delay;
```
vstumpf, line 4110: this comment is fine, but i'd do `delay += tick`.

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+		for (size_t i = 0; i < ms.size(); i++) {
+			// Check if the skill id matches
+			if (ms[i]->skill_id != ms[md.skill_idx]->skill_id) {
+				// Skip other skills
+				continue;
+			}
+
+			// Apply the delay
+			md.skilldelay[i] = delay;
+		}
```
vstumpf, line 4122: The original code made more sense,
for every entry
\>if the skill id matches
\>\>apply delay

now it's
for every entry
\> if it doesnt match
\>\> skip
\> apply delay

if we had to do a bunch of things if we had a match, then sure, we save the extra whitespace
but what was 3 lines of code (excluding braces) with straight-forward logic, is now 4 with (imo) more confusing logic

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+				break;
+			}
+
+			// Check if the skill state applies to any monster skill state
+			if (it->state == MSS_ANY) {
+				// Skill matches and any state applies
+				delay = it->delay;
+				// Use the first delay found
+				break;
+			}
```
vstumpf, line 4096: again, remove the comments

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+
+			// Check if the skill state applies to any target skill states
+			// However do not apply it, if the monster already has a target or when the monster is currently looting
+			if (it->state == MSS_ANYTARGET && md.target_id != 0 && md.state.skillstate != MSS_LOOT) {
+				// Skill matches and any target skill state applies
+				delay = it->delay;
+				// Use the first delay found
```
vstumpf, line 4103: remove comments

src/map/mob.cpp
```diff
@@ -4071,27 +4071,55 @@ void mobskill_delay(mob_data& md, t_tick tick)
...
+			// Check if the skill state applies to any monster skill state
+			if (it->state == MSS_ANY) {
+				// Skill matches and any state applies
+				delay = it->delay;
+				// Use the first delay found
+				break;
+			}
+
+			// Check if the skill state applies to any target skill states
+			// However do not apply it, if the monster already has a target or when the monster is currently looting
```
vstumpf, line 4099: this comment does not match the logic at all

vstumpf (changes requested):
>  But imo there is too many comment now

Agreed, some of these are just obvious

PR #9245, opened 2025-04-14
Battle Damage Target Nullptr Check

Playtester:

* **Server Mode**: Both

- The battle_damage function now exits early when target is a nullptr
  * This would have crashed the map server previously, but luckily there is no call where target is nullptr
- Cleaned up battle_damage function a little knowing that target cannot be nullptr


Atemo (approved):

Lemongrass3110:
Why did you not change it to reference?

Playtester:
> Why did you not change it to reference?

Then you'd need to adjust all calls to it. But I guess we could do that also.

PR #9246, opened 2025-04-14
Config for First Attacker Loot Priority Bonus

Playtester:
* **Addressed Issue(s)**: It's a feature

* **Server Mode**: Both

* **Description of Pull Request**: Added a config to set the loot priority bonus of the first person in the damage log

### DIFF
--- a/conf/battle/drops.conf
+++ b/conf/battle/drops.conf
@@ -13,6 +13,14 @@ item_auto_get: no
 // How long does it take for an item to disappear from the floor after it is dropped? (in milliseconds)
 flooritem_lifetime: 60000
 
+// Bonus to loot priority for first attacker (Note 2)
+// The first person in the damage log of a monster that is alive and on the same map gains a loot priority bonus
+// Officially 30% of the total damage to a monster is added to this person's damage when calculating loot priority
+// If you set this to 0 then loot priority will be solely calculated based on most damage dealt
+// If you set this to 100 (max) then the first person in the damage log will always get first loot priority
+// Loot priority determines how early a person can loot an item (see settings below)
+first_attack_loot_bonus: 30
+
 // Grace time during which only the person who did the most damage to a monster can get the item? (in milliseconds)
 item_first_get_time: 3000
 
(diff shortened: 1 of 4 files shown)


conf/battle/drops.conf
```diff
@@ -13,6 +13,14 @@ item_auto_get: no
 // How long does it take for an item to disappear from the floor after it is dropped? (in milliseconds)
 flooritem_lifetime: 60000
 
+// Bonus to loot priority for first attacker (Note 2)
+// The first person in the damage log of a monster that is alive and on the same map gains a loot priority bonus
```
Lemongrass3110, line 17:
```suggestion
// The first player in the damage log of a monster, that is alive and on the same map, gains a loot priority bonus
```

conf/battle/drops.conf
```diff
@@ -13,6 +13,14 @@ item_auto_get: no
...
+// Officially 30% of the total damage to a monster is added to this person's damage when calculating loot priority
```
Lemongrass3110, line 18:
```suggestion
// Officially 30% of the total damage to a monster is added to this player's damage when calculating loot priority
```

conf/battle/drops.conf
```diff
@@ -13,6 +13,14 @@ item_auto_get: no
...
+// If you set this to 0 then loot priority will be solely calculated based on most damage dealt
```
Lemongrass3110, line 19:
```suggestion
// If you set this to 0, then loot priority will be solely calculated based on most damage dealt
```

conf/battle/drops.conf
```diff
@@ -13,6 +13,14 @@ item_auto_get: no
...
+// If you set this to 100 (max) then the first person in the damage log will always get first loot priority
```
Lemongrass3110, line 20:
```suggestion
// If you set this to 100 (max), then the first player in the damage log will always get first loot priority
```

conf/battle/drops.conf
```diff
@@ -13,6 +13,14 @@ item_auto_get: no
...
+// Loot priority determines how early a person can loot an item (see settings below)
```
Lemongrass3110, line 21:
```suggestion
// Loot priority determines how early a player can loot an item (see settings below)
```

conf/battle/drops.conf
Lemongrass3110: The question is if we should not rephrase this like with the other configs.
Something like: "How many percent does the first player ... get?"

Lemongrass3110: Just look at the config right above xD

Playtester: The config above is not a (Note 2) type config. If you there's no note you need to explain a bit more and we usually phrase it as a question for that, but (Note 2) already makes it clear that you set the percent so no need to ask "How much percent?"
I will stay consistent with other (Note 2) configs.

Issue #1656, opened 2016-10-25
Walking delay when attacked


esu1214:
I have confirmed that monsters cannot be attacked to stop walking in rAthena, but it is very different in Aegis.
Monsters even Mvps can be attacked to stop walking by high ASPD attcker in Aegis. I have just tested it in my Aegis test server.

cydh:
MVP can be hitlocked in idRO, and as some iRO guys said they mostly MVP using ASPD Wug-Ranger for it.
@Atemo said so, he's ASPD Wug-Ranger on iRO

esu1214:
Anyway none-Mvp monsters should be hitlocked by high ASPD attack.

Playtester:
Monsters do stop walking when attack in rAthena too, even bosses.

The actual system of hitlock is weird, though. I tried around with high ASPD snipers for long time and sometimes I could completely stop the monster and sometimes it didn't slow down at all.

I assume it has something to do with the timer intervals, but I never could figure out the exact implementation to simulate it. I guess we'd need to record the exact timestamp of hits and when the client receives monster move notifications from Aegis so we can figure out how it works.

MrAntares:
I know it is somewhat offtopic, but this ridiculous black magic based aegis walk system has been plaguing rA and eA since the beginning. Wouldn't it be worth implementing a properly working one where the server and the client does the same thing without lags and desync? I know it is custom, but there are many improved features already that can be turned on/off, so this could be the same. I prefer vanilla, but honestly I wouldn't care in this case, because trying to emulate the "noone knows how vanilla" is just bad. Also I don't like things like: "_the monster never moves client-sided when hit often, but it certainly moves server-sided_". I think players want to play seamlessly and don't care if it's only 95% vanilla if it's working better.

Playtester:
The problem is that we can't modify the client. The only way to fix it server-sided is by using fixpos packets instead of move packets, but then you lose all the animations and instead have instant movement from cell to cell, like in an old ascii based roguelike. I'm not sure if that's desired.

esu1214:
I can confirmed that it is related by attcker's amotion, monster's dmotion, monster's speed and the most important is the distance between the attacker and the monster.
For example, a Ranger with 190 ASPD or more can hit BAPHOMET to stop walking by normal attack, but the condition is that the distance between them should be the longest range of a bow.

esu1214:
Players in Aegis Server hitlock a Mvp through there mouse-skills, but in emulation, they can never do that.

[...]

PR #9248, opened 2025-04-18
Delay Setting of Attacked ID

Playtester:
* **Addressed Issue(s)**: #1656 (preparation for final fix)

* **Server Mode**: Both

- Implemented official system for setting the attacked ID
  * The attacked ID will now be set only after waiting for the calculated walk delay
  * The monster AI is processed immediately after setting it
- Changed the rules for setting the norm_attacked_id
  * These conditions are checked at the beginning of the attack instead of when the damage is applied or later
  * Monster must be hit by a normal attack
  * Monster needs to be in Berserk state
  * Monster needs to have a different target currently
  * Attacker is within AttackRange+1
  * Setting norm_attacked_id makes monsters switch target in Berserk state (if they have the mode to do so)
- Unified code and fixed the condition for monster skill and attacked events to occur
  * Does no longer require damage to be dealt, but requires the attack to connect
  * Removed duplicate monster skill event triggers (event was triggered 3 times per attack instead of once)
  * Follow-up to 36b9b94, 77b4b3c, 4800b5b and cad5cab
- Related to #1656

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1243,7 +1243,11 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
 		case MSS_BERSERK:
 			if (!(mode&MD_CHANGETARGETMELEE))
 				return 0;
-			if (!(battle_config.mob_ai&0x80) && md->norm_attacked_id != target->id)
+			// If the special normal attacked event occured, always change target in berserk state
+			if (md->norm_attacked_id == target->id)
+				return 1;
+			// If the special setting to switch target even on skills is set, we need to verify the range here
+			if (!(battle_config.mob_ai&0x80))
 				return 0;
 			return (battle_config.mob_ai&0x4 || check_distance_bl(&md->bl, target, md->status.rhw.range+1));
 		case MSS_RUSH:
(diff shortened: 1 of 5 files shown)


src/map/mob.cpp
```diff
@@ -1243,7 +1243,11 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
 		case MSS_BERSERK:
 			if (!(mode&MD_CHANGETARGETMELEE))
 				return 0;
-			if (!(battle_config.mob_ai&0x80) && md->norm_attacked_id != target->id)
+			// If the special normal attacked event occured, always change target in berserk state
+			if (md->norm_attacked_id == target->id)
+				return 1;
+			// If the special setting to switch target even on skills is set, we need to verify the range here
```
vstumpf, line 1249: id change the comment to something like `0x80 is switch target on skills`
alternatively you could give 0x80 a better name and we could do away with the comment altogether

src/map/mob.cpp
```diff
@@ -2232,6 +2236,86 @@ bool mob_ai_sub_hard_attacktimer(mob_data &md, t_tick tick)
...
+			{
+				// Let mobs retaliate against the pet's master
+				md->attacked_id = pd.master->bl.id;
+			}
+			break;
+		}
+		case BL_MOB:
+		{
+			struct mob_data& md2 = *reinterpret_cast<mob_data*>(src);
+			// Let players decide whether to retaliate versus the master or the mob
```
vstumpf, line 2271: let config decide? or let admin decide? not really the player

src/map/skill.cpp
```diff
@@ -4031,22 +4031,6 @@ int64 skill_attack (int32 attack_type, struct block_list* src, struct block_list
 				skill_counter_additional_effect(src, bl, skill_id, skill_lv, dmg.flag, tick);
 		} else
 			battle_delay_damage(tick, dmg.amotion, src, bl, dmg.flag, skill_id, skill_lv, damage, dmg.dmg_lv, dmg.div_, additional_effects, false);
-	} else {
-		// Trigger monster skill condition for damage skills with no amotion.
-		if (bl->type == BL_MOB && src != bl && !status_isdead(*bl)) {
-			if (damage > 0)
-				mobskill_event(BL_CAST(BL_MOB, bl), src, tick, dmg.flag);
```
vstumpf, line 4038: were these duped before? we really don't need it anymore?

Playtester: Yes. basically mobskill_event was called 3 times each time a mob was hit (making it 3 times more likely to use a skill).
I linked like 4 commits that worked together in creating this mess in the description.

Long Story:

Basically the original issue was that Daegaladh moved the mobskill_event to battle_damage, but ground skills didn't reach battle_damage because we called status_damage instead of battle_damage there.

But instead of fixing the call, aleos had added this code here for ground skills.

Then later datawulf added the code directly below, which general calls all mobskill events again. But here he gives damage into the function -> aleos code above was not modified (I assume both were in PR state at the same time and so nobody saw that the automatic merged duplicated the code).
Battle_damage was also modified by datawulf, but only halfway (one call had damage, the second was missing it).
Datawulf modified some events so that they only work with damage, but others didn't have that condition.

So now non-ground skills called the event twice, 1.5/2 times with damage and ground skills called the code twice 1/2 times with damage.

Then in the last linked commit, I finally fixed the status_damage -> battle_damage call, but didn't realize that in the meantime the event calls were duplicated twice. However, I also missed that "additional_effects" is actually false in battle_damage for ground skills (because skill_additional_effect is called directly in skill_attack for those), I incorrectly added the "additional_effects" variable check to the "monster is attacked" event code.

With this PR, I fixed it so that that "monster is attacked" event code in battle_damage is now correctly reached by non-ground and ground skills and also added the missing "damage" parameter. Now everything properly channels through battle_damage, so the code here is no longer needed.

I verified with tests that both non-ground and ground skills now trigger each event exactly once.

src/map/mob.cpp
Playtester: I just moved the code from mob_log_damage, but I guess might as well improve the comments while at it.

src/map/mob.cpp
Playtester: I always used this style of describing it, but maybe I shouldn't even mention the setting - main purpose of this comment is to explain why we check for the distance here again in the code below.

I really don't want to start explaining what each hex value in mob_ai means everywhere we use it.

Giving each value in mob_ai a name should be a separate PR.

vstumpf (approved):

PR #9253, opened 2025-04-21
True Hitlock System / Monster Walk Delay Fixes

### LINKED ISSUE #1656: Walking delay when attacked

Playtester:
* **Addressed Issue(s)**: #1656 

* **Server Mode**: Both

- Made the legacy walk delay system optional
  * Officially, getting hit just stops you on the closest cell center
- Endure no longer causes the "monster attacked" to trigger immediately
  * It still prevents legacy walk delay and getting stopped by hits
- Fixed aggressive mode being disabled too early when hit
- Follow-up to 499d852
- Fixes #1656

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -1642,8 +1642,9 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp, 
 	}
 
 	if( status->hp || (flag&8) ) { // Still lives or has been dead before this damage.
-		if (walkdelay)
-			unit_set_walkdelay(target, gettick(), walkdelay, 0);
+		t_tick tick = gettick();
+		if (walkdelay && !status_isendure(*target, tick, false))
+			unit_set_walkdelay(target, tick, walkdelay, 0);
 		return (int32)(hp+sp+ap);
 	}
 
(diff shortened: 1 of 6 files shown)


src/map/status.cpp
```diff
@@ -1642,8 +1642,9 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp,
 	}
 
 	if( status->hp || (flag&8) ) { // Still lives or has been dead before this damage.
-		if (walkdelay)
-			unit_set_walkdelay(target, gettick(), walkdelay, 0);
+		t_tick tick = gettick();
+		if (walkdelay && !status_isendure(*target, tick, false))
+			unit_set_walkdelay(target, tick, walkdelay, 0);
```
Daegaladh, line 1647:
```suggestion
		if (t_tick tick = gettick(); walkdelay > 0 && !status_isendure(*target, tick, false))
			unit_set_walkdelay(target, tick, walkdelay, 0);
```
Since you're only using the tick on this context

Daegaladh (approved):

Issue #9250, opened 2025-04-19
Full Inventory Trade Issue


Elphin15:
### rAthena Hash

Latest

### Client Date

-

### Server Mode

Renewal

### Result

**Description of Issue:**
When a trade occurs between two players, the system allows the trade to proceed even if one of the players has a full inventory. This behavior leads to unexpected results where only a subset of items is transferred, and the rest are returned to the initiating player. For example:

- **Player One** has a full inventory (100/100).

![Image](https://github.com/user-attachments/assets/f96aa5f0-c5a3-499c-84ae-913e6ae6fb20)

- **Player Two** places multiple items into the trade (e.g., Red Potion, Green Herb and Fluff).

![Image](https://github.com/user-attachments/assets/32eeafcb-faca-44ab-8a08-5b8b35c2a866)

- After the trade:
- **Player One** receives only one item (in this case was Green Herb)
- The remaining items (Knife, Red Potion and Fluff) are returned to **Player Two**.

### Relevant Log Output

```Shell

```

### Expected Result

The deal shouldn't even open, right?

### How to Reproduce

- Ensure Player One has a full inventory.
- Initiate a trade with Player Two.
- Player Two places multiple items into the trade.
- Complete the trade.
- Observe that only one item is transferred to Player One while the remaining items are returned to Player Two.

### Official Information

Check the inventory before opening the trade window.

Playtester:
Tested on official servers: You cannot drag an item to the trade window if weight limit would be exceeded. Will give a "You cannot trade because this character will exceed his weight limit." message.

PR #9260, opened 2025-04-23
Added missing trade checks

Daegaladh:
* **Addressed Issue(s)**: #9250 

* **Server Mode**: Both

- Adds a missing check for full inventory (it shows different messages depending on the client version)
- Adds a missing check for stacked items limit (it shows different messages depending on the client version)
- Removes previously commented message for item trade limit, as I confirmed it's client-side
- Adds a battle config for whether stackable items should be counted separately for inventory space even if the recipient already has that item

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -12197,6 +12197,7 @@ static const struct _battle_data {
 	{ "loot_range",                         &battle_config.loot_range,                      12,     1,      MAX_WALKPATH,   },
 	{ "assist_range",                       &battle_config.assist_range,                    11,     1,      MAX_WALKPATH,   },
 	{ "major_overweight_rate",              &battle_config.major_overweight_rate,           90,     0,      100             },
+	{ "trade_count_stackable",              &battle_config.trade_count_stackable,           1,      0,      1,              },
 
 #include <custom/battle_config_init.inc>
 };
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -11685,9 +11685,11 @@ static const struct _battle_data {
 #ifdef RENEWAL
 	{ "natural_heal_weight_rate",           &battle_config.natural_heal_weight_rate,        70,     0,      100             },
 	{ "open_box_weight_rate",               &battle_config.open_box_weight_rate,            90,     0,      100             },
+	{ "trade_weight_rate",                  &battle_config.trade_weight_rate,               100,     0,      100             },
 #else
 	{ "natural_heal_weight_rate",           &battle_config.natural_heal_weight_rate,        50,     0,      100             },
 	{ "open_box_weight_rate",               &battle_config.open_box_weight_rate,            70,     0,      100             },
+	{ "trade_weight_rate",                  &battle_config.trade_weight_rate,               50,     0,      100             },
 #endif
 	{ "arrow_decrement",                    &battle_config.arrow_decrement,                 1,      0,      2,              },
 	{ "ammo_unequip",                       &battle_config.ammo_unequip,                    1,      0,      1,              },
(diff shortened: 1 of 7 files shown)


src/map/battle.cpp
```diff
@@ -11685,9 +11685,11 @@ static const struct _battle_data {
 #ifdef RENEWAL
 	{ "natural_heal_weight_rate",           &battle_config.natural_heal_weight_rate,        70,     0,      100             },
 	{ "open_box_weight_rate",               &battle_config.open_box_weight_rate,            90,     0,      100             },
+	{ "trade_weight_rate",                  &battle_config.trade_weight_rate,               100,     0,      100             },
 #else
 	{ "natural_heal_weight_rate",           &battle_config.natural_heal_weight_rate,        50,     0,      100             },
 	{ "open_box_weight_rate",               &battle_config.open_box_weight_rate,            70,     0,      100             },
+	{ "trade_weight_rate",                  &battle_config.trade_weight_rate,               50,     0,      100             },
```
Playtester, line 11692: You can trade just fine in pre-renewal when you are at >50% weight.

Playtester (approved):

src/map/trade.cpp
```diff
@@ -413,10 +413,23 @@ void trade_tradeadditem(map_session_data *sd, int16 index, int16 amount)
 	// Locate a trade position
 	ARR_FIND( 0, 10, trade_i, sd->deal.item[trade_i].index == index || sd->deal.item[trade_i].amount == 0 );
 	if( trade_i == 10 ) { // No space left
-		// clif_tradeitemok(*sd, index, EXITEM_ADD_FAILED_OVERWEIGHT); // We do not know if the server respond with this or not since the official client prevents this case client-side.
 		return;
 	}
 
+	// Fail to add the item if the inventory is full
+	if (pc_inventoryblank(target_sd) <= trade_i) {
```
Lemongrass3110, line 420: Should this be moved into line 446?
"New deal item"?

src/map/trade.cpp
```diff
@@ -413,10 +413,23 @@ void trade_tradeadditem(map_session_data *sd, int16 index, int16 amount)
 	// Locate a trade position
 	ARR_FIND( 0, 10, trade_i, sd->deal.item[trade_i].index == index || sd->deal.item[trade_i].amount == 0 );
 	if( trade_i == 10 ) { // No space left
-		// clif_tradeitemok(*sd, index, EXITEM_ADD_FAILED_OVERWEIGHT); // We do not know if the server respond with this or not since the official client prevents this case client-side.
```
Lemongrass3110, line 416: Could @Playtester verify this?
I would like to leave the comment at least "official client prevents this" or not.

src/map/trade.cpp
```diff
@@ -413,10 +413,23 @@ void trade_tradeadditem(map_session_data *sd, int16 index, int16 amount)
...
 	}
 
+	// Fail to add the item if the inventory is full
+	if (pc_inventoryblank(target_sd) <= trade_i) {
+		clif_tradeitemok(*sd, index, EXITEM_ADD_FAILED_OVERCOUNT);
+		return;
+	}
+
+	// Fail to add the item if is stacked over the limit
+	if (itemdb_isstackable(item->nameid)) {
```
Lemongrass3110, line 426: I think to really check this correctly we would need to add a new variable to sd->deal with the required slots.
Here you only calculate the required slots for the current item, but it may be necessary to calculate the required slots for ALL the items in the deal.
See clif_parse_Mail_getattach for a reference.

src/map/trade.cpp
Daegaladh: No, on officials this is checked each time you add an item, and before it's added to sd->deal.item (otherwise it will return the item, basically duplicating them)

src/map/trade.cpp
Daegaladh: I already tested it yesterday on both Aegis pre-re an renewal, but I'd add the comment

src/map/trade.cpp
Daegaladh: That's not how it works, both on officials and rathena it's checked each time you add a new item to the trade, this checks if the other player will exceed 30k of the item with the amount you tried to trade, otherwise is not added to the list and sends the clif_tradeitemok.
And you can't really re-add or update an item to the trade list, we just have the code below (the sd->deal.item checks) to avoid possible exploits.

vstumpf (approved):

PR #9261, opened 2025-04-24
Song/Dance Buff Refresh Behavior

### LINKED ISSUE #1031: Song lingering should not refresh
On Aegis, the status changes that songs give don't overwrite each other. You will have to wait the full time before gaining the effect again or even updating it to another level.

Herc already has implemented something like that, there are too many commits to list them but the final one was here:
https://github.com/csnv/Hercules/commit/19c5f981dc5ce9a2dbea0691c0ba06191685dabc

I recommend we make it optional too.

Playtester:
* **Addressed Issue(s)**: #1031

- Song/Dance buff durations can no longer be refreshed once they become finite
  * When the song/dance buff expires while in the area, it will get restored at that time
  * Added a config to restore previous behavior
- Fixed the song/dance buff status icon showing a fake duration
  * Now the duration will show as unlimited when in the song/dance area
  * Added an option to show real duration once the song/dance area was left
- Fixed an issue that the "onleft" event from dissonance was only invoked on the moving song/dance
  * Now it will also be invoked for the songs/dances that it moved onto
- Song/dance buffs are now correctly reinstated when overlap dissonance ends
- Fixed lingering effects of songs also applying to non-players
- Fixed an issue that pseudo-infinite song buffs ended too early
  * Reduces unnecessary server load
  * This also fixes an issue where the lingering effect would be skipped if you stood in the area the whole time
- Fixes #1031

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -16704,7 +16704,7 @@ static int32 skill_unit_onplace(struct skill_unit *unit, struct block_list *bl, 
 			else if (battle_config.refresh_song == 1 && sce->val4 == 1) { //Readjust timers since the effect will not last long.
 				sce->val4 = 0; //remove the mark that we stepped out
 				delete_timer(sce->timer, status_change_timer);
-				sce->timer = add_timer(tick+sg->limit, status_change_timer, bl->id, type); //put duration back to 3min
+				sce->timer = add_timer(tick + sg->limit + SKILLUNITTIMER_INTERVAL, status_change_timer, bl->id, type);
 				// Update icon duration
 				if (battle_config.refresh_song_icon == 1) {
 					if (auto scdb = status_db.find(type); scdb != nullptr)


src/map/skill.cpp
```diff
@@ -16689,12 +16695,22 @@ static int32 skill_unit_onplace(struct skill_unit *unit, struct block_list *bl,
...
-			else if (sce->val4 == 1) { //Readjust timers since the effect will not last long.
+				sc_start4(ss, bl, type, 100, sg->skill_lv, sg->val1, sg->val2, 0, sg->limit + SKILLUNITTIMER_INTERVAL);
+			else if (battle_config.refresh_song == 1 && sce->val4 == 1) { //Readjust timers since the effect will not last long.
 				sce->val4 = 0; //remove the mark that we stepped out
 				delete_timer(sce->timer, status_change_timer);
 				sce->timer = add_timer(tick+sg->limit, status_change_timer, bl->id, type); //put duration back to 3min
+				// Update icon duration
+				if (battle_config.refresh_song_icon == 1) {
+					std::shared_ptr<s_status_change_db> scdb = status_db.find(type);
+					if (scdb != nullptr)
```
Daegaladh, line 16711:
```suggestion
					if (auto scdb = status_db.find(type); scdb != nullptr)
```

src/map/skill.cpp
```diff
@@ -17871,14 +17887,24 @@ int32 skill_unit_onleft(uint16 skill_id, struct block_list *bl, t_tick tick)
...
 				//NOTE: It'd be nice if we could get the skill_lv for a more accurate extra time, but alas...
 				//not possible on our current implementation.
+				t_tick duration = skill_get_time2(skill_id, 1);
 				sce->val4 = 1; //Store the fact that this is a "reduced" duration effect.
-				sce->timer = add_timer(tick+skill_get_time2(skill_id,1), status_change_timer, bl->id, type);
+				sce->timer = add_timer(tick + duration, status_change_timer, bl->id, type);
+				// Update icon duration
+				if (battle_config.refresh_song_icon == 1) {
+					std::shared_ptr<s_status_change_db> scdb = status_db.find(type);
+					if (scdb != nullptr)
```
Daegaladh, line 17901:
```suggestion
					if (auto scdb = status_db.find(type); scdb != nullptr)
```

Daegaladh (approved):

PR #9262, opened 2025-04-25
Position Synchronization Improvements

Playtester:
* **Addressed Issue(s)**: Follow-up to b7b6812

* **Server Mode**: Both

- The client will now be correctly notified again when it reached the end of the walkpath
  * The scenario of the player already being on the target cell from the start was not considered
  * Follow-up to b7b6812
- Fixed the "end of walkpath" events not being triggered when the end of walkpath was reached by damage
  * This can happen because getting hit moves you to the closest cell center, which could be forward
  * The event is needed here as it checks for a target in attack range and sends a fixpos packet before the attack

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -171,8 +171,13 @@ bool unit_walktoxy_nextcell(block_list& bl, bool sendMove, t_tick tick) {
 		return true;
 
 	// Reached end of walkpath
-	if (ud->walkpath.path_pos >= ud->walkpath.path_len)
+	if (ud->walkpath.path_pos >= ud->walkpath.path_len) {
+		// We need to send the reply to the client even if already at the target cell
+		// This allows the client to synchronize the position correctly
+		if (sendMove && bl.type == BL_PC)
+			clif_walkok(reinterpret_cast<map_session_data&>(bl));
 		return false;
+	}
 
 	// Monsters first check for a chase skill and if they didn't use one if their target is in range each cell after checking for a chase skill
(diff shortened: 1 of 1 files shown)


src/map/unit.cpp
```diff
@@ -171,8 +171,13 @@ bool unit_walktoxy_nextcell(block_list& bl, bool sendMove, t_tick tick) {
 		return true;
 
 	// Reached end of walkpath
-	if (ud->walkpath.path_pos >= ud->walkpath.path_len)
+	if (ud->walkpath.path_pos >= ud->walkpath.path_len) {
+		// We need to send the reply to the client even if already at the target cell
+		// This allows the client to synchronize the position correctly
+		if (sendMove && bl.type == BL_PC)
```
Daegaladh, line 177: Maybe change the order for short-circuit. But I'm not sure if sendMove will be true more often than the other check.

Daegaladh (approved):

src/map/unit.cpp
Playtester: SendMove is a binary check so it should be faster. It's true when it's a new move command, but not true for any following cell moved. A PC moving happens also quite often, so... I don't think it makes a difference. My logic was that the binary check is faster than an integer comparison, but actually not so sure about it?

Daegaladh: Yeah I'm not sure it would make a huge difference, just maybe if this is checked more often for non-players. Anyway it was just a suggestion and probably not even relevant on modern compilers.

PR #9273, opened 2025-04-27
Fixed instance timer

Atemo:
* **Addressed Issue(s)**: -

* **Server Mode**: both

This PR makes it so that the instance timer is recreated when the NPC script is reloaded, assuming that was the original intention.

Example : 
GM used `reloadscript` when a player spent 3 hours inside MD endless tower (`TimeLimit: 14400`)
Before the fix : remaining time in the instance = 4h - 3 h = 1h
After the fix : remaining time in the instance = 4h

### DIFF
--- a/src/map/instance.cpp
+++ b/src/map/instance.cpp
@@ -1235,9 +1235,7 @@ void do_reload_instance(void)
 				instance_addnpc(idata);
 
 			// Create new keep timer
-			std::shared_ptr<s_instance_db> db = instance_db.find(idata->id);
-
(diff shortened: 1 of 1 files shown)


idk-whoami:
after reloadscript, some party members who previously logged out re-entered the game at the last location.

Daegaladh (approved):

Atemo:
> after reloadscript, some party members who previously logged out re-entered the game at the last location.

Nice catch! But I don't see an easy way to fix this right now

PR #9274, opened 2025-04-28
Pre-Renewal Earthquake Damage

### LINKED ISSUE #9272: Earthquake should be affected by NPC_POWERUP
### rAthena Hash

04bb999

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

Earthquake damage is just base ATK multiplied with rate.

### Expected Result

NPC_POWERUP triples Earthquake damage in pre-renewal.

Earthquake also always hits but we have some weird hitrate+20 code that could be removed.

### How to Reproduce

Fight a monster that has both NPC_EARTHQUAKE and NPC_POWERUP, for example RSX.

### Official Information

Official RSX damage:

Earthquake damage without NPC_POWERUP: 29890-52220 (27400-56200 mathematically)
Earthquake damage with NPC_POWERUP: 92550-154170 (82200-168600 mathematically)

I also tested Ifrit Card with Provoke, but Provoke didn't increase its damage, so it seems like not everything affecting ATKpercent works with NPC_EARTHQUAKE. Or maybe it only does for Non-PC.

Confirmed: Provoke increases damage of NPC_EARTHQUAKE, but only for monsters! Same goes for all other skills that use ATKpercent.

Renewal doesn't use this skill anymore, it uses NPC_EARTHQUAKE_K instead.

Impositio increases the damage for both players and monsters.

_No response_

Playtester:
* **Addressed Issue(s)**: #9272 

- Fixed NPC_EARTHQUAKE not being affected by ATKpercent in pre-renewal
- Fixed NPC_EARTHQUAKE damage formula when used by players in pre-renewal (Ifrit Card)
- Added comment about NPC_EARTHQUAKE_K
- Removed useless hitrate bonus code (NPC_EARTHQUAKE is a magic skill)
- Fixes #9272

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -3280,7 +3280,6 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct 
 			case NPC_TELEKINESISATTACK:
 			case NPC_UNDEADATTACK:
 			case NPC_CHANGEUNDEAD:
-			case NPC_EARTHQUAKE:
 			case NPC_POISON:
 			case NPC_BLINDATTACK:
 			case NPC_SILENCEATTACK:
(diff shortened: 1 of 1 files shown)


src/map/battle.cpp
```diff
@@ -8271,10 +8270,22 @@ struct Damage battle_calc_magic_attack(struct block_list *src,struct block_list
...
 					ad.damage = sstatus->str * 2 + battle_calc_weapon_attack(src, target, skill_id, skill_lv, mflag).damage;
-				else
+#else
+					ad.damage = sd->battle_status.batk + sd->battle_status.rhw.atk;
+#endif
+				}
+				else {
 					ad.damage = battle_calc_base_damage(src, sstatus, &sstatus->rhw, sc, tstatus->size, 0);
+#ifndef RENEWAL
+					if (sc != nullptr)
```
Daegaladh, line 8285: I would move this check inside the else, as an else if, since the renewal code didn't check for the sc either,

Daegaladh (approved):

src/map/battle.cpp
Playtester: This won't work. You need to call battle_calc_base_damage even if sc is nullptr.

Daegaladh: Oh I see 👍

PR #9278, opened 2025-04-30
Base Attack Min/Max Limits

### LINKED ISSUE #1364: bBaseAtk,-negative value; ** Not Working **
5481,Hermose_Cap,Hermode Cap,4,20,,1000,,1,,1,0xFFFFFFFF,7,2,256,,1,1,478,{ bonus bAspdRate,10; bonus bBaseAtk,-20; bonus bMatkRate,-10; },{},{}

I notice that when having negative value on bBaseAtk, it did not reduce atk, but if positive, it adds to atk at status window. I think it should reduce.

Playtester:
* **Addressed Issue(s)**: #1364 

* **Server Mode**: Pre-Renewal (mostly)

- Fixed base attack (batk) not allowing negative values in pre-renewal
- Changed type to int32
- Code simplification
- Fixes #1364

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -69,7 +69,7 @@ static uint16 status_calc_wis(struct block_list *, status_change *, int32);
 static uint16 status_calc_spl(struct block_list *, status_change *, int32);
 static uint16 status_calc_con(struct block_list *, status_change *, int32);
 static uint16 status_calc_crt(struct block_list *, status_change *, int32);
-static uint16 status_calc_batk(struct block_list *,status_change *,int32);
+static int32 status_calc_batk(struct block_list *,status_change *,int32);
 static uint16 status_calc_watk(struct block_list *,status_change *,int32);
 static int16 status_calc_hit(struct block_list *,status_change *,int32);
 static int16 status_calc_critical(struct block_list *,status_change *,int32);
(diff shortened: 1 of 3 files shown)


src/map/status.cpp
```diff
@@ -2724,8 +2724,11 @@ void status_calc_misc(struct block_list *bl, struct status_data *status, int32 l
 		status->flee2 = 0;
 
 	if (status->batk) {
-		int32 temp = status->batk + status_base_atk(bl, status, level);
-		status->batk = cap_value(temp, 0, USHRT_MAX);
+		status->batk += status_base_atk(bl, status, level);
+#ifndef RENEWAL
+		if (bl->type != BL_PC)
+			status->batk = cap_value(status->batk, SHRT_MIN, SHRT_MAX);
```
Daegaladh, line 2730: Please add a comment about this being capped for non-players on pre-re on each one of the caps, even if look obvious someone might see it in the future and wonder why we did this

Daegaladh (commented):
Just wondering, batk also allows negative values on Renewal?

src/map/status.cpp
Playtester: I was wondering if we shouldn't just remove capping it altogether. Officially you can't reach values that high or low and if someone adds customized stuff, they probably just want the value to be higher than what short usually allows, what do you think?

Atemo (commented):
Maybe cast/cap value in
https://github.com/rathena/rathena/blob/a658af47ae2d5d7d23f145066f848a1f606860cb/src/map/clif.cpp#L15747
?

Atemo (approved):
Untested but look fine, I'll let you test it ig

PR #9279, opened 2025-05-01
Correction count of inventory slots of a player

Aristeus09:
* **Addressed Issue(s)**:  
Using ```MAX_INVENTORY``` to check player inventory slots is always at maximum even they only have 100 slots as ```default``` in each player.
Related also to https://github.com/rathena/rathena/issues/9250

* **Server Mode**:  Both

Maybe it was missed during the implementation of inventory expansion.

### DIFF
--- a/src/map/atcommand.cpp
+++ b/src/map/atcommand.cpp
@@ -1642,7 +1642,7 @@ ACMD_FUNC(itemreset)
 	int32 i;
 	nullpo_retr(-1, sd);
 
-	for (i = 0; i < MAX_INVENTORY; i++) {
+	for (i = 0; i < sd->status.inventory_slots; i++) {
 		if (sd->inventory.u.items_inventory[i].amount && sd->inventory.u.items_inventory[i].equip == 0 && !itemdb_ishatched_egg(&sd->inventory.u.items_inventory[i])) {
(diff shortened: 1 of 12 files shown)


Daegaladh:
MAX_INVENTORY is already defined as ( INVENTORY_BASE_SIZE + INVENTORY_EXPANSION_SIZE ) so it's taking the expansion into account.
What was your exact issue?

Aristeus09:
My concern is this(please watch the video attached).

As you can see on the video, there's a transaction between the Paladin and Lord Knight.
Whereas the Paladin put 4pcs Knife on the trade and  the Lord Knight put 20m(as payment). The thing is, the Lord Knight's inventory is at 100/100 and yet, when both parties accepted the trade, it prompts as "the trade is successful". The 4pcs Knife from the Paladin returned to his inventory and still received the 20m zeny from the Lord Knight. The trade should prompt as unsuccessful and both parties should get back whatever they put on the trade(items or zeny).

so i assume that ```MAX_INVENTORY``` is not defining the correct inventory slots count of a player.

If what I assume to be the problem is wrong, you can close my Pull request.

https://github.com/user-attachments/assets/d0ad3a2d-b98d-4c3e-a6bd-4270232a74f9

Daegaladh:
That should be fixed in my PR #9260 I already tried a few scenarios and seems to work fine.
Please try it and confirm it works as intended.

PR #9281, opened 2025-05-02
Improved Song/Dance Overlap Dissonance Check

Playtester:
* **Addressed Issue(s)**: Follow-up to fc57498

- Overlap dissonance now also works fully reliably when three or more songs/dances overlapping with each other
- Added an enum e_dance_overlap for the overlap flag
- Improved function comments
- Removed obsolete dissonance code that caused problems
- Fixed an interval error showing on map server even though everything is working as intended

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -15851,26 +15851,26 @@ static int32 skill_dance_overlap_sub(struct block_list* bl, va_list ap)
  * @param flag see enum e_dance_overlap
  * @return Number of skill units that overlap with current skill unit and meet the overlap criteria (excluding current skill unit)
  */
-int32 skill_dance_overlap(struct skill_unit* unit, e_dance_overlap flag)
+int32 skill_dance_overlap(skill_unit& unit, e_dance_overlap flag)
 {
-	if (!unit || !unit->group || !(unit->group->state.song_dance&0x1))
+	if (unit.group == nullptr || !(unit.group->state.song_dance&0x1))
 		return 0;
 
 	if (flag == OVERLAP_REMOVE) {
 		// Nothing to remove, this unit is not overlapped
-		if (!(unit->val2&(1 << UF_ENSEMBLE)))
+		if (!(unit.val2&(1 << UF_ENSEMBLE)))
 			return 0;
 
 		// When the unit is removed, but there are more than one other skill unit that overlaps,
 		// dissonance is only removed from the unit that is removed and not the other units
-		if (int32 count = map_foreachincell(skill_dance_overlap_sub, unit->bl.m, unit->bl.x, unit->bl.y, BL_SKILL, unit, OVERLAP_COUNT); count > 1) {
+		if (int32 count = map_foreachincell(skill_dance_overlap_sub, unit.bl.m, unit.bl.x, unit.bl.y, BL_SKILL, &unit, OVERLAP_COUNT); count > 1) {
 			// Remove dissonance
-			skill_dance_overlap_revert(*unit);
+			skill_dance_overlap_revert(unit);
 			return count;
 		}
 	}
 
-	return map_foreachincell(skill_dance_overlap_sub, unit->bl.m,unit->bl.x,unit->bl.y,BL_SKILL, unit,flag);
+	return map_foreachincell(skill_dance_overlap_sub, unit.bl.m, unit.bl.x, unit.bl.y, BL_SKILL, &unit, flag);
 }
 
 /**
@@ -16939,6 +16939,9 @@ int32 skill_unit_onplace_timer(struct skill_unit *unit, struct block_list *bl, t
 			case UNT_MANHOLE:
 				return 0;
 			default:
+				// Dissonance effect might have killed another bard/dancer and skill returned to normal
+				if (unit->group->state.song_dance&0x1)
+					return 0;
 				ShowError("skill_unit_onplace_timer: interval error (unit id %x)\n", sg->unit_id);
 				return 0;
 		}
@@ -21883,7 +21886,7 @@ skill_unit* skill_initunit(std::shared_ptr<s_skill_unit_group> group, int32 idx,
 			break;
 		default:
 			if (group->state.song_dance&0x1) //Check for dissonance.
-				skill_dance_overlap(unit, OVERLAP_SET);
+				skill_dance_overlap(*unit, OVERLAP_SET);
 			break;
 	}
 
@@ -21910,7 +21913,7 @@ int32 skill_delunit(struct skill_unit* unit)
 		return 0;
 
 	if( group->state.song_dance&0x1 ) //Cancel dissonance effect.
-		skill_dance_overlap(unit, OVERLAP_REMOVE);
+		skill_dance_overlap(*unit, OVERLAP_REMOVE);
 
 	// invoke onout event
 	if( !unit->range )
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -15759,12 +15759,33 @@ int32 skill_castend_map (map_session_data *sd, uint16 skill_id, const char *mapn
 
 static bool skill_dance_switch(struct skill_unit* unit, bool revert);
 
-/// transforms 'target' skill unit into dissonance (if conditions are met)
+/**
+ * Turn a dissonance-overlapped skill unit back to its original form
+ * @param unit: Unit to transform back
+ */
+void skill_dance_overlap_revert(skill_unit* unit) {
```
Lemongrass3110, line 15766: Please make it a reference.

src/map/skill.cpp
```diff
@@ -15759,12 +15759,33 @@ int32 skill_castend_map (map_session_data *sd, uint16 skill_id, const char *mapn
...
+	// Need to delete previous unit on the client as it can't handle unit_id changes
+	clif_skill_delunit(*unit);
+	// Remove overlap signal
+	unit->val2 &= ~(1 << UF_ENSEMBLE);
+	// If the unit is removed because overlap dissonance killed the caster, we need to reset it here
+	skill_dance_switch(unit, true);
+	skill_getareachar_skillunit_visibilty(unit, AREA);
```
Lemongrass3110, line 15773:
```suggestion
	// Respawn the unit with the original unit_id
	skill_getareachar_skillunit_visibilty(unit, AREA);
```

src/map/skill.cpp
```diff
@@ -15759,12 +15759,33 @@ int32 skill_castend_map (map_session_data *sd, uint16 skill_id, const char *mapn
...
+}
+
+/**
+ * Transforms overlapping skill units into dissonance (if conditions are met)
+ * @param bl: Target skill unit found that is on the same cell as source skill unit
+ * @param ap: func* with va_list values:
+ *  src - source skill unit
+ *  flag - Whether to remove dissonance, add dissonance or just return if the overlap conditions are met (OVERLAP_COUNT)
```
Lemongrass3110, line 15781:
```suggestion
 *  flag - see enum e_dance_overlap
```

src/map/skill.hpp
```diff
@@ -442,6 +442,13 @@ enum e_skill_blown	{
 	BLOWN_TARGET_BASILICA		= 0x20, // If target is in Basilica area
 };
 
+// Enum for skill_dance_overlap*
+enum e_dance_overlap {
```
Lemongrass3110, line 446:
```suggestion
enum e_dance_overlap : int32{
```

src/map/skill.hpp
```diff
@@ -442,6 +442,13 @@ enum e_skill_blown	{
...
+	OVERLAP_REMOVE = 0, // Skill unit is about to be removed, remove overlap marker from overlapping units on the cell if applicable
+	OVERLAP_SET = 1, // Skill unit was placed, add overlap marker to overlapping units on the cell
```
Lemongrass3110, line 448:
```suggestion
	OVERLAP_SET, // Skill unit was placed, add overlap marker to overlapping units on the cell
```

src/map/skill.hpp
```diff
@@ -442,6 +442,13 @@ enum e_skill_blown	{
...
+	OVERLAP_COUNT = 2, // Don't change overlap marker, just count units overlapping with skill unit (excluding)
```
Lemongrass3110, line 449:
```suggestion
	OVERLAP_COUNT, // Don't change overlap marker, just count units overlapping with skill unit (excluding)
```

Lemongrass3110: Excluding what?

src/map/skill.hpp
```diff
@@ -442,6 +442,13 @@ enum e_skill_blown	{
 	BLOWN_TARGET_BASILICA		= 0x20, // If target is in Basilica area
 };
 
+// Enum for skill_dance_overlap*
```
Lemongrass3110, line 445:
```suggestion
// Enum for skill_dance_overlap flag parameter
```

src/map/skill.cpp
```diff
@@ -15812,15 +15831,32 @@ static int32 skill_dance_overlap_sub(struct block_list* bl, va_list ap)
...
-//Does the song/dance overlapping -> dissonance check. [Skotlex]
-//When flag is 0, this unit is about to be removed, cancel the dissonance effect
-//When 1, this unit has been positioned, so start the cancel effect.
-int32 skill_dance_overlap(struct skill_unit* unit, int32 flag)
+/**
+ * Does the song / dance overlapping->dissonance check.
+ * @param unit The unit that is about to be removed or was placed
+ * @param flag Whether unit is about to be removed or was placed
+ *   When flag is OVERLAP_REMOVE, this unit is about to be removed, remove the dissonance effect from applicable units on the cell
+ *   When flag is OVERLAP_SET, this unit has been positioned, start the dissonance effect for applicable units on the cell
```
Lemongrass3110, line 15839:
```suggestion
 * @param flag see enum e_dance_overlap
```

src/map/skill.cpp
```diff
@@ -15812,15 +15831,32 @@ static int32 skill_dance_overlap_sub(struct block_list* bl, va_list ap)
...
+ * @return Number of skill units that overlap with current skill unit and meet the overlap criteria (excluding current skill unit)
```
Lemongrass3110, line 15840: Number is never really used on any call.
Consider making the function void instead.

Lemongrass3110 (approved):

src/map/skill.cpp
Playtester: Can do it here, but I think for the actual overlap check it needs to be a pointer (I tried to make it a reference before but it threw errors, so I reverted to pointer, need to check again).

src/map/skill.hpp
Playtester: If you put (excluding) behind something, it refers to excluding counting that (the word before). In this case, it excludes the skill unit for which the overlapping is checked. So if you have 3 skill units overlapping, it will return 2 because 2 other skill units over overlapping with this.

src/map/skill.hpp
Playtester: We were using the * syntax to clarify it's used for all functions that start with that string. It's also used in other places like that. But I'll just push your suggestion because I have no opinion on it either way.

src/map/skill.cpp
Playtester: I was thinking it might be a useful feature for custom / future implementations. Also otherwise one of the possible flag values would do nothing.

src/map/skill.cpp
Playtester: I wanted to expand on the enum explanation here a little bit so developers can see what the flag means in the context of this function. You think it is sufficient to just link to the enum?

src/map/skill.cpp
Playtester: I wanted to expand on the enum explanation here a little bit so developers can see what the flag means in the context of this function. You think it is sufficient to just link to the enum?

src/map/skill.cpp
```diff
@@ -15825,15 +15845,30 @@ static int32 skill_dance_overlap_sub(struct block_list* bl, va_list ap)
...
-//When flag is 0, this unit is about to be removed, cancel the dissonance effect
-//When 1, this unit has been positioned, so start the cancel effect.
-int32 skill_dance_overlap(struct skill_unit* unit, int32 flag)
+/**
+ * Does the song / dance overlapping->dissonance check.
+ * @param unit The unit that is about to be removed or was placed
+ * @param flag see enum e_dance_overlap
+ * @return Number of skill units that overlap with current skill unit and meet the overlap criteria (excluding current skill unit)
+ */
+int32 skill_dance_overlap(struct skill_unit* unit, e_dance_overlap flag)
```
Daegaladh, line 15854: While you're at it, you could update the unit pointer to a reference

Daegaladh (approved):

src/map/skill.cpp
Playtester: Updated

Issue #9283, opened 2025-05-03
Whirlwind, Deluge and Volcano behavior


Daegaladh:
### rAthena Hash

497cb7cd70f058f24dfedab3d30ef7391974b3c5

### Client Date

2022-04-06

### Server Mode

Renewal, Pre-Renewal

### Result

-These skills require a gemstone each time they're casted, even if they're not consumed
-The effect icon shows up to any character over the skill area, showing the remaining time and re-starts each time you step on it. Also it just updates the real left time on each recast.

### Relevant Log Output

```Shell

```

### Expected Result

-The skills should not require to have any gemstone when they're not gonna be consumed
-The effect icon should show the time left of instead of the effect only to the caster and even if it's not over the area. If you already have any unit placed, it uses the remaining time, they're interchangeable.

### How to Reproduce

-With only 1 gemstone, cast any of the 3 skills, then right after, try to cast any of them with no gemstones, it will fail even if there's no gemstone required yet.
-Step on and off the area and look at the effect timer.

### Official Information

Tested on official servers for both pre-re and renewal.

PR #9288, opened 2025-05-03
Whirlwind, Deluge and Volcano behavior

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9283

* **Server Mode**: both

Overall the behavior of these skills remains the same, only the behavior of the icon indicating the time for free cast is updated

### DIFF
--- a/db/re/status.yml
+++ b/db/re/status.yml
@@ -1125,7 +1125,6 @@ Body:
       Shadowweapon: true
       Ghostweapon: true
   - Status: Volcano
-    Icon: EFST_GROUNDMAGIC
     DurationLookup: SA_VOLCANO
     CalcFlags:
       Batk: true
@@ -1136,7 +1135,6 @@ Body:
       NoClearance: true
       RemoveOnHermode: true
   - Status: Deluge
-    Icon: EFST_GROUNDMAGIC
     DurationLookup: SA_DELUGE
     CalcFlags:
       MaxHp: true
@@ -1144,13 +1142,20 @@ Body:
       NoSave: true
       NoClearance: true
   - Status: Violentgale
-    Icon: EFST_GROUNDMAGIC
     DurationLookup: SA_VIOLENTGALE
     CalcFlags:
       Flee: true
     Flags:
       NoSave: true
       NoClearance: true
+  - Status: Groundmagic
+    Icon: EFST_GROUNDMAGIC
+    Flags:
+      # NoRemoveOnDead: true  # todo
+      NoDispell: true
+      NoBanishingBuster: true
+      NoClearance: true
+      NoSave: true
   - Status: Watk_Element
     DurationLookup: MS_MAGNUM
     Flags:
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -14465,6 +14465,7 @@ int32 skill_castend_pos2(struct block_list* src, int32 x, int32 y, uint16 skill_
 				clif_skill_poseffect( *src, skill_id, skill_lv, x, y, tick );
 	}
 
+	// !TODO : early "return" prevents "skill_onskillusage" from being called, intended?
 	switch(skill_id)
 	{
 	case PR_BENEDICTIO:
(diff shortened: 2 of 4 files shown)


src/map/skill.cpp
```diff
@@ -14512,21 +14513,24 @@ int32 skill_castend_pos2(struct block_list* src, int32 x, int32 y, uint16 skill_
...
+						skill_unitsetting(src,skill_id,skill_lv,x,y,0);
+						return 0; // not to consume items
+					}
+					sg2->limit = 0; // Disable it.
+					break;
 			}
-			else
-				sg2->limit = 0; //Disable it.
 		}
+		sc_start(src, src, SC_GROUNDMAGIC, 100, 0, skill_get_time(skill_id, skill_lv));
```
Playtester, line 14530: This should only happen if skill_unitsetting managed to place at least one unit.

src/map/skill.cpp
```diff
@@ -14512,21 +14513,24 @@ int32 skill_castend_pos2(struct block_list* src, int32 x, int32 y, uint16 skill_
...
-				skill_unitsetting(src,skill_id,skill_lv,x,y,0);
-				return 0; // not to consume items
+		// Does not consumes if the skill is already active. [Skotlex]
+		if (std::shared_ptr<s_skill_unit_group> sg2 = skill_locate_element_field(src); sg2 != nullptr) {
+			switch( sg2->skill_id ) {
+				case SA_VOLCANO:
+				case SA_DELUGE:
+				case SA_VIOLENTGALE:
+					if (sg2->limit - DIFF_TICK(gettick(), sg2->tick) > 0) {
+						skill_unitsetting(src,skill_id,skill_lv,x,y,0);
```
Playtester, line 14523: Do we ensure that the duration stays the same?

Also I think the consume check only applies to PC (but maybe that's already handled somewhere else in the code).

src/map/skill.cpp
```diff
@@ -14512,21 +14513,24 @@ int32 skill_castend_pos2(struct block_list* src, int32 x, int32 y, uint16 skill_
...
-	{	//Does not consumes if the skill is already active. [Skotlex]
-		std::shared_ptr<s_skill_unit_group> sg2;
-		if ((sg2= skill_locate_element_field(src)) != nullptr && ( sg2->skill_id == SA_VOLCANO || sg2->skill_id == SA_DELUGE || sg2->skill_id == SA_VIOLENTGALE ))
-		{
-			if (sg2->limit - DIFF_TICK(gettick(), sg2->tick) > 0)
-			{
-				skill_unitsetting(src,skill_id,skill_lv,x,y,0);
-				return 0; // not to consume items
+		// Does not consumes if the skill is already active. [Skotlex]
+		if (std::shared_ptr<s_skill_unit_group> sg2 = skill_locate_element_field(src); sg2 != nullptr) {
```
Playtester, line 14517: Make sure this properly checks for skill_id and also includes a land protector check.

src/map/skill.cpp
Daegaladh: Not sure what you mean, but the 3 skills have the same duration at all levels (5min)

Playtester: What I mean is that if you have 30 seconds left of those 5 minutes and then you recast, not consuming a gemstone, shouldn't the remaining duration still be 30 seconds?

Daegaladh: Yeah, that's exactly how it should work

src/map/skill.cpp
```diff
@@ -16018,22 +16022,20 @@ std::shared_ptr<s_skill_unit_group> skill_unitsetting(struct block_list *src, ui
...
+		if (std::shared_ptr<s_skill_unit_group> old_sg = skill_locate_element_field(src); old_sg != nullptr) {
+			// HelloKitty confirmed that these are interchangeable, so you can change element and not consume gemstones.
+			if (old_sg->limit > 0) {
+				switch( old_sg->skill_id ) {
+					case SA_VOLCANO:
+					case SA_DELUGE:
+					case SA_VIOLENTGALE:
+						// Use the previous limit (minus the elapsed time) [Skotlex]
+						limit = old_sg->limit - DIFF_TICK(gettick(), old_sg->tick);
+						if (limit < 0)	// This can happen...
```
Lemongrass3110, line 16034: This can really happen on AEGIS?
And you get free casting (without gemstones) in such a case too?

src/map/skill.cpp
```diff
@@ -14512,21 +14513,24 @@ int32 skill_castend_pos2(struct block_list* src, int32 x, int32 y, uint16 skill_
...
+		if (std::shared_ptr<s_skill_unit_group> sg2 = skill_locate_element_field(src); sg2 != nullptr) {
+			switch( sg2->skill_id ) {
+				case SA_VOLCANO:
+				case SA_DELUGE:
+				case SA_VIOLENTGALE:
+					if (sg2->limit - DIFF_TICK(gettick(), sg2->tick) > 0) {
+						skill_unitsetting(src,skill_id,skill_lv,x,y,0);
+						return 0; // not to consume items
+					}
+					sg2->limit = 0; // Disable it.
```
Lemongrass3110, line 14526: As @Playtester commented below, even though you disable the unit here, it would still hop from the break to sc_start.

src/map/skill.cpp
Lemongrass3110: Wouldnt it be better to move this sc_start into skill_unit_effect or another appropriate function?

src/map/skill.cpp
```diff
@@ -14512,21 +14513,24 @@ int32 skill_castend_pos2(struct block_list* src, int32 x, int32 y, uint16 skill_
...
-				return 0; // not to consume items
+		// Does not consumes if the skill is already active. [Skotlex]
+		if (std::shared_ptr<s_skill_unit_group> sg2 = skill_locate_element_field(src); sg2 != nullptr) {
+			switch( sg2->skill_id ) {
+				case SA_VOLCANO:
+				case SA_DELUGE:
+				case SA_VIOLENTGALE:
+					if (sg2->limit - DIFF_TICK(gettick(), sg2->tick) > 0) {
+						skill_unitsetting(src,skill_id,skill_lv,x,y,0);
+						return 0; // not to consume items
```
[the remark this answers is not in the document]
Lemongrass3110: You could also use this instead:
flag |= SKILL_NOCONSUME_REQ;

Lemongrass3110 (commented):
I think it would be better, if we would find a way to de-duplicate the code and only have it in skill_unitsetting.

It would also be nice to check (and maybe remove) the code in skill_cell_overlap for these 3 skills.

src/map/skill.cpp
Playtester: Yeah, this is wrong - on Aegis it counts the intact units and both the consumption and the starting of the status change only happen if the count is 0.

It can technically happen because units are removed every 100ms on rAthena, so the sc timer could trigger before the units are removed. Not sure if it can happen on Aegis since the interval is 20ms for everything there.

I would see this code more of a performance optimization as there's not much point to go through all the unit placement code when the unit just disappear within 100ms again.

db/re/status.yml
```diff
@@ -1125,7 +1125,6 @@ Body:
       Shadowweapon: true
       Ghostweapon: true
   - Status: Volcano
-    Icon: EFST_GROUNDMAGIC
```
Playtester, line 1128: These also have an EFST by the way, but it's called EFST_SPECIALZONE. It includes the element in the val.

PR #9290, opened 2025-05-03
Converted viewdata to array

Lemongrass3110:

* **Server Mode**: Both

Required for a future update.
Refactored clif_get_weapon_view into a view_data function.
Corrected some logic issues in clif_changelook:
LOOK_BASE was only updated for players, but it would also be needed for other bl types, if they use player sprites.
SC checks were only done for players.
Replaced duplicated logic with a clif_npc_mayapurple call.
Fixed LOOK_BODY claiming to not send something, but still sending something.


src/map/unit.hpp
```diff
@@ -83,21 +83,12 @@ struct unit_data {
...
-		hair_style,
-		hair_color,
-		cloth_color,
-		body_style;
+	int32 look[LOOK_MAX];
 	char sex;
 	unsigned dead_sit : 2; // 0: Standing, 1: Dead, 2: Sitting
+
+public:
+	void update( map_session_data& sd, _look look );
```
vstumpf, line 91: this should be a method in map_session_data, then you could just do `sd.update(look)`

src/map/unit.cpp
```diff
@@ -75,6 +75,68 @@ static TIMER_FUNC(unit_attack_timer);
 static TIMER_FUNC(unit_walktoxy_timer);
 int32 unit_unattackable(struct block_list *bl);
 
+void view_data::update( map_session_data& sd, _look look ){
+	int32 val = sd.vd.look[look];
+
+	switch( look ){
+		case LOOK_WEAPON:
```
vstumpf, line 82: These two statements look the same, except for the 2h sword/shield check. Can we just keep them together?

Lemongrass3110: Keep them together?
They were separated before in clif.cpp too.
But we can think about mergint them, if they stay readable.
Do you have a good suggestion for merging the two cases together?

src/map/unit.hpp
Lemongrass3110: It should be like this, but actually it should accept a block_list as parameter later on.
I just left it like this now, as right now only the player specific behavior exists in here.
Other cases than just weapon and shield should probably also be merged into this function later on.
Since the view_data objects dont have a link to their "master" - only in the other direction (so from block_list, map_session_data, etc.) updates are possible.
That is why for me it is a function in the view_data class/struct and not a helper/convenience function in the "master" class.

Or do you have a better idea?

vstumpf: if theres anything that's common across all view_datas, then we can put that in view_data, but for player specific, i think we should put it in `map_session_data::updateLook(_look look)`, and then we can call `vd.update(look)` in it

Lemongrass3110: done.

src/map/unit.cpp
Lemongrass3110: I would prefer to keep it separate (like it was).
With this suggestion we make the code less readable and we have to double check a few ifs just to prevent a few lines.

src/map/pc.cpp
```diff
@@ -620,6 +620,68 @@ void PenaltyDatabase::loadingFinished(){
 
 PenaltyDatabase penalty_db;
 
+void map_session_data::update_look( _look look ){
+	int32 val = this->vd.look[look];
```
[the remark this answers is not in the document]
Lemongrass3110: That check is incompatible with the current (server side) code anyway.
PACKETVER < 4 was something before 2004. So if we get a complaint from someone using such a client (we can look into it, not to mention that I think we will have a lot more problems with such old clients, beside just this).

src/map/pc.cpp
```diff
@@ -620,6 +620,68 @@ void PenaltyDatabase::loadingFinished(){
...
+					}
+				}else{
+					// Nothing equipped
+					val = 0;
+				}
+			}
+			break;
+
+		case LOOK_SHIELD:
+			if( this->sc.option&OPTION_COSTUME ){
```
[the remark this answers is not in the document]
Lemongrass3110: I prefer to keep it separate (as it was).

Playtester (approved):
Not tested and I'd prefer if duplicate code was bundles into a function. But otherwise looks fine after reviewing the code. Approving on the assumption you tested it.

PR #9291, opened 2025-05-03
Improved clif_npc_mayapurple

Lemongrass3110:

* **Server Mode**: Both

Changed block_list to reference.
Speeded up by directly reinterpreting the class, instead of looking it up via map_id2nd.
Added some missing calls to the function.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -4092,7 +4092,7 @@ void clif_changelook(struct block_list *bl, int32 type, int32 val) {
 		target = SELF;
 
 #if PACKETVER < 4
-	if( target != SELF && clif_npc_mayapurple( bl ) ){
+	if( target != SELF && clif_npc_mayapurple( *bl ) ){
 		target = SELF;
 	}
 
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -1657,15 +1657,14 @@ void clif_weather(int16 m)
  * @param bl: Block data
  * @return True if NPC is disabled or false otherwise
  */
-static inline bool clif_npc_mayapurple(block_list *bl) {
-	nullpo_retr(false, bl);
-
-	if (bl->type == BL_NPC) {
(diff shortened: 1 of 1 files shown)


src/map/clif.cpp
```diff
@@ -1657,15 +1657,14 @@ void clif_weather(int16 m)
...
-
-	if (bl->type == BL_NPC) {
-		npc_data *nd = map_id2nd(bl->id);
+static inline bool clif_npc_mayapurple( block_list& bl ){
+	if( bl.type == BL_NPC ){
+		npc_data& nd = reinterpret_cast<npc_data&>( bl );
 
 		// TODO: Confirm if waitingroom cause any special cases
-		if (/* nd->chat_id == 0 && */ nd->is_invisible)
+		if( /* nd->chat_id == 0 && */ nd.is_invisible ){
```
Playtester, line 1665: Added {} here seems pointless.

src/map/clif.cpp
```diff
@@ -4087,8 +4092,10 @@ void clif_changelook(struct block_list *bl, int32 type, int32 val) {
 		target = SELF;
 
 #if PACKETVER < 4
-	if (target != SELF && bl->type == BL_NPC && (((TBL_NPC*)bl)->is_invisible))
+	if( target != SELF && clif_npc_mayapurple( bl ) ){
```
Playtester, line 4095: Wait... isn't bl here a pointer? But you expect a reference?

src/map/clif.cpp
```diff
@@ -5029,8 +5040,9 @@ void clif_getareachar_unit( map_session_data* sd,struct block_list *bl ){
 		return;
 
 	// Hide NPC from Maya Purple card
-	if (clif_npc_mayapurple(bl))
+	if( clif_npc_mayapurple( *bl ) ){
 		return;
+	}
```
Playtester, line 5045: More useless {} (also below)

Atemo (approved):

Issue #9294, opened 2025-05-04
Cannot move/interact when teleport user is open while barter shop is open


Wizi-LoM:
### rAthena Hash

50146e9bf544a1124a064cbd7532ce906597d912

### Client Date

2024-08-22

### Server Mode

Pre-Renewal, Renewal

### Result

- When a user has the Trade Shop open and an administrator intercepts them with "@recall" or the "warp" script, etc., the user is intercepted, but cannot move or interact with any other NPCs. @refresh also doesn't work, so the only solution is to log out.

Here's an example of barter I'm using:
```
  - Name: Testing
    Map: prontera
    Sprite: 4_NECORING
    X: 152
    Y: 180
    Direction: East
    Items:
      - Index: 0
        Item: Falchion_
        RequiredItems:
          - Index: 0
            Item: Sticky_Mucus
            Amount: 5
```

### Relevant Log Output

```Shell
No errors in console, neither crash.
```

### Expected Result

When the user has a barter shop opened and is warped, he should be still able to move, interact with NPCs as usual.

### How to Reproduce

- User A opens a barter shop.
- User B use @recallall (for example, but can be done with script).
- User A is warped, however cannot move, neither talk with any NPC.

I've also noticed this behavior on these repro steps:
- User A open an NPC dialog.
- Then do @refresh
- The user cannot move until do @reload script or relog.

I'm warping the user close to the barter in order to test if I can open it again / talk with any other NPC near.

[Link to video](https://drive.google.com/file/d/1mGLq6IcnSgD8oy16gpMp9jnDK3eW499o/view?usp=sharing)

### Official Information

I do not have this info, but I think it's a bug in rAthena.

PR #9297, opened 2025-05-04
Added some state resets to warping

Lemongrass3110:
* **Addressed Issue(s)**: #9294

* **Server Mode**: Both

Added some state resets to warping

Thanks to @Wizi-LoM


Playtester (approved):
Did not test but I see no concerns with resetting these values.

PR #9298, opened 2025-05-04
Added achievements to inter_athena.conf

Lemongrass3110:

* **Server Mode**: Both

The default configuration is missing in the configuration file

### DIFF
--- a/conf/inter_athena.conf
+++ b/conf/inter_athena.conf
@@ -97,6 +97,7 @@ global_acc_reg_num_table: global_acc_reg_num
 global_acc_reg_str_table: global_acc_reg_str
 
 // Char Database Tables
+achievement_table: achievement
 char_db: char
 hotkey_db: hotkey
 scdata_db: sc_data


conf/inter_athena.conf
```diff
@@ -97,6 +97,7 @@ global_acc_reg_num_table: global_acc_reg_num
 global_acc_reg_str_table: global_acc_reg_str
 
 // Char Database Tables
+achievement_table: achievement
```
Playtester, line 100: I'd prefer if you added this add the end so it matches the order in the source code.

Also we usually tend to look for recent additions at the bottom.

Playtester (approved):
Looks like it matches the default value in the src, but I'd prefer if it was added at the bottom instead of the top, see related comment.

conf/inter_athena.conf
Daegaladh: Actually we should change the order to alphabetical in a future PR for easier look up

Issue #9296, opened 2025-05-04
Barter Shop Packet Structure Changed on April 2, 2025 Causing Display Issues and Client Lag


hiphop9:
### rAthena Hash

50146e9

### Client Date

2025-04-16

### Server Mode

Renewal

### Result

Since the April 2, 2025 KRO client update, the packet structure used to open Barter Shops has changed, causing compatibility issues with servers such as rAthena that use the older format.

When a barter item requires 2 or more exchange items, the client fails to display them properly. In some cases, this also causes the client to lag, freeze, or behave abnormally due to packet parsing errors or memory misalignment.

![Image](https://github.com/user-attachments/assets/bdf3baf4-424e-442d-9469-94054e1d5abb)

### Relevant Log Output

```Shell

```

### Expected Result

When opening a Barter shop with multiple required items, all exchange items should be displayed correctly in the UI, and the player should be able to complete the trade without errors.

### How to Reproduce

1. Launch a client dated April 2, 2025 or later.
2. Connect to an rAthena-based server with an older packet structure.
3. Open a Barter Shop NPC that offers an item in exchange for 2 or more required items.
4. Observe that the client display breaks or shows incorrect item requirements

 

### Official Information

On the official kRO server, when opening a Barter shop with multiple required items, all required items are properly displayed in the exchange list.
This was confirmed using the 2025-04-16 kRO client, where trades involving 2 or more items function correctly without UI glitches.
Attached is a screenshot from the official server for reference.

### Modifications that may affect results

No relevant modifications were made to the barter system or packet handling in our `/src/` files.  
This issue occurs on a clean rAthena hash `50146e9`, with no packet-related changes made on the server side

PR #9303, opened 2025-05-04
Added support for refined barter items

Lemongrass3110:
* **Addressed Issue(s)**: #9296

* **Server Mode**: Both

You can now buy refined items from barters.
This will be visible on clients 2025-04-02 and higher.
On older clients you will still get the item with refine, but you will not see the refine level before buying the item.

Thanks to @hiphop9


src/map/npc.cpp
```diff
@@ -608,6 +608,32 @@ uint64 BarterDatabase::parseBodyNode( const ryml::NodeRef& node ){
...
+				}
+
+				int16 refine;
+
+				if( !this->asInt16( itemNode, "Refine", refine ) ){
+					return 0;
+				}
+
+				if( refine > MAX_REFINE ){
+					this->invalidWarning( itemNode["Amount"], "barter_parseBodyNode: Refine %hd is too high, capping to %d.\n", refine, MAX_REFINE );
```
Atemo, line 626:
```suggestion
					this->invalidWarning( itemNode["Refine"], "barter_parseBodyNode: Refine %hd is too high, capping to %d.\n", refine, MAX_REFINE );
```

src/map/npc.cpp
```diff
@@ -608,6 +608,32 @@ uint64 BarterDatabase::parseBodyNode( const ryml::NodeRef& node ){
...
+					refine = MAX_REFINE;
+				}
+
+				item->refine = (int8)refine;
```
Atemo, line 630: Negative values ​​accepted?

src/map/clif.cpp
```diff
@@ -23299,6 +23299,10 @@ void clif_barter_extended_open( map_session_data& sd, struct npc_data& nd ){
 #if PACKETVER_MAIN_NUM >= 20210203 || PACKETVER_RE_NUM >= 20211103
 		item->viewSprite = id->look;
 		item->location = pc_equippoint_sub( &sd, id.get() );
+
+#if PACKETVER_MAIN_NUM >= 20250402
+		item->refine_level = itemPair.second->refine;
```
Atemo, line 23304: Is it ok for left to be unsigned (uint16) and right to be signed (int8)?

Atemo (approved):

src/map/npc.cpp
Lemongrass3110: I just copied the existing variable.
We could revisit all refine variables in the future.

PR #9311, opened 2025-05-08
Poison React

### LINKED ISSUE #7279: Poison React is officially inaccurate
### rAthena Hash

ba9770ce95e7990f79ba30dde77f52fe2514626e

### Client Date

2018-06-21

### Server Mode

Pre-Renewal, Renewal

### Result

After a conversation I had with a friend and testing the latest rAthena backed up with some possible website evidence (unfortunately it's not kRO) there appears to be several inaccuracies with the Poison React skill that deserve a second look from the rAthena Team. Unfortunately I'm incapable of testing these things myself officially to confirm it, but I wish I could. While I did test and investigate these things from a **pre-renewal context**, these problems could be the same on Renewal too according to the provided sources. In the official information section I included 2 warpportal links which go into detail some of these findings to back up these claims:

1) In a PvP context, Poison React's poison counter part should be activated from the attacker's weapon element, not the attacker's armor element. It's uncertain if this is also true for skill damage. With a mob context, it should work as it currently does where it looks at armor element.
2) Poison React should block the target's damage when the poison counter part activates. Right now you take damage but also inflict counter damage on the target. I have no idea if this is true for the Envenom part too (most likely not).
3) Poison React's poison counter damage appears to be much lower than recalled on officials/aegis. This matters in a PvP context as it was recalled that the skill had the potential to deal a significant amount of damage which is no longer as true. Unfortunately, I have no official evidence to back this up, and whether this also includes the Envenom part's damage. The actual skillmod part appears to be fine, so this could very well be something regarding the skill damage formula(s), or the right hand damage potentially, if proven true.

### Relevant Log Output

```Shell
N/A
```

### Expected Result

Poison React should have its officially intended behaviors.

### How to Reproduce

Test the skill in-game with the current rAthena. In this case I tested in a PvP enabled location.

### Official Information

Some warpportal links which discuss how this skill should be working, at least on an iRO context. While I fully understand that kRO is the default standard, this is at least a variant of official software.

Confirms that it's based on weapon element and that it blocks damage if the poison counter activates.
https://forums.warpportal.com/index.php?/topic/49123-is-poison-react-necessary/
https://forums.warpportal.com/index.php?/topic/164840-poison-react/

N/A

Playtester:
* **Addressed Issue(s)**: #7279

* **Server Mode**: Both (most fixes are for pre-re, but skill works almost identical in both)

- Poison React can now counter normal poison attacks (e.g. Poison Knife, Rusty Arrow, Enchant Poison)
  * Poison armor will no longer trigger this counter, but it still triggers on poison-element monsters
- When Poison React counters an attack, that attack is now blocked
- The counter of Poison React is now a damage buff that increases the damage of the next normal attack
  * It will attempt to trigger a counterattack (makes you move closer if not in range)
  * Adds 30% to skill ratio of the next normal attack per level learned, regardless of level used (for PC)
  * The buff also adds a 50% poison chance to the next normal attack (60s base duration)
- Number of Envenom 5 Autocasts is now 1:1:1:2:2:3:3:4:4:5 (applies to both pre-re and re)
  * Level 1 technically has 0, but it activates at least once before the status change ends
- Envenom 5 Autocasts can now also be triggered by weapon skills
- Duration of Poison React is now capped to 60 seconds
- Fixes #7279

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -4649,12 +4649,15 @@ static int32 battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list
 			skillratio += 200;
 #endif
 		if (!skill_id || skill_id == KN_AUTOCOUNTER) {
-			if (sc->getSCE(SC_POISONREACT) != nullptr && sc->getSCE(SC_POISONREACT)->val4 == 1) {
+			if (status_change_entry* sce = sc->getSCE(SC_POISONREACT); sce != nullptr && sce->val4 == 1) {
 				// Damage boost from poison react (bonus depends on level learned)
 				if (sd != nullptr)
 					skillratio += 30 * pc_checkskill(sd, AS_POISONREACT);
 				else
-					skillratio += 30 * sc->getSCE(SC_POISONREACT)->val1;
+					skillratio += 30 * sce->val1;
+				// This attack has a 50% chance to cause poison
+				// TODO: Effect should be delayed by attack motion
+				sc_start2(src, target, SC_POISON, 50, sce->val1, src->id, skill_get_time2(AS_POISONREACT, sce->val1));
 				status_change_end(src, SC_POISONREACT);
 			}
 			if (sc->getSCE(SC_CRUSHSTRIKE)) {
@@ -8034,6 +8037,17 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 			wd.damage2 += battle_calc_cardfix(BF_WEAPON, src, target, nk, right_element, left_element, wd.damage2, 1, wd.flag);
 	}
 
+	// For some reason the Envenom autocast from Poison React happens at this point rather than after attack motion
+	if (tsc != nullptr && wd.damage + wd.damage2 > 0) {
+		if (status_change_entry* sce = tsc->getSCE(SC_POISONREACT); sce != nullptr && rnd_chance_official(sce->val3, 100)) {
+			// Cast Envenom Level 5
+			if (status_check_skilluse(target, src, TF_POISON, 0))
+				skill_attack(BF_WEAPON, target, target, src, TF_POISON, 5, gettick(), 0);
+			if (--sce->val2 <= 0)
+				status_change_end(target, SC_POISONREACT);
+		}
+	}
+
 	// only do 1 dmg to plant, no need to calculate rest
 	if(infdef){
 		battle_calc_attack_plant(&wd, src, target, skill_id, skill_lv);
@@ -10519,8 +10533,7 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
 			sce->val4 = 1;
 
 			// The target will start attacking instead of the source
-			if (tsd != nullptr)
-				clif_movetoattack(*tsd, *src);
+			unit_attack(target, src->id, 0);
 
 			return ATK_BLOCK;
 		}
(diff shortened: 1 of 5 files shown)


src/map/battle.cpp
```diff
@@ -4659,6 +4649,17 @@ static int32 battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list
...
 		if (!skill_id || skill_id == KN_AUTOCOUNTER) {
+			if (status_change_entry* sce = sc->getSCE(SC_POISONREACT); sce != nullptr && sce->val4 == 1) {
+				// Damage boost from poison react (bonus depends on level learned)
+				if (sd != nullptr)
+					skillratio += 30 * pc_checkskill(sd, AS_POISONREACT);
+				else
+					skillratio += 30 * sce->val1;
+				// This attack has a 50% chance to cause poison
+				// TODO: Effect should be delayed by attack motion
+				sc_start2(src, target, SC_POISON, 50, sce->val1, src->id, skill_get_time2(AS_POISONREACT, sce->val1));
```
Lemongrass3110, line 4660:
```suggestion
				// TODO: Effect should be delayed by attack motion
				sc_start2(src, target, SC_POISON, sce->val3, sce->val1, src->id, skill_get_time2(AS_POISONREACT, sce->val1));
```

src/map/battle.cpp
```diff
@@ -4659,6 +4649,17 @@ static int32 battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list
 			skillratio += 200;
 #endif
 		if (!skill_id || skill_id == KN_AUTOCOUNTER) {
+			if (status_change_entry* sce = sc->getSCE(SC_POISONREACT); sce != nullptr && sce->val4 == 1) {
+				// Damage boost from poison react (bonus depends on level learned)
```
Lemongrass3110, line 4653:
```suggestion
				// Damage boost from poison react (for players bonus depends on level learned)
```

src/map/battle.cpp
```diff
@@ -7360,6 +7358,16 @@ static void battle_calc_weapon_final_atk_modifiers(struct Damage* wd, struct blo
 			status_change_end(target, SC_REJECTSWORD);
 	}
 
+	// Poison React Envenom Level 5 Autocast
+	if (tsc != nullptr && wd->damage > 0) {
```
Lemongrass3110, line 7362: Quite a few tsc checks.
Maybe we should combine all of them in the future.

src/map/battle.cpp
```diff
@@ -7360,6 +7358,16 @@ static void battle_calc_weapon_final_atk_modifiers(struct Damage* wd, struct blo
...
+		if (status_change_entry* sce = tsc->getSCE(SC_POISONREACT); sce != nullptr && rnd_chance_official(sce->val3, 100)) {
+			if (status_check_skilluse(target, src, TF_POISON, 0))
+				skill_attack(BF_WEAPON, target, target, src, TF_POISON, 5, gettick(), 0);
+			if (--sce->val2 <= 0)
```
Lemongrass3110, line 7366: Counter is decreased, even if it did not work?

src/map/battle.cpp
```diff
@@ -10517,6 +10525,19 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
 		}
 	}
 
+	// Poison React counter activates on normal poison attacks as well as attacks from poison-element enemies
```
Lemongrass3110, line 10528:
```suggestion
	// Poison React counter activates on attacks from poison-element enemies as well as normal poison attacks
```
Makes it easier to follow your IF logic, if it is in the same order...

src/map/status.cpp
```diff
@@ -10766,12 +10766,9 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
...
 		case SC_POISONREACT:
-#ifdef RENEWAL
-			val2= (val1 + 1) / 2;
-#else
-			val2=(val1+1)/2 + val1/10; // Number of counters [Skotlex]
-#endif
-			val3=50; // + 5*val1; // Chance to counter. [Skotlex]
+			val2 = val1 / 2; // Number of Envenom autocasts
+			val3 = 50; // Chance to autocast Envenom on hit
+			val4 = 0; // 0: Counter Mode; 1: Damage Boost Mode
```
Lemongrass3110, line 10771: Please also document that in /doc/status_change.txt

Lemongrass3110 (approved):

src/map/battle.cpp
Playtester: I was thinking about this one too, but then we are using it for both the Envenom Autocast chance as well as the poison success chance. We should add that to the val3 description then.

src/map/battle.cpp
Playtester: Yeah I didn't want the PR to affect hundreds of lines. And we eventually need to move these effects into the right order anyway, was thinking of doing that at the same time then.

src/map/battle.cpp
Playtester: Yes, that's correct, I confirmed that one.

PR #9317, opened 2025-05-10
Cloaking / Power Maximize on Monsters

### LINKED ISSUE #9316: Cloaking / Power Maximize on Monsters should not be infinite
### rAthena Hash

1ce2acf

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When a monster uses Cloaking / Power Maximize, it never ends, unless it attacks or casts it again.

### Expected Result

When a monster uses Cloaking / Power Maximize, it ends after 10 seconds.

### How to Reproduce

Equip Maya Purple Card, aggro a Hode, when it uses Cloaking, run out of chase range. Then wait until the Cloaking effect ends.

Make a monster cast Power Maximize and measure when the damage returns to normal.

### Official Information

Verified on official servers by testing.

_No response_

Playtester:
* **Addressed Issue(s)**: #9316 

* **Server Mode**: Both

- Fixed duration of Cloaking / Power Maximize being infinite for monsters
  * The duration is now fixed to 10 seconds
- Fixes #9316

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -10748,8 +10748,14 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
 			clif_emotion( *bl, ET_SWEAT );
 			break;
 		case SC_MAXIMIZEPOWER:
-			tick_time = val2 = tick>0?tick:60000;
-			tick = INFINITE_TICK; // Duration sent to the client should be infinite
+			if (sd == nullptr) {
+				// Duration is 10 seconds for monsters on all levels
+				tick = 10000;
+			}
+			else {
+				tick_time = val2 = tick>0?tick:10000;
+				tick = INFINITE_TICK; // Duration sent to the client should be infinite
+			}
 			break;
 		case SC_EDP:
 			val2 = (val1 + 1) / 2 + 2; // Chance to Poison enemies.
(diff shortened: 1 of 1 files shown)


src/map/status.cpp
```diff
@@ -11139,10 +11145,15 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
 			if (map_flag_gvg2(bl->m) || map_getmapflag(bl->m, MF_BATTLEGROUND)) val4 *= 5;
 			break;
 		case SC_CLOAKING:
-			if (!sd) // Monsters should be able to walk with no penalties. [Skotlex]
+			if (sd == nullptr) {
+				// Monsters have no walk penalties and the status change lasts for 10 seconds
```
Lemongrass3110, line 11149:
```suggestion
				// Non-players have no walk penalties and the status change lasts for 10 seconds
```

src/map/status.cpp
```diff
@@ -11139,10 +11145,15 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
...
 				val1 = 10;
-			tick_time = val2 = tick>0?tick:60000; // SP consumption rate.
-			tick = INFINITE_TICK; // Duration sent to the client should be infinite
+				tick = 10000;
+			}
+			else {
+				tick_time = val2 = tick>0?tick:10000; // SP consumption rate.
```
Lemongrass3110, line 11154:
```suggestion
				tick_time = val2 = tick>0?tick:10000; // SP consumption intervall.
```

Lemongrass3110 (approved):
Would be nice to invert the two ifs and not have the "edge" case first.

src/map/status.cpp
Playtester: Actually good point, I should check how other types behave here.

Playtester:
> Would be nice to invert the two ifs and not have the "edge" case first.

Done!

Also I changed it so that all units that can consume SP use the normal skill behavior, rather than using the 10 seconds version for all non-PC types.

Mercenaries, Homunculus and Elementals don't use these skills officially, but it's cleaner for customization like this.

Lemongrass3110:
Even better now :)

PR #9337, opened 2025-05-17
Add support for new macro user report packet

Haw-Krwy:
The file was leaked, but I'm not sure who released it since I've had it in my possession for a long time. I would like to give credit to @awakenedsec
* **Require Exe**: 20230915 or later

* **Addressed Issue(s)**: https://github.com/rathena/rathena/pull/9195

* **Server Mode**: Both systems

* **Description of Pull Request**:

PR #9195, opened 2025-03-24
Added dummy parsing for 0xbe2

Lemongrass3110:
* **Addressed Issue(s)**: #9171

* **Server Mode**: Both

* **Description of Pull Request**: 
Added dummy parsing for 0xbe2

Thanks to @brynner23

PR #9345, opened 2025-05-18
Add support for new macro user report packet (re  follow up #9337)

Vibe-Code-Nullptr:
The file was leaked, but I'm not sure who released it since I've had it in my possession for a long time. I would like to give credit to @awakenedsec & @Asheraf
* **Require Exe**: 20230915 or later

* **Addressed Issue(s)**: https://github.com/rathena/rathena/pull/9195

* **Server Mode**: Both systems

**Sorry for having to reopen it because the original is not master clean.**


Haydrich:
Isnt this also @Asheraf?

Haw-Krwy:
> Isnt this also belong to @Asheraf?

Here's the thing, when I got this diff file, @awakenedsec happened to be the distributor of exe 2023, and someone uploaded those files to the Warp Discord. So, I don't really know who the true owner is. I only credited him because it was just a matter of timing.

And if you disagree and find it inappropriate, I will terminate this PR.

PR #9352, opened 2025-05-23
Defender on Devotee Equip Requirement

### LINKED ISSUE #9351: Defender on Devotee Equip Requirement
### rAthena Hash

10bbf25

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When the devotee changes equip and has no shield, Defender is removed from him.

### Expected Result

The devotee does not require a shield at all, he can freely change equip and Defender status change remains.

However, when the Crusader unequips their shield, it ends on both the Crusader and the devotee.

### How to Reproduce

Cast Devotion, cast Defender, check status icons.
Have devotee unequip their shield, check status icons.
Have Crusader unequip their shield, check status icons.

### Official Information

Was tested by a friend on official servers, he confirmed all of the above. 
Also a second person confirmed that they remember devotees not requiring a shield on official servers.

_No response_

Playtester:
* **Addressed Issue(s)**: #9351 

* **Server Mode**: Both

- Devotees no longer lose the transferred Defender buff when they change equipment and don't have a shield
  * All Crusader-types still require a shield for Defender
  * Devotees still lose Defender when the devoting Crusader-type unequips their shield
- Fixes #9351

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -7363,6 +7363,9 @@ static void pc_checkallowskill(map_session_data *sd)
 		}
 
 		if (flag[SCF_REQUIRESHIELD]) { // Skills requiring a shield
+			// Only crusader-type classes require a shield for Defender
+			if (status == SC_DEFENDER && (sd->class_&MAPID_UPPERMASK) != MAPID_CRUSADER)
+				continue;
 			if (sd->sc.getSCE(status) && sd->status.shield <= 0)
 				status_change_end(&sd->bl, status);
 		}


Daegaladh (approved):

src/map/pc.cpp
```diff
@@ -7363,6 +7363,9 @@ static void pc_checkallowskill(map_session_data *sd)
 		}
 
 		if (flag[SCF_REQUIRESHIELD]) { // Skills requiring a shield
+			// Only crusader-type classes require a shield for Defender
+			if (status == SC_DEFENDER && (sd->class_&MAPID_UPPERMASK) != MAPID_CRUSADER)
+				continue;
```
Atemo, line 7368: It's a bit off topic, but what about other status like autoguard?

Playtester: Auto Guard, Reflect Shield and Endure wouldn't transfer at all through Devotion. They are currently coded incorrectly.

PR #9353, opened 2025-05-23
Charge Attack Knockback Immunity and Distance Calculation

Playtester:
* **Addressed Issue(s)**: -

* **Server Mode**: Both (distance fix only applies to pre-re)

- Fixed Charge Attack moving you incorrectly and causing position lag when you're knockback immune
  * This for example happened when you equipped an RSX-0806 card
  * Removed direction hack that's no longer needed with after this update
- Fixed distance calculation (used for damage, cast time and knockback determination in pre-re)

Hint: The issue was that we used "skill_blown" to move the caster "one cell back" to the right target cell instead of just calculating the correct target coordinates directly.

### DIFF
--- a/src/map/path.hpp
+++ b/src/map/path.hpp
@@ -49,6 +49,10 @@ struct shootpath_data {
 #define check_distance_client_blxy(bl, x1, y1, distance) check_distance_client((bl)->x-(x1), (bl)->y-(y1), distance)
 #define check_distance_client_xy(x0, y0, x1, y1, distance) check_distance_client((x0)-(x1), (y0)-(y1), distance)
 
+#define distance_math_bl(bl1, bl2) distance_math((bl1)->x - (bl2)->x, (bl1)->y - (bl2)->y)
+#define distance_math_blxy(bl, x1, y1) distance_math((bl)->x-(x1), (bl)->y-(y1))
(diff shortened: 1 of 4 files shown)


src/map/path.hpp
```diff
@@ -49,6 +49,10 @@ struct shootpath_data {
 #define check_distance_client_blxy(bl, x1, y1, distance) check_distance_client((bl)->x-(x1), (bl)->y-(y1), distance)
 #define check_distance_client_xy(x0, y0, x1, y1, distance) check_distance_client((x0)-(x1), (y0)-(y1), distance)
 
+#define distance_math_bl(bl1, bl2) distance_math((bl1)->x - (bl2)->x, (bl1)->y - (bl2)->y)
+#define distance_math_blxy(bl, x1, y1) distance_math((bl)->x-(x1), (bl)->y-(y1))
+#define distance_math_xy(x0, y0, x1, y1) distance_math((x0)-(x1), (y0)-(y1))
```
Atemo, line 54: `distance_math_blxy` and `distance_math_xy` are unused?

Atemo (approved):

src/map/path.hpp
Playtester: Yes, I just wanted to make it consistent with the other macros, in case someone wants to use them later.

PR #9362, opened 2025-05-27
Derive map units from block_list

vstumpf:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

All of these classes are block_lists that we reinterpret cast. Let's make them derived from block_list, and then we can use static_cast instead of reinterpret_cast to downcast.

Note: I'm not using dynamic_cast, as we don't want to use RTTI, we already have the type of the block_list set in every one.


StiveRoqer:
In my opinion, it shouldn't be changed to be this complicated. Not changing it doesn't cause any issues, but changing it affects other parts of the code even more. The original implementation using struct block_list X and public: already works just fine. I don't really understand whether these structural changes in the server are based on the official server or just made according to personal preferences or opinions.

Issue #7980, opened 2023-10-24
Slavemobs, homunculus, etc. share master's guild emblem


eppc0330:
### rAthena Hash

master

### Client Date

20221207

### Server Mode

Renewal

### Result

![emblem01](https://github.com/rathena/rathena/assets/47050704/c8a5c666-c010-41dc-85e1-6f7c70db904b)
![emblem02](https://github.com/rathena/rathena/assets/47050704/d53eb611-b080-45f9-b938-e039e1f81b1e)

Summoned creatures(homunculus, slavemobs, etc.) share master's guild emblem.
It is not official behavior

### Expected Result

Shouldn't share guild emblem

### How to Reproduce

Apply guild emblem

### Official Information

Not share guild emblem

aleos89:
Was this a change with the latest clients? I don't recall these types having the guild emblems before by default.

eppc0330:
@aleos89 I've seen this issue since #5731. I have commented about this in #5731 but nobody checks it.
Then I just forgot about this.

aleos89:
Gotcha, so it's a change with the new emblem stuff. I haven't played around with the webservice stuff since it's not needed for debugging. I'll see what I can find.

Kuyaraikub:
Works fine on packet version: 20220406.

https://github.com/rathena/rathena/assets/147054311/421d1133-b391-433e-9903-c358aeff6e44

eppc0330:
@Kuyaraikub I'm not sure you are understanding well. This issue is talking that pet should not share master's guild emblem.
It is not official behavior.

Pokye:
I don't think it's a problem with the client, as I tested the 2022/2023 and 2024 versions and they all continue to show the emblem.
It would be good to add an option in conf/ whether or not to show the emblem (although this is not official)

Haydrich:
> May I know in details about mercenary and others how is it behaves on official ? or its not showing emblems at all ?

It's not showing the emblems at all

PR #9363, opened 2025-05-27
Add configuration for Pets, Mercenaries, Elementals & Homunculi

reunite-ro:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/7980

* **Server Mode**: Both

* **Description of Pull Request**: Add configuration for companion (pets,mercenaries,elementals & homunculi) at conf/battle/client.conf

### DIFF
--- a/conf/battle/client.conf
+++ b/conf/battle/client.conf
@@ -174,6 +174,6 @@ macro_detection_punishment_time: 0
 // Set to 0 to disable
 macrochecker_delay: 600000
 
-// Should companion pets/elementals/homunculi/mercenaries display the guild emblem of their master? (Note 3)
-// Default: 0 (none)
-companion_show_guild_emblem: 0
+// Should pets/homunculi/mercenaries/elementals display the guild emblem? (Note 3)
+// Default: 0
+show_guild_emblem: 0
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -11950,7 +11950,7 @@ static const struct _battle_data {
 	{ "client_limit_unit_lv",               &battle_config.client_limit_unit_lv,            0,      0,      BL_ALL,         },
 	{ "land_protector_behavior",            &battle_config.land_protector_behavior,         0,      0,      1,              },
 	{ "npc_emotion_behavior",               &battle_config.npc_emotion_behavior,            0,      0,      1,              },
-	{ "companion_show_guild_emblem",        &battle_config.companion_show_guild_emblem, BL_NUL,BL_NUL,      BL_ALL,         },
+	{ "show_guild_emblem",                  &battle_config.show_guild_emblem,           	BL_NUL, BL_NUL, BL_ALL,         },
 // BattleGround Settings
 	{ "bg_update_interval",                 &battle_config.bg_update_interval,              1000,   100,    INT_MAX,        },
 	{ "bg_short_attack_damage_rate",        &battle_config.bg_short_damage_rate,            80,     0,      INT_MAX,        },
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -1073,6 +1073,72 @@ static int32 clif_setlevel(struct block_list* bl) {
 }
 
 /*==========================================
+ * Gets guild ID for emblem display purposes only
+ *------------------------------------------*/
+static int32 clif_get_guild_id_for_emblem(struct block_list *bl) {
+	nullpo_ret(bl);
+	switch (bl->type) {
+		case BL_PC:
+			if (battle_config.show_guild_emblem & BL_PC)
+				return ((TBL_PC*)bl)->status.guild_id;
+			break;
+		case BL_PET:
+			if ((battle_config.show_guild_emblem & BL_PET) && ((TBL_PET*)bl)->master)
+				return ((TBL_PET*)bl)->master->status.guild_id;
+			break;
+		case BL_HOM:
+			if ((battle_config.show_guild_emblem & BL_HOM) && ((TBL_HOM*)bl)->master)
+				return ((TBL_HOM*)bl)->master->status.guild_id;
+			break;
+		case BL_MER:
+			if ((battle_config.show_guild_emblem & BL_MER) && ((TBL_MER*)bl)->master)
+				return ((TBL_MER*)bl)->master->status.guild_id;
+			break;
+		case BL_ELEM:
+			if ((battle_config.show_guild_emblem & BL_ELEM) && ((TBL_ELEM*)bl)->master)
+				return ((TBL_ELEM*)bl)->master->status.guild_id;
+			break;
+		default:
+			// For MOB, NPC, SKILL etc., use the original function
+			return status_get_guild_id(bl);
+	}
+	return 0;
+}
+
+/*==========================================
+ * Gets emblem ID for emblem display purposes only
+ *------------------------------------------*/
+static int32 clif_get_emblem_id_for_emblem(struct block_list *bl) {
+	nullpo_ret(bl);
+	switch (bl->type) {
+		case BL_PC:
+			if (battle_config.show_guild_emblem & BL_PC)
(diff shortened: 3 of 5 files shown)


reunite-ro:
> It just an opinion, I would like if you added a value for each companion, so i can apply it for Pet, or Homun or both without the Elementals.

understood, ill check the bitmask and update soon

src/map/battle.cpp
```diff
@@ -11950,7 +11950,7 @@ static const struct _battle_data {
 	{ "client_limit_unit_lv",               &battle_config.client_limit_unit_lv,            0,      0,      BL_ALL,         },
 	{ "land_protector_behavior",            &battle_config.land_protector_behavior,         0,      0,      1,              },
 	{ "npc_emotion_behavior",               &battle_config.npc_emotion_behavior,            0,      0,      1,              },
-	{ "companion_show_guild_emblem",        &battle_config.companion_show_guild_emblem,     1,      0,      1,              },
+	{ "companion_show_guild_emblem",        &battle_config.companion_show_guild_emblem,     0,     0,      540,            },
```
Playtester, line 11953: Please use the corresponding BL constants. I'd just put BL_ALL even though not all apply.

conf/battle/client.conf
```diff
@@ -174,6 +174,12 @@ macro_detection_punishment_time: 0
...
 macrochecker_delay: 600000
 
-// Should companion pets/elementals/homunculi/mercenaries display the guild emblem of their master? (Note 1)
-// Default: no
-companion_show_guild_emblem: no
\ No newline at end of file
+// Should companion pets/elementals/homunculi/mercenaries display the guild emblem of their master? (Note 3)
+// Uses unit types bit field values:
+// 0: None - Hide guild emblems for all companions
+// 4: Pets
```
Playtester, line 180: What each number means is already covered by says (Note 3).
It would be sufficient to just list to which types it applies in a single line.

reunite-ro: but elemental isnt covered, so should we add there too ?

// Note 3: Value is a bit field. If no description is given,
//         assume unit types (1: Pc, 2: Mob, 4: Pet, 8: Homun, 16: Mercenary)

src/map/battle.cpp
```diff
@@ -11950,6 +11950,7 @@ static const struct _battle_data {
 	{ "client_limit_unit_lv",               &battle_config.client_limit_unit_lv,            0,      0,      BL_ALL,         },
 	{ "land_protector_behavior",            &battle_config.land_protector_behavior,         0,      0,      1,              },
 	{ "npc_emotion_behavior",               &battle_config.npc_emotion_behavior,            0,      0,      1,              },
+	{ "companion_show_guild_emblem",        &battle_config.companion_show_guild_emblem,     0,     0,      540,            },
```
Lemongrass3110, line 11953:
```suggestion
	{ "companion_show_guild_emblem",        &battle_config.companion_show_guild_emblem, BL_NUL,BL_NUL,      BL_ALL,         },
```

conf/battle/client.conf
```diff
@@ -173,3 +173,13 @@ macro_detection_punishment_time: 0
...
+
+// Should companion pets/elementals/homunculi/mercenaries display the guild emblem of their master? (Note 3)
+// Uses unit types bit field values:
+// 0: None - Hide guild emblems for all companions
+// 4: Pets
+// 8: Homunculus
+// 16: Mercenary
+// 512: Elemental
+// Default: 0
+companion_show_guild_emblem: 0
```
Lemongrass3110, line 185:
```suggestion
companion_show_guild_emblem: 0

```
Missing empty newline at the end of the file.

conf/battle/client.conf
Lemongrass3110: Feel free to add elemental (512 is correct) at the top of the file, but please remove the list here.

src/map/status.cpp
```diff
@@ -9133,7 +9133,7 @@ int32 status_get_guild_id(struct block_list *bl)
 		case BL_PC:
 			return ((TBL_PC*)bl)->status.guild_id;
 		case BL_PET:
-			if (battle_config.companion_show_guild_emblem && ((TBL_PET*)bl)->master)
+			if ((battle_config.companion_show_guild_emblem & 4) && ((TBL_PET*)bl)->master)
```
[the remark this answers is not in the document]
Lemongrass3110: Attention! This function is not only for the emblem!
It is also used for other purposes, such as enemy calculation.
If you only want to show the emblem, this has to be done only at the 4 calls to this function in clif.cpp, but not here in this function for all purposes.

Lemongrass3110 (changes requested):

conf/battle/client.conf
```diff
@@ -173,3 +173,7 @@ macro_detection_punishment_time: 0
 // Macrochecker delay (per map)
 // Set to 0 to disable
 macrochecker_delay: 600000
+
+// Should companion pets/elementals/homunculi/mercenaries display the guild emblem of their master? (Note 3)
+// Default: 0 (none)
+companion_show_guild_emblem: 0
```
Playtester, line 179: Here is actually another thought... why don't we call this option just "show_guild_emblem" and set the default to 1. Then users could also configure if the emblem should be shown on the player. Or is it not possible to not show it on PC?

reunite-ro: > Here is actually another thought... why don't we call this option just "show_guild_emblem" and set the default to 1. Then users could also configure if the emblem should be shown on the player. Or is it not possible to not show it on PC?

in current status.cpp there isnt bl)->master at case BL_PC

src/map/clif.cpp
```diff
@@ -1072,6 +1072,72 @@ static int32 clif_setlevel(struct block_list* bl) {
...
+			// For MOB, NPC, SKILL etc., use the original function
+			return status_get_guild_id(bl);
+	}
+	return 0;
+}
+
+/*==========================================
+ * Gets emblem ID for emblem display purposes only
+ *------------------------------------------*/
+static int32 clif_get_emblem_id_for_emblem(struct block_list *bl) {
```
Playtester, line 1111: Is this needed? Do you need to get the emblem ID for non-emblem purposes?

src/map/clif.cpp
```diff
@@ -1072,6 +1072,72 @@ static int32 clif_setlevel(struct block_list* bl) {
 	return lv;
 }
 
+/*==========================================
+ * Gets guild ID for emblem display purposes only
+ *------------------------------------------*/
+static int32 clif_get_guild_id_for_emblem(struct block_list *bl) {
```
Playtester, line 1078: Much better now. But it looks to me that whenever the guild_id is required for an emblem, the emblem_id is also required.

Why not make it a function that sets both using references?

Something like:
```cpp
static void clif_get_emblem_data(struct block_list *bl, int32& guild_id, int32& emblem_id) {
```
And then you can set it for example in the default case like this:
```cpp
guild_id = status_get_guild_id(bl);
emblem_id = status_get_emblem_id(bl);
```

Also please make sure to use the modern function comment style that explains all parameters and return value (if exists).
```cpp
/**
 * Description
 * @param bl [...]
 * @param guild_id [...]
 * @param emblem_id [...]
 */
```

src/map/battle.cpp
```diff
@@ -11950,6 +11950,7 @@ static const struct _battle_data {
 	{ "client_limit_unit_lv",               &battle_config.client_limit_unit_lv,            0,      0,      BL_ALL,         },
 	{ "land_protector_behavior",            &battle_config.land_protector_behavior,         0,      0,      1,              },
 	{ "npc_emotion_behavior",               &battle_config.npc_emotion_behavior,            0,      0,      1,              },
+	{ "show_guild_emblem",                  &battle_config.show_guild_emblem,           	BL_NUL, BL_NUL, BL_ALL,         },
```
Playtester, line 11953: Default here would be BL_PC.

PR #9366, opened 2025-05-27
Added a hasSCE function

Lemongrass3110:

* **Server Mode**: Both

This prevents to always call getSCE and have to check it for nullptr.
It should be used when it is only necessary to know, if a BL has a status active or not.
If you want to access values of a SCE you should still use getSCE and store the returned pointer.


Lemongrass3110:
> I really like this one, but still need to adjust all the checks. We just do that one-by-one?

Whenever we touch something or when somebody is motivated to go over a file.

PR #9371, opened 2025-05-28
Meteor Timings (Wizard + Summoner)

### LINKED ISSUE #9370: Meteor Timings (Wizard + Summoner)
### rAthena Hash

bcba190

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

Meteors deal damage 1000ms after they start dropping.

Summoner meteors drop every 500ms.

### Expected Result

Meteors deal damage 700ms after they start dropping.

Summoner meteors drop every 700ms.

### How to Reproduce

Cast Meteor Storm / Summoner Meteor (forgot name). Packet capture makes timings clear.

### Official Information

Verified through packet captures.

_No response_

Playtester:
* **Addressed Issue(s)**: #9370 

* **Server Mode**: Both

- All meteors now deal damage exactly 700ms after they appeared
- Catnip meteors now appear every 700ms instead of every 500ms
- Meteor Storm Level 11 now summons 10 meteors in a 33x33 area
- Fixes #9370

### DIFF
--- a/db/pre-re/skill_db.yml
+++ b/db/pre-re/skill_db.yml
@@ -3067,7 +3067,7 @@ Body:
       - Level: 10
         Area: 3
       - Level: 11
-        Area: 14
+        Area: 16
     CopyFlags:
       Skill:
         Plagiarism: true
@@ -3115,6 +3115,8 @@ Body:
         Time: 6000
       - Level: 10
         Time: 7000
+      - Level: 11
+        Time: 10000
     Duration2: 5000
     Requires:
       SpCost:
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -3130,7 +3130,7 @@ Body:
       - Level: 10
         Area: 3
       - Level: 11
-        Area: 14
+        Area: 16
     CopyFlags:
       Skill:
         Plagiarism: true
@@ -3178,6 +3178,8 @@ Body:
         Time: 6000
       - Level: 10
         Time: 7000
+      - Level: 11
+        Time: 10000
     Duration2: 4500
     Cooldown: 7000
     FixedCastTime:
--- a/doc/mob_skill_db_powerskill.txt
+++ b/doc/mob_skill_db_powerskill.txt
@@ -31,7 +31,7 @@ Skill | rAthena Lv | Explanation
 57,KN_BRANDISHSPEAR   |10| 562% damage at all ranges.
 60,KN_TWOHANDQUICKEN  |30| Increases ASPD by 70%. Duration 300 seconds.
 80,WZ_FIREPILLAR      |10| Damage multiplied by number of hits.
-83,WZ_METEOR          |11| Meteors drop in 29x29 AoE. 15 hits per meteor.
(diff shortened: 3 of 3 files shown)


src/map/skill.cpp
```diff
@@ -22593,6 +22593,10 @@ static int32 skill_unit_timer_sub(DBKey key, DBData *data, va_list ap)
 							else
 								clif_skill_poseffect( *src, group->skill_id, group->skill_lv, bl->x, bl->y, tick );
 							group->val2 = 1;
+
+							// All meteors hit 700ms after the effect starts, regardless of their interval
+							if (group->skill_id == WZ_METEOR || group->skill_id == SU_CN_METEOR || group->skill_id == SU_CN_METEOR2)
+								group->limit = group->limit - group->interval + 700;
```
Lemongrass3110, line 22599:
```suggestion
								group->limit -= group->interval + 700;
```

src/map/skill.cpp
```diff
@@ -22593,6 +22593,10 @@ static int32 skill_unit_timer_sub(DBKey key, DBData *data, va_list ap)
 							else
 								clif_skill_poseffect( *src, group->skill_id, group->skill_lv, bl->x, bl->y, tick );
 							group->val2 = 1;
+
+							// All meteors hit 700ms after the effect starts, regardless of their interval
+							if (group->skill_id == WZ_METEOR || group->skill_id == SU_CN_METEOR || group->skill_id == SU_CN_METEOR2)
```
Atemo, line 22598: Is it relevant to include SU_CN_METEOR/SU_CN_METEOR2 since you changed their interval to 700 ms ?

Playtester: I wanted to make sure it still works if someone wants to customize the interval value.

src/map/skill.cpp
Playtester: I don't think that works. Wouldn't that substract 1700 instead of 300 (assuming interval is 1000).

Daegaladh:
```suggestion
								group->limit += 700 - group->interval;
```
I think this would be the correct optimization here

Playtester: Is it just me? I didn't want to use += or -= here because I found it hard to understand. Wanted to make it clear that we substract the interval and then add the 700ms (time for the meteor to hit the ground).

Daegaladh: I don't really care, as the compiler optimizes it anyway, I just corrected the math, so chose whatever you think it's easier to read

Lemongrass3110: For me it is hard to understand in general, why interval is substracted here and not set already when the unit is being placed (most likely in skill_unitsetting).

But I must say that whole default block, that is not actually a default block, but rather a 99% specific code block, makes me want to change much more here.

Playtester: The reason for that is that we made the implementation quite dynamic by calculating the number of meteors by calculating duration/interval.

Officially the units placed just last `(interval * count) + 700` and 700ms before expiration the animation starts and then when the unit expires, it deals damage.

Since we want users to be able to freely adjust the duration and interval for more or less meteors, we coded it so that it will work without code changes.

Issue #1141, opened 2016-04-06
Frozen / Stone and Multiple hits


Paoly28:
if attack is fast enough,  the 1st hit  and succeeding hits will behave as if the target is still water or earth, should not be the case as only the 1st hit should receive the status changes. can easily replicate with Overbrand, the 2 multi swings will deal 1.75x in which the second swing should only deal 1x another case would be blastmine and frozen in which the succeeding 4~5 mines would interact as if the 1st hit didn't break the status.

Playtester:
Very hard to fix. The problem is that currently we break Freeze/Stone when the damage is applied. We need to do this because we have to remember the target was frozen/stoned because in that case it should not be possible re-freeze or re-stone the target.

Official servers do that quite differently. Just like damage, whether a status change happens or not is calculated already before the damage delay. Nothing that happens in between is able to influence that anymore. And the actual status change break happens immediately after the damage was calculated, so that it's impossible for a second attack to benefit from it.

To implement that properly, we'd need to implement the official status change delay timer system where each status change does not only have a duration but also a pre-delay (which is btw substracted from the duration).

PR #9373, opened 2025-05-29
Status Change Start Delay Timer

Playtester:
* **Addressed Issue(s)**: Related to various issues reported (doesn't fix any, but opens the way to fix them)

For example related to: #1141 

* **Server Mode**: Both

- It is now possible to delay the start of any status change
  * Split status_change_start into status_change_start and status_change_start_post_delay
  * status_change_start handles everything that needs to happen before the delay
  * status_change_start_post_delay handles everything that needs to happen after the delay
  * Added struct s_delay_status to pass all status change data to a timer
  * Added status_change_start_timer to handle the delay
  * OPT1 and OPT2 status changes have most their conditions checked before the delay
  * If the target dies during the delay, the status change is not applied
  * Bleeding always ignores the delay
  * In renewal the delay is subtracted from the total duration
- Code improvements

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -38,8 +38,6 @@
 
 using namespace rathena;
 
-static struct eri* delay_status_ers; // For delayed status structures
-
 // Regen related flags.
 enum e_regen {
 	RGN_NONE = 0x00,
@@ -50,6 +48,23 @@ enum e_regen {
 };
 
 static struct status_data dummy_status;
+
+/// Delayed Status Structure
+struct s_delay_status {
+	int32 src_id;
+	int32 bl_id;
+	sc_type type;
+	int32 val1;
+	int32 val2;
+	int32 val3;
+	int32 val4;
+	int32 tick;
+	uint8 flag;
+};
+
+int32 delay_status_index = 0;
+// delay_status : delay_status_index -> data
+std::unordered_map<uint32, std::shared_ptr<s_delay_status>> delay_status;
 
 int16 current_equip_item_index; /// Contains inventory index of an equipped item. To pass it into the EQUP_SCRIPT [Lupus]
 uint32 current_equip_combo_pos; /// For combo items we need to save the position of all involved items here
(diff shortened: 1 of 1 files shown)


src/map/status.cpp
```diff
@@ -10173,28 +10212,81 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
...
 		case SC_STONE:
 		case SC_STONEWAIT:
 		case SC_FREEZE:
 			// Undead are immune to Freeze/Stone
-			if (undead_flag && !(flag&SCSTART_NOAVOID))
+			if (battle_check_undead(status->race, status->def_ele) != 0 && !(flag&SCSTART_NOAVOID))
 				return 0;
+			else if (type == SC_STONEWAIT) {
+				// Stonewait has a unique handling where the delay is actually duration until stone kicks in
+				val3 = max(1, tick - delay); // Petrify time
```
Daegaladh, line 10225:
```suggestion
				val3 = std::max<int32>(1, tick - delay); // Petrify time
```
Probably would need to change it to t_tick eventually but in the meantime...

src/map/status.cpp
```diff
@@ -10173,28 +10212,81 @@ int32 status_change_start(struct block_list* src, struct block_list* bl,enum sc_
...
+	status_data* status = status_get_status_data(*bl);
+	int32 undead_flag = battle_check_undead(status->race,status->def_ele);
+	int32 tick_time = 0;
+	bool sc_isnew = true;
+	std::shared_ptr<s_status_change_db> scdb = status_db.find(type);
+
+	// Check for immunities / sc fails
+	switch (type) {
+		case SC_VACUUM_EXTREME:
+			if (sc && sc->getSCE(SC_VACUUM_EXTREME_POSTDELAY) && sc->getSCE(SC_VACUUM_EXTREME_POSTDELAY)->val2 == val2) // Ignore post delay from other vacuum (this will make stack effect enabled)
```
Daegaladh, line 10287:
```suggestion
			if (sc != nullptr && sc->hasSCE(SC_VACUUM_EXTREME_POSTDELAY) && sc->getSCE(SC_VACUUM_EXTREME_POSTDELAY)->val2 == val2) // Ignore post delay from other vacuum (this will make stack effect enabled)
```

src/map/status.cpp
```diff
@@ -10071,9 +10073,52 @@ void status_display_remove(struct block_list *bl, enum sc_type type) {
...
+
+/**
+ * Timer for delayed status changes
+ * Triggers at the end of the delay
+ * Checks if target is still valid and then proceeds with applying the status change
+ * @param data delay_status struct that contains all required status data
+ * @return 0
+ */
+TIMER_FUNC(status_change_start_timer) {
+	struct delay_status* dat = reinterpret_cast<delay_status*>(data);
```
Daegaladh, line 10099:
```suggestion
	delay_status* dat = reinterpret_cast<delay_status*>(data);
```
Just a nitpick

Daegaladh (approved):

PR #9378, opened 2025-05-31
Randomtarget Monster Mode Behavior

### LINKED ISSUE #8104: Random Target Mode - Mobs Casting Buffs on Players
### rAthena Hash

673ef066fb261985b7be4526b444e8b735e7ed95

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

Unexpected behavior with mobs mode "RandomTarget: true". Instead of casting buffs on themselves, mobs are casting buffs on players. 

### Relevant Log Output

_No response_

### Expected Result

Mobs in "Random Target" mode should not cast buffs on players but should use their skills for offensive skill or self/friendly support.

### How to Reproduce

Add the following modes to 1769, 1770, 1918 mob IDs:
```
Modes:
  RandomTarget: true
```
`@monster 1769 / 1770 / 1918`

For mob ID 1796, reduce its HP to 50% to trigger AL_HEAL. The MG_SAFETYWALL is also incorrectly cast on a player.

For mob ID 1770, wait until you acquire BS_MAXIMIZE.

For mob ID 1918, reduce its HP to 80% and wait until the mob casts NPC_GRANDDARKNESS on you then failed and the error in the console
`[Warning]: skill_castend_nodamage_id: missing code case for skill NPC_GRANDDARKNESS (339)`

### Official Information

none

none

Playtester:
* **Addressed Issue(s)**: #8104 

* **Server Mode**: Both (wrong modes only in pre-re)

- The randomtarget monster mode no longer causes skills to have a random target (fixes #8104)
  * Target skills will always go for the original target
  * Normal attacks will pick a random target, but will not change the monster's original target
  * Self/Support skills will now properly target self or friends
- Fixed an issue where a monster would continue moving when it found a new target in attack range
  * This was caused by setting target_to to 0 without stopping the movement
  * Follow up to f272ca6
- Added the correct "randomtarget" modes to pre-re monsters
  * These are already correctly set in renewal, but were not set in pre-re when the mode was implemented

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1263,6 +1263,24 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
 	}
 }
 
+/**
+ * Randomizes the target ID of a monster if it has the given mode
+ * @param bl Unit that is going to attack
+ * @param target_id Target ID to modify
+ */
+void mob_randomtarget(mob_data& md, int32& target_id) {
+	if (!(md.status.mode&MD_RANDOMTARGET))
+		return;
+
+	int32 search_size = md.status.rhw.range;
+	if (md.sc.hasSCE(SC_BLIND))
+		search_size = 1;
+
+	block_list* target = battle_getenemy(&md.bl, DEFAULT_ENEMY_TYPE((&md)), search_size);
+	if (target != nullptr)
+		target_id = target->id;
+}
+
 /*==========================================
  * Determination for an attack of a monster
  *------------------------------------------*/
@@ -2119,8 +2137,8 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 	// Normal attack / berserk skill is only used when target is in range
 	if (battle_check_range(&md->bl, tbl, md->status.rhw.range))
 	{
-		// Make sure there is no chase target when already in attack range
-		md->ud.target_to = 0;
+		// Stop and make sure there is no chase target when already in attack range
+		unit_stop_walking(&md->bl, USW_FIXPOS|USW_RELEASE_TARGET);
 
 		// Hiding is a special case because it prevents normal attacks but allows skill usage
 		// TODO: Some other states also have this behavior and should be investigated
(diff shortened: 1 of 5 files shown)


src/map/mob.cpp
```diff
@@ -1263,6 +1263,24 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
 	}
 }
 
+/**
+ * Randomizes the target ID of a monster if it has the given mode
+ * @param bl Unit that is going to attack
+ * @param target_id Target ID to modify
+ */
+void mob_randomtarget(mob_data& md, int32& target_id) {
+	if (!(md.status.mode&MD_RANDOMTARGET))
```
Daegaladh, line 1272:
```suggestion
	if (!status_has_mode(md.status, MD_RANDOMTARGET))
```

Daegaladh (approved):

src/map/mob.cpp
```diff
@@ -1263,6 +1263,24 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
 	}
 }
 
+/**
+ * Randomizes the target ID of a monster if it has the given mode
+ * @param bl Unit that is going to attack
+ * @param target_id Target ID to modify
+ */
+void mob_randomtarget(mob_data& md, int32& target_id) {
```
Daegaladh, line 1271: Why do you pass the target_id as a reference instead of just passing the value?

Playtester: So that we can update it. Saves you having to reassign the value and also saves memory.

src/map/mob.cpp
Playtester: Here we already have md. I thought status_has_mode is only needed when the bl type is unknown?

src/map/unit.cpp
```diff
@@ -2885,6 +2885,12 @@ int32 unit_attack(struct block_list *src,int32 target_id,int32 continuous)
 
 	nullpo_ret(ud = unit_bl2ud(src));
 
+	// Check for special monster random target mode
+	if (src->type == BL_MOB) {
+		mob_data& md = *reinterpret_cast<mob_data*>(src);
```
vstumpf: this should now be a `*static_cast<mob_data*>(src);`

Playtester: Done!

PR #9381, opened 2025-06-01
Change reinterpret_casts of block_list to static_casts

vstumpf:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

static_cast is safer than reinterpret_cast, we will get compiler errors if the cast can't be made.


Singe-Horizontal:
The crashes in the builds are caused by missing placement news that didn't initialize the vtables.
I found 2 of them so far, also we can now call block_list directly in map_destroyblock since its destructor is now virtual.
https://github.com/Singe-Horizontal/rathena/commit/6617ba60301967d3841a7d535229adc117fcf0a5

PR #9389, opened 2025-06-02
Official Wink Charm Implementation

### LINKED ISSUE #9379: Wink Charm Heart Spam
### rAthena Hash

e535dde

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

Monsters that are charmed spam the heart emotion hundreds of times per second.

### Expected Result

Monsters that are charmed show the heart emotion every 2250ms.

### How to Reproduce

Use Wink Charm on a Demi-Human, Demon or Angel.

### Official Information

Official information available. Confirmed by tests.

What Wink Charm actually does:
- Prevents skill casting on non-players
- If non-player attacks the caster, then the attack will be cancelled, but it does not unlock target

Players can actually be affected by it if SC_CONFUSION is successful, but they will only do the heart emotion. 
In renewal, players have a 100% base chance to get SC_CONFUSION and are always affected by SC_HALLUCINATION for 10 seconds instead.

Counts as Exp Tap, see also: https://github.com/rathena/rathena/issues/9209

There is no requirement to be higher level than the monster either, but the rate becomes 0 when the monster is 40 or more levels higher than the caster.

Cast on invalid target just gives skill has failed message. No SP consumption or delay.

Works on bosses, but cannot target bosses. (Only autocast works.)

Wink Charm draws aggro on monsters without a target or monster that change target on skill, but when Wink Charm ends, any passive monster will drop target.

_No response_

Playtester:
* **Addressed Issue(s)**: #9379 

* **Server Mode**: Both

- Wink Charm no longer spams the client with hundreds of emotions and move commands
- Complete re-implementation of Wink Charm
  * Wink Charm blocks skill casting for non-players
  * Wink Charm blocks attacks on the caster for non-players
  * Wink Charm no longer makes the target move towards the caster, but can trigger aggro like a damage skill
  * Wink Charm can trigger additional effects just like a ranged weapon attack
  * Wink Charm can no longer target bosses, but bosses can now be affected by it (Aliza Card)
  * When Wink Charm fails due to invalid target, it will no longer consume SP nor set aftercast delay
  * When the target is valid, Wink Charm now always shows the animation, even if the status is resisted
  * When Wink Charm ends, the affected monster will drop its target
- Wink Charm can still cause Confusion to players at low chance
  * When Confusion succeeds, players will also do the heart emotion for 10 seconds
  * Confusion now lasts 30s in pre-re and 18+2s in re
- Properly integrated SC_WINKCHARM and SC_VOICEOFSIREN into the interval status change system
  * Both skills show the heart emotion every 2250ms
- Pang Voice shares the boss target block logic with Wink Charm
- Fixes #9379

**TODO**: Check why confusion is not working, exp tap, cast cancel, check renewal

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -9450,23 +9450,20 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, 
 
 	case DC_WINKCHARM:
 		if( dstsd ) {
-			clif_skill_nodamage(src,*bl,skill_id,skill_lv, sc_start(src,bl,SC_CONFUSION,10,7,skill_get_time2(skill_id,skill_lv)));
+			if (sc_start(src, bl, SC_CONFUSION, 10, 7, skill_get_time2(skill_id, skill_lv)))
+				sc_start(src, bl, type, 100, skill_lv, skill_get_time(skill_id, skill_lv));
 #ifdef RENEWAL
 			sc_start(src, bl, SC_HALLUCINATION, 30, skill_lv, skill_get_time(skill_id, skill_lv)); // TODO: Confirm success rate and duration
 #endif
 		} else
 		if( dstmd )
 		{
-			if( status_get_lv(src) > status_get_lv(bl)
-			&&  (tstatus->race == RC_DEMON || tstatus->race == RC_DEMIHUMAN || tstatus->race == RC_PLAYER_HUMAN || tstatus->race == RC_PLAYER_DORAM || tstatus->race == RC_ANGEL)
-			&&  !status_has_mode(tstatus,MD_STATUSIMMUNE) )
-				clif_skill_nodamage(src,*bl,skill_id,skill_lv, sc_start2(src,bl,type,(status_get_lv(src) - status_get_lv(bl)) + 40, skill_lv, src->id, skill_get_time(skill_id, skill_lv)));
-			else
-			{
-				clif_skill_nodamage(src,*bl,skill_id,skill_lv,false);
-				if(sd) clif_skill_fail( *sd, skill_id );
+			if (sc_start2(src, bl, type, (status_get_lv(src) - status_get_lv(bl)) + 40, skill_lv, src->id, skill_get_time(skill_id, skill_lv))) {
+				// This triggers a 0 damage event and might make the monster switch target to caster
+				battle_damage(src, bl, 0, 1, skill_lv, 0, ATK_DEF, BF_WEAPON|BF_LONG|BF_NORMAL, true, tick, false);
 			}
 		}
+		clif_skill_nodamage(src, *bl, skill_id, skill_lv);
 		break;
 
 #ifdef RENEWAL
@@ -13722,6 +13719,16 @@ static int8 skill_castend_id_check(struct block_list *src, struct block_list *ta
 				status_data* tstatus = status_get_status_data(*target);
 				if (!battle_check_undead(tstatus->race, tstatus->def_ele))
 					return USESKILL_FAIL_MAX;
+			}
+			break;
+		case DC_WINKCHARM:
+			{
+				status_data* tstatus = status_get_status_data(*target);
+				if (tstatus->race == RC_DEMON || tstatus->race == RC_DEMIHUMAN || tstatus->race == RC_ANGEL)
+					break;
+				if (tstatus->race == RC_PLAYER_HUMAN || tstatus->race == RC_PLAYER_DORAM)
+					break;
+				return USESKILL_FAIL_LEVEL;
 			}
 			break;
 		case PR_LEXDIVINA:
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -2115,19 +2115,6 @@ bool status_check_skilluse(struct block_list *src, struct block_list *target, ui
 		)
 			return false;
 
-		if (sc->getSCE(SC_WINKCHARM) && target && !flag) { // Prevents skill usage
-			block_list *wink_target = map_id2bl(sc->getSCE(SC_WINKCHARM)->val2);
-
-			if (wink_target != nullptr) {
-				unit_data *wink_ud = unit_bl2ud(src);
-				if (wink_ud != nullptr && wink_ud->walktimer == INVALID_TIMER)
-					unit_walktobl(src, map_id2bl(sc->getSCE(SC_WINKCHARM)->val2), 3, 1);
-				clif_emotion( *src, ET_THROB );
-				return false;
-			} else
-				status_change_end(src, SC_WINKCHARM);
-		}
-
 		if (sc->getSCE(SC_BLADESTOP)) {
 			switch (sc->getSCE(SC_BLADESTOP)->val1) {
 				case 5:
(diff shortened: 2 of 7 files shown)


src/map/skill.cpp
```diff
@@ -13724,6 +13721,16 @@ static int8 skill_castend_id_check(struct block_list *src, struct block_list *ta
 					return USESKILL_FAIL_MAX;
 			}
 			break;
+		case DC_WINKCHARM:
+			{
+				status_data* tstatus = status_get_status_data(*target);
+				if (tstatus->race == RC_DEMON || tstatus->race == RC_DEMIHUMAN || tstatus->race == RC_ANGEL)
+					break;
+				if (tstatus->race == RC_PLAYER_HUMAN || tstatus->race == RC_PLAYER_DORAM)
```
Daegaladh: Why you don't combine the 2 ifs?

Playtester: Didn't want to make it too long, Lemon also prefers if we don't make too large ifs (though he would probably advocate to have an extra line for each check).

src/map/skill.cpp
```diff
@@ -13724,6 +13724,16 @@ static int8 skill_castend_id_check(struct block_list *src, struct block_list *ta
...
+					break;
+				return USESKILL_FAIL_LEVEL;
+			}
```
Atemo, line 13739:
```suggestion
				switch (status_get_status_data(*target)->race) {
				case RC_DEMON:
				case RC_DEMIHUMAN:
				case RC_ANGEL:
				case RC_PLAYER_HUMAN:
				case RC_PLAYER_DORAM:
					break;
				default:
					return USESKILL_FAIL_LEVEL;
				}
```
or `switch (status_data* tstatus = status_get_status_data(*target); tstatus->race) {` if you need tstatus in the switch

Daegaladh: Yeah I also like the switch more

src/map/skill.cpp
```diff
@@ -9450,23 +9450,23 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
 
 	case DC_WINKCHARM:
 		if( dstsd ) {
-			clif_skill_nodamage(src,*bl,skill_id,skill_lv, sc_start(src,bl,SC_CONFUSION,10,7,skill_get_time2(skill_id,skill_lv)));
 #ifdef RENEWAL
-			sc_start(src, bl, SC_HALLUCINATION, 30, skill_lv, skill_get_time(skill_id, skill_lv)); // TODO: Confirm success rate and duration
+			// Causes Confusion and Hallucination to 100% base chance in renewal
```
```suggestion
			// In Renewal it causes Confusion and Hallucination to 100% base chance
```

src/map/skill.cpp
```diff
@@ -9450,23 +9450,23 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
...
+			sc_start(src, bl, SC_CONFUSION, 100, skill_lv, skill_get_time2(skill_id, skill_lv));
+			sc_start(src, bl, SC_HALLUCINATION, 100, skill_lv, skill_get_time(skill_id, skill_lv));
+#else
+			if (sc_start(src, bl, SC_CONFUSION, 10, skill_lv, skill_get_time2(skill_id, skill_lv)))
```
Lemongrass3110, line 9459:
```suggestion
			// In Pre-Renewal it only causes Wink Charm, if Confusion was successfully started
			if (sc_start(src, bl, SC_CONFUSION, 10, skill_lv, skill_get_time2(skill_id, skill_lv)))
```
Helps a little to know why here "type" is used and not hardcoded SC names.

src/map/skill.cpp
```diff
@@ -9450,23 +9450,23 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
...
-			&&  (tstatus->race == RC_DEMON || tstatus->race == RC_DEMIHUMAN || tstatus->race == RC_PLAYER_HUMAN || tstatus->race == RC_PLAYER_DORAM || tstatus->race == RC_ANGEL)
-			&&  !status_has_mode(tstatus,MD_STATUSIMMUNE) )
-				clif_skill_nodamage(src,*bl,skill_id,skill_lv, sc_start2(src,bl,type,(status_get_lv(src) - status_get_lv(bl)) + 40, skill_lv, src->id, skill_get_time(skill_id, skill_lv)));
-			else
-			{
-				clif_skill_nodamage(src,*bl,skill_id,skill_lv,false);
-				if(sd) clif_skill_fail( *sd, skill_id );
+			if (sc_start2(src, bl, type, (status_get_lv(src) - status_get_lv(bl)) + 40, skill_lv, src->id, skill_get_time(skill_id, skill_lv))) {
+				// This triggers a 0 damage event and might make the monster switch target to caster
+				battle_damage(src, bl, 0, 1, skill_lv, 0, ATK_DEF, BF_WEAPON|BF_LONG|BF_NORMAL, true, tick, false);
```
Lemongrass3110, line 9468: According to your explanation shouldnt this also be done for players (Pre-Renewal), when Wink Charm SC is started?
Should this maybe be be called from within sc_start instead?

src/map/status.cpp
```diff
@@ -12045,9 +12041,10 @@ static bool status_change_start_post_delay(block_list* src, block_list* bl, sc_t
 		case SC_HARMONIZE:
 			val2 = 5 + 5 * val1;
 			break;
+		case SC_WINKCHARM:
```
Lemongrass3110: Both could actually be moved up below SC_KILLING_AURA and removed here.

src/map/skill.cpp
```diff
@@ -9450,23 +9450,23 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
...
 		{
-			if( status_get_lv(src) > status_get_lv(bl)
-			&&  (tstatus->race == RC_DEMON || tstatus->race == RC_DEMIHUMAN || tstatus->race == RC_PLAYER_HUMAN || tstatus->race == RC_PLAYER_DORAM || tstatus->race == RC_ANGEL)
-			&&  !status_has_mode(tstatus,MD_STATUSIMMUNE) )
-				clif_skill_nodamage(src,*bl,skill_id,skill_lv, sc_start2(src,bl,type,(status_get_lv(src) - status_get_lv(bl)) + 40, skill_lv, src->id, skill_get_time(skill_id, skill_lv)));
-			else
-			{
-				clif_skill_nodamage(src,*bl,skill_id,skill_lv,false);
-				if(sd) clif_skill_fail( *sd, skill_id );
+			if (sc_start2(src, bl, type, (status_get_lv(src) - status_get_lv(bl)) + 40, skill_lv, src->id, skill_get_time(skill_id, skill_lv))) {
```
Lemongrass3110, line 9466:
```suggestion
			// For monsters it causes Wink Charm with a chance depending on the level difference
			if (sc_start2(src, bl, type, (status_get_lv(src) - status_get_lv(bl)) + 40, skill_lv, src->id, skill_get_time(skill_id, skill_lv))) {
```

Lemongrass3110 (approved):

src/map/skill.cpp
Playtester: No, the damage tap to get aggro is only needed for monsters. Players don't have an AI.

src/map/status.cpp
Playtester: Will check it.

Playtester: Done.

PR #9394, opened 2025-06-06
Skid Trap Default Direction

Playtester:
* **Addressed Issue(s)**: -

* **Server Mode**: Both

- Fixed Skid Trap's default knockback direction being inverted
  * This only applies when the target is standing on the exact cell the caster was on when placing the trap
  * Default direction in this scenario is now "West" instead of "East"

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -17171,7 +17171,7 @@ int32 skill_unit_onplace_timer(struct skill_unit *unit, struct block_list *bl, t
 		case UNT_SKIDTRAP: {
 				//Knockback away from position of user during placement [Playtester]
 				skill_blown(unit,bl,skill_get_blewcount(sg->skill_id,sg->skill_lv),
-					(map_calc_dir_xy(sg->val1>>16,sg->val1&0xFFFF,bl->x,bl->y,6)+4)%8,BLOWN_NONE);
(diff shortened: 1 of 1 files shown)


src/map/skill.cpp
```diff
@@ -17171,7 +17171,7 @@ int32 skill_unit_onplace_timer(struct skill_unit *unit, struct block_list *bl, t
 		case UNT_SKIDTRAP: {
 				//Knockback away from position of user during placement [Playtester]
 				skill_blown(unit,bl,skill_get_blewcount(sg->skill_id,sg->skill_lv),
-					(map_calc_dir_xy(sg->val1>>16,sg->val1&0xFFFF,bl->x,bl->y,6)+4)%8,BLOWN_NONE);
+					map_calc_dir_xy(bl->x,bl->y,sg->val1>>16,sg->val1&0xFFFF,6)%8,BLOWN_NONE);
```
```suggestion
					map_calc_dir_xy(bl->x,bl->y,sg->val1>>16,sg->val1&0xFFFF,DIR_EAST),BLOWN_NONE);
```
6 = DIR_EAST
%8 = %DIR_MAX can be skipped, as it is safe to assume, that this function always returns a valid number.

Lemongrass3110 (approved):

PR #9397, opened 2025-06-06
Random Target Mode vs. Blind

### LINKED ISSUE #9396: Random Target vs. Blind
### rAthena Hash

2d3dfd7

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When a monster with random target mode is blind and there is at least one target next to it, it will only attack the targets next to it.

### Expected Result

When a monster with random target mode is blind and there is at least one target next to it, it can still attack targets further away with random attacks as long as they are not hidden and not behind an obstacle.

The main target of the monster will still switch to the target next to it (relevant for skill targetting).

### How to Reproduce

Summon Ragged Zombie.

Test 1:
Player A stand 5 cells away from Ragged Zombie and is the main target of the Ragged Zombie.
Player B stands 1 cell away from Ragged Zombie.
Now one of the two players cast "Status Recovery" on Ragged Zombie to blind it.

Test 2:
Player A stand 5 cells away from Ragged Zombie and is the main target of the Ragged Zombie.
Player B stands 1 cell away from Ragged Zombie.
Now one of the two players cast places Flasher next to Ragged Zombie and uses Arrow Shower to knock the trap into Ragged Zombie to blind it (without damaging).

### Official Information

Official information available. Verified through tests.

In both test cases above, Ragged Zombie will still attack both player A and player B with normal attacks. Before being blinded, all skills hit Player A. After being blinded, all skills hit player B (even after blind ends).

_No response_

Playtester:
* **Addressed Issue(s)**: #9396 

* **Server Mode**: Both

- Monsters can now target random targets outside their view range (1 if blind) with normal attacks
  * When blinded, a monster will still switch its main target (for skills) to the target next to it
- When the random target picked is not in shoot range, the monster will stay in its current mode
- Minor code optimizations
- Follow-up to 4d68187 and fb6fdd1
- Fixes #9396

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1271,7 +1271,7 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
  */
 bool mob_randomtarget(mob_data& md, int32& target_id) {
 	if (!status_has_mode(&md.status, MD_RANDOMTARGET))
-		return true;
+		return true; // Keep current target
 
 	// Pick a random visible target
 	block_list* target = battle_getenemy(&md, DEFAULT_ENEMY_TYPE((&md)), md.status.rhw.range);
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -1267,18 +1267,23 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
  * Randomizes the target ID of a monster if it has the given mode
  * @param bl Unit that is going to attack
  * @param target_id Target ID to modify
+ * @return Whether a target was found (true) or not (false)
  */
-void mob_randomtarget(mob_data& md, int32& target_id) {
+bool mob_randomtarget(mob_data& md, int32& target_id) {
 	if (!status_has_mode(&md.status, MD_RANDOMTARGET))
-		return;
+		return true;
 
-	int32 search_size = md.status.rhw.range;
-	if (md.sc.hasSCE(SC_BLIND))
-		search_size = 1;
+	// Pick a random visible target
+	block_list* target = battle_getenemy(&md, DEFAULT_ENEMY_TYPE((&md)), md.status.rhw.range);
+	if (target == nullptr)
+		return false;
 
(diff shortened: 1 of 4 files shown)


src/map/mob.cpp
```diff
@@ -1267,18 +1267,23 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
  * Randomizes the target ID of a monster if it has the given mode
  * @param bl Unit that is going to attack
  * @param target_id Target ID to modify
+ * @return Whether a target was found (true) or not (false)
  */
-void mob_randomtarget(mob_data& md, int32& target_id) {
+bool mob_randomtarget(mob_data& md, int32& target_id) {
 	if (!status_has_mode(&md.status, MD_RANDOMTARGET))
-		return;
+		return true;
```
Lemongrass3110: True means that a target was found according to your doc?
May need a little extra comment, why it is different in this case.

src/map/mob.cpp
```diff
@@ -1267,18 +1267,23 @@ static int32 mob_can_changetarget(struct mob_data* md, struct block_list* target
...
 
-	int32 search_size = md.status.rhw.range;
-	if (md.sc.hasSCE(SC_BLIND))
-		search_size = 1;
+	// Pick a random visible target
+	block_list* target = battle_getenemy(&md, DEFAULT_ENEMY_TYPE((&md)), md.status.rhw.range);
+	if (target == nullptr)
```
Lemongrass3110, line 1278: Doesnt that return too early, if a target would be in shoot range?

Lemongrass3110 (approved):

src/map/mob.cpp
Playtester: Well, the target was found, it's the current target for monsters without random target mode. But I guess I can add that as comment.
```suggestion
		return true; // Keep current target
```
Like this.

src/map/mob.cpp
Playtester: If you don't have a target, you don't need to check if it's in shoot range.

This is how it works officially too. It first fetches a random visible target and if that random target is then not in shoot range, it just cancels out of the process, it doesn't try other targets.

PR #9399, opened 2025-06-07
Random Target Fail and Wink Charm Movement

### LINKED ISSUE #9398: Random Target Fail and Wink Charm should not stop monster
### rAthena Hash

d6ab4e6

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

When a monster with random target mode fails to find a random target, it will stop.

A monster affected by Wink Charm stops once it gets into attack range of its charmer.

### Expected Result

When a monster with random target mode fails to find a random target, it will continue moving on its walkpath until it reaches the end of it.

A monster affected by Wink Charm will not stop when it comes into attack range of its charmer. It will move to the end of its walkpath.

Note: In both cases monsters only start moving again once you walk all the way outside the attack range.

### How to Reproduce

Ragged Zombie -> make it chase you -> Hide.

Goblin Archer -> Wink Charm it -> Move out of its attack range.

### Official Information

Official information available. Verified by tests on official servers. Both cases described above have been verified.

_No response_

Playtester:
* **Addressed Issue(s)**: #9398 

* **Server Mode**: Both

- A monsters that fails finding a random target will no longer stop moving
- A monster affect by Wink Charm getting into attack range of its charmer will no longer stop moving
- Fixes #9398

### DIFF
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -118,11 +118,17 @@ bool unit_update_chase(block_list& bl, t_tick tick, bool fullcheck) {
 
 	// Reached destination, start attacking
 	if (tbl != nullptr && tbl->type != BL_ITEM && tbl->m == bl.m && ud->walkpath.path_pos > 0 && check_distance_bl(&bl, tbl, ud->chaserange)) {
-		// We need to make sure the walkpath is cleared here so a monster doesn't continue walking in case it unlocks its target
-		unit_stop_walking(&bl, USW_FIXPOS|USW_FORCE_STOP|USW_RELEASE_TARGET);
+		e_unit_attack atk_result = ATTACK_FAIL;
+
 		if (ud->state.attack_continue)
-			unit_attack(&bl, tbl->id, ud->state.attack_continue);
-		return true;
+			atk_result = unit_attack(&bl, tbl->id, ud->state.attack_continue);
+
+		// Stop, unless the attack was skipped
+		if (atk_result != ATTACK_SKIP) {
+			// We need to make sure the walkpath is cleared here so a monster doesn't continue walking in case it unlocked its target
+			unit_stop_walking(&bl, USW_FIXPOS|USW_FORCE_STOP|USW_RELEASE_TARGET);
+			return true;
+		}
 	}
 	// Cancel chase
 	else if (tbl == nullptr || (fullcheck && !status_check_visibility(&bl, tbl, (bl.type == BL_MOB)))) {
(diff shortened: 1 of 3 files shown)


src/map/unit.cpp
```diff
@@ -1688,9 +1694,14 @@ bool unit_stop_walking( block_list* bl, int32 type, t_tick canmove_delay ){
 		return false;
 	}
 
-	ud = unit_bl2ud(bl);
+	if (ud = unit_bl2ud(bl); ud == nullptr)
```
Lemongrass3110: Please dont put the assignment into the if, unless you declare the variable in this scope.

src/map/unit.cpp
```diff
@@ -2882,43 +2890,46 @@ int32 unit_unattackable(struct block_list *bl)
...
+ * @return How the unit should stop; see e_unit_stop_walking
  */
 int32 unit_attack(struct block_list *src,int32 target_id,int32 continuous)
 {
 	struct block_list *target;
 	struct unit_data  *ud;
 	int32 range;
 
-	nullpo_ret(ud = unit_bl2ud(src));
+	if (ud = unit_bl2ud(src); ud == nullptr)
```
Lemongrass3110: Please dont put the assignment into the if, unless you declare the variable in this scope.

src/map/unit.cpp
```diff
@@ -2882,43 +2890,46 @@ int32 unit_unattackable(struct block_list *bl)
...
 	if( !unit_can_attack(src, target_id) ) {
 		unit_stop_attack(src);
-		return 0;
+		return USW_NONE; // Continue walking
 	}
 
 	if( battle_check_target(src,target,BCT_ENEMY) <= 0 || !status_check_skilluse(src, target, 0, 0) ) {
 		unit_unattackable(src);
-		return 1;
+		return stop_flag;
```
Lemongrass3110, line 2932: Why is it not USW_NONE here, like above?

Lemongrass3110 (changes requested):

src/map/unit.cpp
Playtester: Ok, I thought that is the preferred style. But no problem, can change.

src/map/unit.cpp
Playtester: If the target itself is invalid, is goes to unit_unattackable() which makes the monster switch to Idle. In this case the unit should always stop, even if called before the code where Aegis would usually stop the attacker.

Only when the code happens before the Aegis stop code and it returns "success" for the attack, the unit will continue moving, which is only the case when the monster itself is "charmed" (or in other words: it can move but cannot attack).

src/map/unit.cpp
Playtester: Fixed.

src/map/unit.cpp
Playtester: Fixed.

Lemongrass3110 (approved):

PR #9408, opened 2025-06-09
Prevent Freeing Data Too Early

### LINKED ISSUE #9407: freed-data is changed
### rAthena Hash

a658af47ae2d5d7d23f145066f848a1f606860cb

### Client Date

2024-08-22

### Server Mode

Pre-Renewal, Renewal

### Result

When I spam Reject Sword (stalker/rogue), and I kill some monsters, then spawn again, an error appears in the mapserver console:

`Memory manager: freed-data is changed. (freed in PATH\rathena\src\map\map.cpp line 287).`

I have no custom mods in this rAthena build, the line 287 in map.cpp is the function:
"map_destroyblock".
Line: 287: aFree( bl );

**EDIT:**
I can see this error with all monsters, I spawned 10 familiars, then killed some of them with only reject sword, and then repeat the command to summon other 10 familiars, the error will show.

### Relevant Log Output

```Shell
Memory manager: freed-data is changed. (freed in PATH\rathena\src\map\map.cpp line 287).
```

### Expected Result

I believe it should not throw any error.

### How to Reproduce

1. Summon 10 familiars
2. Spam Reject Sword Skill, and kill some of them
3. Summon the 10 familiars again
4. The error will appear

[Familiar Video](https://streamable.com/4swc17)
[Custom Mob video](https://streamable.com/ihqmxi)

### Official Information

N/A

I have no custom mods, it's a clean build.

Playtester:
* **Addressed Issue(s)**: #9407 

* **Server Mode**: Both

- Fixed a few issues where the code didn't consider that an attacker could die from reflect damage
- Ensure we only call map_freeblock_unlock right before we return from a function
  * This prevents the risk of accessing pointers that point to already freed data
- Fixes #9407

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2144,6 +2144,9 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 	{
 		int32 stop_flag = USW_FIXPOS|USW_RELEASE_TARGET;
 
+		// Source may die due to reflect damage
+		map_freeblock_lock();
+
 		// Hiding is a special case because it prevents normal attacks but allows skill usage
 		// TODO: Some other states also have this behavior and should be investigated
 		// TODO: EFST_AUTOCOUNTER, EFST_BLADESTOP, NPC_SR_CURSEDCIRCLE
@@ -2165,6 +2168,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 			unit_stop_walking(md, stop_flag);
 
 		//Target still in attack range, no need to chase the target
+		map_freeblock_unlock();
 		return true;
 	}
 
--- a/src/map/unit.cpp
+++ b/src/map/unit.cpp
@@ -120,6 +120,9 @@ bool unit_update_chase(block_list& bl, t_tick tick, bool fullcheck) {
 	if (tbl != nullptr && tbl->type != BL_ITEM && tbl->m == bl.m && ud->walkpath.path_pos > 0 && check_distance_bl(&bl, tbl, ud->chaserange)) {
 		int32 stop_flag = USW_FIXPOS|USW_RELEASE_TARGET;
 
+		// Source may die due to reflect damage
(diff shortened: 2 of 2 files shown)


src/map/mob.cpp
```diff
@@ -2165,6 +2168,7 @@ static bool mob_ai_sub_hard(struct mob_data *md, t_tick tick)
 			unit_stop_walking(md, stop_flag);
 
 		//Target still in attack range, no need to chase the target
+		map_freeblock_unlock();
```
Lemongrass3110, line 2170: Please move this above the comment

src/map/unit.cpp
```diff
@@ -3329,6 +3338,7 @@ static int32 unit_attack_timer_sub(struct block_list* src, int32 tid, t_tick tic
 	if( sd && battle_config.prevent_logout_trigger&PLT_ATTACK )
 		sd->canlog_tick = gettick();
 
+	map_freeblock_unlock();
```
Lemongrass3110, line 3344: The lock is applied within an if. This means you may unlock too often now.

Lemongrass3110 (changes requested):

src/map/unit.cpp
Playtester: Hmm true, that means we need to lock earlier actually.

src/map/mob.cpp
Playtester: Fixed

src/map/unit.cpp
Playtester: Fixed

Lemongrass3110: Not really sure how it can die here, but okay.

Lemongrass3110 (approved):

src/map/unit.cpp
Playtester: The source dies from reflect damage when calling:
```
ud->attacktarget_lv = battle_weapon_attack(src,target,tick,0);
```

PR #9412, opened 2025-06-10
Delete Map Session Data Typo

Haydrich:
* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/src/map/atcommand.cpp
+++ b/src/map/atcommand.cpp
@@ -674,7 +674,7 @@ ACMD_FUNC(where)
  *------------------------------------------*/
 ACMD_FUNC(jumpto)
 {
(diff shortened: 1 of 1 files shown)


src/map/atcommand.cpp
```diff
@@ -10543,7 +10543,7 @@ ACMD_FUNC(langtype)
 
 ACMD_FUNC(vip) {
 #ifdef VIP_ENABLE
-	map_session_data* pl_sd = nullptr;;
+	map_session_data* pl_sd = nullptr;
```
Atemo, line 10546: There is more extra ";" in this file could you please removed them?

Lemongrass3110 (approved):

PR #9419, opened 2025-06-11
Fixed typos EndlessTower.txt

brnkvalcante:
* **Addressed Issue(s)**:  N/A

* **Server Mode**:  both

* **Description of Pull Request**:  fixed some typos.


Lemongrass3110:
@Atemo do we even need to specify the names here or should we simply use the ones from mob db?

Atemo (approved):
It seems the mob name is the same in mob DB (need more check, can be done later).

Issue #9360, opened 2025-05-26
Hindsight


Playtester:
### rAthena Hash

d9607d2

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

Hindsight max level and skill level are determined when you cast it and then stored in the status change. It uses the values determined during cast for all autocasts.

When linked, Hindsight makes bolt autocasts use the level learned.

When not linked, Fireball autocasts can only be level 1 at both Hindsight level 8.

### Relevant Log Output

```Shell

```

### Expected Result

When cast, Hindsight only stores the selected skill (and duration) with the status change. Each time the status change activates, the levels are determined. That means if you increase the skill's level while it is active or get Soul Linked, the higher level will be used immediately without requiring a recast.

When linked, Hindsight makes all autocasts use the level learned (not just bolt skills).

When not linked, Fireball autocasts can be level 1 or 2 at both Hindsight level 8 and 9.

### How to Reproduce

Cast Hindsight, get linked, attack monster and check damage of each autocast.

Learn Hindsight 8 and check damage of Fireball autocasts.

### Official Information

Verified by tests for pre-re.

Playtester:
In renewal, it will always autocast half of the Hindsight level learned. When you get linked after already casting Hindsight, you will autocast at Hindsight level instead. But in both cases you never cast above level learned.

Example:

Cold Bolt 10 and Hindsight 5:

Cast Hindsight - Autocast Cold Bolt Level 2
Linked - Autocast Cold Bolt Level 5
Learn Hindsight 6 - Autocast Cold Bolt Level 6
Link ends - Autocast Cold Bolt Level 3

PR #9420, opened 2025-06-14
Fixed Pre-renewal SA_AUTOSPELL

Singe-Horizontal:
* **Addressed Issue(s)**: 
#9360

* **Server Mode**: 
Pre-Renewal

 Now only stores selected skill and status duration in its status entry.
 Autospell level, the autocasted skill level and soul link status are now checked on each proc attempt.
 Corrected chances of activation for each level.
 When linked, all autocasts will use their learnt level instead of just element bolts.
 Added validity checks for skill_id and skill_lv

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -10797,117 +10797,104 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
 	
 	if(sc && sc->getSCE(SC_AUTOSPELL)){
 		[&]{
+			uint16 autospell_lv{};
+
+			if(sd != nullptr)
+				autospell_lv = pc_checkskill(sd,SA_AUTOSPELL);
+			else
+				autospell_lv = sc->getSCE(SC_AUTOSPELL)->val1; //For non-players, we keep autospell level in the status entry
+
+			if(autospell_lv == 0)
+				return;
 #ifdef RENEWAL
-			if(rnd()%100 >= sc->getSCE(SC_AUTOSPELL)->val4)
-				return;
-
-			uint16 skill_id = sc->getSCE(SC_AUTOSPELL)->val2;
-			if(skill_id == 0)
-				return;
-			uint16 skill_lv = sc->getSCE(SC_AUTOSPELL)->val3;
-			if(skill_lv == 0)
-				return;
-			int32 i = rnd()%100;
-			if(sc->getSCE(SC_SPIRIT) && sc->getSCE(SC_SPIRIT)->val2 == SL_SAGE)
-				i = 0; //Max chance, no skill_lv reduction. [Skotlex]
-		//reduction only for skill_lv > 1
-			if(skill_lv > 1){
-				if(i >= 50) skill_lv /= 2;
-				else if(i >= 15) skill_lv--;
-			}
+			uint32 base_chance = 0;
 #else
-			uint16 autospell_level{};
-			if(sd != nullptr){
-				autospell_level = pc_checkskill(sd,SA_AUTOSPELL);
-
-				if(autospell_level == 0)
-					return;
-
-				if(rnd()%100 >= (5 + autospell_level * 2))
-					return;
-			}else{
-				if(rnd()%100 >= sc->getSCE(SC_AUTOSPELL)->val4)
-					return;
-			}
-
-			uint16 skill_id = sc->getSCE(SC_AUTOSPELL)->val2;
-			if(skill_id == 0)
+			uint32 base_chance = 5;
+#endif
+			if(rnd()%100 >= (base_chance + autospell_lv * 2))
 				return;
 
-			uint16 learned_level{};
-			if(sd){
-				learned_level = pc_checkskill(sd,skill_id);
-				if(learned_level == 0)
-					return;
-			}
+			uint16 proced_skill_id = sc->getSCE(SC_AUTOSPELL)->val2;
+			if(proced_skill_id == 0)
+				return;
 
-			uint16 skill_lv{};
-			if(sd == nullptr)
-				skill_lv = sc->getSCE(SC_AUTOSPELL)->val3;
-			else if(sc->hasSCE(SC_SPIRIT) && sc->getSCE(SC_SPIRIT)->val2 == SL_SAGE)
-				skill_lv = learned_level;
-			else{			
-				uint16 max_lvl = 1;
-				switch(skill_id){
+			uint16 learned_lv{};
+			if(sd != nullptr)
+				learned_lv = pc_checkskill(sd,proced_skill_id);
+			else
+				learned_lv = skill_get_max(proced_skill_id); //For non-players, we use the maximum level of the proc'ed skill
+
+			if(learned_lv == 0)
+				return;
+
+			uint16 proced_skill_lv{};
+			if(sd && sc->hasSCE(SC_SPIRIT) && sc->getSCE(SC_SPIRIT)->val2 == SL_SAGE)
+				proced_skill_lv = learned_lv;
+			else{
+#ifdef RENEWAL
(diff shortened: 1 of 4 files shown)


src/map/battle.cpp
```diff
@@ -10794,44 +10794,138 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
-		int32 sp = 0;
-		uint16 skill_id = sc->getSCE(SC_AUTOSPELL)->val2;
-		uint16 skill_lv = sc->getSCE(SC_AUTOSPELL)->val3;
-		int32 i = rnd()%100;
-		if (sc->getSCE(SC_SPIRIT) && sc->getSCE(SC_SPIRIT)->val2 == SL_SAGE)
-			i = 0; //Max chance, no skill_lv reduction. [Skotlex]
+	
+	if(sc && sc->getSCE(SC_AUTOSPELL)){
+		[&]{
+#ifdef RENEWAL
```
Playtester: I think we might want to fix this for renewal at the same time so that we don't need to overcomplicate the code just to only fix it for pre-renewal. Of course someone needs to check how it works in renewal.

src/map/battle.cpp
```diff
@@ -10794,44 +10794,138 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
+						else if(autospell_level == 6)
+							skill_lv = 2;
+						else if(autospell_level >= 7)
+							skill_lv = 3;
+						break;
+					case MG_FIREBALL:
+						if(autospell_level == 8)
+							skill_lv = 1;
+						else if(autospell_level >= 9)
+							skill_lv = 2;
```
Playtester: You call this variable `skill_lv`, but it reality it's more like a chance pattern. Also for Fireball it should always be 2, no? See linked bug report.

Please also give it a default value, then you can remove a lot of code here and ensure it is set.

src/map/battle.cpp
```diff
@@ -10794,44 +10794,138 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
+					default:
+						return;
+				}
+
+				skill_lv = min(max_level,skill_lv);
+				if(skill_lv > 1){
+					uint32 i = rnd()%1000;
+					switch(skill_lv){
+						case 2:
+							if(i >= 235)
```
Playtester: How did you conclude these numbers? Seems overcomplicating it compared to how it works on Aegis. At least explain in a comment how the math works here.

src/map/battle.cpp
```diff
@@ -10794,44 +10794,138 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
-				if (DIFF_TICK(ud->canact_tick, tick + autospell_tick) < 0) {
-					ud->canact_tick = i64max(tick + autospell_tick, ud->canact_tick);
-					if (battle_config.display_status_timers && sd)
-						clif_status_change(src, EFST_POSTDELAY, 1, autospell_tick, 0, 0, 0);
+			uint16 max_level = sd ? pc_checkskill(sd,skill_id) : sc->getSCE(SC_AUTOSPELL)->val3;
+
+			if(max_level == 0)
+				return;
+
+			if(!sd || (sd && sd->sc.getSCE(SC_SPIRIT) && sd->sc.getSCE(SC_SPIRIT)->val2  ==  SL_SAGE))
```
Playtester: Why are you getting the sc from "sd" here, when you already have it in a variable? You only checked that sc is set. Also please use "hasSCE" to check if an sce is active. There are also too many spaces near the ==.

Please also note that all pointer checks should be `== nullptr` or `!= nullptr`. Also in other parts of this PR.

You don't need to check for sd again when already had `!sd` checked with an `||` behind since it would already stop checking if the first part was true.

src/map/battle.cpp
```diff
@@ -10794,44 +10794,138 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
+				skill_lv = max_level;
+			else{
+				switch(skill_id){
+					case MG_NAPALMBEAT:
+						skill_lv = 3;
+						break;
+					case MG_COLDBOLT:
+					case MG_FIREBOLT:
+						[[fallthrough]];
```
Playtester: No need for the fallthrough here.

src/map/battle.cpp
```diff
@@ -10794,44 +10794,138 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
+				if(rnd()%100 >= (5 + autospell_level * 2))
+					return;
+			}else{
+				if(rnd()%100 >= sc->getSCE(SC_AUTOSPELL)->val4)
+					return;
 			}
-			if (ud) {
-				int32 autospell_tick = skill_delayfix(src, skill_id, skill_lv);
+			uint16 skill_lv{};
+			int32 sp{};
```
Playtester: Please declare variables only at the point where you first use them.

src/map/status.cpp
```diff
@@ -10968,7 +10968,7 @@ static bool status_change_start_post_delay(block_list* src, block_list* bl, sc_t
 #ifdef RENEWAL
 			val4 = val1 * 2; // Chance of casting
 #else
-			val4 = 5 + val1*2; // Chance of casting
+			val4 = 5 + val1*2; // Chance of casting for non-players
```
Playtester: Does this even work? It looks to me like you always pass 0 as val1 in pre-re? How do you set it for non-players?

Singe-Horizontal: It's set at cast end for non sd
https://github.com/rathena/rathena/blob/66a4a2ab68ce85d781514c3c000433f981d6842f/src/map/skill.cpp#L10250

src/map/skill.cpp
```diff
@@ -20896,32 +20893,18 @@ int32 skill_autospell(map_session_data *sd, uint16 skill_id)
 		maxlv = 10; //Soul Linker bonus. [Skotlex]
 	else
 		maxlv = skill_lv / 2; // Half of Autospell's level unless player learned a lower level (capped below)
```
Playtester: We could confirm now that in renewal, the level should also be determined per attack and not during casting of Hindsight, so this logic needs to be moved to the other part as well. Then the code here can just be shared between pre-re and renewal.

To clarify:
- In renewal you always cast `skill_lv / 2` except when linked, then you cast `skill_lv`. But in both cases never higher than learned level, there's no randomness to it

src/map/battle.cpp
```diff
@@ -10797,117 +10797,104 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
 					case MG_FIREBALL:
-						max_lvl = 2;
+						max_proced_lv = 2;
 						break;
 				}
-
-				max_lvl = min(learned_level,max_lvl);
-				skill_lv = 1;
-				if(max_lvl > 1){
+				max_proced_lv = min(learned_lv,max_proced_lv);
```
Playtester, line 10861: The capping here seems to be at the wrong place. It should be at the end after the level was already picked, no?

Could have shared code between renewal and pre-renewal for it too then.

src/map/battle.cpp
```diff
@@ -10797,117 +10797,104 @@ enum damage_lv battle_weapon_attack(struct block_list* src, struct block_list* t
...
+				proced_skill_lv = 1;
+				if(max_proced_lv > 1){
 					uint32 i = rnd()%1000;
```
Playtester, line 10864: I'd still prefer if we just used the official way of having a penalty value that included in the math.

Woud also allow to use `rnd_chance_official` to get an even more accurate result.

Edit: Actually, I think we need to use `rnd_value` here instead. As such as you could just make either a max value or min value instead of using a penalty.

Because essentially what you have is either 0-99, or 15-99 or 50-99.

PR #9424, opened 2025-06-19
Skill DB clean-up

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: both

Removed fields that use the default value.
Fields moved to where we expect them.

### DIFF
--- a/db/pre-re/skill_db.yml
+++ b/db/pre-re/skill_db.yml
@@ -9239,7 +9239,7 @@ Body:
       IsNpc: true
     Hit: Single
     HitCount: 1
-    Duration1: 30000
+    Duration: 30000
     Status: WeaponBreaker
   - Id: 344
     Name: NPC_ARMORBRAKE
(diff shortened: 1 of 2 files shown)


db/pre-re/skill_db.yml
```diff
@@ -2106,6 +2106,7 @@ Body:
     SplashArea: 2
     ActiveInstance: 5
     Knockback: 2
+    CastCancel: false
```
Playtester, line 2109: This seems like a useless change as the order shouldn't matter, does it?

db/pre-re/skill_db.yml
```diff
@@ -22691,19 +22678,11 @@ Body:
     Hit: Single
     HitCount: 1
     Element:
-      - Level: 1
-        Element: Neutral
       - Level: 2
         Element: Holy
```
Playtester, line 22682: Only defining the element for level 2?

db/pre-re/skill_db.yml
Atemo: We could call it that, although keeping the file organized slightly improves the readability. This is a PR to prepare for the cooldown and cast time update.

db/pre-re/skill_db.yml
Atemo: Default should be neutral, right?

PR #9441, opened 2025-06-28
Updated renewal skill delay data

Atemo:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

Updated CastTime, FixedCastTime, AfterCastActDelay and Cooldown according to data from KRO client.

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -1,5 +1,5 @@
 # This file is a part of rAthena.
-#   Copyright(C) 2021 rAthena Development Team
+#   Copyright(C) 2025 rAthena Development Team
 #   https://rathena.org - https://github.com/rathena
 #
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


db/re/skill_db.yml
```diff
@@ -1644,6 +1604,7 @@ Body:
         - Level: 10
           Amount: 50
     Status: Angelus
+    Cooldown: 30000
```
Playtester: Didn't want to keep a consistent order of attributes? Pretty sure Status usually comes after Cooldown.

db/re/skill_db.yml
```diff
@@ -2064,7 +2025,11 @@ Body:
         Time: 3200
       - Level: 3
         Time: 1600
+      - Level: 4
+        Time: 0
```
Playtester, line 2029: Didn't you just remove those? (0 is default I guess?)

Atemo: When?

db/re/skill_db.yml
Atemo: I wanted to keep it consistent, but I thought the order didn't matter to you so I'm following your will, or did I misunderstand?

Playtester: I'd say if we have to change something anyway, we can make it consistent, but we shouldn't just change something for th sake of making it consistent.

However, didn't we discuss it with Lemongrass and said let's make it consistent in a separate PR and then set to ignore the commit so it doesn't show up in blame. So I thought that was the way to go.

db/re/skill_db.yml
Playtester: Thought you wanted to remove all default values. I just didn't get around to test how it behaves exactly.
I personally prefer it like this though.

Playtester (approved):
Did not check data for validity, but format-wise it looks fine.

Issue #9417, opened 2025-06-11
Skill "Flying Kick" and "Back Stab" not working in script "bAutoSpellOnSkill"


Jaikayray:
### rAthena Hash

a5c3785

### Client Date

2024-10-30

### Server Mode

Renewal

### Result

Skill "Flying Kick" and "Back Stab" not working in script "bAutoSpellOnSkill"
In this combo item "Red Commando Beret" ID Item "400478 and 400479" and "Shadow Ring" ID Item "490267 and 490268" from script 
".@r = getequiprefinerycnt(EQI_HEAD_TOP);
bonus4 bAutoSpellOnSkill,"TK_JUMPKICK","RG_BACKSTAP",max(getskilllv("RG_BACKSTAP"),1),1000;
bonus4 bAutoSpellOnSkill,"RG_BACKSTAP","RG_STEALCOIN",max(getskilllv("RG_STEALCOIN"),1),50+(.@r*10);"

![Image](https://github.com/user-attachments/assets/2b8d40a1-6633-4412-b56a-a1326af5631d)

### Relevant Log Output

```Shell

```

### Expected Result

Skill "Flying Kick" and "Back Stab" not working in script "bAutoSpellOnSkill"
In this combo item "Red Commando Beret" ID Item "400478 and 400479" and "Shadow Ring" ID Item "490267 and 490268" from script 
".@r = getequiprefinerycnt(EQI_HEAD_TOP);
bonus4 bAutoSpellOnSkill,"TK_JUMPKICK","RG_BACKSTAP",max(getskilllv("RG_BACKSTAP"),1),1000;
bonus4 bAutoSpellOnSkill,"RG_BACKSTAP","RG_STEALCOIN",max(getskilllv("RG_STEALCOIN"),1),50+(.@r*10);"

### How to Reproduce

I don't know how to doing for this, can you help me fix that?

### Official Information

in thROC is working 100% when use skill "Flying Kick" auto spell on skill "Back Stab"

gasstation51:
src/map/pc.hpp


struct s_qi_display {
	bool is_active;
	e_questinfo_types icon;
	e_questinfo_markcolor color;
};

class map_session_data : public block_list {
public:
	+struct block_list* last_autospell_target = nullptr; // 최초 트리거 대상 저장용 필드
	+
	struct unit_data ud;
	struct view_data vd;
	struct status_data base_status, battle_status;
	status_change sc;
	struct regen_data regen;
[...]

PR #9444, opened 2025-06-30
Fix auto spell that cast back stab

kaninhot004:
* **Addressed Issue(s)**: 

https://github.com/rathena/rathena/issues/9417

* **Server Mode**: 

Renewal

https://github.com/user-attachments/assets/31530ebe-d845-4a86-b857-ab045d7b5818

Don't know If I fix it in wrong way or not. But it was working. So I just open PR for now.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -5582,7 +5582,7 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 
 	case RG_BACKSTAP:
 		{
-			if (!check_distance_bl(src, bl, 0)) {
+			if (distance_bl(src, bl) <= 1) {
 #ifdef RENEWAL
 				uint8 dir = map_calc_dir(src, bl->x, bl->y);
 				int16 x, y;


src/map/skill.cpp
```diff
@@ -5582,7 +5582,7 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 
 	case RG_BACKSTAP:
 		{
-			if (!check_distance_bl(src, bl, 0)) {
+			if (distance_bl(src, bl) <= 1) {
```
Playtester: Shouldn't this still be check_distance_bl just with 1 as range?

kaninhot004: When using Flying Kick -> Trigger auto spell Backstab the distance is 0
When using Backstab the distance is 1
That's why I used distance_bl instead.

Playtester: I don't get the explanation. Check distance is already defined like this:
```
	if (dx < 0) dx = -dx;
	if (dy < 0) dy = -dy;
	return ((dx<dy?dy:dx) <= distance);
```

kaninhot004: Done fixing in latest commit. Thanks!

src/map/skill.cpp
```diff
@@ -5582,7 +5582,7 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 
 	case RG_BACKSTAP:
 		{
-			if (!check_distance_bl(src, bl, 0)) {
+			if (check_distance_bl(src, bl, 1)) {
```
Playtester: Hmm, the problem here is that the previous check was there because if you are on the exact same cell as the target, you cannot calculate the direction to the target.

But I'm also not sure how it's supposed to work in renewal. It seems to move you to behind the target?

But how does it fail on autocast? You mean like a skill moved you on the same cell as the target and that skill is then supposed to autocast Backstab?

kaninhot004: Flying Kick made caster go inside target and distance was 0 that cause cast fail for backstab.

After fix the result was good. (Caster cast Backstab and move to behind target)

Playtester: I tested around a bit and this is how it should work:
- Pre-re: If you are on the same cell as the the target, then Backstab only succeeds if the target is facing northeast or southwest
- Re: They removed the check that you need to be behind the target, instead you are now actually moving to behind the target by the skill

That means the distance check shouldn't be here at all. Instead we need a check for pre-re that if target is on same cell as the source, then it only succeeds when the target is facing northeast or southwest.

It's also illustrated on Kokotewa here: https://ro.kokotewa.com/db/skl_info?id=RG_BACKSTAP

Playtester: Sorry small correction, Kokotewa isn't fully correct, northwest and southwest also works. I'm highlighting it here:
grafik

kaninhot004: I've added skip check for renewal and use same code as before for pre-renewal (I can't code the direction check like you explained in photo.).

src/map/skill.cpp
```diff
@@ -5582,7 +5582,11 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 
 	case RG_BACKSTAP:
 		{
-			if (!check_distance_bl(src, bl, 0)) {
+			bool is_skip_check = false;
+			#ifdef RENEWAL
+			is_skip_check = true;
+			#endif
+			if (is_skip_check || !check_distance_bl(src, bl, 0)) {
```
Playtester:
```suggestion
#ifndef RENEWAL
			if (check_distance_bl(src, bl, 0))
				break;
#endif
```
I feel like this is a way more simple way to put it.

But you will also need to remove the indentation on the lines below.

That makes it easier to later put it the correct direction code.

kaninhot004: Done fixing. Thanks for better coding suggestion.

src/map/skill.cpp
```diff
@@ -5583,42 +5583,44 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 
 	case RG_BACKSTAP:
 		{
-			if (!check_distance_bl(src, bl, 0)) {
+#ifndef RENEWAL
+		if (check_distance_bl(src, bl, 0))
+			break;
```
Playtester: Hmm, now that I look at it, might be better to just move this code into the "else" part so that you don't need to ifdef twice.

Playtester: Also you removed two tabs instead of just one. So there no indent after the first { here and the last } is on the same height as the code, which makes it a bit confusing.

kaninhot004: Done fixing indent and no more 2 IFDEF.

Playtester (approved):
Looks good to me now, even though it doesn't include the official implementation of the skill. Can be done later.

Playtester:
@kaninhot004 Can you test if the skill is still working as expected ingame?

kaninhot004:
> @kaninhot004 Can you test if the skill is still working as expected ingame?

Yes working.


https://github.com/user-attachments/assets/10dedd3a-aa9c-4779-be01-1cba8f62b867


https://github.com/user-attachments/assets/492939b6-6ff0-46a5-ade7-7aab0e54b0b7

Issue #9446, opened 2025-07-03
Many Navi Generation (navi_map, navi_npc, navi_link) improvements/flaw fixes


TBlazeWarriorT:
### Is your request related to a missing official feature?

Yes

### Describe the solution you'd like

- [navi_link] You cannot register true warps (type 200) to client. I'd like to be able to.
- [navi_npc] Some warp NPCs (JT_WARPNPC) still get registered to navi_npc. I believe they should be filtered
- [navi_npc] Empty-name NPCs still get registered to navi_npc. They probably won't show up ingame, but still I believe they should be filtered just in case and to save space
- [navi_map] The generated navi_map is half broken. It flags maps as indoors for no reason, such as wolfvill or comodo, and this causes the client to not render their minimap. (also, I've had to make a python script to fix the generated map names using SystemEN/mapInfo right after generating, but I guess this can't be fixed server-side for 
 the lack of mapInfo)

### Describe alternatives you've considered

I've made changes to my fork to address the missing features and design flaws above. Will likely pull request
My proposed changes:
[navi_link]: naviregisterwarp("", ...) registers the warp as type 200, which doesn't show the NPC name and tells you to walk to it instead of talking. Good for registering fake warps.
[navi_npc]: filtered out JT_WARPNPC and empty display name entries.
[navi_map]: disabled type 5003 (indoors), making it default to 5001 instead. It causes more issues than helps, and the detection is broken.

PR #9447, opened 2025-07-03
Many Navi Generation (navi_map, navi_npc, navi_link) improvements/flaw fixes

TBlazeWarriorT:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9446

* **Server Mode**:  Both

**[navi_link]:** naviregisterwarp("", ...) (empty first arg) registers the warp as type 200, which doesn't show the NPC name and tells you to walk to it instead of talking. Good for registering fake warps.

**[navi_npc]:** filtered out JT_WARPNPC (in addition to JT_HIDDEN_WARP_NPC) and empty display name entries.

**[navi_map]:** disabled type 5003 (indoors), making it default to 5001 instead. It causes more issues than helps, and the detection is broken.

I believe the code works as expected, confirm if needed. The changes are decently simple, make changes if needed. Thanks.

### DIFF
--- a/db/pre-re/item_db_equip.yml
+++ b/db/pre-re/item_db_equip.yml
@@ -23635,7 +23635,7 @@ Body:
     ArmorLevel: 1
     EquipLevelMin: 90
     Script: |
-      bonus bMdef,2;
+      bonus bMdef,1;
       bonus bAgi,2;
       bonus bDex,1;
   - Id: 2751
--- a/db/pre-re/status.yml
+++ b/db/pre-re/status.yml
@@ -3993,6 +3993,7 @@ Body:
       BlEffect: true
       DisplayPc: true
       StopAttacking: true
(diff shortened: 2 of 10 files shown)


src/map/navi.cpp
```diff
@@ -333,9 +333,9 @@ int32 map_type(const struct map_data * m) {
 
 	if (segmented && has_mob) {
 		return 5005;
-	} else if (segmented) {
-		return 5003;
-	}
+	}/* else if (segmented) {
```
vstumpf, line 336: Does this still work for a map like prontera_in, with the refinement npc? or pyramids b1?
https://github.com/rathena/rathena/pull/6023#issuecomment-1065989706

it might also be packetver-dependent

Lemongrass3110: Can you check if it works more correct, if you apply #9551?

Issue #9452, opened 2025-07-09
[Pull Request] Fix Hat Effects to work on NPCs and mobs


TBlazeWarriorT:
### Is your request related to a missing official feature?

Maybe?

### Describe the solution you'd like

It was hardcoded to only work on players for seemingly no reason, it always bothered me.
I've managed to implement it on my local fork, and the code was inspired by Tokei's code on a forum post from a few years ago, which I rewrote.
I did not run into any issues on my fork so far, and the code seems clean to me, should be a direct upgrade hopefully.

**Pull Request link**: https://github.com/rathena/rathena/pull/9453

PR #9453, opened 2025-07-09
Add hat effects to all units

TBlazeWarriorT:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9452

* **Server Mode**: Both

Changes hat effects (including script commands) to also work on NPCs and mobs. The client seems fully capable of rendering hat effects on npcs, moving npcs and mobs, and this finally makes it usable.

_"It was hardcoded to only work on players for seemingly no reason, it always bothered me.
I've managed to implement it on my local fork, and the code was inspired by Tokei's code on a forum post from a few years ago, which I rewrote.
I did not run into any issues on my fork so far, and the code seems clean to me, should be a direct upgrade hopefully."_

image


TBlazeWarriorT:
Working on fixing conflicts rn, my 'master' was outdated by accident when I created the pull request. File preview was looking fine, but I need to apply the newest commits from master.

src/map/clif.cpp
```diff
@@ -21390,58 +21394,49 @@ void clif_navigateTo(map_session_data *sd, const char* mapname, uint16 x, uint16
 
 /// Send hat effects to the client.
 /// 0A3B <Length>.W <AID>.L <Status>.B { <HatEffectId>.W } (ZC_EQUIPMENT_EFFECT)
-void clif_hat_effects( map_session_data& sd, block_list& bl, enum send_target target ){
+void clif_hat_effects(struct block_list* src, struct block_list* bl, bool enable, enum send_target target) {
```
HopeItBuilds:
```suggestion
void clif_hat_effects(struct block_list* src, struct block_list* bl, bool enable, enum send_target target) {
```

doc/script_commands.txt
```diff
@@ -11211,12 +11211,17 @@ solution rather than sending the map and the monster_id.
...
-*hateffect(<Hat Effect ID>,<State>);
+*hateffect(<hat effect id>,<state>{,<target id>});
 
-This will set a Hat Effect onto the player. The state field allows you to
-enable (true) or disable (false) the effect on the player.
-The Hat Effect constants can be found in 'src/map/script_constants.hpp' starting
-with HAT_EF_*.
+Enable or disables a Hat Effect onto the specified target unit.
+
+The Hat Effect constants can be found in 'src/map/script.hpp' starting with HAT_EF_*.
```
HopeItBuilds:
```suggestion
The Hat Effect constants can be found in 'src/map/script_constants.hpp' starting with HAT_EF_*.
```

src/map/script.cpp
```diff
@@ -24757,39 +24757,31 @@ BUILDIN_FUNC(recalculatestat) {
...
 	int16 effectID = script_getnum(st,2);
 	bool enable = script_getnum(st,3) ? true : false;
 
-	if( effectID <= HAT_EF_MIN || effectID >= HAT_EF_MAX ){
-		ShowError( "buildin_hateffect: unsupported hat effect id %d\n", effectID );
-		return SCRIPT_CMD_FAILURE;
-	}
-
-	auto it = util::vector_get( sd->hatEffects, effectID );
+	// This is unecessary and annoying half the time
```
HopeItBuilds: I don't think disabling this is correct

TBlazeWarriorT: I think I trust Tokei on this, this was from his original source mod.
I just tested an NPC and a mob with
```
hateffect 4123976, 1, getnpcid(0);
hateffect -1, 1, getnpcid(0);
```
And it causes no errors. Just silently fails if the client does not know that hateffect. Seems fine to me. If there's a side effect, I couldn't find it.

TBlazeWarriorT:
Seems like player hat effects aren't rendering after this, currently investigating. NPCs/Mob changes worked just fine so far.

TBlazeWarriorT:
Nvm, the fix was a single character. Should be good now.

src/map/script.cpp
```diff
@@ -24770,24 +24770,31 @@ BUILDIN_FUNC(hateffect){
...
+	if( ud == nullptr ){
+		ShowError( "buildin_hateffect: unsupported unit type %d\n", bl->type );
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	auto it = util::vector_get( ud->hatEffects, effectID );
 
 	if( enable ){
-		if( it != sd->hatEffects.end() ){
+		if( it != ud->hatEffects.end() ){
```
Atemo: Unrelated to the PR but `vector_exists` could be used instead of `vector_get`

src/map/script.cpp
```diff
@@ -24770,24 +24770,31 @@ BUILDIN_FUNC(hateffect){
...
 			return SCRIPT_CMD_SUCCESS;
 		}
 
-		sd->hatEffects.push_back( effectID );
+		ud->hatEffects.push_back( effectID );
 	}else{
-		if( it == sd->hatEffects.end() ){
+		if( it == ud->hatEffects.end() ){
```
Atemo: Unrelated to the PR but `vector_exists` could be used instead of `vector_get`

src/map/clif.cpp
```diff
@@ -21390,58 +21393,53 @@ void clif_navigateTo(map_session_data *sd, const char* mapname, uint16 x, uint16
...
-
-	if( target == SELF ){
-		tsd = BL_CAST(BL_PC,&bl);
-		tbl = &sd;
-	}else{
-		tsd = &sd;
-		tbl = &bl;
-	}
-
-	if( tsd == nullptr ){
```
Atemo, line 21406: Didn't we change the logic a bit by deleting the lines above?

src/map/script.cpp
Lemongrass3110: Fixed

src/map/script.cpp
Lemongrass3110: Fixed

src/map/clif.cpp
Lemongrass3110: Yes, but I guess intentional.
The caller should know how to call it and provide the correct arguments.
I cleaned it up a little more.

Issue #9209, opened 2025-03-28
Some nodamage skills should have partial damage traits


Playtester:
### rAthena Hash

2954154

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

The skills:
MO_ABSORBSPIRITS (on success vs. mob)
SA_DEATH
PF_MINDBREAKER
RG_STEALCOIN (on success)
DC_WINKCHARM (on success vs. mob)
SM_PROVOKE 

Don't trigger various additional effects (some work but not all).

### Relevant Log Output

```Shell

```

### Expected Result

The skills:
MO_ABSORBSPIRITS (on success vs. mob)
SA_DEATH
PF_MINDBREAKER
RG_STEALCOIN (on success)
DC_WINKCHARM (on success vs. mob)
SM_PROVOKE 

Trigger the "attacked" event for normal attacks with damage=0, which results in the following:

1.
Resets: EFFECTSTATE_BURROW, EFFECTSTATE_HIDING, EFST_NOACTION (NOACT_PROVOCATION, NOACT_SMOKING), EFST_SIEGEMODE
The same function also resets EFST_DC_WINKCHARM, but only if damage > 1, so Winkcharm and the other skills listed above don't end Winkcharm.

2.
Sets target on MONSTER_TYPE_06 (I think it's for Alchemist Marine Sphere trigger).
Damage log entry with "0 damage" (which causes Exp Tap bonus).

3.
Causes: Coma, SP Destruction (Dark Priest Card), Polymorph, Autospell (due to being flagged as normal attack), Auto Bonus (Vanberk Card, etc.)

4.
Stops movement on target.

5.
Removes Freeze, Stone and Sleep.

What it does NOT cause are effect from "bAddEff" bonuses. The code for those is only applied when damage is modified, which is not the case for above skills.

### How to Reproduce

Freeze a monster, use one of those skills -> Freeze is removed.
Use one of those skills, kill monster with another character -> 25% more Exp
etc.

### Official Information

After implementing 2954154 we noticed that some skills do give Exp tap bonus on official servers even though they deal no damage. Reason for this is that they invoke the "attacked" event.

We verified that all skills above can trigger several of the events mentioned in the expected results.

Playtester:
Important extra note: Provoke usually triggers autospells both offensive as well as counter - but the chance is halved due to it being a ranged skill. It cannot trigger Poison React autocast as that happens in the damage calculation code which is skipped here.

[...]

PR #9454, opened 2025-07-09
Zero damage Skills Breaking SC

Playtester:
* **Addressed Issue(s)**: Related to #9209 

* **Server Mode**: Both

- Zero damage skills like Wink Charm can now break status changes that break on damage
- Zero damage skills like Wink Charm now count as Exp Tap even if disabled for "Nodamage" skills
- Unified special behavior of Wink Charm, Voice of Siren and Deep Sleep
  * For PC: These status changes end on zero damage skills
  * For Non-PC: These status changes only end when actual damage was dealt
- Ensured that existing behavior for damage skills is not affected
  * When an attack misses, is blocked or is fully resisted, no damage event will be triggered
  * SP/AP damage will never break status changes that break on damage
- Currently the only zero damage skill implemented is Wink Charm, but can easily be expanded to other skills
- SP/AP damage will no longer trigger anything (status effects, autocasts, etc.)
- Related to #9209

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2777,7 +2777,7 @@ void mob_log_damage(mob_data* md, block_list* src, int64 damage, int64 damage_ta
 //Call when a mob has received damage.
 void mob_damage(struct mob_data *md, struct block_list *src, int32 damage)
 {
-	if (src && damage > 0) { //Store total damage...
+	if (src) { //Store total damage...
 		//Log damage
 		mob_log_damage(md, src, static_cast<int64>(damage));
 	}
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2777,7 +2777,7 @@ void mob_log_damage(mob_data* md, block_list* src, int64 damage, int64 damage_ta
 //Call when a mob has received damage.
 void mob_damage(struct mob_data *md, struct block_list *src, int32 damage)
 {
-	if (src) { //Store total damage...
+	if (src != nullptr) { //Store total damage...
 		//Log damage
 		mob_log_damage(md, src, static_cast<int64>(damage));
 	}
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -1537,7 +1537,7 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp, 
 		ap = status->ap;
 	}
 
-	if (!hp && !sp && !ap)
+	if (hp == 0 && sp == 0 && ap == 0 && (flag&1))
 		return 0;
 
 	if( !status->hp )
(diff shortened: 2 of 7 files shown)


src/map/mob.cpp
```diff
@@ -2777,7 +2777,7 @@ void mob_log_damage(mob_data* md, block_list* src, int64 damage, int64 damage_ta
 //Call when a mob has received damage.
 void mob_damage(struct mob_data *md, struct block_list *src, int32 damage)
 {
-	if (src && damage > 0) { //Store total damage...
+	if (src) { //Store total damage...
```
Lemongrass3110:
```suggestion
	if (src != nullptr) { //Store total damage...
```

src/map/status.hpp
```diff
@@ -3521,11 +3521,11 @@ static int32 status_fix_damage( struct block_list *src, struct block_list *targe
 }
```
Lemongrass3110, line 3521: I think it will be confusing, if status_fix_damage is active damage and status_fix_spdamage+status_fix_apdamage are passive damage.
Is that really what we want?

src/map/status.cpp
```diff
@@ -1547,13 +1547,18 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp,
 	if (hp && battle_config.invincible_nodamage && src && sc && sc->getSCE(SC_INVINCIBLE))
 		hp = 1;
 
-	if( hp && !(flag&1) ) {
+	if (!(flag&1)) {
```
Lemongrass3110, line 1552:
```suggestion
	// If the damage is not passive
	if (!(flag&1)) {
```

src/map/status.cpp
```diff
@@ -1547,13 +1547,18 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp,
...
 		if( sc ) {
 			struct status_change_entry *sce;
 
 			for (const auto &it : status_db) {
 				sc_type type = static_cast<sc_type>(it.first);
 
+				// Wink Charm checks for damage so it doesn't end itself
+				// Other status changes end even on 0 damage (e.g. Wink Charm ends Freeze)
+				if (type == SC_WINKCHARM && hp == 0)
+					continue;
```
Lemongrass3110: Should be moved into the next if

src/map/status.cpp
```diff
@@ -1537,7 +1537,7 @@ int32 status_damage(struct block_list *src,struct block_list *target,int64 dhp,
 		ap = status->ap;
 	}
 
-	if (!hp && !sp && !ap)
+	if (hp == 0 && sp == 0 && ap == 0 && (flag&1))
```
Lemongrass3110, line 1541:
```suggestion
	// If no damage is dealt and the damage was passive
	if (hp == 0 && sp == 0 && ap == 0 && (flag&1))
```

src/map/status.hpp
Playtester: Flag&1 was already defined as to mean to not end status changes.

Previously that was the case because the check we did was:
`if( hp && !(flag&1) ) {`

So the description was already a bit off because even if you called status damage without the "do not end status changes" flag, it still didn't end the status changes.

Now I fixed it and the flag means exactly what it says. Going by this, we need to ensure that that flag is always set to 1 for SP and AP damage. I thought the easiest way to ensure this is to make it impossible through the function calls in the first place.

Alternatively we could also completely remove these functions and just use `status_zap`. What do you think?

Playtester: After confirming that SP and AP damage don't trigger anything that would require a "src", I removed those damage function completely and use status_zap instead now.

Lemongrass3110 (approved):

PR #9458, opened 2025-07-15
Add has_hateffect

Milchdealer:
* **Addressed Issue(s)**: 

* **Server Mode**:  Both

* **Description of Pull Request**: Add script command which checks if a given hat effect is active on a player.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -24795,6 +24795,36 @@ BUILDIN_FUNC(hateffect){
 }
 
 /**
+ * Check if hat effect is enabled on player
+ * *has_hateffect(<effect_id>{,<char_id>});
+ */
+BUILDIN_FUNC(has_hateffect) {
+#if PACKETVER_MAIN_NUM >= 20150507 || PACKETVER_RE_NUM >= 20150429 || defined(PACKETVER_ZERO)
+	map_session_data* sd;
+
+	if (!script_charid2sd(3, sd))
+		return SCRIPT_CMD_FAILURE;
+
+	int16 effectID = script_getnum(st, 2);
+
+	if (effectID <= HAT_EF_MIN || effectID >= HAT_EF_MAX) {
+		ShowError("buildin_has_hateffect: unsupported hat effect id %d\n", effectID);
+		script_pushint(st, 0);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	auto it = util::vector_get(sd->hatEffects, effectID);
+
+	if (it != sd->hatEffects.end()) {
+		script_pushint(st, 1);
+	} else {
+		script_pushint(st, 0);
+	}
+#endif
+	return SCRIPT_CMD_SUCCESS;
+}
+
+/**
 * Retrieves param of current random option. Intended for random option script only.
 * getrandomoptinfo(<type>);
 * @author [secretdataz]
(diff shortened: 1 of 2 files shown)


src/map/script.cpp
```diff
@@ -24794,6 +24794,36 @@ BUILDIN_FUNC(hateffect){
...
+
+	if (effectID <= HAT_EF_MIN || effectID >= HAT_EF_MAX) {
+		ShowError("buildin_has_hateffect: unsupported hat effect id %d\n", effectID);
+		script_pushint(st, 0);
+		return SCRIPT_CMD_FAILURE;
+	}
+
+	auto it = util::vector_get(sd->hatEffects, effectID);
+
+	if (it != sd->hatEffects.end()) {
```
Atemo: Uses `vector_exists` instead

src/map/script.cpp
```diff
@@ -24794,6 +24794,36 @@ BUILDIN_FUNC(hateffect){
...
+		script_pushint(st, 1);
+	} else {
+		script_pushint(st, 0);
+	}
+#endif
+	return SCRIPT_CMD_SUCCESS;
```
Atemo:
```suggestion
	return SCRIPT_CMD_SUCCESS;
#else
	ShowError( "buildin_has_hateffect: This command requires PACKETVER 2015-05-27 or newer.\n" );
	return SCRIPT_CMD_FAILURE;
#endif
```

src/map/script.cpp
```diff
@@ -24794,6 +24794,36 @@ BUILDIN_FUNC(hateffect){
...
+/**
+ * Check if hat effect is enabled on player
+ * *has_hateffect(<effect_id>{,<char_id>});
+ */
+BUILDIN_FUNC(has_hateffect) {
+#if PACKETVER_MAIN_NUM >= 20150507 || PACKETVER_RE_NUM >= 20150429 || defined(PACKETVER_ZERO)
+	map_session_data* sd;
+
+	if (!script_charid2sd(3, sd))
+		return SCRIPT_CMD_FAILURE;
```
Atemo, line 24798: SCRIPT_CMD_FAILURE doesn't stop the script. 0 must be pushed into the stack

rabbit-lunatic:
A silly question, wouldn't this affect those who use costumes equipment with effects?
Example: https://www.divine-pride.net/database/item/5979

Milchdealer:
> A silly question, wouldn't this affect those who use costumes equipment with effects? Example: https://www.divine-pride.net/database/item/5979

What do you mean by affect? It would just be returned as `true`

src/map/script.cpp
Milchdealer: Do we want it to stop? Also, does `script_pushint(st, 0)` stop it. I thought it's `st->state = END;`

Lemongrass3110:
I would like to merge #9453 first before working on this one.
Because then it is not only about a player, but units in general.
Introducing a player specific command now and changing it to unit a few weeks later, does not really make sense to me.

Milchdealer:
> I would like to merge #9453 first before working on this one. Because then it is not only about a player, but units in general. Introducing a player specific command now and changing it to unit a few weeks later, does not really make sense to me.

I have a script command that does this. I just need to noodle it out of my custom server. I can contribute it maybe this weekend. Alternatively I can wait for #9453 to merge and then adjust this PR

Lemongrass3110 (approved):

PR #9336, opened 2025-05-16
Add support for new stylist packet

Balferian:
* **Addressed Issue(s)**: #8729 
* **Server Mode**: Renewal
* **Description of Pull Request**: 
This update allow you to use stylist UI again for new clients.
  - Added packet 0x0bf7.
  - Updated stylist db.

Thanks to @kianzack for help.

PR #9467, opened 2025-07-26
Initial version of 2024 bodystyle support

Lemongrass3110:

* **Server Mode**: Both

Adds basic support for bodystyle in 2024+ clients.
Feedback is welcome.
Once this PR has been merged, we can add more newer features.


src/map/atcommand.cpp
```diff
@@ -1957,34 +1957,33 @@ ACMD_FUNC(model)
...
+		clif_displaymessage( fd, msg_txt( sd, 739 ) ); // Please enter a body style (usage: @bodystyle <job ID>).
 		return -1;
 	}
 
-	if (body_style >= MIN_BODY_STYLE && body_style <= MAX_BODY_STYLE) {
-		pc_changelook(sd, LOOK_BODY2, body_style);
-		clif_displaymessage(fd, msg_txt(sd,36)); // Appearence changed.
-	} else {
-		clif_displaymessage(fd, msg_txt(sd,37)); // An invalid number was specified.
+	if( body_style != sd->status.class_ && !util::vector_exists( job->alternate_outfits, body_style ) ){
```
Atemo, line 1977: shouldn't it be 
```suggestion
	if( body_style == sd->status.class_ || !util::vector_exists( job->alternate_outfits, body_style ) ){
```
?

Lemongrass3110: I think with the new logic I also need to change it to allow changing back to the normal job

rabbit-lunatic:
Have you checked if it also works with clones?

Lemongrass3110:
> Have you checked if it also works with clones?

@rabbit-lunatic just tried and didnt encounter any problem. are you encountering a problem with clones in this PR?

brynner23:
Nice it works fine with other jobs, even when switching to jobs below 3rd class. no palette errors using old stylist npc.

rabbit-lunatic:
> > Have you checked if it also works with clones?
> 
> @rabbit-lunatic just tried and didnt encounter any problem. are you encountering a problem with clones in this PR?

I tested it here and it works, clone and disguise.
However, I tested it in StylistUI and it gave me a packet error.

```
[Warning]: clif_parse: Received unsupported packet (packet 0x0bf7, 14 bytes received), disconnecting session #3.
```

Lemongrass3110:
> > > Have you checked if it also works with clones?
> > 
> > 
> > @rabbit-lunatic just tried and didnt encounter any problem. are you encountering a problem with clones in this PR?
> 
> I tested it here and it works, clone and disguise. However, I tested it in StylistUI and it gave me a packet error.
> 
> ```
> [Warning]: clif_parse: Received unsupported packet (packet 0x0bf7, 14 bytes received), disconnecting session #3.
> ```

Stylist is not supported. This will be handled in and added with #9336 separate from this PR.
But thank you for testing and the positive feedback. :-)

src/map/atcommand.cpp
Lemongrass3110:
```suggestion
	// If it is not the actual job class and not any of the alternate outfits available for that job
	if( body_style != sd->status.class_ && !util::vector_exists( job->alternate_outfits, body_style ) ){
```
@Atemo the check is correct in my opinion, should i make it more clear with a comment?
Basically it means allow you to change your appearance back to your "normal" job or to one of the alternate outfits (defined).

Atemo: Ha yes, maybe we could also check
```
if( body_style == sd->status.body ){
```
somewhere

src/map/pc.cpp
```diff
@@ -11127,7 +11125,9 @@ void pc_changelook(map_session_data *sd,int32 type,int32 val) {
 		sd->setlook_robe = val;
 		break;
 	case LOOK_BODY2:
-		val = cap_value(val, MIN_BODY_STYLE, MAX_BODY_STYLE);
+		if( !job_db.exists( val ) ){
+			return;
```
Atemo, line 11138: Should we add a showerror at this point?

db/re/job_outfits.yml
```diff
@@ -0,0 +1,171 @@
...
+      ArchBishop_2nd: true
+      Mechanic_2nd: true
+      Guillotine_Cross_2nd: true
+      Royal_Guard_2nd: true
+      Sorcerer_2nd: true
+      Minstrel_2nd: true
+      Wanderer_2nd: true
+      Sura_2nd: true
+      Genetic_2nd: true
+      Shadow_Chaser_2nd: true
```
aleos89: Does this group not have an alternate? Do they have to be defined if the list is empty by default?

Lemongrass3110: Hmm actually they could be removed.
I thought I had implemented it, so that those jobs have to be "available" in the job db.
But I just look up the constant and if it exists, I allow it.

src/map/pc.cpp
Lemongrass3110: This function doesnt give any error messages, that is why I did not add one.
The whole process with changelook, setlook, etc needs more refactoring in the future, so i would leave this for then.
See also my TODO comment in BUILDIN_FUNC(setlook)

src/map/atcommand.cpp
Lemongrass3110: Done!

Introvert-RO:
This look good, wait for this PR get merged. Thank for your hardwork.

Kanzero-Dev:
setunitdata UMOB_CLASS doesn't work, it's still Novice.

db/re/job_outfits.yml
brynner23: If you remove this, it won’t change to the alternative outfit. You can only change to 3rd job outfit and 4th job but not for the alternative outfit.image

sinju92:
> setunitdata UMOB_CLASS doesn't work, it's still Novice.

UNPC_CLASS have same issue too, when npc walking they become novice when npc stop back to origin sprite should be

gidzdlcrz:
Can we update this PR to latest? Thanks!

gidzdlcrz:
To anyone who wants to try this, I have updated this PR to match the latest commit. 
You can simply run `git apply bodystyle2024.diff` since this PR is very tedious to apply manually.

Download Here: [bodystyle2024.zip](https://github.com/user-attachments/files/23004623/bodystyle2024.zip)

mazvi:
waiting :)

Issue #9464, opened 2025-07-20
Refactor Skills to use Strategy pattern


Danil0v3s:
### Is your request related to a missing official feature?

No

### Describe the solution you'd like

> [!WARNING]
> Please refer to the actual Pull Requests [here](https://github.com/rathena/rathena/milestone/34) if you wish to collaborate or understand what's happening.

<details>
  <summary>Old very long AI generated plan</summary>
Been thinking about this idea for long now. Decide to ask some LLMs to help me think through it and trace a plan. The idea is to create an abstraction around the Skill class in order to remove the gigantic switch cases we have all around battle.cpp... Following is the plan:

Here is the complete plan for your review and approval.

### **Refactoring Plan: From `switch` to Strategy Pattern**

This plan outlines the refactoring of the skill system in [`battle.cpp`](src/map/battle.cpp:1) to a more maintainable and extensible object-oriented design using the Strategy pattern.

#### 1. The Problem: Large `switch` Statements

The current implementation, particularly in functions like [`battle_calc_attack_skill_ratio`](src/map/battle.cpp:4638), relies on a large `switch` statement to handle different skills. This makes the code difficult to maintain and extend.

#### 2. The Solution: A `Skill` Abstraction

I propose a new abstract base class, `Skill`, which will define a common interface for all skills. This will allow us to treat all skills uniformly, regardless of their specific implementation.

Here is the proposed `skill.hpp`:

```cpp
#ifndef SKILL_HPP
#define SKILL_HPP

#include <memory>

// Forward declarations
class Battle;
struct block_list;
struct Damage;

class Skill {
public:
    virtual ~Skill() = default;

    virtual bool can_use(const block_list& source, const block_list& target) const = 0;
    virtual void calculate_damage(Damage& wd, const block_list& source, const block_list& target, int skill_lv) const = 0;
    virtual void apply_effects(Battle& battle, block_list& source, block_list& target, int tick) const = 0;
    virtual void execute(Battle& battle, block_list& source, block_list& target, int skill_lv, int tick) const;
};

#endif // SKILL_HPP
```

#### 3. Concrete Skill Implementations

Each skill will be implemented as a concrete subclass of `Skill`. This encapsulates the logic for each skill in its own class. For example:

*   `NormalAttackSkill`
*   `BashSkill`
*   `HealSkill`
*   `ShieldBoomerangSkill`

#### 4. Refactoring `battle.cpp`

[...]

PR #9468, opened 2025-07-26
Add a Strategy Pattern for skills

Danil0v3s:
* **Addressed Issue(s)**: 
* Gigantic switch cases on the codebase, namely on battle.cpp and skill.cpp

* **Server Mode**: N/D

* This aims to create the boilerplate necessary to have the skills using a strategy pattern where we define classes/files for each skill.
* The implementation had some changes since I presented it at https://github.com/rathena/rathena/issues/9464. @Lemongrass3110 suggested I kept it to a minimal on this PR and then create more methods as we need them
* This implements SM_BASH and MS_BASH as an example

https://github.com/user-attachments/assets/66851242-158d-442b-a169-2c10487fb535

### DIFF
--- a/src/map/skills/skill_factory.cpp
+++ b/src/map/skills/skill_factory.cpp
@@ -9,13 +9,13 @@
 #include "./mercenary/skill_factory_mercenary.hpp"
 #include "./swordsman/skill_factory_swordsman.hpp"
 
-SkillFactoryImpl::SkillFactoryImpl() : factories_{
-	std::make_shared<SkillFactoryMercenary>(),
-	std::make_shared<SkillFactorySwordsman>()
-} {}
-
 std::unique_ptr<const SkillImpl> SkillFactoryImpl::create(const e_skill skill_id) const {
-	for (const std::shared_ptr<SkillFactory>& factory : factories_) {
+	static const std::vector<std::shared_ptr<SkillFactory>> factories = {
+		std::make_shared<SkillFactoryMercenary>(),
+		std::make_shared<SkillFactorySwordsman>()
+	};
+
+	for (const std::shared_ptr<SkillFactory>& factory : factories) {
 		if (std::unique_ptr<const SkillImpl> impl = factory->create(skill_id); impl != nullptr) {
 			return impl;
 		}
--- a/src/map/skills/skill_factory.hpp
+++ b/src/map/skills/skill_factory.hpp
@@ -21,9 +21,5 @@ public:
 
 class SkillFactoryImpl : public SkillFactory, public rathena::util::Singleton<SkillFactoryImpl, SkillFactory> {
 public:
-	SkillFactoryImpl();
 	virtual std::unique_ptr<const SkillImpl> create(const e_skill skill_id) const override;
-
-private:
-	std::vector<std::shared_ptr<SkillFactory>> factories_{};
 };


src/map/skills/weapon/bash.hpp
```diff
@@ -0,0 +1,93 @@
...
+ */
+class BashSkill : public WeaponSkill {
+public:
+    // Core identification
+    uint16 get_skill_id() const override { 
+        return SM_BASH; // Note: MS_BASH shares same implementation
+    }
+    
+    const char* get_skill_name() const override { 
+        return "Bash";
```
vstumpf: This should come from skilldb, no?

src/map/skills/weapon/bash.hpp
```diff
@@ -0,0 +1,93 @@
...
+            status_change_start(src, target, SC_STUN, stun_chance, skill_lv, 0, 0, 0,
+                              skill_get_time2(get_skill_id(), skill_lv), SCSTART_NONE);
+        }
+    }
+
+    // ========== VALIDATION ==========
+    bool can_use(const block_list& source, const block_list& target) const override {
+        // Bash requires a melee weapon (not bare hands)
+        if (source.type == BL_PC) {
+            const map_session_data* sd = (const map_session_data*)&source;
```
vstumpf: here you probably want something like
`auto sd = static_cast<map_session_data&>(bl);`

src/map/skills/weapon/bash.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include "pc.hpp"
+
+/**
+ * BashSkill - Implementation of SM_BASH and MS_BASH
+ * A basic melee attack skill that deals increased damage and has stun effect
+ */
+class BashSkill : public WeaponSkill {
+public:
+    // Core identification
+    uint16 get_skill_id() const override {
```
vstumpf: All skills have a skill id, so make the skill id a member of BattleSkill, add it to the constructor of all the base skills.
something like https://github.com/vstumpf/rathena/compare/refactor/cmake...vstumpf:rathena:refactor/skills#diff-d259011f9c18d10972c5036f69f7aee74575862f7ce3123ad3ea7527b759d711R11
https://github.com/vstumpf/rathena/compare/refactor/cmake...vstumpf:rathena:refactor/skills#diff-590e433f1e9ec7966ddb6a25dc976716c8e3c19261e4b517c2c657c59f8848ceR24

src/map/battle_skill_factory.hpp
```diff
@@ -0,0 +1,110 @@
...
+ * Uses singleton pattern with thread-safe skill registration and caching
+ */
+class BattleSkillFactory
+{
+private:
+    // Skill creators - functions that create skill instances
+    std::unordered_map<uint16, std::function<std::shared_ptr<BattleSkill>()>> skill_creators;
+
+    // Skill cache - stores created instances for performance
+    mutable std::unordered_map<uint16, std::shared_ptr<BattleSkill>> skill_cache;
```
vstumpf: I don't get this, why do we need a skill_id -> "creator" func map, and a skill_id -> shared_ptr<skill> map? Why doesn't registering the skill just create the shared_ptr<skill> and we're done with it?

src/map/battle_skill_factory.cpp
```diff
@@ -0,0 +1,80 @@
...
+    std::lock_guard<std::mutex> lock(cache_mutex);
+    skill_cache.clear();
+}
+
+void BattleSkillFactory::register_all_skills()
+{
+    register_weapon_skills();
+    register_magic_skills();
+    register_misc_skills();
+    register_passive_skills();
```
vstumpf: personally i disagree with this split, i think we should split it by jobs instead of the type of skill. a job has 10/20 skills, which makes it easier to find

src/map/battle_skill.hpp
```diff
@@ -0,0 +1,234 @@
+#ifndef BATTLE_SKILL_HPP
+#define BATTLE_SKILL_HPP
+
+#include <memory>
+#include "battle.hpp"
+#include "status.hpp"
+
+// Forward declarations
+
+class BattleSkill {
```
vstumpf: Why BattleSkill, why not just `Skill`?

src/map/battle_skill_factory.hpp
Danil0v3s: Probably forgot to remove these, cursor suggested to have some "mechanism" of measuring cache hits/ad-hoc creations, i deleted some of it but not these it sesms. Will fix 👍🏼

src/map/skills/weapon/bash.hpp
Danil0v3s: Will do. Couldn't also find where to plug this method, searched both battle and skill. Where would I use this one? 🤔

src/map/battle_skill.hpp
Danil0v3s: Because then we would have to create a skill.cpp and map already has it. Could get confusing?

src/map/battle_skill_factory.cpp
Danil0v3s: I understand and looking aat your implementation it seems to me your approach went a couple steps before mine, actually going into the initial step of the skills...

My fear is that by creating these aggregators with the actual implementation these could grow big with 10/20 skills including everything from clif to calculations... Not to mention I could easily forget about something when going that route (eg: when i look at some methods there's a truck load of ad-hoc checks before the switches). Also going that approach we would have each skill implementation being responsible for their clif which i'd argue it's not a skill responsibility. 
I would say the skill would be just the "pipeline" which base damage and status etc goes through, then other method is responsible to actually sending that back.

What do you say?

Danil0v3s: What I could do to meet in the middle though is create the "Jobs" class as an aggregator, without the actual skill implementation, just a proxy. How does it sound?

vstumpf: I think we're misunderstanding each other
I think the WeaponSkill, MagicSkill, RangedWeaponSkill, PassiveSkill, etc, classes are good.
I just think registering the skills (and the files themselves honestly) should be organized by jobs

src/map/battle_skill.hpp
vstumpf: Ah makes sense, that's why I had it all in a subdir

src/map/battle_skill_factory.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void BattleSkillFactory::register_weapon_skills()
+{
+    register_skills<BashSkill>({SM_BASH, MS_BASH});
```
vstumpf: Personally I don't like this, I think SM_BASH and MS_BASH should be two different classes
You could make BashSkill and MercenaryBashSkill inherit BashSkill

src/map/battle_skill_factory.hpp
```diff
@@ -0,0 +1,98 @@
...
+    // Destructor
+    ~BattleSkillFactory() = default;
+
+    /**
+     * Register a skill type with template for type safety
+     * @tparam SkillType The skill class type
+     * @param skill_id The skill ID to register
+     */
+    template <typename SkillType>
+    void register_skill(uint16 skill_id)
```
vstumpf: Can't we just take a shared_ptr here?
```
void register_skill(uint16 skill_id, const std::shared_ptr<BattleSkill>& skill) {
  skill_db[skill_id] = skill;
}
```
this way, we can save the compiler work having to generate this function for EVERY skill

src/map/battle_skill_factory.hpp
```diff
@@ -0,0 +1,98 @@
...
+    {
+        static_assert(std::is_base_of_v<BattleSkill, SkillType>, "SkillType must inherit from BattleSkill");
+
+        if (skill_db.find(skill_id) != skill_db.end())
+        {
+            throw SkillAlreadyImplementedException();
```
vstumpf: I think we should overwrite, imagine if someone wants to replace the default SM_BASH with `MyCustomBash`, we should let them!

vstumpf (commented):
Also, as a general rule, I think we're using camel case for member functions, like `getSkill` or `canUse`

src/map/battle_skill_factory.hpp
Danil0v3s: I've added at [f59db4e](https://github.com/rathena/rathena/pull/9468/commits/f59db4e6ce60d13dc7eabea5f60ed0ffce52e8a1)

src/map/battle.cpp
```diff
@@ -3270,12 +3271,13 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
 #endif
 
 	if(skill_id) {
+		auto skill = BattleSkillFactory::instance().getSkill(skill_id);
+        if (skill) {
```
Lemongrass3110:
```suggestion
        if (skill != nullptr) {
```

src/map/battle.cpp
```diff
@@ -3270,12 +3271,13 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
...
+            skill->modifyHitRate(hitrate, src, target, skill_lv);
+            return (rnd() % 100) < hitrate;
```
Lemongrass3110:
```suggestion
```
Should simply continue with the existing code.

src/map/battle.cpp
Lemongrass3110: indentation looks wrong?

src/map/battle.cpp
```diff
@@ -4696,11 +4698,12 @@ static int32 battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list
 		}
 	}
 
+	auto skill = BattleSkillFactory::instance().getSkill(skill_id);
+	if (skill) {
+		return skill->calculateSkillRatio(wd, src, target, skill_lv, skillratio);
```
Lemongrass3110:
```suggestion
		skill->calculateSkillRatio(wd, src, target, skill_lv, skillratio);
```
Should simply continue with the existing code.

src/map/battle.cpp
```diff
@@ -4696,11 +4698,12 @@ static int32 battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list
 		}
 	}
 
+	auto skill = BattleSkillFactory::instance().getSkill(skill_id);
+	if (skill) {
```
Lemongrass3110:
```suggestion
	if (skill != nullptr) {
```

src/map/battle_skill_factory.cpp
Danil0v3s: Would this be enough? [9cf82ed](https://github.com/rathena/rathena/pull/9468/commits/9cf82ed242c22d49dd0ff6eadf0ca2bd2de37f34)

src/map/skill.cpp
```diff
@@ -1310,6 +1311,8 @@ int32 skill_additional_effect( struct block_list* src, struct block_list *bl, ui
 		}
 
 		if( skill_id ) {
+
+
```
```suggestion
```
Looks like a leftover

src/map/map.cpp
```diff
@@ -52,6 +52,7 @@
 #include "quest.hpp"
 #include "storage.hpp"
 #include "trade.hpp"
+#include "./skills/battle_skill_factory.hpp"
```
Lemongrass3110: Unused?

src/map/battle.cpp
```diff
@@ -33,6 +33,7 @@
 #include "pc.hpp"
 #include "pc_groups.hpp"
 #include "pet.hpp"
+#include "./skills/battle_skill_factory.hpp"
```
Lemongrass3110:
```suggestion
```
Not needed anymore?

src/map/CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ if( BUILD_SERVERS )
 message( STATUS "Creating target map-server" )
 file(GLOB MAP_HEADERS ${MAP_SOURCE_DIR}/*.hpp)
 file(GLOB MAP_SOURCES ${MAP_SOURCE_DIR}/*.cpp)
+file(GLOB_RECURSE MAP_SKILL_HEADERS ${MAP_SOURCE_DIR}/skills/*.hpp)
+file(GLOB_RECURSE MAP_SKILL_SOURCES ${MAP_SOURCE_DIR}/skills/*.cpp)
```
Lemongrass3110: @vstumpf doesnt like GLOBs, so you need to add all by hand here.

src/map/battle.cpp
```diff
@@ -3270,12 +3271,13 @@ static bool is_attack_hitting(struct Damage* wd, struct block_list *src, struct
 #endif
 
 	if(skill_id) {
+		auto skill = skill_db.find(skill_id)->impl;
```
Lemongrass3110: Not really a friend of auto...
Also you need to first check if the skill was found (!= nullptr), otherwise you might crash.

src/map/battle.cpp
```diff
@@ -4696,11 +4698,12 @@ static int32 battle_calc_attack_skill_ratio(struct Damage* wd, struct block_list
 		}
 	}
 
+	auto skill = skill_db.find(skill_id)->impl;
```
Lemongrass3110: Not really a friend of auto...
Also you need to first check if the skill was found (!= nullptr), otherwise you might crash.

src/map/skill.cpp
```diff
@@ -1336,6 +1339,12 @@ int32 skill_additional_effect( struct block_list* src, struct block_list *bl, ui
 	if( dmg_lv < ATK_DEF ) // no damage, return;
 		return 0;
 
+	auto skill = skill_db.find(skill_id)->impl;
```
Lemongrass3110: Not really a friend of auto...
Also you need to first check if the skill was found (!= nullptr), otherwise you might crash.

src/map/skill.cpp
```diff
@@ -1336,6 +1339,12 @@ int32 skill_additional_effect( struct block_list* src, struct block_list *bl, ui
...
+    if (skill) {
```
```suggestion
    if (skill != nullptr) {
```

src/map/skill.cpp
```diff
@@ -1336,6 +1339,12 @@ int32 skill_additional_effect( struct block_list* src, struct block_list *bl, ui
...
+        skill->applyAdditionalEffects(src, bl, skill_lv, tick, attack_type, dmg_lv);
+        return 0; // Let legacy code continue for other effects
```
```suggestion
```
Should simply continue with the existing code.

src/map/skills/battle_skill.cpp
```diff
@@ -0,0 +1,145 @@
...
+}
+
+void BattleSkill::applyPostDefenseModifiers(Damage* wd, const block_list* src, const block_list* target, uint16 skill_lv) const {
+    // Default implementation - no post-defense modification
+}
+
+void BattleSkill::execute(const block_list* src, block_list* target, uint16 skill_lv, t_tick tick) const {
+    // Default implementation - basic skill execution
+    // This would typically call the damage calculation pipeline
+}
```
Lemongrass3110: Missing empty newline

src/map/skills/battle_skill_categories.cpp
```diff
@@ -0,0 +1,21 @@
...
+int32 MagicSkill::getMagicElement(const block_list *src, const block_list *target, uint16 skill_lv, int32 mflag) const {
+    // Default magic skill element behavior
+    return ELE_NEUTRAL; // Can be overridden by specific magic skills
+}
+
+// MiscSkill implementations
+int32 MiscSkill::getMiscElement(const block_list *src, const block_list *target, uint16 skill_lv, int32 mflag) const {
+    // Default misc skill element behavior
+    return ELE_NEUTRAL; // Can be overridden by specific misc skills
+}
```
Lemongrass3110: Missing empty newline

src/map/skills/battle_skill_factory.cpp
```diff
@@ -0,0 +1,10 @@
+#include "battle_skill_factory.hpp"
+
+#include "./swordsman/swordsman.hpp"
+#include "./mercenary/mercenary.hpp"
+
+void BattleSkillFactory::registerAllSkills()
+{
+    Swordsman::register_skills();
+    Mercenary::register_skills();
+}
```
Lemongrass3110: Missing empty newline

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,31 @@
...
+public:
+    // Destructor
+    ~BattleSkillFactory() = default;
+
+    static void registerSkill(uint16 skill_id, const std::shared_ptr<BattleSkill>& skill)
+    {
+        auto skill_entry = skill_db.find(skill_id);
+        if (skill_entry) {
+            skill_entry->impl = skill;
+        }
```
Lemongrass3110: else: error message

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,31 @@
...
+ * Factory class for creating and managing BattleSkill instances
+ * Uses singleton pattern with thread-safe skill registration and caching
+ */
+class BattleSkillFactory
+{
+public:
+    // Destructor
+    ~BattleSkillFactory() = default;
+
+    static void registerSkill(uint16 skill_id, const std::shared_ptr<BattleSkill>& skill)
```
Lemongrass3110: Please only define it here and move the logic to the .cpp file

src/map/skills/mercenary/mercenary_bash.hpp
```diff
@@ -0,0 +1,91 @@
...
+    bool usesAmmo() const override
+    {
+        return false; // Bash doesn't consume ammo
+    }
+
+    int32 getRangeType(const block_list *src, const block_list *target, uint16 skill_lv) const override
+    {
+        return BF_SHORT; // Always short range
+    }
+};
```
Lemongrass3110: Missing empty newline

src/map/skills/swordsman/bash.hpp
```diff
@@ -0,0 +1,91 @@
...
+    bool usesAmmo() const override
+    {
+        return false; // Bash doesn't consume ammo
+    }
+
+    int32 getRangeType(const block_list *src, const block_list *target, uint16 skill_lv) const override
+    {
+        return BF_SHORT; // Always short range
+    }
+};
```
Lemongrass3110: Missing empty newline

src/map/skills/swordsman/swordsman.hpp
```diff
@@ -0,0 +1,13 @@
...
+#include "bash.hpp"
+
+class Swordsman
+{
+public:
+    static void register_skills()
+    {
+        BattleSkillFactory::registerSkill(SM_BASH, std::make_shared<BashSkill>());
+    }
+};
```
Lemongrass3110: Missing empty newline

src/map/skills/mercenary/mercenary.hpp
```diff
@@ -0,0 +1,13 @@
...
+#include "mercenary_bash.hpp"
+
+class Mercenary
+{
+public:
+    static void register_skills()
+    {
+        BattleSkillFactory::registerSkill(MS_BASH, std::make_shared<MercenaryBashSkill>());
+    }
+};
```
Lemongrass3110: Missing empty newline

src/map/battle.cpp
Danil0v3s: If i hit auto format, everything is going to be changed 😂 it looks correct to me?
Screenshot 2025-08-10 at 23 41 07

src/map/skills/mercenary/mercenary_bash.hpp
```diff
@@ -0,0 +1,45 @@
...
+    {
+        map_session_data *sd = getPlayerData(src);
+
+        // Stun effect only if skill level > 5 and has Fatal Blow
+        if (sd && skill_lv > 5 && pc_checkskill(sd, SM_FATALBLOW) > 0)
+        {
+            int32 stun_chance = (skill_lv - 5) * sd->status.base_level * 10;
+            status_change_start(src, target, SC_STUN, stun_chance, skill_lv, 0, 0, 0, skill_get_time2(getSkillId(), skill_lv), SCSTART_NONE);
+        }
+    }
```
Lemongrass3110:
```suggestion
```
This was not in the original file.

src/map/skills/mercenary/mercenary_bash.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "../battle_skill_categories.hpp"
+#include "../../battle.hpp"
+#include "../../pc.hpp"
+#include "../../skill.hpp"
+
+class MercenaryBashSkill : public WeaponSkill
+{
+public:
+    MercenaryBashSkill() : WeaponSkill(SM_BASH) {}
```
Lemongrass3110:
```suggestion
    MercenaryBashSkill() : WeaponSkill(MS_BASH) {}
```

src/map/skills/mercenary/mercenary_bash.hpp
```diff
@@ -0,0 +1,45 @@
...
+    // Core identification
+    uint16_t getSkillId() const override
+    {
+        return skill_id_;
+    }
+
+    const char *getSkillName() const override
+    {
+        return skill_get_name(skill_id_);
+    }
```
Lemongrass3110: Should be in the base class or maybe interlinked during factory initialization.

src/map/skills/battle_skill.cpp
```diff
@@ -0,0 +1,39 @@
...
+status_data* BattleSkill::getStatusData(const block_list* bl) {
+    if (bl) {
+        return status_get_status_data(const_cast<block_list&>(*bl));
+    }
+    return nullptr;
+}
+
+status_change* BattleSkill::getStatusChange(const block_list* bl) {
+    return status_get_sc(const_cast<block_list*>(bl));
+}
```
Lemongrass3110: These helpers are not really needed.
The respective skill can still use the normal macros/functions for these kind of things.

src/map/skills/mercenary/mercenary_bash.hpp
Danil0v3s: I'm removing the getSkillName altogether and moving the getSkillId to the default impl of the BattleSkill.cpp at [03075c8](https://github.com/rathena/rathena/pull/9468/commits/03075c80a08b9607e3a0f2d5cafbff8a46a1f746)

src/map/skills/swordsman/bash.hpp
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    void modifyHitRate(int16 &hit_rate, const block_list *src, const block_list *target, uint16 skill_lv) const override
+    {
+        hit_rate += hit_rate * 5 * skill_lv / 100; // +5% hit per level
+    }
+
+    void applyAdditionalEffects(block_list *src, block_list *target, uint16 skill_lv, t_tick tick, int32 attack_type, enum damage_lv dmg_lv) const override
+    {
+        map_session_data *sd = nullptr;
```
Lemongrass3110:
```suggestion
        map_session_data *sd = BL_CAST( BL_PC, src );
```

Danil0v3s: [36103c6](https://github.com/rathena/rathena/pull/9468/commits/36103c612eb84ccaa5c0db1ba65995b12597f64a)

src/map/CMakeLists.txt
Danil0v3s: @vstumpf should I? The lines above are using GLOBs, so do I change it on this PR or can I assume this change is out of the scope?

vstumpf: just leave it here, we'll fix it in https://github.com/rathena/rathena/pull/7722

src/map/skill.cpp
```diff
@@ -1310,6 +1311,7 @@ int32 skill_additional_effect( struct block_list* src, struct block_list *bl, ui
 		}
 
 		if( skill_id ) {
+
```
Lemongrass3110: Unchanged, please revert.

src/map/skills/battle_skill.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <memory>
+#include "../battle.hpp"
+#include "../status.hpp"
+
+class BattleSkill
+{
+public:
+    virtual ~BattleSkill() = default;
```
Lemongrass3110: Please change the order: constructor before destructor.

src/map/skills/battle_skill.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+    explicit BattleSkill(e_skill skill_id) : skill_id_(static_cast<uint16_t>(skill_id)) {}
```
Lemongrass3110: Please put the constructor logic into battle_skill.cpp

src/map/skills/battle_skill.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+    virtual e_battle_flag getAttackType() const = 0; // BF_WEAPON, BF_MAGIC, BF_MISC
```
Lemongrass3110:
```suggestion
    virtual e_battle_flag getAttackType() const = 0;
```
You already use the enum, no need to list the values of the enum here.

src/map/skills/battle_skill.hpp
```diff
@@ -0,0 +1,34 @@
...
+     */
+    virtual void modifyHitRate(int16 &hit_rate, const block_list *src, const block_list *target, uint16 skill_lv) const;
+
+    /**
+     * Apply additional effects after damage - called from skill_additional_effect
+     */
+    virtual void applyAdditionalEffects(block_list *src, block_list *target, uint16 skill_lv, t_tick tick, int32 attack_type, enum damage_lv dmg_lv) const;
+
+protected:
+    uint16_t skill_id_;
```
Lemongrass3110:
```suggestion
    uint16 skill_id_;
```

src/map/skills/battle_skill.cpp
```diff
@@ -0,0 +1,27 @@
+#include "battle_skill.hpp"
+#include <common/cbasetypes.hpp>
+#include "../battle.hpp"
+#include "../status.hpp"
+#include "../pc.hpp"
+#include "../map.hpp"
+#include <bitset>
+
+uint16_t BattleSkill::getSkillId() const
```
Lemongrass3110:
```suggestion
uint16 BattleSkill::getSkillId() const
```

src/map/skills/battle_skill.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+class BattleSkill
+{
+public:
+    virtual ~BattleSkill() = default;
+
+    explicit BattleSkill(e_skill skill_id) : skill_id_(static_cast<uint16_t>(skill_id)) {}
+
+    virtual e_battle_flag getAttackType() const = 0; // BF_WEAPON, BF_MAGIC, BF_MISC
+    uint16_t getSkillId() const;
```
Lemongrass3110:
```suggestion
    uint16 getSkillId() const;
```

src/map/skills/battle_skill.cpp
```diff
@@ -0,0 +1,27 @@
+#include "battle_skill.hpp"
+#include <common/cbasetypes.hpp>
+#include "../battle.hpp"
+#include "../status.hpp"
+#include "../pc.hpp"
+#include "../map.hpp"
+#include <bitset>
```
Lemongrass3110:
```suggestion
```
Unused?

src/map/skill.hpp
```diff
@@ -307,6 +308,8 @@ struct s_skill_db {
 	int32 abra_probability[MAX_SKILL_LEVEL];
 	uint16 improvisedsong_rate;
 	sc_type sc;									///< Default SC for skill
+
+	std::shared_ptr<BattleSkill> impl;
```
Lemongrass3110:
```suggestion
	std::shared_ptr<const BattleSkill> impl;
```

src/map/skills/battle_skill_categories.hpp
```diff
@@ -0,0 +1,51 @@
+#pragma once
```
Lemongrass3110: I would go with only the BattleSkill first and if we see the need to have "skill type" specific functions in one of such a base class, we can create it in the relevant follow up commit.

src/map/skills/battle_skill_factory.cpp
```diff
@@ -0,0 +1,24 @@
+#include <common/showmsg.hpp>
+
+#include "battle_skill_factory.hpp"
```
Lemongrass3110:
```suggestion
#include "battle_skill_factory.hpp"

#include <common/showmsg.hpp>

```

src/map/skills/battle_skill_factory.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "./swordsman/swordsman.hpp"
+#include "./mercenary/mercenary.hpp"
+
+void BattleSkillFactory::registerSkill(uint16 skill_id, const std::shared_ptr<BattleSkill> &skill)
+{
+    std::shared_ptr<s_skill_db> skill_entry = skill_db.find(skill_id);
+    if (skill_entry)
```
Lemongrass3110:
```suggestion
    if (skill_entry != nullptr)
```

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,24 @@
...
+ * Factory class for creating and managing BattleSkill instances
+ * Uses singleton pattern with thread-safe skill registration and caching
+ */
+class BattleSkillFactory
+{
+public:
+    // Destructor
+    ~BattleSkillFactory() = default;
+
+    static void registerSkill(uint16 skill_id, const std::shared_ptr<BattleSkill>& skill);
```
Lemongrass3110:
```suggestion
    static void registerSkill(const e_skill skill_id, const std::shared_ptr<BattleSkill>& skill);
```

src/map/skills/battle_skill_factory.cpp
```diff
@@ -0,0 +1,24 @@
+#include <common/showmsg.hpp>
+
+#include "battle_skill_factory.hpp"
+#include "./swordsman/swordsman.hpp"
+#include "./mercenary/mercenary.hpp"
+
+void BattleSkillFactory::registerSkill(uint16 skill_id, const std::shared_ptr<BattleSkill> &skill)
```
Lemongrass3110:
```suggestion
void BattleSkillFactory::registerSkill(const e_skill skill_id, const std::shared_ptr<const BattleSkill> &skill) const
```

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <memory>
```
Lemongrass3110: Lots of headers here.
Are really all of them needed?

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
```
Lemongrass3110:
```suggestion

#include "battle_skill.hpp"

```

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <memory>
+#include <unordered_map>
+#include <functional>
+#include <mutex>
+#include <vector>
+#include <common/cbasetypes.hpp>
+#include "battle_skill.hpp"
```
Lemongrass3110:
```suggestion

```

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <memory>
+#include <unordered_map>
+#include <functional>
+#include <mutex>
+#include <vector>
```
Lemongrass3110:
```suggestion
#include <vector>

```

src/map/skills/battle_skill_factory.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <common/cbasetypes.hpp>
+#include "battle_skill.hpp"
+#include "../skill.hpp"
+
+/**
+ * Factory class for creating and managing BattleSkill instances
+ * Uses singleton pattern with thread-safe skill registration and caching
```
Lemongrass3110:
```suggestion
```
It doesnt use that pattern anymore, right?

src/map/skills/mercenary/mercenary.hpp
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "../battle_skill_factory.hpp"
+#include "mercenary_bash.hpp"
+
+class Mercenary
```
Lemongrass3110: Shouldnt this maybe derive from BattleSkillFactory?
If yes, simply make registerSkill final in BattleSkillFactory.

src/map/skills/mercenary/mercenary_bash.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "../../skill.hpp"
+
+class MercenaryBashSkill : public WeaponSkill
+{
+public:
+    MercenaryBashSkill() : WeaponSkill(MS_BASH) {}
+
+    void calculateSkillRatio(const Damage *wd, const block_list *src, const block_list *target, uint16 skill_lv, int32 &base_skillratio) const override
+    {
+        base_skillratio += 30 * skill_lv; // Base 100% + 30% per level
```
Lemongrass3110:
```suggestion
        // Base 100% + 30% per level
        base_skillratio += 30 * skill_lv;
```

src/map/skills/mercenary/mercenary_bash.hpp
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    void modifyHitRate(int16 &hit_rate, const block_list *src, const block_list *target, uint16 skill_lv) const override
+    {
+        hit_rate += hit_rate * 5 * skill_lv / 100; // +5% hit per level
```
Lemongrass3110:
```suggestion
        // +5% hit per level
        hit_rate += hit_rate * 5 * skill_lv / 100;
```

src/map/skills/swordsman/bash.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include "../../skill.hpp"
+
+class BashSkill : public WeaponSkill
+{
+public:
+    BashSkill() : WeaponSkill(SM_BASH) {}
+
+    void calculateSkillRatio(const Damage *wd, const block_list *src, const block_list *target, uint16 skill_lv, int32 &base_skillratio) const override
+    {
+        base_skillratio += 30 * skill_lv; // Base 100% + 30% per level
```
Lemongrass3110:
```suggestion
        // Base 100% + 30% per level
        base_skillratio += 30 * skill_lv;
```

src/map/skills/swordsman/bash.hpp
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    void modifyHitRate(int16 &hit_rate, const block_list *src, const block_list *target, uint16 skill_lv) const override
+    {
+        hit_rate += hit_rate * 5 * skill_lv / 100; // +5% hit per level
```
Lemongrass3110:
```suggestion
        // +5% hit per level
        hit_rate += hit_rate * 5 * skill_lv / 100;
```

src/map/skill.cpp
```diff
@@ -1401,14 +1408,6 @@ int32 skill_additional_effect( struct block_list* src, struct block_list *bl, ui
 			}
 			break;
 
-	case SM_BASH:
-		if( sd && skill_lv > 5 && pc_checkskill(sd,SM_FATALBLOW)>0 ){
-			//BaseChance gets multiplied with BaseLevel/50.0; 500/50 simplifies to 10 [Playtester]
```
Lemongrass3110, line 1406: Please restore this comment in the new file.

src/map/skills/swordsman/bash.hpp
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    void applyAdditionalEffects(block_list *src, block_list *target, uint16 skill_lv, t_tick tick, int32 attack_type, enum damage_lv dmg_lv) const override
+    {
+        map_session_data *sd = BL_CAST(BL_PC, src);
+
+        // Stun effect only if skill level > 5 and has Fatal Blow
+        if (sd && skill_lv > 5 && pc_checkskill(sd, SM_FATALBLOW) > 0)
```
Lemongrass3110:
```suggestion
        if (sd != nullptr && skill_lv > 5 && pc_checkskill(sd, SM_FATALBLOW) > 0)
```
Either like this or leave the check out completely, since pc_checkskill has a nullpointer check anyway.

src/map/skills/battle_skill_factory.cpp
```diff
@@ -0,0 +1,24 @@
...
+{
+    std::shared_ptr<s_skill_db> skill_entry = skill_db.find(skill_id);
+    if (skill_entry)
+    {
+        skill_entry->impl = skill;
+    }
+    else
+    {
+        ShowError("registerSkill: could not find s_skill_db with %d", skill_id);
```
Lemongrass3110:
```suggestion
        ShowError("registerSkill: skill ID %hu not found in skill DB", skill_id);
```

src/map/skills/mercenary/mercenary.hpp
Danil0v3s: Not sure, I tried to accomodate nitrous suggestion to gather skills by classes. Ideally what I would do is have a class/interface with a common method, then ask DI to inject all the impl of said class but since we need to manually create the classes and assign them i dont know

Danil0v3s:
The latest changes seem to not have impacted, just tested and its still working as before

src/map/skills/skill_factory.cpp
```diff
@@ -8,17 +8,14 @@
...
+	if( std::unique_ptr<const SkillImpl> impl = SkillFactoryMercenary::getInstance().create( skill_id ); impl != nullptr ){
+		return impl;
+	}
 
-	if (skill_entry != nullptr) {
-		skill_entry->impl = std::move(skill);
-	} else {
-		ShowError("registerSkill: skill ID %hu not found in skill DB", skill_id);
+	if( std::unique_ptr<const SkillImpl> impl = SkillFactorySwordsman::getInstance().create( skill_id ); impl != nullptr ){
+		return impl;
```
Danil0v3s: Any way we can hide this boilerplate with an inline function or some sort of generics? Or both idk

Lemongrass3110 (approved):

PR #9480, opened 2025-08-05
Song/Dance SP Handling

### LINKED ISSUE #9479: Song/Dance SP Handling
### rAthena Hash

ad04a42

### Client Date

2021-11-03

### Server Mode

Pre-Renewal

### Result

When using Amp, it consumes 1 SP.
When using Encore, it consumes 1 SP.
When using the previous song after using Encore, the SP cost is halved.
A song ends when it tries to consume SP and you don't have enough.

### Relevant Log Output

```Shell

```

### Expected Result

When using Amp, it consumes 0 SP.
When using Encore, it consumes 0 SP.
When using the previous song after using Encore, you pay the full SP cost.
A song ends immediately when your SP reaches 0.

### How to Reproduce

Use Amp, check SP lost.
Use Encore when not having enough SP for repeating the song, check SP lost.
Use Poem of Bragi, then use Encore, then use Poem of Bragi again, check SP lost on last cast.
Keep playing a song until SP reaches 0. Check if song ends.

### Official Information

Verified by tests.

_No response_

Playtester:
* **Addressed Issue(s)**: #9479

- Amp and Encore no longer consume SP (but still require 1 SP to be used)
- When using the same song again after using Encore, the SP cost is no longer halved
  * Cost is only halved when using Encore to repeat the song
- Songs/Dances now immediately end when SP reaches 0
- Fixes #9479

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -10679,8 +10679,11 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, 
 		break;
 	case BD_ENCORE:
 		clif_skill_nodamage(src,*bl,skill_id,skill_lv);
-		if(sd)
+		if (sd) {
 			unit_skilluse_id(src,src->id,sd->skill_id_dance,sd->skill_lv_dance);
+			// Need to remove remembered skill to prevent permanent halving of SP cost
+			sd->skill_id_old = 0;
+		}
 		break;
 
 	case TR_RETROSPECTION:
@@ -19808,6 +19811,8 @@ void skill_consume_requirement(map_session_data *sd, uint16 skill_id, uint16 ski
 		switch( skill_id ) {
 			case CG_TAROTCARD: // TarotCard will consume sp in skill_cast_nodamage_id [Inkfish]
 			case MC_IDENTIFY:
+			case BD_ADAPTATION:
+			case BD_ENCORE:
 				require.sp = 0;
 				break;
 			case AL_HOLYLIGHT:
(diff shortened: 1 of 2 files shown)


src/map/skill.cpp
```diff
@@ -10679,8 +10679,11 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl,
 		break;
 	case BD_ENCORE:
 		clif_skill_nodamage(src,*bl,skill_id,skill_lv);
-		if(sd)
+		if (sd) {
```
```suggestion
		if (sd != nullptr) {
```

src/map/status.cpp
```diff
@@ -14436,7 +14436,7 @@ TIMER_FUNC(status_change_timer){
 				if (sc->getSCE(SC_LONGING))
 					sp*= 3;
 #endif
-				if (!status_charge(bl, 0, sp))
+				if (!status_charge(bl, 0, sp) || status->sp == 0)
```
Lemongrass3110:
```suggestion
				if (!status_charge(bl, 0, sp) || (sp == 0 && status->sp == 0))
```
If I understood you correctly. In other cases this is already handled by status_charge/status_damage

Playtester: `sp` is the SP damage here, not the current value.

So if your current SP is 1 and status_charge reduces it from 1 to 0, it should end.

`status_charge` only returns false if the cost is higher than your current value, not if it's equal.

Lemongrass3110: Ah then please put that in a separate line below with a comment. Like this it isnt really obvious.

Playtester: Done.

PR #9443, opened 2025-06-30
Fix Kvasir Sonata on Class 3 skills

kaninhot004:
* **Addressed Issue(s)**: 

https://github.com/rathena/rathena/issues/9393

* **Server Mode**: 

Renewal

* **Description of Pull Request**: 

Official

https://github.com/user-attachments/assets/c3dd40e0-ab9c-4e5d-9040-1dbbe1380996

Fix in rAthena

https://github.com/user-attachments/assets/fd3bcffa-de1a-430b-b7c1-9e76dc2b8504

Issue #9393, opened 2025-06-06
Skill "Kvasir Sonata" bug when use ensemble skill in class Clown & Gypsy / Troubadour & Trouvere


Jaikayray:
### rAthena Hash

a5c3785

### Client Date

2024-10-30

### Server Mode

Renewal

### Result

When use skill "Kvasir Sonata" in job Clown & Gypsy / Troubadour & Trouvere it's bug can't use all skill ensemble in class Minstrel & Wanderer 

List skill can not use when use skill "Kvasir Sonata"
- CG_MOONLIT
- WM_SONG_OF_MANA
- WM_DANCE_WITH_WUG
- WM_SOUND_OF_DESTRUCTION
- WM_SATURDAY_NIGHT_FEVER
- WM_LERADS_DEW
- WM_MELODYOFSINK
- WM_BEYOND_OF_WARCRY
- WM_UNLIMITED_HUMMING_VOICE

![Image](https://github.com/user-attachments/assets/fbb92082-f1ac-4e49-8e54-77bc6d3e607e)

![Image](https://github.com/user-attachments/assets/0d9e16fb-9686-4386-9f3c-39d482361e3b)

### Relevant Log Output

```Shell

```

### Expected Result

This skill it's can use all skill ensemble in list but only use in ensemble on job Bard & Dancer only

### How to Reproduce

I don't know how to doing for this, can you help me fix that?

### Official Information

In official kRO server can use all skill ensemble when use skill "Kvasir Sonata"

PR #9484, opened 2025-08-08
Fix Kvasir Sonata on Class 3 skills

Lemongrass3110:
* **Addressed Issue(s)**: #9393

* **Server Mode**: Renewal

Added the same check like for ensembles.

Continuation of #9443, since it was closed and the underlying repository was deleted.


Playtester (approved):
According to @Haydrich this change is correct.

Issue #9461, opened 2025-07-17
Map Drop not working when use equipment drop% or item drop% and not affected by base level difference in "Illusion Map"


Jaikayray:
### rAthena Hash

a5c3785

### Client Date

2025-07-17

### Server Mode

Renewal

### Result

"Map Drop System" is not working when player equipment item drop percent (Example Zarathustra (ID : 19429)) or use item drop percent (Example HE Bubble Gum (ID : 12412))

<img width="678" height="710" alt="Image" src="https://github.com/user-attachments/assets/2fd1889e-bdab-4bee-b065-94ad6bbfd665" />

<img width="593" height="518" alt="Image" src="https://github.com/user-attachments/assets/a4c72311-f21c-4ae0-9d2f-19a2a07ea8c7" />

<img width="1368" height="109" alt="Image" src="https://github.com/user-attachments/assets/fcb83f4e-1f47-4b31-bcad-ec191db5550c" />

I test with enchant Zarathustra all slot equipment (Drop Item 720000%) (Picture 1) and test "Map Drop System" in Rudus 4th floor it's not working (Picture 2), in fact I test to find "Etel Dust" map drop in Rudus 4th floor is 54,007.5% when equip drop item 720000% (In picture 3 example to find Etel Dust)

And "Map Drop System" can working when player in "Illusion Map" (Example Illusion of Moonlight)
<Item received from Illusion monsters are not affected by base level difference>

### Relevant Log Output

```Shell

```

### Expected Result

"Map Drop System" can working when use equipment drop% or item drop% and not affected by base level difference in "Illusion Map"

### How to Reproduce

Sorry I don't know how to doing for this, can you help me fix that?

### Official Information

In kRO not have news patch "Map Drop System" but in ROGGT have news patch "Map Drop System" (see this link "https://ro.gnjoy.in.th/map-drop-system/")
Why am I explain for this? Because ROGGT client is from kRO (Official)

<img width="756" height="223" alt="Image" src="https://github.com/user-attachments/assets/ff5f81d6-24da-43c5-afab-f7e1a4d1083c" />

Tips :
- "Map Drop System" can not use skill "Steal" for steal item monster
- Equipment and item buff increase drop item from monster enable "Map Drop System"
- Drop rate penalties are affected by the content of the map, for example "Illusion Map" will not have a drop rate reduction penalty, but normal maps will have the normal drop rate penalties

PR #9493, opened 2025-08-12
Fixed Drop Rate Bonus affecting mapdrop.

StrawberriesX1:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9461

* **Server Mode**: Renewal

* **Description of Pull Request**: Added bonus calculation formula to map_drops_rate.

### DIFF
--- a/conf/battle/drops.conf
+++ b/conf/battle/drops.conf
@@ -207,7 +207,6 @@ drop_rate_cap_vip: 9000
 // Displays a colored pillar effect for items dropped by monsters that contain random options.
 rndopt_drop_pillar: on
 
-// Drop rate bonus can increase drop rate from map_drops.
-// On official servers, level difference penalties do not affect map drops.
+// Drop rate bonus can increase drop rate from map drops.
 // Default: on
 enable_bonus_map_drops: on
--- a/conf/battle/drops.conf
+++ b/conf/battle/drops.conf
@@ -207,6 +207,6 @@ drop_rate_cap_vip: 9000
 // Displays a colored pillar effect for items dropped by monsters that contain random options.
 rndopt_drop_pillar: on
 
-// Drop rate bonus can increase drop rate from map drops.
-// Default: on
-enable_bonus_map_drops: on
+// Do drop rate bonus increase the drop rate of map drops? (Note 1)
+// Default: yes
+enable_bonus_map_drops: yes
--- a/conf/battle/drops.conf
+++ b/conf/battle/drops.conf
@@ -206,3 +206,8 @@ drop_rate_cap_vip: 9000
 
 // Displays a colored pillar effect for items dropped by monsters that contain random options.
 rndopt_drop_pillar: on
+
+// Drop rate bonus can increase drop rate from map_drops.
+// On official servers, level difference penalties do not affect map drops.
+// Default: on
+enable_bonus_map_drops: on
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2813,9 +2813,11 @@ void mob_damage(struct mob_data *md, struct block_list *src, int32 damage)
  * @param mob: Monster data
  * @param base_rate: Base drop rate
  * @param drop_modifier: RENEWAL_DROP level modifier
+ * @param md: the actual monster killed
+ * @param factor: factor which is applied to all multiplicative bonuses and upper bound caps
  * @return Modified drop rate
  */
-int32 mob_getdroprate(struct block_list *src, std::shared_ptr<s_mob_db> mob, int32 base_rate, int32 drop_modifier, mob_data* md)
+int32 mob_getdroprate(struct block_list *src, std::shared_ptr<s_mob_db> mob, int32 base_rate, int32 drop_modifier, mob_data* md, int32 factor)
 {
 	int32 drop_rate = base_rate;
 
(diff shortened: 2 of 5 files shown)


conf/battle/drops.conf
```diff
@@ -206,3 +206,8 @@ drop_rate_cap_vip: 9000
 
 // Displays a colored pillar effect for items dropped by monsters that contain random options.
 rndopt_drop_pillar: on
+
+// Drop rate bonus can increase drop rate from map_drops.
+// On official servers, level difference penalties do not affect map drops.
+// Default: on
+enable_bonus_map_drops: on
```
Playtester: I'm not sure if that's actually true. Drop rate modifiers might actually work for map drops officially. I'm not sure if there's an easy way to test this, though.

StrawberriesX1: I think an experiment I could try myself would be to kill 10000~100000 monsters in the Illusion of Moonlight, like you did when finding the rate of Finding Ore skill.

Playtester: @StrawberriesX1 We could verify how it works. It's kinda similar to normal drop rates, except it's on a base of 100000.

The logic is:
```cpp
if (drop_rate < 90000)
  drop_rate = min(drop_rate * worldfactor_droprate / 100.0 + drop_rate * drop_bonus / 100.0, 90000);
```
So yes, by default the drop rate bonuses such as bubblegum should apply. And similar to normal drop rates, it cannot raise the rate above 90% if the original rate was lower.

conf/battle/drops.conf
```diff
@@ -206,3 +206,8 @@ drop_rate_cap_vip: 9000
 
 // Displays a colored pillar effect for items dropped by monsters that contain random options.
 rndopt_drop_pillar: on
+
+// Drop rate bonus can increase drop rate from map_drops.
+// On official servers, level difference penalties do not affect map drops.
```
Lemongrass3110:
```suggestion
```
Based on your own description in the issue, this is mapflag dependent:
```
Drop rate penalties are affected by the content of the map, for example "Illusion Map" will not have a drop rate reduction penalty, but normal maps will have the normal drop rate penalties
``` 
This was also confirmed by us.

src/map/mob.cpp
```diff
@@ -3375,9 +3375,18 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 		if( mapdrops != nullptr ){
 			// Process map wide drops
 			for( const auto& it : mapdrops->globals ){
-				if( rnd_chance( it.second->rate, 100000u ) ){
+				uint32 final_rate;
+
+				if ( battle_config.enable_bonus_map_drops ) {
+					int32 drop_rate = mob_getdroprate(first_sd, md->db, it.second->rate, 100, md);
```
Lemongrass3110:
```suggestion
					final_rate = mob_getdroprate(first_sd, md->db, it.second->rate / 10, drop_modifier, md) * 10;
```
I think we need divide by 10 here, otherwise all the capping and settings in mob_getdroprate will not work, since they are all based on n/10000 instead of n/100000.

We also confirmed that drop_modifier has to be taken into account.

src/map/mob.cpp
```diff
@@ -3388,7 +3397,16 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 
 			if( specific != mapdrops->specific.end() ){
 				for( const auto& it : specific->second ){
-					if( rnd_chance( it.second->rate, 100000u ) ){
+					uint32 final_rate;
+
+					if ( battle_config.enable_bonus_map_drops ) {
+						int32 drop_rate = mob_getdroprate(first_sd, md->db, it.second->rate, 100, md);
```
Lemongrass3110:
```suggestion
						final_rate = mob_getdroprate(first_sd, md->db, it.second->rate / 10, drop_modifier, md) * 10;
```

conf/battle/drops.conf
```diff
@@ -206,3 +206,8 @@ drop_rate_cap_vip: 9000
 
 // Displays a colored pillar effect for items dropped by monsters that contain random options.
 rndopt_drop_pillar: on
+
+// Drop rate bonus can increase drop rate from map_drops.
```
Lemongrass3110:
```suggestion
// Drop rate bonus can increase drop rate from map drops.
```

src/map/mob.cpp
```diff
@@ -3375,9 +3375,18 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 		if( mapdrops != nullptr ){
 			// Process map wide drops
 			for( const auto& it : mapdrops->globals ){
-				if( rnd_chance( it.second->rate, 100000u ) ){
+				uint32 final_rate;
+
+				if ( battle_config.enable_bonus_map_drops ) {
+					int32 drop_rate = mob_getdroprate(first_sd, md->db, it.second->rate, 100, md);
+					final_rate = (uint32)((int64)drop_rate * 100000 / 10000);
```
Lemongrass3110:
```suggestion
```

src/map/mob.cpp
```diff
@@ -3388,7 +3397,16 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 
 			if( specific != mapdrops->specific.end() ){
 				for( const auto& it : specific->second ){
-					if( rnd_chance( it.second->rate, 100000u ) ){
+					uint32 final_rate;
+
+					if ( battle_config.enable_bonus_map_drops ) {
+						int32 drop_rate = mob_getdroprate(first_sd, md->db, it.second->rate, 100, md);
+						final_rate = (uint32)((int64)drop_rate * 100000 / 10000);
```
Lemongrass3110:
```suggestion
```

Lemongrass3110 (approved):

src/map/mob.cpp
StrawberriesX1: With this suggestion, this block can use a simple statement
```
if ( battle_config.enable_bonus_map_drops )
						final_rate = mob_getdroprate(first_sd, md->db, it.second->rate / 10, drop_modifier, md) * 10;
					else
						final_rate = it.second->rate;
```
Should we still use the Braced form (with { }): in the following format:
```
if ( battle_config.enable_bonus_map_drops ) {
final_rate = mob_getdroprate(first_sd, md->db, it.second->rate / 10, drop_modifier, md) * 10;
} else {
final_rate = it.second->rate;
}
```
Or should we use a simple statement format instead?
What style does the rAthena team prefer for this change?

conf/battle/drops.conf
```diff
@@ -206,3 +206,7 @@ drop_rate_cap_vip: 9000
 
 // Displays a colored pillar effect for items dropped by monsters that contain random options.
 rndopt_drop_pillar: on
+
+// Drop rate bonus can increase drop rate from map drops.
+// Default: on
+enable_bonus_map_drops: on
```
Playtester: I wonder if we shouldn't stick to yes/no by default. Earlier I mistook "on" as "no".

Issue #9496, opened 2025-08-15
Incorrect "Blitz Beat" skillratio


StrawberriesX1:
### rAthena Hash

02ee461

### Client Date

2025-03-19

### Server Mode

Renewal

### Result

1.rAthena used :
```
(sstatus->dex / 10 + sstatus->agi / 2 + skill * 3 + 40) * 2;
RE_LVL_MDMOD(100);
```
This makes the damage dealt higher than the actual amount.

Stats used in the experiment :

<img width="1025" height="622" alt="Image" src="https://github.com/user-attachments/assets/9401d2b0-75b3-4089-ab57-d1d05ed00194" />

<img width="1014" height="623" alt="Image" src="https://github.com/user-attachments/assets/6d73fe98-0148-49a5-bfa7-02ff8f6d2ca7" />

Character :

Agi : 100+12
Dex : 120+18
Blitz Beat : Level 5
Steel Crow : Level 9

Damage :
```
md.damage = ( 138 / 10 + 112 / 2 + 9 * 3 + 40) * 2; => 272
RE_LVL_MDMOD(100); = md.damage * caster_lv / 100 => 272 * 175 / 100 = 476
```

https://github.com/user-attachments/assets/7d3cfca7-1eab-4ab9-a2de-d06ce52e29e8

### Relevant Log Output

```Shell

```

### Expected Result

Damage :
```
SC                                   KRO         			RA
None                                 292                  476
Improve                             308                        504
Falcon Eye                          314                       514
```

Stats used in the experiment :

<img width="592" height="418" alt="Image" src="https://github.com/user-attachments/assets/0485035c-e142-494e-8179-a0c633b05459" />

<img width="572" height="438" alt="Image" src="https://github.com/user-attachments/assets/6743f244-ce14-4a7e-92e7-7ef082392671" />

<img width="305" height="260" alt="Image" src="https://github.com/user-attachments/assets/710cefec-02e5-43ef-867c-0a89a5a0e0c8" />

Character :

Agi : 100+12
Dex : 120+18
Blitz Beat : Level 5
Steel Crow : Level 9

https://github.com/user-attachments/assets/f9f91134-77f9-4596-9e04-d86212005429

https://github.com/user-attachments/assets/b1d66b99-56f9-4ef7-a959-52feeb7642e5

Formula
```
skill_lv * 20 + skill * 6 + ((sstatus->agi / 2) *2) + ((sstatus->dex / 10) *2);
```

Result damage :

```
[...]

PR #9497, opened 2025-08-15
Corrected "Blitz Beat" skillratio

StrawberriesX1:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9496

* **Server Mode**: Renewal

The damage dealt by the Blitz Beat skill is correct according to the Official kRO & thRO
Thanks to @Haydrich

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -9726,8 +9726,7 @@ struct Damage battle_calc_misc_attack(struct block_list *src,struct block_list *
 				if(!sd || !(skill = pc_checkskill(sd,HT_STEELCROW)))
 					skill = 0;
 #ifdef RENEWAL
-				md.damage = (sstatus->dex / 10 + sstatus->agi / 2 + skill * 3 + 40) * 2;
-				RE_LVL_MDMOD(100);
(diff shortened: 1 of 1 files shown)


Atemo (approved):

src/map/battle.cpp
```diff
@@ -9726,8 +9726,7 @@ struct Damage battle_calc_misc_attack(struct block_list *src,struct block_list *
 				if(!sd || !(skill = pc_checkskill(sd,HT_STEELCROW)))
 					skill = 0;
 #ifdef RENEWAL
-				md.damage = (sstatus->dex / 10 + sstatus->agi / 2 + skill * 3 + 40) * 2;
-				RE_LVL_MDMOD(100);
+				md.damage = skill_lv * 20 + skill * 6 + ((sstatus->agi / 2) *2) + ((sstatus->dex / 10) *2);
```
Playtester, line 9729:
```suggestion
				md.damage = (sstatus->dex / 10 + sstatus->agi / 2 + skill * 3 + skill_lv * 10) * 2;
```
Wouldn't it be better to write it like this? So it's easier to read and compare with the pre-re formula.

Atemo: It doesn't matter to me. Note that the formula as it is now match irowiki formula https://irowiki.org/wiki/Blitz_Beat

Issue #9501, opened 2025-08-18
Use SQL db does not work anymore


CDelgrange:
### rAthena Hash

f7c8edc46dfdf2d25746171151420a89ff52387f

### Client Date

20220406

### Server Mode

Renewal, Pre-Renewal

### Result

Hello everyone !

This bug is neither related to the client version nor the server mode.

The issue comes from this commit: https://github.com/rathena/rathena/commit/51bbb132144e2ec6bfd1df9d6236e4dd9208fb4d (if you look at the `mob.cpp` file changes).
The `groupid` and `title` have been added to the SQL query, however in the `mob_read_sqldb_sub` method, the selected fields order of the query matter (because we use an index to get the row values).
Because of this, the `std::stoi(...)` call does not work on this line: 
```cpp
modes["Looter"] << (std::stoi(str[index]) ? "true" : "false");
```

### Relevant Log Output

```Shell

```

### Expected Result

Expect the map-server to run

### How to Reproduce

To reproduce it, you just need to clone the repo, and set `use_sql_db: yes` in `inter_athena.conf`. After that, you won't be able to run the map-server.

### Official Information

N/A

PR #9502, opened 2025-08-18
Added missing fields in mob_read_sqldb_sub

CDelgrange:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/9501

* **Server Mode**: 
Both

Fix the order of the selected fields in the sql query of the mob.cpp

> You can ignore white spaces when reviewing the request, as they are just removed from line endings

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -5663,6 +5663,10 @@ static bool mob_read_sqldb_sub(std::vector<std::string> str) {
 		node["DamageMotion"] << str[index];
 	if (!str[++index].empty())
 		node["DamageTaken"] << str[index];
+	if (!str[++index].empty())
+		node["GroupId"] << str[index];
+	if (!str[++index].empty())
+		node["Title"] << str[index];	
 	if (!str[++index].empty() && strcmp(str[index].c_str(), "06") != 0)
 		node["Ai"] << str[index];
 	if (!str[++index].empty() && strcmp(str[index].c_str(), "Normal") != 0)


src/map/mob.cpp
```diff
@@ -265,7 +265,7 @@ static bool mobdb_searchname_sub(uint16 mob_id, const char * const str, bool ful
 
 	if (mob == nullptr)
 		return false;
-
```
Playtester: Please don't change lines that have nothing to do with the issue, otherwise it's hard to track later.

CDelgrange: Ok, done :heavy_check_mark:

src/map/mob.cpp
```diff
@@ -5803,10 +5803,11 @@ static int32 mob_read_sqldb(void)
 			"`racegroup_bio5_swordman_thief`,`racegroup_bio5_acolyte_merchant`,`racegroup_bio5_mage_archer`,`racegroup_bio5_mvp`,`racegroup_clocktower`,`racegroup_thanatos`,`racegroup_faceworm`,`racegroup_hearthunter`,`racegroup_rockridge`,`racegroup_werner_lab`,`racegroup_temple_demon`,`racegroup_illusion_vampire`,"
 			"`racegroup_malangdo`,`racegroup_ep172alpha`,`racegroup_ep172beta`,`racegroup_ep172bath`,`racegroup_illusion_turtle`,`racegroup_rachel_sanctuary`,`racegroup_illusion_luanda`,`racegroup_illusion_frozen`,`racegroup_illusion_moonlight`,`racegroup_ep16_def`,`racegroup_edda_arunafeltz`,`racegroup_lasagna`,`racegroup_glast_heim_abyss`,"
 			"`racegroup_destroyed_valkyrie_realm`,`racegroup_encroached_gephenia`,"
-			"`element`,`element_level`,`walk_speed`,`attack_delay`,`attack_motion`,`damage_motion`,`damage_taken`,`groupid`,`title`,`ai`,`class`,"
+			"`element`,`element_level`,`walk_speed`,`attack_delay`,`attack_motion`,`damage_motion`,`damage_taken`,`ai`,`class`,"
```
Atemo: Aren't the nodes "GroupId" and "Title" just missing in https://github.com/rathena/rathena/blob/master/src/map/mob.cpp#L5665 ?

CDelgrange: You're right, @Lemongrass3110 told me this too. 

I've just pushed the change :heavy_check_mark:

PR #9504, opened 2025-08-18
Koshei The Immortal Quest Fix

### LINKED ISSUE #9503: Koshei the Immortal quest softlock
### rAthena Hash

f7c8edc

### Client Date

20220406

### Server Mode

Pre-Renewal

### Result

After defeating the boss, Koshei, the Immortal, the npc called "Wall" is disabled and the player cannot advance in the quest (unless @reloadscript is used).

### Relevant Log Output

```Shell

```

### Expected Result

The npc called "Wall" should become visible after the boss' defeat in order to let the player interact and continue the quest.

### How to Reproduce

Simply follow the quest: https://irowiki.org/classic/Koschei_the_Immortal
The step 11 is where it's acting weird.

### Official Information

No official information but the quest is clearly bugged.

_No response_

Playtester:
* **Addressed Issue(s)**: #9503 

* **Server Mode**: Both

- Fixed "Koshei The Immortal" quest getting soft-locked after defeated Koshei
- Fixes #9503
- Fix done by @Akieros90 

Please note that I only pushed the fix for @Akieros90, I did not test nor review this script.

### DIFF
--- a/npc/quests/quests_moscovia.txt
+++ b/npc/quests/quests_moscovia.txt
@@ -7102,7 +7102,6 @@ mosk_dun01,45,259,0	script	Wall#rus04	HIDDEN_NPC,{
 	}
 
 OnInit:
-	enablenpc "Wall#rus04";
 	$@rus_req02 = 0;
 	end;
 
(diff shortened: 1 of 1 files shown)


npc/quests/quests_moscovia.txt
```diff
@@ -7102,17 +7102,23 @@ mosk_dun01,45,259,0	script	Wall#rus04	HIDDEN_NPC,{
 	}
 
 OnInit:
+	enablenpc "Wall#rus04";
```
Atemo:
```suggestion
```

npc/quests/quests_moscovia.txt
```diff
@@ -7102,17 +7102,23 @@ mosk_dun01,45,259,0	script	Wall#rus04	HIDDEN_NPC,{
...
 OnEnable:
 	enablenpc "Wall#rus04";
 	end;
 
+OnDisable:
+	disablenpc "Wall#rus04";
+	end;
+
 OnMyMobDead:
+	donpcevent "Koshei#rus47::OnDisable";
```
Lemongrass3110, line 7117: Doesnt calling the npc event do the same logic like falling through to the old OnDisable code?
Is the order really so important in this case?

npc/quests/quests_moscovia.txt
```diff
@@ -7102,17 +7102,23 @@ mosk_dun01,45,259,0	script	Wall#rus04	HIDDEN_NPC,{
...
 	$@rus_req02 = 0;
+	end;
+
 OnEnable:
 	enablenpc "Wall#rus04";
 	end;
 
+OnDisable:
+	disablenpc "Wall#rus04";
+	end;
```
Lemongrass3110, line 7114: Same code like the old OnDisable event.
If we do not want to fallthrough we could simply add an "end;" at the end of OnMyMobDead instead

npc/quests/quests_moscovia.txt
Lemongrass3110: I can only agree with @Atemo here.
If anything, then this is the only real change in this PR.

npc/quests/quests_moscovia.txt
Playtester: Atemo explained it on Discord, this is actually calling an event on another NPC. So this is the actual fix.

Lemongrass3110 (approved):

Lemongrass3110:
I was able to finish the quest without any problems with this fix.

PR #9511, opened 2025-08-23
Add an RAII class to handle freeblock locks

vstumpf:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

Use a class to lock and unlock the freeblock list. This will help with the skill implementations introduced in #9468

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -346,30 +346,32 @@ TIMER_FUNC(map_freeblock_timer){
 }
 
 FreeBlockLock::FreeBlockLock(bool startLocked) {
+	this->locked = false;
+
 	if (startLocked) {
-		lock();
+		this->lock();
 	}
 }
 
 FreeBlockLock::~FreeBlockLock() {
-	if (locked) {
-		unlock();
+	if (this->locked) {
+		this->unlock();
 	}
 }
 
 void FreeBlockLock::lock() {
-	if (!locked) {
+	if (!this->locked) {
 		map_freeblock_lock();
-		locked = true;
+		this->locked = true;
 	} else {
 		ShowWarning("FreeBlockLock is already locked.\n");
 	}
 }
 
 void FreeBlockLock::unlock() {
-	if (locked) {
+	if (this->locked) {
 		map_freeblock_unlock();
-		locked = false;
+		this->locked = false;
 	} else {
 		ShowWarning("FreeBlockLock is not locked.\n");
 	}
--- a/src/map/map.hpp
+++ b/src/map/map.hpp
@@ -1097,7 +1097,7 @@ public:
 	FreeBlockLock(const FreeBlockLock&) = delete;
 	FreeBlockLock& operator=(const FreeBlockLock&) = delete;
 private:
-	bool locked{false};
+	bool locked;
 };
 
 
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -3614,6 +3614,7 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 	if( md->can_summon )
 		mob_deleteslave(md);
 
+	// TODO: Check if this is called to early. md is still used below.
 	freeLock.unlock();
 
 	if( !rebirth ) {
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -7506,6 +7506,7 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 	if( sc && sc->getSCE(SC_CURSEDCIRCLE_ATKER) ) //Should only remove after the skill has been casted.
 		status_change_end(src,SC_CURSEDCIRCLE_ATKER);
 
+	// TODO: Check if this unlock call is too early. Compare with skill_castend_nodamage_id, there it seems to be called correctly via deconstructor.
 	freeLock.unlock();
 
 	if( sd && !(flag&1) )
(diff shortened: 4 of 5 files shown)


src/map/map.cpp
```diff
@@ -345,6 +345,31 @@ TIMER_FUNC(map_freeblock_timer){
...
+
+FreeBlockLock::~FreeBlockLock() {
+	unlock();
+}
+
+void FreeBlockLock::lock() {
+	if (!locked) {
+		map_freeblock_lock();
+		locked = true;
+	}
```
Lemongrass3110, line 368: Shouldnt we warn/error if it is called twice?

src/map/map.cpp
```diff
@@ -345,6 +345,31 @@ TIMER_FUNC(map_freeblock_timer){
...
+}
+
+void FreeBlockLock::unlock() {
+	if (locked) {
+		map_freeblock_unlock();
+		locked = false;
+	}
```
Lemongrass3110, line 377: Shouldnt we warn/error if it was not even locked?

src/map/map.hpp
```diff
@@ -1087,6 +1087,22 @@ inline bool map_flag_gvg2_no_te(int16 m) {
...
+	FreeBlockLock(bool startLocked = true);
+	~FreeBlockLock();
+
+	void lock();
+	void unlock();
+
+	FreeBlockLock(const FreeBlockLock&) = delete;
+	FreeBlockLock& operator=(const FreeBlockLock&) = delete;
+private:
+	bool locked{false};
```
Lemongrass3110:
```suggestion
	bool locked;
```

src/map/map.cpp
```diff
@@ -345,6 +345,31 @@ TIMER_FUNC(map_freeblock_timer){
 	return 0;
 }
 
+FreeBlockLock::FreeBlockLock(bool startLocked) {
+	if (startLocked) {
```
Lemongrass3110, line 351:
```suggestion
	locked = false;

	if (startLocked) {
```

src/map/mob.cpp
```diff
@@ -3616,7 +3614,7 @@ int32 mob_dead(struct mob_data *md, struct block_list *src, int32 type)
 	if( md->can_summon )
 		mob_deleteslave(md);
 
-	map_freeblock_unlock();
+	freeLock.unlock();
```
Lemongrass3110, line 3618: Unrelated to this PR, but isnt this too early here?
"md" is still used below here and may be freed by calling unlock here.

src/map/mob.cpp
```diff
@@ -4379,7 +4377,8 @@ bool mobskill_use(struct mob_data *md, t_tick tick, int32 event, int64 damage)
 
 		if (!flag)
 			continue; //Skill requisite failed to be fulfilled.
-
+		
+		FreeBlockLock freeLock;
```
Lemongrass3110, line 4382: Nice improvement!

src/map/script.cpp
```diff
@@ -11490,9 +11490,8 @@ BUILDIN_FUNC(killmonster)
 		}
 	}
 
-	map_freeblock_lock();
+	FreeBlockLock freeLock;
```
Lemongrass3110:
```suggestion
```
map_foreachinmap already has a FreeBlockLock included.
This can safely be removed.

src/map/skill.cpp
```diff
@@ -7505,14 +7501,13 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
...
 			skill_id, skill_lv, skill_get_hit(skill_id) );
-		map_freeblock_unlock();
 		return 1;
 	}
 
 	if( sc && sc->getSCE(SC_CURSEDCIRCLE_ATKER) ) //Should only remove after the skill has been casted.
 		status_change_end(src,SC_CURSEDCIRCLE_ATKER);
 
-	map_freeblock_unlock();
+	freeLock.unlock();
```
Lemongrass3110, line 7510: Unrelated to this PR, but isnt this too early here?
"sd" is still used below here and may be freed by calling unlock here.
(See also skill_castend_nodamage_id, there it is called at the correct location, or so to say "not called" now anymore, since we use the scope destructor now).

src/map/status.cpp
```diff
@@ -14149,6 +14148,8 @@ TIMER_FUNC(status_change_timer){
 		sce->timer = add_timer(t, status_change_timer, bl->id, data);
 	};
 	
+	FreeBlockLock freeLock{false};
```
Lemongrass3110:
```suggestion
	FreeBlockLock freeLock(false);
```

Lemongrass3110 (changes requested):
Nice improvement!

src/map/map.cpp
vstumpf: Good idea, I'll make the change for `lock`

src/map/map.cpp
vstumpf: I don't think so, because we could choose to unlock it manually, and then when it's destroyed it'll try to unlock again, which is fine.

Lemongrass3110: Shouldnt the destructor check the "locked" status and only call unlock, when necessary?

src/map/mob.cpp
vstumpf: I think that would be up to the next PR to see if we need to move it. I'd like to keep it here to lower the amount of changes.

src/map/skill.cpp
vstumpf: I think that would be up to the next PR to see if we need to move it. I'd like to keep it here to lower the amount of changes.

Lemongrass3110 (approved):

PR #9505, opened 2025-08-20
Added skill factories for all jobs

Lemongrass3110:
* **Addressed Issue(s)**: None

* **Server Mode**: Both

* **Description of Pull Request**: 
Renamed Swordsman => Swordman
Added skill factories for all jobs

PR #9516, opened 2025-08-24
Split up swordman skills

Danil0v3s:
* **Addressed Issue(s)**: 

* **Server Mode**: 

* **Description of Pull Request**: 

Following #9505 here's the implementation of Swordman (not any of its subclasses):

- [x] Autoberserk
- [x] Endure
- [x] Magnum Break
- [x] Provoke
- [x] Self Provoke

PR #9520, opened 2025-08-26
Split up Novice skills 

kreanlnwza:
Addressed Issue(s):
Server Mode:
Description of Pull Request:
Following https://github.com/rathena/rathena/pull/9505

* [x] Basic Skill (Not sure if I need to move.)
* [X]  First Aid
* [X]  Play Dead


src/map/skills/novice/callbaby.hpp
```diff
@@ -0,0 +1,11 @@
...
+// For more information, see LICENCE in the main folder
+
+#pragma once
+
+#include "../skill_impl.hpp"
+
+class SkillCallBaby : public SkillImpl {
+public:
+	SkillCallBaby(e_skill skill_id);
+};
```
Haydrich:
```suggestion
};

```

src/map/skills/novice/firstaid.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#include "firstaid.hpp"
+
+SkillFirstAid::SkillFirstAid(e_skill skill_id) : SkillImpl(skill_id) {
+}
```
Haydrich:
```suggestion
}

```

src/map/skills/novice/firstaid.hpp
```diff
@@ -0,0 +1,11 @@
...
+// For more information, see LICENCE in the main folder
+
+#pragma once
+
+#include "../skill_impl.hpp"
+
+class SkillFirstAid : public SkillImpl {
+public:
+	SkillFirstAid(e_skill skill_id);
+};
```
Haydrich:
```suggestion
};

```

src/map/skills/novice/trickdead.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#include "trickdead.hpp"
+
+SkillTrickDead::SkillTrickDead(e_skill skill_id) : SkillImpl(skill_id) {
+}
```
Haydrich:
```suggestion
}

```

src/map/skills/novice/trickdead.hpp
```diff
@@ -0,0 +1,11 @@
...
+// For more information, see LICENCE in the main folder
+
+#pragma once
+
+#include "../skill_impl.hpp"
+
+class SkillTrickDead : public SkillImpl {
+public:
+	SkillTrickDead(e_skill skill_id);
+};
```
Haydrich:
```suggestion
};

```

src/map/skills/novice/basic.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#include "basic.hpp"
+
+SkillBasic::SkillBasic(e_skill skill_id) : SkillImpl(skill_id) {
+}
```
Haydrich:
```suggestion
}

```

src/map/skills/novice/basic.hpp
```diff
@@ -0,0 +1,11 @@
...
+// For more information, see LICENCE in the main folder
+
+#pragma once
+
+#include "../skill_impl.hpp"
+
+class SkillBasic : public SkillImpl {
+public:
+	SkillBasic(e_skill skill_id);
+};
```
Haydrich:
```suggestion
};

```

src/map/skills/novice/callbaby.cpp
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#include "callbaby.hpp"
+
+SkillCallBaby::SkillCallBaby(e_skill skill_id) : SkillImpl(skill_id) {
+}
```
Haydrich:
```suggestion
}

```

src/map/skills/novice/skill_factory_novice.cpp
```diff
@@ -3,13 +3,23 @@
...
 	switch( skill_id ){
+		case NV_BASIC:
+			return std::make_unique<SkillBasic>(skill_id);
+		case NV_FIRSTAID:
+			return std::make_unique<SkillFirstAid>(skill_id);
+		case NV_TRICKDEAD:
+			return std::make_unique<SkillTrickDead>(skill_id);
+		case WE_CALLBABY:
+			return std::make_unique<SkillCallBaby>(skill_id);
+
```
Danil0v3s: Notice how we are creating the instances here: https://github.com/rathena/rathena/pull/9516/files#diff-57eb9662dceda1152a51c5c4fb197dfa92830ff06fe1fc72206a577d2329172fR15-R26.

The concrete skill impl doesn't take in the ID, just the abstraction

Danil0v3s (changes requested):
The goal is to migrate skills that implement any of the declared methods at skill_impl.hpp and omit the ones that don't.

From novice it seems to be:
- NV_FIRSTAID
- NV_TRICKDEAD

The rest I believe you can omit since they aren't present on any of the methods we declared at the base skill_impl.hpp

kreanlnwza:
> The goal is to migrate skills that implement any of the declared methods at skill_impl.hpp and omit the ones that don't.
> 
> From novice it seems to be:
> 
> * NV_FIRSTAID
> * NV_TRICKDEAD
> 
> The rest I believe you can omit since they aren't present on any of the methods we declared at the base skill_impl.hpp

Since we've already started, I think it would be better if we separate the classes.

src/map/skills/novice/skill_factory_novice.cpp
kreanlnwza: I read and looked at the code you made, and I'm very confused, but I will try to work on it anyway.

Danil0v3s:
> > The goal is to migrate skills that implement any of the declared methods at skill_impl.hpp and omit the ones that don't.
> > From novice it seems to be:
> > 
> > * NV_FIRSTAID
> > * NV_TRICKDEAD
> > 
> > The rest I believe you can omit since they aren't present on any of the methods we declared at the base skill_impl.hpp
> 
> Since we've already started, I think it would be better if we separate the classes.



> > The goal is to migrate skills that implement any of the declared methods at skill_impl.hpp and omit the ones that don't.
> > From novice it seems to be:
> > 
> > * NV_FIRSTAID
> > * NV_TRICKDEAD
> > 
> > The rest I believe you can omit since they aren't present on any of the methods we declared at the base skill_impl.hpp
> 
> Since we've already started, I think it would be better if we separate the classes.

Not really, we could've created the empty files for all classes at once but chose to implement by class as we move forward.

src/map/skill.cpp
```diff
@@ -14031,7 +14028,7 @@ TIMER_FUNC(skill_castend_id){
 					}
 				}
 				break;
-			case WE_CALLBABY:
+
```
Danil0v3s: Revert this as it doesnt belong to any of the methods we're migrating (see skill_impl.hpp)

src/map/skill.cpp
```diff
@@ -16221,9 +16218,7 @@ std::shared_ptr<s_skill_unit_group> skill_unitsetting(struct block_list *src, ui
 		 	val2 = sd->status.mother;
 		}
 		break;
-	case WE_CALLBABY:
-		if (sd) val1 = sd->status.child;
-		break;
+
```
Danil0v3s: Revert this as it doesnt belong to any of the methods we're migrating (see skill_impl.hpp)

src/map/skills/novice/trickdead.hpp
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#pragma once
+
+#include "../skill_impl.hpp"
+
+class SkillTrickDead : public SkillImpl {
+public:
+	SkillTrickDead(e_skill skill_id);
```
Danil0v3s, line 10: You removed this skill from skill.cpp but did not add the implementation here

src/map/skills/novice/firstaid.hpp
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#pragma once
+
+#include "../skill_impl.hpp"
+
+class SkillFirstAid : public SkillImpl {
+public:
+	SkillFirstAid(e_skill skill_id);
```
Danil0v3s, line 10: You removed this skill from skill.cpp but did not add the implementation here

src/map/skills/novice/callbaby.hpp
Danil0v3s, line 1: This does not implement any of the methods we declared at skill_impl.hpp

src/map/skills/novice/basic.hpp
Danil0v3s, line 1: This does not implement any of the methods we declared at skill_impl.hpp

Danil0v3s (commented):
Not really

Danil0v3s:
Basically the migration should be:

1. You see which class skills' implement any of the methods declared at skill_impl.hpp
  - castendNoDamageId
  - castendDamageId
  - calculateSkillRatio
  - modifyHitRate
  - applyAdditionalEffects
 2. You add the implementation of each method a skill implement and then move the code from the switch blocks from the original methods to the smaller class
 3. The code you move should be a 1:1 copy, no enhancements or anything as it will make code reviewing faster
 4. Skills that don't implement at least one of these methods should be omitted, exceptions are skills that implement only what's defined at:
   - StatusSkillImpl
   - WeaponSkillImpl

For point 4, you can register them as the abstract class if the default method is the only thing implemented, like SM_ENDURE for instance:
```
case SM_ENDURE:
    return std::make_unique<StatusSkillImpl>(SM_ENDURE);
```

PR #9526, opened 2025-08-31
Add default implementations for SkillImpl that throw an exception

vstumpf:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

Add default implementations for castendDamageId and castendNoDamageId that throw exceptions, so we can cleanly exit skill_castend_damage_id and skill_castsend_nodamage_id, respectively.

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -7496,7 +7496,12 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 
 	default:
 		if (std::shared_ptr<s_skill_db> skill = skill_db.find(skill_id); skill != nullptr && skill->impl != nullptr) {
-			skill->impl->castendDamageId(src, bl, skill_lv, tick, flag);
+			try {
+				skill->impl->castendDamageId(src, bl, skill_lv, tick, flag);
+			} catch (const SkillImplException& e) {
+				ShowWarning(e.what());
+				return 1;
+			}
 			break;
 		}
 
@@ -13671,7 +13676,12 @@ int32 skill_castend_nodamage_id (struct block_list *src, struct block_list *bl, 
 		std::shared_ptr<s_skill_db> skill = skill_db.find(skill_id);
 
 		if( skill != nullptr && skill->impl != nullptr ){
-			skill->impl->castendNoDamageId( src, bl, skill_lv, tick, flag );
+			try {
+				skill->impl->castendNoDamageId( src, bl, skill_lv, tick, flag );
+			} catch(const SkillImplException& e) {
+				ShowWarning(e.what());
+				return 1;
+			}
 			break;
 		}
 
--- a/src/map/skills/skill_impl.cpp
+++ b/src/map/skills/skill_impl.cpp
@@ -3,6 +3,9 @@
 
 #include "skill_impl.hpp"
 
+#include "../clif.hpp"
+#include "../status.hpp"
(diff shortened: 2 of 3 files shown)


src/map/skill.cpp
```diff
@@ -7496,7 +7496,12 @@ int32 skill_castend_damage_id (struct block_list* src, struct block_list *bl, ui
 
 	default:
 		if (std::shared_ptr<s_skill_db> skill = skill_db.find(skill_id); skill != nullptr && skill->impl != nullptr) {
-			skill->impl->castendDamageId(src, bl, skill_lv, tick, flag);
+			try {
+				skill->impl->castendDamageId(src, bl, skill_lv, tick, flag);
+			} catch (const SkillImplException& e) {
```
vstumpf, line 7501: not married to this implementation, but the things we need are:
1) return 1 (error) when the skill doesn't implement the specific method
2) A warning to the user that it's not right
3) We need to call `clif_skill_(no)damage` even if there's no implementation, to not mess up the clients

Lemongrass3110: Isnt that exactly what we want?
That the user doesnt have to implement/override useless functions?

vstumpf: I just meant that i couldn't really think of anything other than throwing an exception and catching it, so if anyone had any other suggestions I'm all ears.

Danil0v3s: How would this behave if I migrate a skill that does not have these methods? Is that even possible?

vstumpf: it's possible. if someone calls `skill_castend_nodamage_id(SM_BASH, ...)`, right now we don't error at all. This will imitate the original code that would complain in the log "Unknown skill in skill_castend_nodamage_id!"

Lemongrass3110: If this is truly necessary, simply add it as error message (not warning) in the default implementation.
No need to overcomplicate it with exception logic.

But with the current class design that means:
All skills have to implement nodamage and damage.

This would not really be necessary in my opinion.

src/map/skills/skill_impl.cpp
```diff
@@ -11,11 +14,15 @@ e_skill SkillImpl::getSkillId() const {
 }
 
 void SkillImpl::castendNoDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32 flag) const {
-	// no-op
+	clif_skill_nodamage(src, *target, getSkillId(), skill_lv);
```
Lemongrass3110, line 17: I wouldnt add this, even if the old code did it.
The skill execution was not successful, so why simulate that part visually on the client?

src/map/skills/skill_impl.cpp
```diff
@@ -11,11 +14,15 @@ e_skill SkillImpl::getSkillId() const {
...
+	throw SkillImplException("castendNoDamageId is not implemented for skill " + std::to_string(getSkillId()));
 }
 
 void SkillImpl::castendDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32 flag) const {
-	// no-op
+	clif_skill_damage(*src, *target, tick, status_get_amotion(src), status_get_dmotion(target), 0,
```
Lemongrass3110, line 22: I wouldnt add this, even if the old code did it.
The skill execution was not successful, so why simulate that part visually on the client?

PR #9530, opened 2025-09-02
Invincibility Times After Warp

### LINKED ISSUE #9523: Player invincibility time after warp should be 3s on normal maps
### rAthena Hash

ed68f79

### Client Date

2021-11-03

### Server Mode

Pre-Renewal, Renewal

### Result

Player invincibility time is 5 seconds on normal and PVP maps.
Player invincibility time is 10 seconds on GVG maps.

### Expected Result

Player invincibility time is 3 seconds on normal and PVP maps.
Player invincibility time is 10 seconds on GVG maps.

### How to Reproduce

Warp to a map, then wait until a monster starts attacking you.

### Official Information

Verified by testing on official servers.

_No response_

Playtester:
* **Addressed Issue(s)**: #9523 

* **Server Mode**: Both

- On normal maps you are now protected from damage and monster-aggro after warp for only 3 seconds
- Added a mapflag "invincible_time" that allows to set this time for individual maps
- Optimized code so that the invincibility time logic is only in one central place
- Fixed the "nosave.txt" mapflag file not being loaded in pre-re
- Fixes #9523


src/map/pc.cpp
```diff
@@ -831,9 +831,20 @@ static TIMER_FUNC(pc_invincible_timer){
 	return 0;
 }
 
-void pc_setinvincibletimer(map_session_data* sd, int32 val) {
+void pc_setinvincibletimer(map_session_data* sd) {
```
Atemo: how about
```suggestion
void pc_setinvincibletimer(map_session_data& sd) {
```

Playtester: This only works when you're guaranteed that sd is not nullptr. I would need to check again but I thought that was not the case everywhere.

src/map/map.cpp
```diff
@@ -4976,6 +4976,16 @@ bool map_setmapflag_sub(int16 m, enum e_mapflag mapflag, bool status, union u_ma
 			}
 			mapdata->setMapFlag(mapflag, status);
 			break;
+		case MF_INVINCIBLE_TIME:
+			if (status) {
+				nullpo_retr(false, args);
+
+				// TODO: do we really want a default value? and one that is different from the default value of the battle conf?
+				// Default is 5 seconds.
+				mapdata->setMapFlag(mapflag, (args->flag_val <= 0) ? 5000 : args->flag_val);
```
Playtester: Should default to 0 here, right?

src/map/pc.cpp
Atemo: `nullpo_retv(sd);` is called here, that means we expect sd not nullptr, right?

Lemongrass3110: Yes, we expect it to not be nullptr.
If it is called with nullptr in debug mode this will print a debug information.
In release mode it will do nothing and the code will continue and this will cause a crash.

doc/mapflags.txt
```diff
@@ -396,6 +396,15 @@ prtg_cas01	mapflag	skill_duration	HT_ANKLESNARE,400
 
 ---------------------------------------
 
+*invincible_time	{<duration>}
```
Lemongrass3110:
```suggestion
*invincible_time	<duration>
```
{} is for showing that this is optional. I made it mandatory.

npc/pre-re/mapflag/invincible_time.txt
```diff
@@ -0,0 +1,37 @@
+//============================================================
```
Lemongrass3110, line 9:
```suggestion
//===== rAthena Script =======================================
//= Mapflag: invincible time map setting.
//===== Description: ========================================= 
//= Sets the invincible time for when a player loads onto a map.
//===== Additional Comments: ================================= 
//= 1.0 Initial script. [Playtester]
//============================================================

//============================================================
```

PR #9532, opened 2025-09-03
Implementation of Normal Atk Rate Bonus

Haydrich:

* **Server Mode**: Renewal

* **Description of Pull Request**: Added the new bonus bNormalAtkRate https://www.divine-pride.net/forum/index.php?/topic/10404-kro-main-changelog-september-03-2025/

### DIFF
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -4398,7 +4398,7 @@ static void battle_calc_skill_base_damage(struct Damage* wd, block_list *src,blo
 						ATK_ADDRATE(wd->damage, wd->damage2, sd->bonus.crit_atk_rate);
 					}
 				}
-				else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_id == 0 || skill_tmp != nullptr && !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
+				else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_tmp == nullptr || !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
 					// custom, officially non_crit_atk_rate did not exist on pre-renewal
 					if (sd->bonus.non_crit_atk_rate > 0) {
 						ATK_ADDRATE(wd->damage, wd->damage2, sd->bonus.non_crit_atk_rate);
@@ -7961,7 +7961,7 @@ static struct Damage battle_calc_weapon_attack(block_list *src, block_list *targ
 						wd.damage2 += (int64)floor((float)(wd.damage2 * sd->bonus.crit_atk_rate / 100));
 				}
 			}
-			else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_id == 0 || skill_tmp != nullptr && !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
+			else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_tmp == nullptr || !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
 				ATK_ADDRATE(wd.damage, wd.damage2, sd->bonus.non_crit_atk_rate);
 			}
 


src/map/map.hpp
```diff
@@ -580,7 +580,8 @@ enum _sp {
 	SP_LONG_SP_GAIN_VALUE, SP_LONG_HP_GAIN_VALUE, SP_SHORT_ATK_RATE, SP_MAGIC_SUBSIZE, SP_CRIT_DEF_RATE, // 2093-2097
 	SP_MAGIC_SUBDEF_ELE, SP_REDUCE_DAMAGE_RETURN, SP_ADD_ITEM_SPHEAL_RATE, SP_ADD_ITEMGROUP_SPHEAL_RATE, // 2098-2101
 	SP_WEAPON_SUBSIZE, SP_ABSORB_DMG_MAXHP2, // 2102-2103
-	SP_SP_IGNORE_RES_RACE_RATE, SP_SP_IGNORE_MRES_RACE_RATE, SP_EMATK_HIDDEN, SP_SKILL_RATIO // 2104-2107
+	SP_SP_IGNORE_RES_RACE_RATE, SP_SP_IGNORE_MRES_RACE_RATE, SP_EMATK_HIDDEN, SP_SKILL_RATIO, // 2104-2107
+	SP_NORMAL_ATK_RATE, //2108
```
Haydrich:
```suggestion
	SP_NORMAL_ATK_RATE //2108
```

db/re/skill_db.yml
```diff
@@ -29429,9 +29450,11 @@ Body:
     Description: Bind Trap
     MaxLevel: 5
     Type: Misc
-    TargetType: Ground
+    TargetType: Self
```
Playtester: This change seems unrelated to this PR.

src/map/battle.cpp
```diff
@@ -8139,6 +8139,13 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 
 	battle_do_reflect(BF_WEAPON,&wd, src, target, skill_id, skill_lv); //WIP [lighta]
 
+	if (std::shared_ptr<s_skill_db> skill = skill_db.find(skill_id);
+		sd != nullptr
+		&& !(wd.type == DMG_CRITICAL || wd.type == DMG_MULTI_HIT_CRITICAL)
+		&& !(skill != nullptr && skill->inf2[INF2_IGNORENORMALATKBONUS])) {
+		ATK_ADDRATE(wd.damage, wd.damage2, sd->bonus.normal_atk_rate);
```
Playtester: Is it correct to place this at the very end of the calculation? For example now reflect damage will not be affected by it. Also mastery damage bonuses will be increased by it.

Lemongrass3110: That is the only thing i am not sure about with this PR, rest looks good for me.

src/map/script_constants.hpp
```diff
@@ -842,6 +842,7 @@
 	export_constant2("bAddItemSPHealRate", SP_ADD_ITEM_SPHEAL_RATE);
 	export_constant2("bAddItemGroupSPHealRate", SP_ADD_ITEMGROUP_SPHEAL_RATE);
 	export_constant2("bWeaponSubSize", SP_WEAPON_SUBSIZE);
+	export_constant2("bNormalAtkRate", SP_NORMAL_ATK_RATE);
```
Lemongrass3110: Missing in /doc/item_bonus.txt

db/re/skill_db.yml
```diff
@@ -4225,6 +4225,7 @@ Body:
       IsTrap: true
       AlterRangeResearchTrap: true
       IgnoreHovering: true
+      IgnoreNormalAtkBonus: true
```
Lemongrass3110: Do we need to add that to pre-renewal too?
Or should this bonus be exclusive to renewal?

src/map/battle.cpp
Atemo: Maybe the bonus should be around https://github.com/rathena/rathena/blob/bdffa63956f7c70b2391aa280b514d58641181cf/src/map/battle.cpp#L8007

For reference https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=8115&curpage=1
```
> The increased physical damage per hit effect applies to regular physical attacks that do not trigger a critical hit, or to attacks that trigger a critical hit but do not inflict critical damage. 
> Does not apply to magic skills, fixed damage skills, skills that do not receive a melee/ranged physical damage increase effect, and some exceptional skills.
```

db/re/skill_db.yml
Haydrich: Since i moved it inside `#ifdef RENEWAL` so no need to add it to pre-re section, it's exclusive to renewal.

src/map/battle.cpp
Haydrich: Updated accordingly to Atemo's informatrion

doc/item_bonus.txt
```diff
@@ -129,6 +129,7 @@ bonus bAtk,n;      			ATK + n (unofficial)
 bonus bAtk2,n;     			ATK2 + n
 bonus bAtkRate,n;  			ATK + n% that won't interfere with Damage modifier and SC_EDP (renewal mode only)
 bonus bWeaponAtkRate,n; 	Weapon ATK + n%
+bonus bNormalAtkRate,n;		Normal ATK + n% (renewal, the damage bonus doesn't increase critical damage)
```
Haydrich:
```suggestion
bonus bNormalAtkRate,n;		Normal ATK + n% (renewal mode only, the damage bonus doesn't increase critical damage)
```

src/map/battle.cpp
```diff
@@ -8005,6 +8005,12 @@ static struct Damage battle_calc_weapon_attack(struct block_list *src, struct bl
 		if (tsd && tsd->bonus.crit_def_rate != 0)
 			ATK_ADDRATE(wd.damage, wd.damage2, -tsd->bonus.crit_def_rate);
 	}
+	if (std::shared_ptr<s_skill_db> skill = skill_db.find(skill_id);
```
Playtester: If you use "else if" here you don't need to do the critical checks again.

Playtester: Actually looking at this again "crate" is actually a bonus to the critical bonus multiplier in renewal.

Where you should actually place this code to be consistent is where we use `bCritAtkRate` instead:
```cpp
			// CritAtkRate modifier
			if (wd.type == DMG_CRITICAL || wd.type == DMG_MULTI_HIT_CRITICAL) {
				if (skill_id > 0) {
					wd.damage += (int64)floor((float)(wd.damage * sd->bonus.crit_atk_rate / 200));
					if (is_attack_left_handed(src, skill_id))
						wd.damage2 += (int64)floor((float)(wd.damage2 * sd->bonus.crit_atk_rate / 200));
				}
				else {
					wd.damage += (int64)floor((float)(wd.damage * sd->bonus.crit_atk_rate / 100));
					if (is_attack_left_handed(src, skill_id))
						wd.damage2 += (int64)floor((float)(wd.damage2 * sd->bonus.crit_atk_rate / 100));
				}
			}
```
There's another place where this is implemented for pre-renewal where we could add it too, then it's consistent.

doc/item_bonus.txt
Playtester: Maybe place it below ` bCritAtkRate` and phrase it similar to it, e.g.:
```
bonus bNormalAtkRate,n;     		Increases non-critical damage by +n%
```
I recommend implementing it for pre-re also (see my other comment), we already get enough complaints about bAtkRate not working in pre-re.

satter001:
hi, i've found the randomoption lub file and it has DAMAGE_HIT_TARGET (increases hit damage on enemy) and DAMAGE_HIT_USER (reduces hit damage from enemy) lines, can you add item_randomopt_db.yml too?

eppc0330:
I suggest to change bonus name to `bHitAtkRate`
based on following two reasons :

 - The opposite concept of critical in Ro is HIT, and this bonus is also created to support jobs which do not have critical attack skills.
 - The name `NormalAtkRate` can be easily misundersood as the bonus strengthening normal attack, not a skill.

Playtester:
> I suggest to change bonus name to `bHitAtkRate` based on following two reasons :
> 
>     * The opposite concept of critical in Ro is HIT, and this bonus is also created to support jobs which do not have main attack skills with critical ability.
> 
>     * The name `NormalAtkRate` can be easily misundersood as the bonus strengthening normal attack, not a skill.

I haven't heard "Hit" in the context of non-critical attacks before. I would rather understand "Hit" as hit rate. Maybe just use "NonCriticalAtkRate" then.

db/re/skill_db.yml
Atemo:
```suggestion
    TargetType: Ground
```
Later, in a new PR

Atemo:
The bonus is called `AddDamage_HIT` in the client.
But it seems to me than naming it NonCriticalAtkRate would be more explicit.

src/map/battle.cpp
Atemo: Indeed it might make more sense to place it near crit_atk_rate, although this still needs to be checked on official server

eppc0330:
> > I suggest to change bonus name to `bHitAtkRate` based on following two reasons :
> > ```
> > * The opposite concept of critical in Ro is HIT, and this bonus is also created to support jobs which do not have main attack skills with critical ability.
> > 
> > * The name `NormalAtkRate` can be easily misundersood as the bonus strengthening normal attack, not a skill.
> > ```
> 
> I haven't heard "Hit" in the context of non-critical attacks before. I would rather understand "Hit" as hit rate. Maybe just use "NonCriticalAtkRate" then.

The bonus is called `명중 물리 데미지` in Korean, and it means HIT PHYSICAL DAMAGE. 
But yeah, `bNonCriticalAtkRate` looks clear.

Atemo:
According to tests the damage does not match KRO. It matches when the bonus is applied before the defense reduction (which isn't the case here), so the bonus should probably be around crit_atk_rate

src/map/battle.cpp
```diff
@@ -4393,9 +4393,18 @@ static void battle_calc_skill_base_damage(struct Damage* wd, block_list *src,blo
...
 
 #ifndef RENEWAL
-				if (sd->bonus.crit_atk_rate && (bflag & BDMG_CRIT)) { // add +crit damage bonuses here in pre-renewal mode [helvetica]
-					ATK_ADDRATE(wd->damage, wd->damage2, sd->bonus.crit_atk_rate);
+				if (bflag & BDMG_CRIT) { // add +crit damage bonuses here in pre-renewal mode [helvetica]
+					if (sd->bonus.crit_atk_rate > 0) {
+						ATK_ADDRATE(wd->damage, wd->damage2, sd->bonus.crit_atk_rate);
+					}
+				}
+				else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_id == 0 || skill_tmp != nullptr && !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
```
Lemongrass3110:
```suggestion
				else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_tmp == nullptr || !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
```

src/map/battle.cpp
```diff
@@ -7952,6 +7961,9 @@ static struct Damage battle_calc_weapon_attack(block_list *src, block_list *targ
 						wd.damage2 += (int64)floor((float)(wd.damage2 * sd->bonus.crit_atk_rate / 100));
 				}
 			}
+			else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_id == 0 || skill_tmp != nullptr && !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
```
Lemongrass3110:
```suggestion
			else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_tmp == nullptr || !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
```

src/map/battle.cpp
```diff
@@ -4393,9 +4393,18 @@ static void battle_calc_skill_base_damage(struct Damage* wd, block_list *src,blo
...
 #ifndef RENEWAL
-				if (sd->bonus.crit_atk_rate && (bflag & BDMG_CRIT)) { // add +crit damage bonuses here in pre-renewal mode [helvetica]
-					ATK_ADDRATE(wd->damage, wd->damage2, sd->bonus.crit_atk_rate);
+				if (bflag & BDMG_CRIT) { // add +crit damage bonuses here in pre-renewal mode [helvetica]
+					if (sd->bonus.crit_atk_rate > 0) {
+						ATK_ADDRATE(wd->damage, wd->damage2, sd->bonus.crit_atk_rate);
+					}
+				}
+				else if (std::shared_ptr<s_skill_db> skill_tmp = skill_db.find(skill_id); skill_id == 0 || skill_tmp != nullptr && !skill_tmp->inf2[INF2_IGNORENONCRITATKBONUS]) {
+					// custom, officially non_crit_atk_rate did not exist on pre-renewal
```
Lemongrass3110, line 4402: Since we add this "custom" logic here: We need to add the skill flags to pre-renewal too?

Atemo: I would prefer to let the user add the flag if they want to. Adding it means we'll keep the flag updated in skill_db pre-renewal as well, which I don't want.

Lemongrass3110: Ok 👍

gidzdlcrz: > I would prefer to let the user add the flag if they want to. Adding it means we'll keep the flag updated in skill_db pre-renewal as well, which I don't want.

It would be awesome to add this in pre-renewal as well.

Atemo: The flag is also available for pre-renewal, feel free to customize your server as you wish.

PR #9538, opened 2025-09-06
Change C style ERS to moden C++ style ERS pool template for safer usage.

inhyositsu:

* **Server Mode**: Both

- Reason for Change:
  The original ERS (especially the Create() method) exhibited different behavior on macOS compared to Linux/Windows, which could lead to unexpected results. With this update, initialization is always delegated to the class constructor itself, ensuring consistent behavior across platforms without platform-specific discrepancies.
- Summary: Replace the legacy C-style ERS (entry reuse system) with a modern, ty
ped, RAII C++ pool: `ERS<T>`.
  - Typed pool per T: Each `ERS<T>` manages pages of `T` and a freelist of retur
ned objects.
  - RAII: Pages are released when the pool object goes out of scope; no global t
eardown needed.
  - Not thread-safe: Same as the legacy ERS; cross-thread usage must be guarded
by callers (or use per-thread pools).
  - Opt-out switch: `DISABLE_ERS` macro makes `ERS<T>` fall back to plain `new`/
`delete` (no pooling), useful for debugging or bisects.

- New API (example):
  - Construction: `ERS<MyType> pool("name", /*chunkSize=*/64);`
  - Allocate: `auto* p = pool.alloc(args...);`
  - Free: `pool.free(p);`
  - Tuning: `pool.setChunkSize(n);` (alias: `set_chunk_size(n)`)
  - Info: `pool.entry_size() == sizeof(MyType);`
  - Cleanup: Destructor automatically frees backing pages (also available via `f
inalize()`).

- Removed/Changed:
  - Legacy `ERSOptions` and related behaviors (e.g. `ERS_OPT_CLEAR`, `ERS_OPT_CL
EAN`) are removed. Initialization/zeroing should be handled by callers as needed
.
  - No global report/final hooks (`ers_report`, `ers_final`). Pool management is
 local and RAII-driven.

- Migration notes:
  - `p = ers_alloc(pool, struct type);` → `p = pool.alloc();`
  - `ers_free(pool, p);` → `pool.free(p);`
  - `ers_chunk_size(pool, n);` → `pool.setChunkSize(n);`
  - `ers_destroy(pool);` → usually unnecessary with RAII; if heap-allocated, `de
lete pool;`
  - If any modules relied on global report/final, please remove those calls or r
eplace with module-local accounting.

- Performance (single-thread microbench, ops/s; higher is better):
  - 64B payload:
    - Immediate: Legacy ~534M vs `ERS<T>` ~374M
    - Bulk alloc: ~100M vs ~98M
    - Bulk free: Legacy ~128M vs `ERS<T>` ~211M
    - Churn (K=10k): ~44.2M vs ~42.6M
  - 256B payload:
    - Immediate: ~203M vs ~168M
    - Bulk alloc: ~33.3M vs ~30.1M
    - Bulk free: ~65.7M vs ~177.5M
    - Churn (K=10k): ~38.9M vs ~36.2M
  - Takeaway: The new `ERS<T>` is significantly faster on bulk free (high teardo
wn scenarios), and comparable in other patterns. A small benchmark harness can b
e provided under `tools/bench` if desired.

- Testing/Build:
  - Builds with modern compilers (Clang/GCC). No game logic changes; only memory
 management internals are affected.
  - Map server boots and runs. If maintainers prefer, we can stage adoption (add
 `ERS<T>` first, migrate call sites next, remove legacy last).

### DIFF
--- a/src/common/ers.hpp
+++ b/src/common/ers.hpp
@@ -79,7 +79,6 @@ public:
     void free(T* obj) { delete obj; }
 
     void setChunkSize(uint32) {}
-    void set_chunk_size(uint32 n) { setChunkSize(n); }
     std::size_t entry_size() const { return sizeof(T); }
     void finalize() {}
 
@@ -109,7 +108,6 @@ public:
     }
 
     void setChunkSize(uint32 n) { if (n > 0) chunkSize_ = n; }
-    void set_chunk_size(uint32 n) { setChunkSize(n); }
     std::size_t entry_size() const { return sizeof(T); }
 
     void finalize() {


src/common/ers.hpp
```diff
@@ -1,180 +1,138 @@
...
+public:
+    explicit ERS(const char* /*name*/, uint32 /*chunk*/ = 64) {}
+    ~ERS() {}
+
+    template <typename... Args>
+    T* alloc(Args&&... args) { return new T(std::forward<Args>(args)...); }
+    void free(T* obj) { delete obj; }
+
+    void setChunkSize(uint32) {}
+    void set_chunk_size(uint32 n) { setChunkSize(n); }
```
vstumpf: Do we need both of these? Why not just `setChunkSize()`?

inhyositsu:
> The changes look great! I know we want to move the DBMaps to unordered_maps, maybe in the future we can reuse this to make pool allocators for those containers

I'm planning to make those change too!

src/common/ers.hpp
inhyositsu: You are right, we don't need this, let me fix this.
Thank you!

PR #9545, opened 2025-09-08
Split up gunslinger skills

Danil0v3s:
* **Addressed Issue(s)**: 

* **Server Mode**: 


src/map/skills/gunslinger/skill_factory_gunslinger.cpp
```diff
@@ -3,13 +3,69 @@
...
+#include "spreadattack.hpp"
+#include "tracking.hpp"
+#include "tripleaction.hpp"
+
 std::unique_ptr<const SkillImpl> SkillFactoryGunslinger::create(const e_skill skill_id) const {
-#if 0
-	switch( skill_id ){
+	switch (skill_id) {
+		case GS_ADJUSTMENT:
+			return std::make_unique<SkillAdjustment>();
```
Atemo: Shouldn't it be
```suggestion
			return std::make_unique<StatusSkillImpl>(GS_ADJUSTMENT);
```
?

src/map/skills/gunslinger/skill_factory_gunslinger.cpp
```diff
@@ -3,13 +3,69 @@
...
+		case GS_GATLINGFEVER:
+			return std::make_unique<SkillGatlingfever>();
+		case GS_GLITTERING:
+			return std::make_unique<SkillGlittering>();
+		case GS_GROUNDDRIFT:
+			return std::make_unique<SkillGroundDrift>();
+		case GS_INCREASING:
+			return std::make_unique<SkillIncreasing>();
+		case GS_MADNESSCANCEL:
+			return std::make_unique<SkillMadnessCancel>();
```
Atemo: Shouldn't it be
```suggestion
			return std::make_unique<StatusSkillImpl>(GS_MADNESSCANCEL);
```
?

src/map/skills/gunslinger/skill_factory_gunslinger.cpp
```diff
@@ -3,13 +3,69 @@
...
+		case GS_FULLBUSTER:
+			return std::make_unique<SkillFullBuster>();
+		case GS_GATLINGFEVER:
+			return std::make_unique<SkillGatlingfever>();
+		case GS_GLITTERING:
+			return std::make_unique<SkillGlittering>();
+		case GS_GROUNDDRIFT:
+			return std::make_unique<SkillGroundDrift>();
+		case GS_INCREASING:
+			return std::make_unique<SkillIncreasing>();
```
Atemo:
```suggestion
			return std::make_unique<StatusSkillImpl>(GS_INCREASING);
```

src/map/skills/gunslinger/skill_factory_gunslinger.cpp
```diff
@@ -3,13 +3,69 @@
...
+		case GS_MADNESSCANCEL:
+			return std::make_unique<SkillMadnessCancel>();
+		case GS_MAGICALBULLET:
+			return std::make_unique<SkillMagicalBullet>();
```
Atemo: Couldn't we define the skill like this?
```suggestion
#ifdef RENEWAL
			return std::make_unique<StatusSkillImpl>(GS_MAGICALBULLET);
#else
			return std::make_unique<WeaponSkillImpl>(GS_MAGICALBULLET);
#endif
```

src/map/skills/gunslinger/skill_factory_gunslinger.cpp
Danil0v3s: Not really, check [this](https://github.com/rathena/rathena/blob/master/src/map/skills/archer/skill_factory_archer.cpp#L12). We feed in the interface and the compiler figures out the concrete implementation. Since the concrete implementation also has the SkillID at its constructor we can safely omit it here

PR #9546, opened 2025-09-08
Split up Ninja skills

Danil0v3s:
* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -1748,7 +1748,6 @@ int32 skill_additional_effect( block_list* src, block_list *bl, uint16 skill_id,
 	case KO_JYUMONJIKIRI:
 		sc_start(src,bl,SC_JYUMONJIKIRI,100,skill_lv,skill_get_time(skill_id,skill_lv));
 		break;
-	case SP_SOULEXPLOSION:
 	case KO_SETSUDAN: // Remove soul link when hit.
 		status_change_end(bl, SC_SPIRIT);
 		status_change_end(bl, SC_SOULGOLEM);
@@ -1860,12 +1859,6 @@ int32 skill_additional_effect( block_list* src, block_list *bl, uint16 skill_id,
 	case SJ_STAREMPEROR:
 		sc_start(src, bl, SC_SILENCE, 50 + 10 * skill_lv, skill_lv, skill_get_time(skill_id, skill_lv));
 		break;
-	case SP_SHA:
-		sc_start(src, bl, SC_SP_SHA, 100, skill_lv, skill_get_time(skill_id, skill_lv));
-		break;
-	case ABC_UNLUCKY_RUSH:
-		sc_start(src, bl, SC_HANDICAPSTATE_MISFORTUNE, 30 + 10 * skill_lv, skill_lv, skill_get_time(skill_id, skill_lv));
-		break;
 	case TR_ROSEBLOSSOM:// Rose blossom seed can only bloom if the target is hit.
 		sc_start4(src, bl, SC_ROSEBLOSSOM, 100, skill_lv, TR_ROSEBLOSSOM_ATK, src->id, 0, skill_get_time(skill_id, skill_lv));
 		[[fallthrough]];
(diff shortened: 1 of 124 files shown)


src/map/skill.cpp
```diff
@@ -5081,8 +5064,7 @@ int32 skill_castend_damage_id (block_list* src, block_list *bl, uint16 skill_id,
 			else
 				y = 0;
 			// Ashura Strike still has slide effect in GVG
-			if ((mbl == src || (!map_flag_gvg2(src->m) && !map_getmapflag(src->m, MF_BATTLEGROUND))) &&
-				unit_movepos(src, mbl->x + x, mbl->y + y, 1, 1)) {
+			if (unit_movepos(src, src->x + x, src->y + y, 1, 1)) {
```
Lemongrass3110: Nice catch... Insane that it was done like this...

src/map/skills/ninja/finalstrike.cpp
```diff
@@ -0,0 +1,70 @@
...
+	else
+		x = 0;
+	if (dir > 2 && dir < 6)
+		y = -i;
+	else if (dir == 7 || dir < 2)
+		y = i;
+	else
+		y = 0;
+
+	if ((!map_flag_gvg2(src->m) && !map_getmapflag(src->m, MF_BATTLEGROUND)) && unit_movepos(src, target->x + x, target->y + y, 1, 1)) {
```
Lemongrass3110:
```suggestion
	// No slide effect in GVG
	if (!map_flag_gvg2(src->m) && !map_getmapflag(src->m, MF_BATTLEGROUND) && unit_movepos(src, target->x + x, target->y + y, 1, 1)) {
```
Removed useless parantheses and added a small comment.

It should be considered to use the same code as in renewal (skill_check_unit_movepos)

src/map/skill.cpp
```diff
@@ -5081,8 +5064,7 @@ int32 skill_castend_damage_id (block_list* src, block_list *bl, uint16 skill_id,
 			else
 				y = 0;
 			// Ashura Strike still has slide effect in GVG
```
```suggestion
```
Can be removed now or if not it should be just something like "// Show slide effect"

src/map/skills/ninja/finalstrike.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+SkillFinalStrike::SkillFinalStrike() : WeaponSkillImpl(NJ_ISSEN) {
+}
+
+void SkillFinalStrike::castendDamageId(block_list *src, block_list *target, uint16 skill_lv, t_tick tick, int32 &flag) const {
+	int16 x, y;
+	int16 dir = map_calc_dir(src, target->x, target->y);
+
+#ifdef RENEWAL
+	// Move 2 cells (From target)
```
Lemongrass3110: x/y calculation is the same in renewal and pre-renewal and should be moved above.

src/map/skills/ninja/kamaitachi.cpp
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#include "kamaitachi.hpp"
+
+#include "map/pc.hpp"
```
Lemongrass3110, line 8:
```suggestion
#include "map/battle.hpp"
#include "map/map.hpp"
#include "map/pc.hpp"
```
battle_config + map_foreachinpath

src/map/skills/ninja/mirrorimage.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include "map/status.hpp"
+
+SkillMirrorImage::SkillMirrorImage() : StatusSkillImpl(NJ_BUNSINJYUTSU) {
+}
+
+void SkillMirrorImage::castendNoDamageId(block_list *src, block_list *target, uint16 skill_lv, t_tick tick, int32& flag) const {
+	status_change_end(target, SC_BUNSINJYUTSU); // on official recasting cancels existing mirror image [helvetica]
+	StatusSkillImpl::castendNoDamageId(src, target, skill_lv, tick, flag);
+	status_change_end(target, SC_NEN);
```
Lemongrass3110, line 15: Both status_change_end could be put into status.yml
Then this would be default StatusSkillImpl and could exist without .cpp/.hpp file

src/map/skills/ninja/vanishingslash.cpp
```diff
@@ -0,0 +1,24 @@
...
+#ifdef RENEWAL
+	base_skillratio += 20 * skill_lv;
+#else
+	base_skillratio += 10 * skill_lv;
+#endif
+}
+
+void SkillVanishingSlash::castendDamageId(block_list *src, block_list *target, uint16 skill_lv, t_tick tick, int32& flag) const {
+	if (skill_attack(BF_WEAPON,src,src,target,getSkillId(),skill_lv,tick,flag) > 0)
+		sc_start(src,src,SC_HIDING,100,skill_lv,skill_get_time(getSkillId(),skill_lv));
```
Lemongrass3110, line 24: Should be moved to applyAdditionalEffects instead.
This already covers the logic that it is only called when there was a hit.

Lemongrass3110 (approved):

PR #9548, opened 2025-09-08
Split up mage skills

Danil0v3s:
* **Addressed Issue(s)**: 

* **Server Mode**: 


src/map/skills/mage/fireball.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "fireball.hpp"
+
+#include "map/pc.hpp"
+
+SkillFireBall::SkillFireBall() : SkillImplRecursiveDamageSplash(MG_FIREBALL) {
+}
+
+void SkillFireBall::calculateSkillRatio(const Damage *wd, const block_list *src, const block_list *target, uint16 skill_lv, int32 &base_skillratio, int32 mflag) const {
+#ifdef RENEWAL
+	base_skillratio += 40 + 20 * skill_lv;
```
Atemo, line 13: Perhaps `base_skillratio` should be renamed `skillratio` to use https://github.com/rathena/rathena/blob/master/src/config/const.hpp#L95 - even through we don't use `RE_LVL_DMOD` here

Lemongrass3110: We can do that independent of this PR.
However the macro should actually not hide which parameters from outside are being used.
We were also thinking if we should hand over the base ratio via reference or if we should return an integer instead of void for this function

Lemongrass3110 (approved):

PR #9551, opened 2025-09-13
Fixed two iterator checks in navi.cpp

Lemongrass3110:

* **Server Mode**: Both

std::find_if was used to iterate over warps_outof, but the end was compared to warps_into


src/map/navi.cpp
```diff
@@ -323,8 +323,8 @@ int32 map_type(const struct map_data * m) {
 		return std::find_if(m->navi.warps_outof.begin(), m->navi.warps_outof.end(), [&link](const navi_link* link2) {
```
vstumpf, line 323: Actually I think one needs to be `into` and one `outof`, the idea is to see if one in_warp can't be reached by a out_warp

vstumpf (changes requested):

src/map/navi.cpp
Lemongrass3110: Feel free to push to the branch and overwrite my changes.

The comment "// find if any two warps in a map cannot be reached" was not clear to me if it should be the warps into the map or the warps out of the map

PR #9570, opened 2025-09-16
map: migrate DBMap* nick_db to STL unordered_map containers; fix requester tracking

inhyositsu:

* **Server Mode**: Both

* 
- Replace linked-list `charid_request` with `std::set<uint32> charid_list` in `c harid2nick` to dedupe and simplify pending requester tracking.
- Change `charid2nick::nick` from `char[NAME_LENGTH]` to `std::string`.
- Replace `static DBMap* nick_db` with `static std::unordered_map<uint32, charid 2nick> nick_db`.
- Remove DBMap-dependent helpers (`create_charid2nick`, `nick_db_final`) and ini t/final calls; rely on RAII.
- Update `map_addnickdb`, `map_delnickdb`, `map_reqnickdb`, and `map_charid2nick ` to use the new containers and `std::string` (use `c_str()` for clif).
- Fix latent bug where requester ID was never stored (previously `req->charid` u ninitialized). The new `charid_list` reliably tracks requesters.


src/map/map.cpp
```diff
@@ -2389,18 +2366,15 @@ chat_data* map_id2cd(int32 id){
 /// Returns the nick of the target charid or nullptr if unknown (requests the nick to the char server).
 const char* map_charid2nick(int32 charid)
 {
-	struct charid2nick *p;
-	map_session_data* sd;
-
-	sd = map_charid2sd(charid);
+	map_session_data* sd = map_charid2sd(charid);
 	if( sd )
```
Lemongrass3110:
```suggestion
	if(map_session_data* sd = map_charid2sd(charid); sd != nullptr)
```

src/map/map.cpp
```diff
@@ -2389,18 +2366,15 @@ chat_data* map_id2cd(int32 id){
...
-		return sd->status.name;// character is online, return it's name
+		return sd->status.name; // character is online, return its name
 
-	p = (struct charid2nick*)idb_ensure(nick_db, charid, create_charid2nick);
-	if( *p->nick )
-		return p->nick;// name in nick_db
+	auto it = nick_db.find(static_cast<uint32>(charid));
```
Lemongrass3110:
```suggestion
	auto it = nick_db.find(charid);
```
Instead of making a static_cast here, better change the function parameter to unsigned.

src/map/map.cpp
```diff
@@ -2161,16 +2141,13 @@ void map_reqnickdb(map_session_data * sd, int32 charid)
...
+	if (!entry.nick.empty()) {
+		clif_solved_charname(*sd, charid, entry.nick.c_str());
 		return;
 	}
-	// not in cache, request it
-	CREATE(req, struct charid_request, 1);
-	req->next = p->requests;
-	p->requests = req;
+	// not in cache, request it and subscribe
+	entry.charid_list.insert((uint32)sd->status.char_id);
```
Lemongrass3110:
```suggestion
	entry.charid_list.insert(sd->status.char_id);
```

src/map/map.cpp
```diff
@@ -2161,16 +2141,13 @@ void map_reqnickdb(map_session_data * sd, int32 charid)
 		return;
 	}
 
-	p = (struct charid2nick*)idb_ensure(nick_db, charid, create_charid2nick);
-	if( *p->nick )
-	{
-		clif_solved_charname( *sd, charid, p->nick );
+	auto &entry = nick_db[static_cast<uint32>(charid)];
```
Lemongrass3110:
```suggestion
	auto &entry = nick_db[charid];
```
Instead of making a static_cast here, better change the function parameter to unsigned.

src/map/map.cpp
```diff
@@ -2086,70 +2087,49 @@ int32 map_addflooritem(struct item *item, int32 amount, int16 m, int16 x, int16
...
 
 /// Removes the nick of charid from nick_db.
 /// Sends name to all pending requests on charid.
 void map_delnickdb(int32 charid, const char* name)
 {
-	struct charid2nick* p;
-	DBData data;
-
-	if (!nick_db->remove(nick_db, db_i2key(charid), &data) || (p = (struct charid2nick*)db_data2ptr(&data)) == nullptr)
+	auto it = nick_db.find(static_cast<uint32>(charid));
```
Lemongrass3110:
```suggestion
	auto it = nick_db.find(charid);
```
Instead of making a static_cast here, better change the function parameter to unsigned.

src/map/map.cpp
```diff
@@ -2086,70 +2087,49 @@ int32 map_addflooritem(struct item *item, int32 amount, int16 m, int16 x, int16
...
 {
-	struct charid2nick* p;
-	
 	if( map_charid2sd(charid) )
-		return;// already online
+		return; // already online
 
-	p = (struct charid2nick*)idb_ensure(nick_db, charid, create_charid2nick);
-	safestrncpy(p->nick, nick, sizeof(p->nick));
+	auto &entry = nick_db[static_cast<uint32>(charid)];
```
Lemongrass3110:
```suggestion
	auto &entry = nick_db[charid];
```
Instead of making a static_cast here, better change the function parameter to unsigned.

src/map/map.cpp
```diff
@@ -2086,70 +2087,49 @@ int32 map_addflooritem(struct item *item, int32 amount, int16 m, int16 x, int16
...
-		if( sd != nullptr )
-			clif_solved_charname( *sd, charid, p->nick );
-		aFree(req);
+	if (!entry.charid_list.empty()) {
+		for (uint32 requester_id : entry.charid_list) {
+			if (map_session_data* sd = map_charid2sd((int32)requester_id); sd != nullptr)
+				clif_solved_charname(*sd, charid, entry.nick.c_str());
+		}
+		entry.charid_list.clear();
 	}
```
Lemongrass3110: This logic is the same in map_addnickdb and map_delnickdb.
Can you put this into a function of the struct charid2nick please?
void charid2nick::notify() maybe?

src/map/map.cpp
```diff
@@ -2086,70 +2087,49 @@ int32 map_addflooritem(struct item *item, int32 amount, int16 m, int16 x, int16
...
-	struct charid2nick* p;
-	
 	if( map_charid2sd(charid) )
-		return;// already online
+		return; // already online
 
-	p = (struct charid2nick*)idb_ensure(nick_db, charid, create_charid2nick);
-	safestrncpy(p->nick, nick, sizeof(p->nick));
+	auto &entry = nick_db[static_cast<uint32>(charid)];
+	entry.nick.assign(nick ? nick : "");
```
Lemongrass3110:
```suggestion
	entry.nick.assign(nick != nullptr ? nick : "");
```
Or maybe only assign a value, if nick != nullptr.
Empty string should be the default value from the constructor anyway.

src/map/map.cpp
```diff
@@ -143,15 +145,14 @@ bool agit2_flag = false;
...
 int32 night_flag = 0; // 0=day, 1=night [Yor]
 
-struct charid_request {
-	struct charid_request* next;
-	int32 charid;// who want to be notified of the nick
-};
 struct charid2nick {
-	char nick[NAME_LENGTH];
-	struct charid_request* requests;// requests of notification on this nick
+	std::string nick;             // cached nickname for offline char
```
Lemongrass3110:
```suggestion
	std::string nick; // cached nickname for offline char
```

src/map/map.cpp
inhyositsu: I think a better way is to change map_addnickdb(int32, const char*) to map_addnickdb(uint32, const char*) to match the same type from the original source mmo_charstatus::char_id

src/map/map.cpp
inhyositsu: I think a better way is to change map_addnickdb(int32, const char*) to map_addnickdb(uint32, const char*) to match the same type from the original source mmo_charstatus::char_id

src/map/map.cpp
inhyositsu: nick is now std::string type, so it's not possible to be nullptr itself.
But I think change it to
`entry.nick = nick;`
is better reading here

inhyositsu:
> Thank you for splitting it up!
> 
> Looks very promising! :-)

Hello, I’m not very familiar with GitHub, so I apologize if I didn’t make the changes correctly or request another review properly.
I’ve made the suggested changes. Could you please review them again whenever you have time?

Thank you!

PR #9578, opened 2025-09-23
Fixed Songs Not Preventing Attacks

Playtester:
* **Addressed Issue(s)**: Follow-up to c8fdc58

- Fixed small mistake that sneaked in and suddenly allowed attacks on all regular songs
- Follow-up to c8fdc58

### DIFF
--- a/src/map/status.cpp
+++ b/src/map/status.cpp
@@ -5714,8 +5714,8 @@ void status_calc_state( block_list& bl, status_change& sc, std::shared_ptr<s_sta
 							break;
 						}
 #endif
-						restriction = true;
 					}
+					restriction = true;
 					break;
 
 				case SC_LONGING:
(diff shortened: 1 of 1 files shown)


src/map/status.cpp
```diff
@@ -5707,15 +5707,9 @@ void status_calc_state( block_list& bl, status_change& sc, std::shared_ptr<s_sta
...
 			switch( type ){
 				case SC_DANCING:
-					if( sce.val4 != 0 ){
-#ifndef RENEWAL
-						// In Pre-Renewal you can attack when SC_LONGING is active
-						if( sc.getSCE( SC_LONGING ) != nullptr ){
-							break;
-						}
-#endif
+					// Normal songs prevent attacks; Ensembles prevent attacks as long Longing is not active
```
Lemongrass3110, line 5710: Did you confirm that the behavior is the same on pre-renewal and renewal?
Rest looks good.

Playtester: There's no "NoAttackCond" for SC_DANCING in Renewal, so this code was never reached in Renewal anyway, so it can't be impacted.

Lemongrass3110 (approved):

Issue #9583, opened 2025-09-27
Stat increase cost is not updated when changing jobs


vE5li:
### rAthena Hash

928e1cb53ea7f8306b5d7bcb951645e692d94549

### Client Date

Korangar v0.1.0-2025-06-08

### Server Mode

Renewal

### Result

Stat increase cost is not updated when changing jobs.

This seems to be an old and known issue, see https://forums.warpportal.com/index.php?/topic/62978-stuck-at-99-after-3rd-job-change/

### Relevant Log Output

```Shell

```

### Expected Result

The stat increase cost should be updated when changing jobs.

### How to Reproduce

- Create a new character
- Set a stat to 99
- Change to a job that has a higher maximum stats (for example `cardinal`)
- Client is not informed of the new stat increase cost, making increasing the stat impossible until reconnecting or leveling up

### Official Information

This seems to be the same behavior as the official servers but it's very annoying for players so I hope we can improve rather than copy.

### Modifications that may affect results

I am not using the official client but rather [Korangar](https://github.com/vE5li/korangar) but I am confident that this is unrelated to the issue.

PR #9584, opened 2025-09-27
Update cost indicator when changing jobs

vE5li:
* **Addressed Issue(s)**: 
* https://github.com/rathena/rathena/issues/9583

* **Server Mode**: 
* Both

* Update cost indicator when changing jobs so that players don't have to manually reconnect.

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -10915,7 +10915,7 @@ bool pc_jobchange(map_session_data *sd,int32 job, char upper)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/map/pc.cpp
```diff
@@ -10915,6 +10915,14 @@ bool pc_jobchange(map_session_data *sd,int32 job, char upper)
 		}
 	}
 
+	// Update increase cost indicator (vE5li)
```
Lemongrass3110:
```suggestion
	// Update increase cost indicator
```

PR #9591, opened 2025-10-11
Corrected run_tomb function

gidzdlcrz:
* **Server Mode**: 
Both

  * The key change is adding `sd->npc_id = nd->id;`. This attaches the player to the NPC, which prevents them from moving or performing other actions while the npc dialog is displayed.

  * Also, I modified the ``npc_scriptcont`` block to handle tomb NPCs without script states. This handles tomb NPCs that don't have script states by resetting sd->npc_id when the player closes the dialog.

### DIFF
--- a/src/map/npc.cpp
+++ b/src/map/npc.cpp
@@ -2186,6 +2186,9 @@ void run_tomb(map_session_data* sd, npc_data* nd)
 	char buffer[200];
 	char time[10];
 
+	// Attach player to NPC to prevent movement/actions  
+	sd->npc_id = nd->id;  
+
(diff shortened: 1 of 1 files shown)


YgorCollpy:
Both changes caused an error when compiling the emulator.
Did they work on your emulator?

gidzdlcrz:
> Both changes caused an error when compiling the emulator. Did they work on your emulator?

Tried and tested to clean rA. Update your emulator.

YgorCollpy:
Ok, in my case im still using struct:
	// Attach player to NPC to prevent movement/actions
	sd->npc_id = nd->bl.id;
And works, thanks Gidz o/

PR #9599, opened 2025-10-23
Config to give MVP Reward to Loot Priority Player

Playtester:
* **Addressed Issue(s)**: It's a feature

* **Server Mode**: Both

- Added a custom config that makes the MVP reward go to the player with the first loot priority instead
  * Officially it goes to the player with the highest value for [damage dealt]+[damage tanked] (the MVP)
  * This restores legacy Athena behavior that makes it easier to see who got loot priority
  * However, it will make the MVP reward no longer go to the actual MVP, but to the one with loot priority
  * This will have a strong bias towards the first attacker unless you also change first_attack_loot_bonus

### DIFF
--- a/src/map/mob.cpp
+++ b/src/map/mob.cpp
@@ -2892,12 +2892,17 @@ int32 mob_getdroprate(block_list *src, std::shared_ptr<s_mob_db> mob, int32 base
 /**
  * Returns the MVP player based on the monster's damage log
  * This player has the highest value when damage dealt and damage tanked are added together
+ * @param first_sd: Player with loot priority (for custom legacy behavior)
  * @return The MVP player
  */
-map_session_data* mob_data::get_mvp_player() {
+map_session_data* mob_data::get_mvp_player(map_session_data* first_sd) {
 	// There cannot be an MVP player if the monster is not an MVP
 	if (this->get_bosstype() != BOSSTYPE_MVP)
 		return nullptr;
+
+	// Skip MVP player determination and return the player with loot priority instead
+	if (battle_config.mvp_to_loot_priority)
+		return first_sd;
 
 	int64 mvp_damage = 0;
 	map_session_data* mvp_sd = nullptr;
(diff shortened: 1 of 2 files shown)


Atemo (approved):

lunamoonf:
Tested this and encounter some problem if i set mvp_to_loot_priority to yes
every time i kill monster the MvP effect always show above my character head

kill normal monster like poring mvp effect will show

Playtester:
> Tested this and encounter some problem if i set mvp_to_loot_priority to yes every time i kill monster the MvP effect always show above my character head
> 
> kill normal monster like poring mvp effect will show

Good spot, I need to rewrite it.

Playtester:
> Tested this and encounter some problem if i set mvp_to_loot_priority to yes every time i kill monster the MvP effect always show above my character head
> 
> kill normal monster like poring mvp effect will show

Try now.

src/map/mob.cpp
```diff
@@ -2892,13 +2892,18 @@ int32 mob_getdroprate(block_list *src, std::shared_ptr<s_mob_db> mob, int32 base
...
  * @return The MVP player
  */
-map_session_data* mob_data::get_mvp_player() {
+map_session_data* mob_data::get_mvp_player(map_session_data* first_sd) {
 	// There cannot be an MVP player if the monster is not an MVP
 	if (this->get_bosstype() != BOSSTYPE_MVP)
 		return nullptr;
 
+	// Skip MVP player determination and return the player with loot priority instead
+	if (battle_config.mvp_to_loot_priority)
```
Lemongrass3110, line 2904:
```suggestion
	if (battle_config.mvp_to_loot_priority != 0 && first_sd != nullptr)
```

Playtester: Should we really add this check? That would make it look for the MVP player if nobody has loot priority, which is pointless. Better just return nullptr here.

PR #9605, opened 2025-10-29
Fix for Unable to restore stack! Double Continuation

gidzdlcrz:
* **Addressed Issue(s)**:
 https://github.com/rathena/rathena/issues/8625

* **Server Mode**: 
Both

  * A temporary fix to restrict item usage during NPC Dialogs to prevent previous script loss.


Lemongrass3110:
Sorry this would just be a cosmetic fix, that would hide the underlying issue.
I was finally able to reproduce this and will fix this with #9606

However: Thank you for your contribution!

PR #9606, opened 2025-10-29
Fixed an issue with item_enabled_npc

Lemongrass3110:
* **Addressed Issue(s)**: #8625

* **Server Mode**: Both

By default only pre-renewal has been affected.
However when changing the battle config item_enabled_npc this issue may also occur on renewal.
 
Thanks to @thanna


sinju92:
Bug when callfunc npc with Next; get stuck

Issue #9617, opened 2025-11-04
clif_getareachar_unit function bug for admin accounts


gidzdlcrz:
### rAthena Hash

Latest

### Client Date

2025

### Server Mode

Pre-Renewal

### Result

I noticed a bug with monsters that have cloaking skills. ~~I think the refresh mechanism~~ is not properly handling invisible or cloaked units, causing their shadows to become visible on the ground.

**EDIT**
It seems that refresh mechanism isnt the culprit so I will be editing the title. It's the `clif_getareachar_unit` function.

**EDIT 2**
It seems the bug occurs only if you have `<aid></aid>` tag is present in clientinfo.xml. Thanks to @Daraen1 for pointing it out.

### Relevant Log Output

```Shell
N/A
```

### Expected Result

When you use `@refresh` the monster shadows who are in cloaking state should not appear.

### How to Reproduce

1. Make sure that you are a YELLOW name (admin account in clientside)
2. Spawn a monster with Cloaking Skill. In my case i used Whisper (1185)
3. Use `@refresh` command
4. See result

OR

1. Make sure that you are a YELLOW name (admin account in clientside)
2. Spawn a monster with Cloaking Skill. In my case i used Whisper (1185)
3. Move away from the monster frame
4. Go back to last known position of the monsters you spawn
5. See result

Screen Record:
FOR YELLOW NAMES

https://github.com/user-attachments/assets/b8af6a6a-411c-4cee-8271-a19001de7195

### Modifications that may affect results

N/A

Playtester:
Cannot reproduce this on a 2021-11-03 client.

Daraen1:
Also can you check if you have  `<aid>YourAccountID</aid>` in your clientinfo. IIRC you can seen hiden unit with these tags.
(Maybe it's not related, just trying to help)

gidzdlcrz:
> Also can you check if you have `<aid>YourAccountID</aid>` in your clientinfo. IIRC you can seen hiden unit with these tags. (Maybe it's not related, just trying to help)

This is interesting. Yes my character is yellow. I will try with normal account. Thank you!

**EDIT**
Confirmed. This bug only occurs when the `<aid></aid>` tag is present in clientinfo.xml. Kudos @Daraen1

gidzdlcrz:
Closing, This was fixed in https://github.com/rathena/rathena/commit/1052cc30c19a17360b13346d45e93449ab133c5c

PR #9621, opened 2025-11-04
Adds visibility check for cloaked units in clif_getareachar_unit

gidzdlcrz:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9617

* **Server Mode**: Both

  * The fix needs to be in **clif_getareachar_unit** to catch all cases where units enter your view range, whether through movement, refresh, or any other mechanism.

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -5031,7 +5031,7 @@ void clif_getareachar_unit( map_session_data* sd,block_list *bl ){
 		return;
 	}
 
-	// Prevents visual bug where monster shadows appear when entering view range
+	// Prevents visual bug where unit shadows appear when entering view range
 	if (battle_config.hide_cloaked_units & bl->type) {
 		if (status_change* sc = status_get_sc(bl); sc != nullptr && sc->option & (OPTION_HIDE | OPTION_CLOAK | OPTION_CHASEWALK | OPTION_INVISIBLE)) {
 			return; // Hide this unit type


src/map/clif.cpp
```diff
@@ -5031,6 +5031,14 @@ void clif_getareachar_unit( map_session_data* sd,block_list *bl ){
 		return;
 	}
 
+	// Prevents visual bug where monster shadows appear when entering view range
+	if (bl->type == BL_MOB) {
+		status_change* sc = status_get_sc(bl);
+		if (sc && sc->option & (OPTION_HIDE | OPTION_CLOAK | OPTION_CHASEWALK | OPTION_INVISIBLE)) {
+			return; // Don't show cloaked monsters
```
Playtester: This might create all kinds of bugs. Officially the position of even hidden monsters is transferred to the client as far as I know. Did you test how this works with NPC_EMOTION? Do you still see it at the right position for example?

Also I assume this should probably be a custom config, unless we have some packet captures that show that cloaking monster positions are no longer transferred officially.

Lemongrass3110: And if we do that, we should not only do that for monsters, most likely.

gidzdlcrz: > This might create all kinds of bugs. Officially the position of even hidden monsters is transferred to the client as far as I know. Did you test how this works with NPC_EMOTION? Do you still see it at the right position for example?
> 
> Also I assume this should probably be a custom config, unless we have some packet captures that show that cloaking monster positions are no longer transferred officially.

No I did not since emotions is really visible even the unit is in HIDING or CLOAKING state. Tho, I conducted another test where monster uses emotion while there's this setting. It worked as normal.

image

There's no shadow showing yet the monster emotion still function. I even checked with box of sunlight.


> And if we do that, we should not only do that for monsters, most likely.

I specifically chooses only BL_MOB since that's where I saw the problem. If the community wants I can also make a battle config for this PR.

Initial Idea is like this:
```
// Should cloaked monsters be completely hidden from client? (Note 1)  
// Official behavior may send position data even for cloaked monsters  
// 0: Send position with OPTION_CLOAK (official suspected behavior)  
// 1: Don't send spawn packet at all (custom behavior, fixes shadow bug)  
// Default: 0 (Off)
hide_cloaked_monsters: 1  
```
Else, we can add flags beside monster only.

Playtester: Why not make a (Note 3) type config? By default 0 (used for no type). If you put it to 2 it would be mob-only. 3 would be PC+mob, etc.

> // Note 3: Value is a bit field. If no description is given,
> //         assume unit types (1: Pc, 2: Mob, 4: Pet, 8: Homun, 16: Mercenary, 128: NPC, 512: Elemental)

gidzdlcrz: > Why not make a (Note 3) type config? By default 0 (used for no type). If you put it to 2 it would be mob-only. 3 would be PC+mob, etc.
> 
> > // Note 3: Value is a bit field. If no description is given,
> > //         assume unit types (1: Pc, 2: Mob, 4: Pet, 8: Homun, 16: Mercenary, 128: NPC, 512: Elemental)

An excellent suggestion. I will check whether my skills can do this.

gidzdlcrz: made some changes. Please check. Thanks!

src/map/clif.cpp
```diff
@@ -5031,6 +5031,17 @@ void clif_getareachar_unit( map_session_data* sd,block_list *bl ){
 		return;
 	}
 
+	// Prevents visual bug where monster shadows appear when entering view range
+	if (battle_config.hide_cloaked_units) {  
+		status_change* sc = status_get_sc(bl);  
+		if (sc && sc->option & (OPTION_HIDE | OPTION_CLOAK | OPTION_CHASEWALK | OPTION_INVISIBLE)) {
```
Daegaladh:
```suggestion
		if (status_change* sc = status_get_sc(bl);  sc != nullptr && sc->option & (OPTION_HIDE | OPTION_CLOAK | OPTION_CHASEWALK | OPTION_INVISIBLE)) {  
```

Daegaladh (changes requested):

conf/battle/client.conf
```diff
@@ -171,3 +171,16 @@ macro_detection_punishment_time: 0
...
+// Default (Official): 0
+// 0: Show all cloaked units (default behavior)  
+// 1: Hide cloaked PCs (BL_PC = 0x001)  
+// 2: Hide cloaked Mobs (BL_MOB = 0x002)  
+// 4: Hide cloaked Pets (BL_PET = 0x004)  
+// 8: Hide cloaked Homunculus (BL_HOM = 0x008)  
+// 16: Hide cloaked Mercenaries (BL_MER = 0x010)  
+// 128: Hide cloaked NPCs (BL_NPC = 0x080)  
+// 512: Hide cloaked Elementals (BL_ELEM = 0x200)  
+// Values can be combined (e.g., 2 = Mob + NPC, 130 = Mob + NPC)
```
Playtester: You don't need to explain all of this here. (Note 3) already explains that.

src/map/clif.cpp
```diff
@@ -5031,6 +5031,16 @@ void clif_getareachar_unit( map_session_data* sd,block_list *bl ){
 		return;
 	}
 
+	// Prevents visual bug where monster shadows appear when entering view range
+	if (battle_config.hide_cloaked_units) {
```
Playtester: You can just check `battle_config.hide_cloaked_units&bl->type` here directly, then you don't need the if below and save processing time too.

conf/battle/client.conf
gidzdlcrz: Alright.

src/map/clif.cpp
gidzdlcrz: like this? 

```
if (battle_config.hide_cloaked_units & bl->type) {  
    if (status_change* sc = status_get_sc(bl); sc != nullptr && sc->option & (OPTION_HIDE | OPTION_CLOAK | OPTION_CHASEWALK | OPTION_INVISIBLE)) {  
        return;  
    }  
}
```

Playtester: Yes.

gidzdlcrz: Updated. Thanks!

src/map/battle.cpp
```diff
@@ -12378,6 +12378,7 @@ static const struct _battle_data {
 	{ "major_overweight_rate",              &battle_config.major_overweight_rate,           90,     0,      100             },
 	{ "trade_count_stackable",              &battle_config.trade_count_stackable,           1,      0,      1,              },
 	{ "enable_bonus_map_drops",             &battle_config.enable_bonus_map_drops,          1,      0,      1,              },
+	{ "hide_cloaked_units",                 &battle_config.hide_cloaked_units,              0,      0,      INT_MAX,        },
```
Playtester: Maximum should be BL_ALL I think. Otherwise I think it's good now.

gidzdlcrz: Oh yeah! Since It was changed to BL types. Thank you!

Playtester (approved):

Playtester:
@gidzdlcrz Did you test after all the changes and can confirm the option is still working as you intended?

gidzdlcrz:
> @gidzdlcrz Did you test after all the changes and can confirm the option is still working as you intended?

Yes. Tests done in cloaking monster (whisper) and npc that has ``hideonnpc`` script. Shadows no longer appear when have 130 in config.

Issue #6451, opened 2022-01-07
Episode 16.2 : Weekend Memorial Dungeon


Everade:
**Missing Official Content**
- [x] 2016.06.29 - Weekend Memorial Dungeon					https://ro.gnjoy.com/news/update/View.asp?seq=185&curpage=4

PR #9625, opened 2025-11-06
Removed spawn of monster Pouring in mal_dun01

Elphin-l:
* **Addressed Issue(s)**: The monster Pouring can currently be found on mal_dun01 and it drops the item Celebrate Egg (ID: 23724).

* **Server Mode**: Renewal

### DIFF
--- a/npc/re/mobs/dungeons/mal_dun.txt
+++ b/npc/re/mobs/dungeons/mal_dun.txt
@@ -23,4 +23,3 @@ mal_dun01,0,0	monster	Aster	1266,20,5000
 mal_dun01,0,0	monster	Red Eruma	2197,60,5000
 mal_dun01,0,0	monster	Siorava	2199,45,5000
 mal_dun01,0,0	monster	Wild Rider	2208,1,5000
-mal_dun01,0,0	monster	Pouring	1894,30,5000


Ossa88:
So digging into this was rather interesting in a variety of ways.


The fact that Pouring is on mal_dun01 is due to iRO. As referenced on https://db.irowiki.org/db/monster-info/1894/
But, not listed on Divine as mentioned above. Meaning that on iRO the pouring is possibly a natural spawning creature.

YET, as the above link mentions Pourings on iRO only drop unidentified fish.
On kRO they drop the current re/mob_db.yml drop list.

The update to re/mob_db.yml is due to a change from 15 months ago the undentified_fish -> celebrate_egg
This was done because on kRO pouring around 2020 the pouring dropped Celebrate_egg for a limited time event. The Pouring no longer spawns naturally from what it seems according to Divine.

YET AGAIN, unknown_fish ID #6451 is a required quest item for the quests_malangdo.txt
The quests_malangdo.txt is part of another event quest involving the island and various Grade Coins, E in this case, for the re/merchants/coin_exhchange.txt


Meaning that the malangdo_quest.txt is (more than likely) following whatever iRO did, which differed from that of kRO.


If someone could check the map in both offical servers, and then test the loot if the mob exists, we would get the exact answer.
Then of which the question comes down to, follow kRO or to follow iRO.


Edit: Further I did not realize this earlier... The version here has 999 DEX which would make it the kRO version, but still I am assuming that its old drop being part of of the malangdo_quest.txt is an iRO thing.

Atemo (approved):

Issue #9629, opened 2025-11-07
Invalid getwaitingroomstate return type


Singe-Horizontal:
### rAthena Hash

f2dad3f153

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

When calling getwaitingroomstate and no npc is found or it has is no chatroom, the command always returns -1 although a string type is expected for a few cases.

### Relevant Log Output

```Shell

```

### Expected Result

Those cases must return an empty string instead.

### How to Reproduce

Call `getwaitingroomstate` on an npc without chatroom for type 4,5 or 16 and assign it to a string variable.

### Official Information

N/A

PR #9630, opened 2025-11-07
Fixed getwaitingroomstate invalid return type

Singe-Horizontal:
* **Addressed Issue(s)**: 
#9629

* **Server Mode**: 
 Both

 A few information types are expecting a string, which wasn't checked when the npc or its chatroom is null.
 This adds a check to return `"null"` string instead of `-1` in those cases.

### DIFF
--- a/doc/script_commands.txt
+++ b/doc/script_commands.txt
@@ -7865,7 +7865,6 @@ The valid information types are:
  WIT_OVERTRIGGER - Whether the amount of users in the waiting room is higher than the trigger
 			       number.
 
-This function returns -1 if the type is unknown.
 If no npc is found or npc has no chatroom, return -1 when expecting a number
 or "null" for strings.
 


src/map/script.cpp
```diff
@@ -13248,7 +13248,14 @@ BUILDIN_FUNC(getwaitingroomstate)
...
 	if( nd == nullptr || (cd=(chat_data *)map_id2bl(nd->chat_id)) == nullptr )
 	{
-		script_pushint(st, -1);
+		switch(type){
+			case 4:
+			case 5:
+			case 16:
+				script_pushconststr(st,"null");
+				break;
+			default: script_pushint(st,-1);break;
```
Atemo: Please update the doc https://github.com/rathena/rathena/blob/master/doc/script_commands.txt#L7846 to document the return value based on the type

Lemongrass3110: Could we list the "known" types and (at least) give a warning when an unknown value is used?
I mean I would prefer to return SCRIPT_CMD_FAILURE in this case.
It is okay to return -1 for known types.

Lemongrass3110 (approved):

src/map/script.cpp
```diff
@@ -13244,26 +13244,50 @@ BUILDIN_FUNC(getwaitingroomstate)
 	if( script_hasdata(st,3) )
 		nd = npc_name2id(script_getstr(st, 3));
 	else
-		nd = (npc_data *)map_id2bl(st->oid);
+		nd = static_cast<npc_data*>(map_id2bl(st->oid));
```
Lemongrass3110:
```suggestion
		nd = map_id2nd(st->oid);
```

src/map/script.cpp
```diff
@@ -13244,26 +13244,50 @@ BUILDIN_FUNC(getwaitingroomstate)
...
+		case WIT_USERCOUNT:
+		case WIT_MAXUSERS:
+		case WIT_HASTRIGGER:
+		case WIT_ISDISABLED:
+		case WIT_TITLE:
+		case WIT_PASSWORD:
+		case WIT_EVENTNAME:
+		case WIT_ISFULL:
+		case WIT_OVERTRIGGER:
+			if(nd == nullptr || (cd=static_cast<chat_data*>(map_id2bl(nd->chat_id))) == nullptr){
```
Lemongrass3110:
```suggestion
			if(nd == nullptr || (cd=map_id2cd(nd->chat_id)) == nullptr){
```

src/map/script.cpp
```diff
@@ -13244,26 +13244,50 @@ BUILDIN_FUNC(getwaitingroomstate)
...
+		case WIT_USERCOUNT: script_pushint(st,cd->users); break;
+		case WIT_MAXUSERS: script_pushint(st,cd->limit); break;
+		case WIT_HASTRIGGER: script_pushint(st,cd->trigger&0x7f); break;
+		case WIT_ISDISABLED:  script_pushint(st,((cd->trigger&0x80)!=0)); break;
+		case WIT_TITLE: script_pushstrcopy(st,cd->title); break;
+		case WIT_PASSWORD: script_pushstrcopy(st,cd->pass); break;
+		case WIT_EVENTNAME: script_pushstrcopy(st,cd->npc_event);break;
+		case WIT_ISFULL: script_pushint(st,(cd->users >= cd->limit)); break;
+		case WIT_OVERTRIGGER: script_pushint(st,(cd->users >= cd->trigger)); break;
+		default: script_pushint(st,-1); break;
```
Lemongrass3110:
```suggestion
```
Can safely be removed here, now that all those cases are covered above.

src/map/script.hpp
```diff
@@ -428,6 +428,18 @@ enum e_questinfo_markcolor : uint8 {
 	QMARK_MAX
 };
 
+enum waitingroom_info_type{
```
Lemongrass3110:
```suggestion
enum e_waitingroom_info_type{
```

doc/script_commands.txt
```diff
@@ -7851,19 +7851,23 @@ any.
...
+				0 otherwise.
+ WIT_TITLE - The Title of the waiting room (string)
+ WIT_PASSWORD - Password of the waiting room, if any. Pointless, since there is no way to
+				set a password on a waiting room right now.
+ WIT_EVENTNAME - Event name of the waiting room (string)
+ WIT_ISFULL - Whether or not the waiting room is full.
+ WIT_OVERTRIGGER - Whether the amount of users in the waiting room is higher than the trigger
+			       number.
+
+This function returns -1 if the type is unknown.
```
Lemongrass3110:
```suggestion
```
It does not anymore!
Thank you for that!!!

src/map/script.cpp
```diff
@@ -13244,26 +13244,50 @@ BUILDIN_FUNC(getwaitingroomstate)
 	if( script_hasdata(st,3) )
```
Lemongrass3110, line 13244: The in-source function documentation needs an update.
Maybe most of it should be moved into the enum?

Issue #8729, opened 2024-10-09
The outfit change system has not been implemented through the Styling Shop NPC


hiphop9:
### Is your request related to a missing official feature?

Yes

### Describe the solution you'd like

Since December 20, 2023, due to the body change patch, an issue has occurred where all classes appear as Novice. Additionally, when changing the body through the Styling NPC, an error occurs, causing the server to shut down!
https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=7830&curpage=14
![image](https://github.com/user-attachments/assets/3385814d-a96a-4c6c-835f-c98e52d00f81)
![image](https://github.com/user-attachments/assets/9f36741d-5fae-4848-b638-8e76acb8dc42)
![image](https://github.com/user-attachments/assets/00e86766-1bf6-4939-8f11-81799f4f51a4)

Beginerxxxxx:
I did not encounter this problem.

https://github.com/user-attachments/assets/33aa2996-b1d8-4a30-9a7e-9bc534965dbf

Haydrich:
> I did not encounter this problem.
> 
>  Facebook.mp4

Read the description. He is speaking about the new UI in December 2023.

Beginerxxxxx:
In my opinion, rA has been updated continuously. If your system is not updated, it may not work properly.

hiphop9:
I am already using the latest rAthena and using the default settings!!
![image](https://github.com/user-attachments/assets/cc59c5d6-2a60-49d3-8a56-80963d3a1384)

hiphop9:
> 제 생각에는 rA는 지속적으로 업데이트되었습니다. 시스템이 업데이트되지 않았다면 제대로 작동하지 않을 수 있습니다. 테스트를 시도해 보세요. 1- 마스터에서 서버 사용 2- EXE에서 사용자 지정 스타일리스트 사용 안 함 3- client.conf에서 기본 hair_style / cloth_color / body_style 사용

I am already using the latest rAthena and using the default settings!!
![image](https://github.com/user-attachments/assets/82243e45-3830-4e3e-afe8-d01b74813849)

Beginerxxxxx:
Please post your error in map server.

hiphop9:
> 맵 서버에 오류를 게시해 주세요.

![image](https://github.com/user-attachments/assets/e90ae6b2-2488-4251-8dd3-10375768f735)

Beginerxxxxx:
[...]

PR #9644, opened 2025-11-14
Add bodystyle support for all jobs in PACKETVER_MAIN_NUM >= 20231220 issue #8729

mikelancev2:
**Addressed Issue(s):** Feature enhancement for packet version 20231220+

**Server Mode:** Both (Pre-Renewal and Renewal)

**Description of Pull Request:**

This patch implements full bodystyle support for all job classes in `PACKETVER_MAIN_NUM >= 20231220`, removing the previous restriction that limited bodystyle functionality to all class jobs only.

**Motivation:**
Starting from packet version 20231220, the official Ragnarok Online client allows all job classes to use the bodystyle system. This patch aligns rAthena with the official server behavior while maintaining backward compatibility.

**Changes:**

- **`char.cpp`**: Set `body` field to `start_job` for new characters in packet version >= 20231220 (maintains `body = 0` for older versions)
- **`atcommand.cpp`**: Remove job restriction check for `@bodystyle` command in new packet version, allowing all jobs to change bodystyle via GM command
- **`clif.cpp`**: Adjust `LOOK_BODY2` handling to not override with class ID when using `OPTION_COSTUME` in new packet version
- **`pc.cpp`**: 
  - Update `pc_makesavestatus()` to set `body = class_` instead of 0 when not saving body style
  - Update `pc_jobchange()` to properly handle body field during job changes
  - Use `sd->status.body` directly in `clif_changelook()` calls
- **`status.cpp`**: Conditional `LOOK_BODY2` update in costume status changes (old versions update explicitly, new versions handle automatically)

**Backward Compatibility:**
All changes are wrapped in `#if PACKETVER_MAIN_NUM >= 20231220` conditionals to maintain full backward compatibility with:
- Pre-20231220 packet versions
- Servers using older clients  
- Existing functionality for  all class jobs in older versions


Lemongrass3110:
Could you please remove all those discord bot changes?
Then it would be easier to review.

mikelancev2:
> Could you please remove all those discord bot changes? Then it would be easier to review.

Sorry, it was an accident, that Discord change shouldn't even be public.

amm0nition:
How is this supposed to work?
When I tried to do `@bodystyle <JOB ID>`, it fails and returns "An invalid number was specified." message.

Issue #8017, opened 2023-11-21
Extend whodrops command to include map drops


laziem:
### Is your request related to a missing official feature?

No

### Describe the solution you'd like

Since there's no way for player to check map drops in games, request to include map drops on `@whodrops` command.

Suggested format:

```
Item: <Apple> (ID: 512)
Common mobs with highest drop chance (only max %d are listed):
- Muka Ringleader (2741): 30%
- Muka (3400): 20%
Map drops:
- Muka (1055): 10% - (prt_fild08)
- Muka (1055): 5% - (moc_fild03)
- Muka (1055): 1% - (moc_fild01, moc_fild02)
- Celine Kimi (2996): 1% - (1@xm_d)
```

For map drops on instance map, map name can be the instance name from instance_db.yml
`- Celine Kimi (2996): 1% - (Horror Toy Factory)`

IReSpI:
Up

Manuuuu21:
![Image](https://github.com/user-attachments/assets/7fd97d9a-d850-4603-8802-791a2453c715)

It wont displaying the map where it drops the item (969) even I recompiled my emulator base your updated extended whordrops commit.

IchigoHikariRO:
@YlenXWalker 

Great contribution!!

But I found a small problem.
If an item is repeated in many maps, using the command kills the map_server.

Is it possible to add a limiter to prevent this?

And if it's possible to add the option to add any map, that would be fantastic.
They added it here: https://rathena.org/board/files/file/4518-anymapdrops/

Likewise, thank you so much for this wonderful contribution!

PR #9649, opened 2025-11-15
Extend whodrops command #8017

YlenXWalker:
* **Extend whodrops command to include map drops**

* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/8017

* **Server Mode**: Renewal/Pre-renewal

* **Description of Pull Request**: Add map-specific drops display to @whodrops command

image

image


Atemo (commented):
Should there be a max limit of map and/or mobs displayed?

YlenXWalker:
> Should there be a max limit of map and/or mobs displayed?

I mean yeah we can set a limit. We don't want it flooding with 30+ of entries every time the command is called.

Lemongrass3110:
I think it should be done in the same way like with normal drops:

1. Calculated once only during loading (see mob_drop_ratio_adjust and mob_reload_itemmob_data)
2. Only MAX_SEARCH entries
3. Output should be sorted

Additionally this error message should not be printed, when the item is dropped from map drops:
`strcpy(atcmd_output, msg_txt(sd,1286)); //  - Item is not dropped by mobs.`

Could you include those changes?

YlenXWalker:
> I think it should be done in the same way like with normal drops:
> 
> 1. Calculated once only during loading (see mob_drop_ratio_adjust and mob_reload_itemmob_data)
> 2. Only MAX_SEARCH entries
> 3. Output should be sorted
> 
> Additionally this error message should not be printed, when the item is dropped from map drops: `strcpy(atcmd_output, msg_txt(sd,1286)); // - Item is not dropped by mobs.`
> 
> Could you include those changes?

I have updated my pull request.

Lemongrass3110:
@YlenXWalker it is not necessary to update your branch / merge master into your branch, if there are no conflicts.

Issue #9637, opened 2025-11-12
New bodystyle support crashes older clients


hayu-hayu:
### rAthena Hash

048c210d5521ff7fd0e0e3ac05814244ebd8d2b2

### Client Date

2022-04-06

### Server Mode

Pre-Renewal

### Result

Since the new bodystyle support (#9467) was added, older clients will crash for other players for certain jobs.
This happens because there are checks in the char server to override the "body" field with 0 or 1, so if you only test it on a single client it might look fine. But there are some checks missing in the map server. The client will try to display a costume sprite that does not exist and crash.

Thanks to _llchrisll_ and _MyRO_ from Discord who helped me tracking this down. _MyRO_ also posted some more information in #source-support:

> [17:26] MyRO: In pc_jobchange , the newly added
> `clif_changelook( sd, LOOK_BODY2, sd->vd.look[LOOK_BODY2] );`
> causes players to crash others around them, while they can walk around just fine (on my server, at least :< ).
>
> [17:27] MyRO: Putting it back in comment (as it was before the 2024 bodystyle rework) fixes it.
>
> [17:27] MyRO: Additionally, sending jobid (4336, ex) as BODY2 for 2022 clients shows jRO for the player, but not for others in area. They see normal 3rd.
> Sending 1 as value as before still works fine.
> 
> https://github.com/rathena/rathena/commit/048c210d5521ff7fd0e0e3ac05814244ebd8d2b2
> 
> The addition to clif_sprite_change() doesn't cover all cases (walking, idle, .. clif calls)
> 
> I've added a bandaid:
> `p.body = (vd->look[LOOK_BASE] >= JOB_RUNE_KNIGHT && vd->look[LOOK_BASE] <= JOB_MECHANIC_T2 && vd->look[LOOK_BODY2] > JOB_SECOND_JOB_START && vd->look[LOOK_BODY2] < JOB_SECOND_JOB_END) ? 1 : 0;`
> 
> jobchange also first changes LOOK_BASE and only then LOOK_BASE2, sending an updated LOOK_BASE with the client still having a LOOK_BASE2 will cause the client to crash.

### Relevant Log Output

n/a

### Expected Result

The client should no crash.

### How to Reproduce

- Start a fresh server after 048c210d5521ff7fd0e0e3ac05814244ebd8d2b2
- Use a client without bodystyle support (like 2022-04-06). 
- Reset your database. That is important as characters created with older server versions will not crash the client unless you run the migration script.
[...]

PR #9657, opened 2025-11-24
Fixed bodystyle on older clients

Lemongrass3110:
* **Addressed Issue(s)**: #9637 

* **Server Mode**: Both

Restores proper handling of bodystyle in old clients (before 2023-12-20) The correct logic was already included into the char-server, but not merged into the map-server

Thanks to @hayu-hayu and @llchrisll

### DIFF
--- a/src/map/clif.cpp
+++ b/src/map/clif.cpp
@@ -1178,7 +1178,7 @@ static void clif_set_unit_idle( block_list* bl, bool walking, send_target target
 		p.isBoss = BOSSTYPE_NONE;
 	}
 #endif
-#if PACKETVER >= 20231220
+#if PACKETVER >= 20231220 && !defined(PACKETVER_ZERO)
 	p.body = vd->look[LOOK_BODY2];
 #elif PACKETVER >= 20150513
 	if( vd->look[LOOK_BODY2] > JOB_SECOND_JOB_START && vd->look[LOOK_BODY2] < JOB_SECOND_JOB_END ){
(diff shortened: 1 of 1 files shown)


llchrisll:
Why am I getting mentioned? xD
I think I only pointed to the possible reason, which I also only knew about from issues from other ppl.

So I didn't really contribute to any direct solution imo.

Lemongrass3110:
@llchrisll you were mentioned in the issue

eyzi:
thought i'd mention it here. i have a server running on top of commit 1987cd7 and im affected by the issue of the client trying to display a missing costume for the job and crashing. i [applied this commit](https://github.com/aeternox/server/commits/master/) but the issue persists for me.

gmragnarok:
How will this continue?

src/map/clif.cpp
```diff
@@ -1178,8 +1178,14 @@ static void clif_set_unit_idle( block_list* bl, bool walking, send_target target
 		p.isBoss = BOSSTYPE_NONE;
 	}
 #endif
-#if PACKETVER >= 20150513
+#if PACKETVER >= 20231220
```
AoShinRO:
```suggestion
#if PACKETVER >= 20231220 && !defined(PACKETVER_ZERO)
```

src/map/clif.cpp
```diff
@@ -1421,8 +1427,14 @@ static void clif_set_unit_walking( block_list& bl, map_session_data* tsd, struct
...
```
AoShinRO:
```suggestion
#if PACKETVER >= 20231220 && !defined(PACKETVER_ZERO)
```

Issue #9668, opened 2025-11-30
Restoring old reload bahavior that uses strncmp


gidzdlcrz:
### Is your request related to a missing official feature?

No

### Describe the solution you'd like

Before `@reload` command works as a separate, unified `@reload` command that uses strncmp to match partial subcommand names.

### Describe alternatives you've considered

Restore the deleted code but retain the split type codes. So they co exist.

### Additional context

N/A

PR #9669, opened 2025-11-30
Restored atcommand reload

gidzdlcrz:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9668

* **Server Mode**: 
Both

  * Works as a separate, unified `@reload` command that uses **strncmp** to match partial subcommand names.

  * The key insight is that this is a parent command pattern - `@reload` is its own command that dispatches to different reload operations based on its argument, rather than trying to modify the existing individual commands.
  
  * Also, restores old doc information and added logconf entry.

### DIFF
--- a/conf/atcommands.yml
+++ b/conf/atcommands.yml
@@ -1132,7 +1132,7 @@ Body:
   - Command: reload
     Help: |
       Params: <type>
-      Reload a database or configuration file.
+      Reload a database or a configuration file.
       itemdb               mobdb               skilldb
       atcommand         battleconf          statusdb
       pcdb                  motd                 script


src/map/atcommand.cpp
```diff
@@ -4339,6 +4339,149 @@
...
+		achievement_db_reload();
+		clif_displaymessage(fd, msg_txt(sd,771)); // Achievement database has been reloaded.
+	} else if (strstr(command, "attendancedb") || strncmp(message, "attendancedb", 4) == 0) {
+		attendance_db.reload();
+		clif_displaymessage(fd, msg_txt(sd, 795)); // Attendance database has been reloaded.
+	}else if( strstr( command, "barterdb" ) || strncmp( message, "barterdb", 4 ) == 0 ){
+		barter_db.reload();
+		clif_displaymessage(fd, msg_txt(sd, 830)); // Barter database has been reloaded.
+	} else if (strstr(command, "logconf") || strncmp(message, "logconf", 3) == 0) {
+		log_config_read(LOG_CONF_NAME);
```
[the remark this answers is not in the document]
gidzdlcrz: made changes to addressed this check failure.

gidzdlcrz:
I can't fix the false positive github checks. Tho I compiled on both windows and linux there's no warning/error and it behaves the way I wanted it to be.

conf/atcommands.yml
```diff
@@ -1168,6 +1177,8 @@ Body:
     Help: |
       Reload message configuration.
   - Command: reloadnpcfile
+    Aliases:
+      - reloadnpc
```
Lemongrass3110, line 1181: This is unrelated to this PR

conf/atcommands.yml
```diff
@@ -1129,6 +1129,15 @@ Body:
   - Command: reject
     Help: |
       Automatically reject duel invitations.
+  - Command: reload
+    Help: |
+      Params: <type>
+      Reload a database or configuration file.
```
Lemongrass3110:
```suggestion
      Reload a database or a configuration file.
```

Lemongrass3110 (approved):

conf/atcommands.yml
gidzdlcrz: I thought I should re add this because this was removed as well from https://github.com/rathena/rathena/pull/8913

Lemongrass3110: You are correct, this was accidentally removed in 754a078 and partially restored in 589ddd6

PR #9673, opened 2025-11-30
Corrected changesex new bodystyle behavior

AoShinRO:
* **Addressed Issue(s)**: 

* **Server Mode**: 

* **Description of Pull Request**:  fixed new client bodystyle missing sprite error when changing sex with BARDDANCER and KAGEROUOBORO

### DIFF
--- a/src/char/char_logif.cpp
+++ b/src/char/char_logif.cpp
@@ -459,7 +459,7 @@ void chlogif_parse_change_sex_sub(int32 sex, int32 acc, int32 char_id, int32 cla
 	if (SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `equip` = '0', `equip_switch` = '0' WHERE `char_id` = '%d'", schema_config.inventory_db, char_id))
 		Sql_ShowDebug(sql_handle);
 
-	if (SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `class` = '%d', `weapon` = '0', `shield` = '0', `head_top` = '0', `head_mid` = '0', `head_bottom` = '0', `robe` = '0', `sex` = '%c' WHERE `char_id` = '%d'", schema_config.char_db, class_, sex == SEX_MALE ? 'M' : 'F', char_id))
+	if (SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `class` = '%d', `body` = '%d', `weapon` = '0', `shield` = '0', `head_top` = '0', `head_mid` = '0', `head_bottom` = '0', `robe` = '0', `sex` = '%c' WHERE `char_id` = '%d'", schema_config.char_db, class_, class_, sex == SEX_MALE ? 'M' : 'F', char_id))
 		Sql_ShowDebug(sql_handle);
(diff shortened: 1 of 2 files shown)


src/char/char_logif.cpp
```diff
@@ -459,7 +459,7 @@ void chlogif_parse_change_sex_sub(int32 sex, int32 acc, int32 char_id, int32 cla
 	if (SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `equip` = '0', `equip_switch` = '0' WHERE `char_id` = '%d'", schema_config.inventory_db, char_id))
 		Sql_ShowDebug(sql_handle);
 
-	if (SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `class` = '%d', `weapon` = '0', `shield` = '0', `head_top` = '0', `head_mid` = '0', `head_bottom` = '0', `robe` = '0', `sex` = '%c' WHERE `char_id` = '%d'", schema_config.char_db, class_, sex == SEX_MALE ? 'M' : 'F', char_id))
+	if (SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `class` = '%d', `body` = '%d', `weapon` = '0', `shield` = '0', `head_top` = '0', `head_mid` = '0', `head_bottom` = '0', `robe` = '0', `sex` = '%c' WHERE `char_id` = '%d'", schema_config.char_db, class_, class_, sex == SEX_MALE ? 'M' : 'F', char_id))
```
rabbit-lunatic, line 462: 2 class_,  ?
```suggestion
	if (SQL_ERROR == Sql_Query(sql_handle, "UPDATE `%s` SET `class` = '%d', `body` = '%d', `weapon` = '0', `shield` = '0', `head_top` = '0', `head_mid` = '0', `head_bottom` = '0', `robe` = '0', `sex` = '%c' WHERE `char_id` = '%d'", schema_config.char_db, class_, sex == SEX_MALE ? 'M' : 'F', char_id))
```

AoShinRO: > 2 class_, ?

you missed body style update?
actually bodystyle is same of class values or class costume value, when changing sex looktypes are reseted and need to be same at this point because on SQL bodystyle persist with other sex class and make client give a sprite error and close on relogin

gidzdlcrz, line 462: Can we do if packever here to support old clients?

src/map/chrif.cpp
```diff
@@ -1001,6 +1001,7 @@ int32 chrif_changedsex(int32 fd) {
 			else// Changed from male version of job.
 				sd->status.class_ += 1;
 			//sd->class_ Does not need to be updated as both versions of the job are the same.
+			sd->status.body = sd->status.class_;
```
gidzdlcrz, line 1004: also here

src/char/char_logif.cpp
AoShinRO: @Lemongrass3110 has a pull request that corrects the bodystyle for older packet versions; there's no need to worry about it.

gidzdlcrz: Thanks!

PR #9702, opened 2025-12-25
Reworked Parts of The Job MAPID System / Added Job Group Checks

Rytech2:

* **Server Mode**:  Renewal

This is another part of my series of updates that aims to fix many long standing issues with the handling of certain jobs, checks, and making things more clean, efficient, and future proof. Sorry if its a lot to read. I love explaining all the technical stuff on things like this.

P.S. This update needed to be done first before I could do the item/script fixes.

Reworked parts of the job MAPID system.
Back when 3rd jobs became a thing, Athena code was still using regular C and the enum for MAPID
was on signed short which limited the hex values up to 0x8000. So having JOBL_THIRD share the
same bit with JOBL_UPPER and JOBL_BABY was the only way to make things work and for the time it
did. But with the adding of 4th jobs and many other expanded jobs it has created some what of a
mess. Especially for MAPID_THIRDMASK and MAPID_FOURTHMASK use for checks.

This update focuses on reworking all of this now that the enum is set as unsigned long which
gives plenty of room to do things correctly. This will allow for things to be more organized,
cleaner checks, cleaner code, adding new features without issues, and future proof the system.

JOBL_THIRD/JOBL_FOURTH are now set to their own digits.
Because these are basicly job levels or tiers, they should be in their own digits and not mixed
in with JOBL_UPPER/JOB_BABY which are used to set alternate versions of a job. Its the same
with 1st and 2nd jobs being on their own digits.

JOBL_UPPER/JOBL_BABY are now the last digit of the hex value.
These are special values that set alternate versions of a job. Because of this they should be
kept as the last digit in the hex value. This is to prevent including them in certain checks
when checking for the base of any jobs.

MAPID_THIRDMASK/MAPID_FOURTHMASK is redone to be cleaner and only allow for checking the base
versions of jobs just like how MAPID_BASEMASK and MAPID_UPPERMASK does.

MAPID_BASEMASK/MAPID_UPPERMASK is renamed to MAPID_FIRSTMASK/MAPID_SECONDMASK.
For some people (including me in the late 2000's), JOBL_UPPER and MAPID_UPPERMASK sounded like
their related somehow and yet they are not. JOBL_UPPER is a value that sets a job as a rebirth
job, but MAPID_UPPERMASK is a filter used to check for the base of a 2nd job and has nothing to
do with rebirths. I had enough of this confusion and renamed them to be as easy to recognize as
MAPID_THIRDMASK/MAPID_FOURTHMASK is.

Reordered the 1st jobs.
I always wanted to do this but in the eAthena days the 1st/2nd jobs MAPID hex's affected what
value you use in the item_db to set what jobs can equip it. So I couldn't. But in rAthena I can
since thats no longer the case. All 1st jobs are now listed first with the costumes listed after.
The order of the job are also done in a way where jobs with branches ending in 4th jobs, then
3rd, then 2nd are listed in order. This made it possible to remove a lot of value defines in
the different groups of jobs below, leading to less code and a cleaner look.

4th jobs are now handled as base jobs.
MAPID_FOURTHMASK existed for a few years but wasn't really used except for a check in skills.cpp.
But I ran into some issues when I tried it with some experimental code for scripts in items.
Expanded jobs that are a 4th job on the branch work with the check but primary 4th jobs didn't.
Turns out the system needs a base version of a job for it to work with the check. So I made
this change. But to keep the bonus HP/SP/Stat Points functions I used the pc_is_primary_fourth
job group check. Its also best to do this since official ended the base/rebirth version stuff
with primary 3rd jobs being the last to use it.

Added Job Group Checks
Because every primary job branch has the same number of jobs but every expanded job branch has
a different number of jobs, using JOBL_THIRD and JOBL_FOURTH job checks for certain things
won't cut it. To deal with this issue, a bunch of different checks were created to check for
certain groups of jobs and their successors. This will not only correct some issues, but also
future proof things should we get more successor jobs in the future.

Here are the checks....
pc_is_primary_third - Checks if the player is a primary 3rd job or higher.
pc_is_primary_fourth - Checks if the player is a primary 4th job or higher.
pc_is_upper_expanded_first - Checks if the player is a 1st upper expanded job or higher.
pc_is_upper_expanded_second - Checks if the player is a 2nd upper expanded job or higher.
pc_is_renewal_job - Checks if the player is a renewal era job or higher. (Primary 3rd / 1st Upper Expanded)
pc_is_trait_job - Checks if the player is a trait era job or higher. (Primary 4th / 2nd Upper Expanded)

Examples of their use can be found with pc_is_primary_fourth and pc_is_trait_job.

### DIFF
--- a/db/re/item_db_usable.yml
+++ b/db/re/item_db_usable.yml
@@ -57152,7 +57152,7 @@ Body:
       NoMail: true
       NoAuction: true
     Script: |
-      if ((eaclass()&EAJ_ASSASSIN_CROSS)==EAJ_ASSASSIN_CROSS) {
+      if ((eaclass()&EAJ_ASSASSIN_CROSS)==EAJ_ASSASSIN_CROSS || (eaclass()&EAJ_THIRDMASK)==EAJ_GUILLOTINE_CROSS) {
          sc_start SC_DPOISON,60000,0;
          sc_start SC_ASPDPOTION3,60000,9;
       }


db/re/item_db_usable.yml
```diff
@@ -1824,7 +1824,7 @@ Body:
     Flags:
       BuyingStore: true
     Script: |
-      if ((eaclass()&EAJ_ASSASSIN_CROSS)==EAJ_ASSASSIN_CROSS) {
+      if ((eaclass()&EAJ_ASSASSIN_CROSS)==EAJ_ASSASSIN_CROSS || (eaclass()&EAJ_THIRDMASK)==EAJ_GUILLOTINE_CROSS) {
```
Atemo, line 1827: Poison Bottle Id 22549 need to be updated too

doc/ea_job_system.txt
```diff
@@ -86,17 +87,24 @@ The next step is to either become a 2-1 or a 2-2 job:
...
+	EAJ_SWORDMAN|EAJL_2_2|EAJL_THIRD|EAJL_FOURTH	-> EAJ_IMPERIAL_GUARD
+
+Getting out the rebirth versions of a swordman:
+
+	EAJ_SWORDMAN|EAJL_UPPER 			-> EAJ_SWORDMAN_HIGH
+	EAJ_SWORDMAN|EAJL_2_1|EAJL_UPPER		-> EAJ_LORD_KNIGHT
+	EAJ_SWORDMAN|EAJL_2_2|EAJL_UPPER		-> EAJ_PALADIN
+	EAJ_SWORDMAN|EAJL_2_1|EAJL_THIRD|EAJL_UPPER	-> EAJ_RUNE_KNIGHT_T
+	EAJ_SWORDMAN|EAJL_2_2|EAJL_THIRD|EAJL_UPPER	-> EAJ_ROYAL_GUARD_T
+
```
Atemo, line 107: Similar to the example with the swordsman can you add 2 examples for extended classes, one for Spirit Handler and another for Soul ascetic?

Rytech2: The Swordman example explains everything. Is there a reason you want examples of expanded jobs? It would just show Spirit Handler is 2nd job and Soul Ascetic is 4th job. Nothing special.

Jason59762:
After the update, when the character is an Arch_Mage, some skills cannot be learned normally. For example: HW_MAGICCRASHER, HW_GANBANTEIN, HW_GRAVITATION, HW_MAGICCRASHER, HW_NAPALMVULCAN. It is necessary to allocate all the third-job skill points before these skills can be learned, which is illogical. No similar issues have been found with other classes so far

gasstation51:
After the victory, you cannot learn some of the skills of the priest class.

gidzdlcrz:
> After the update, when the character is an Arch_Mage, some skills cannot be learned normally. For example: HW_MAGICCRASHER, HW_GANBANTEIN, HW_GRAVITATION, HW_MAGICCRASHER, HW_NAPALMVULCAN. It is necessary to allocate all the third-job skill points before these skills can be learned, which is illogical. No similar issues have been found with other classes so far

> After the victory, you cannot learn some of the skills of the priest class.

Can you guys submit an issue with this. Thanks!

Rytech2:
> After the update, when the character is an Arch_Mage, some skills cannot be learned normally. For example: HW_MAGICCRASHER, HW_GANBANTEIN, HW_GRAVITATION, HW_MAGICCRASHER, HW_NAPALMVULCAN. It is necessary to allocate all the third-job skill points before these skills can be learned, which is illogical. No similar issues have been found with other classes so far

I can confirm the issue. It looks like a well hidden bug got past me. From the testing I can learn the trans skills normally during my tests while doing normal job changing, going through Swordsman all the way to Dragon Knight but keeping in mind your normally required to use all your skill points before job changing. Everything appears to work fine. But its after I reset my skills and reach the point of investing points into the 2nd job skills the issue happens. It looks like just having skills inherited in the skill tree isn't enough. I think its a hidden bug in the normalize skill tree code and am currently looking for a way to fix it.

Jason59762:
> > After the update, when the character is an Arch_Mage, some skills cannot be learned normally. For example: HW_MAGICCRASHER, HW_GANBANTEIN, HW_GRAVITATION, HW_MAGICCRASHER, HW_NAPALMVULCAN. It is necessary to allocate all the third-job skill points before these skills can be learned, which is illogical. No similar issues have been found with other classes so far
> 
> I can confirm the issue. It looks like a well hidden bug got past me. From the testing I can learn the trans skills normally during my tests while doing normal job changing, going through Swordsman all the way to Dragon Knight but keeping in mind your normally required to use all your skill points before job changing. Everything appears to work fine. But its after I reset my skills and reach the point of investing points into the 2nd job skills the issue happens. It looks like just having skills inherited in the skill tree isn't enough. I think its a hidden bug in the normalize skill tree code and am currently looking for a way to fix it.

You're right, it's a problem that only appears after resetting skill points. Thank you for your selfless contribution. ^_^

Issue #9689, opened 2025-12-15
character deletion bug using old client [Error]: Received unknown packet 0x0068


bgamez23:
### rAthena Hash

4979c180dfea393818087925df4af41a3a861e2a

### Client Date

2010-07-30

### Server Mode

Pre-Renewal

### Result

character deletion failed and you will get [Error]: Received unknown packet 0x0068

### Relevant Log Output

```Shell

```

### Expected Result

character deletion failed and you will get [Error]: Received unknown packet 0x0068

### How to Reproduce

delete existing character. im using old birthday format for character deletion. 

https://vimeo.com/1146489342?share=copy&fl=sv&fe=ci

### Official Information

character deletion failed and you will get [Error]: Received unknown packet 0x0068

PR #9705, opened 2025-12-29
Fixed char deletion with 0x68

Lemongrass3110:
* **Addressed Issue(s)**: #9689

* **Server Mode**: Both

Both packets have to be supported in parallel, since some client diffs force-send old packets.

Thanks to @bgamez23

### DIFF
--- a/src/char/char_clif.cpp
+++ b/src/char/char_clif.cpp
@@ -1353,14 +1353,8 @@ void chclif_delchar( int32 fd ){
 	socket_send( fd, p );
 }
 
-bool chclif_parse_delchar( int fd, struct char_session_data& sd ){
-	const PACKET_CH_DELETE_CHAR* p = reinterpret_cast<PACKET_CH_DELETE_CHAR*>( RFIFOP( fd, 0 ) );
-
-	char email[40];
-	uint32 cid = p->CID;
-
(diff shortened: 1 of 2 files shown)


bgamez23:
thanks if works now i can delete the char using email. but not using the birthday (YYMMDD)
or  does this option not work on all client?

> // Restrict character deletion by email address or birthdate.
> // This restricts players from changing the langtype and deleting characters.
> // Defaults based on client date.
> // 1: Email address
> // 2: Birthdate
> // 3: Email address or Birthdate
> // IMPORTANT!
> // - This config only works for clients that send 0x0068 or 0x01fb for delete request.
> // - Use langtype 1 for newer clients (2013+), to use 0x01fb.
> // - Clients that are not using 0x0068 or 0x01fb, only use birthdate (YYMMDD) as default.
> char_del_option: 3

Lemongrass3110:
@bgamez23 did you lose a "not" in your sentence? So is deleting via birthdate working or not? This is a bit unclear to me now.

I did not touch the deletion logic itself, it still hands over the config value (charserv_config.char_config.char_del_option) to the other function.

bgamez23:
> @bgamez23 did you lose a "not" in your sentence? So is deleting via birthdate working or not? This is a bit unclear to me now.
> 
> I did not touch the deletion logic itself, it still hands over the config value (charserv_config.char_config.char_del_option) to the other function.

oh yes sorry. character deletion using birthday is not working.

PR #9717, opened 2026-01-06
Item Classes Inherit / BaseThird And BaseFourth Checks

Rytech2:

* **Server Mode**: Both

Reworked The Classes Setting For Items
Back in the eAthena days before third jobs, this setting was used to set what versions of a job
can use the item and was set using a hex digit. 0x0 for normal (base), 0x1 for upper (trans),
and 0x2 for babys. It basicly relied on that special JOBL_ hex digit for UPPER/BABY to do checks.
But rAthena changed the system which no longer relies on this bit entirely, removing the limits on how
we can do checks with this setting. Things have also become more complex with primary jobs
getting 3rd and 4th jobs and expandeds being mixed with each branch having 4, 3, or 2 jobs total.

This update focuses on reworking the classes setting which fixes a bunch of issues with it while
also allowing for cleaner checks, cleaner code, adding new kinds of checks, and future proof the system.

ITEMJ_UPPER reworked.
Items set with this setting will check if the player is a transcedent job. If not, it then checks
if the player is a primary 3rd job. This 2nd check uses pc_is_primary_job to check for the base
version of primary 3rd jobs which makes it work for trans/baby versions too and is inherited by
primary 4th jobs. If 5th jobs come to exist, they will inherit this as well. No need to update.

ITEMJ_BABY corrected.
Why was this setup to fail for third job babies? It should only care about if a job is a baby job.
This change makes it work for all baby jobs so we don't need other things like ITEMJ_THIRD_BABY
and ITEMJ_ALL_BABY. Plus there's no items exclusive to the baby jobs as far as im aware.
I thought about removing this check entirely but decided to keep it in for baby checks.

ITEMJ_RENEWAL_JOB and ITEMJ_TRAIT_JOB added.
This adds special checks for certain groups of jobs according to the era they were released in.
ITEMJ_RENEWAL_JOB checks for Primary 3rd and 1st Upper Expanded jobs and is inherited by successor
jobs like Primary 4th and 2nd Upper Expanded Jobs. ITEMJ_TRAIT_JOB checks for Primary 4th and
2nd Upper Expanded jobs and is inherited by any future jobs that may come to exist.

ITEMJ_THIRD, ITEMJ_THIRD_UPPER, ITEMJ_THIRD_BABY, ITEMJ_FOURTH
These classes checks are marked for deprecation due to their issues. For example, A Rune Knight
is a 3rd jobs, but Rebellion is a 2nd job even tho its limits are on pair with the Rune Knight.
But Thats because their in the same area of renewal era jobs. So a 3rd job check won't work here.
The same can be said for fourth job checks like Dragon Knight to a 3rd job Night Watch being trait
era jobs that are on pair with each other. Finally, ITEMJ_RENEWAL_JOB works on all versions of
primary 3rd jobs which makes ITEMJ_THIRD_UPPER and ITEMJ_THIRD_BABY pointless.

ITEMJ_NORMAL
Because of the changes with the system, this check is pointless and has been pointless since
rAthena's changes with the system. I would remove this but its still used in pre-re item databases.
Will come back to this one soon.

ITEMJ_ALL_UPPER, ITEMJ_ALL_BABY, ITEMJ_ALL_THIRD
Also pointless due to ITEMJ_RENEWAL_JOB check which can do it all in one check.
ITEMJ_ALL_THIRD is in use but the other are not. Will work to get these removed soon.

BaseFirst and BaseSecond constants added.
These were made to switch over from the BaseClass/BaseJob constants which by name are very
confusing. But for now they will exist as a 2nd set of constants next to the old ones while we
work to update all scripts over to the new ones to avoid breaking anything.

SP_BASETHIRD/SP_BASEFOURTH added. / BaseThird/BaseFourth constants added.
These were added to support proper and efficant checks for 3rd and 4th jobs in scripts.
They work just like BaseFirst/BaseSecond but for 3rd and 4th jobs. Nothing more to say.

item_db and script_commands documents were updated to give more info on all of the above changes.
Please take the time to check them to get fimilar with how to use classes check on items and
proper job checks in scripts.

================================

Here's some examples of how to setup job checks and restrictions with the system.

Item 1100
The old method uses "eaclass()&EAJL_THIRD && BaseJob == Job_Alchemist" which checks if player is a
3rd job Alchemist. While this works, its not a clean way to check. Using "BaseThird == Job_Genetic"
instead does a clean check to see if your 3rd job base is a Genetic. This base check passes for
any job that has Genetic as its 3rd job base like trans/baby versions and even Biolo since
Genetic is its 3rd job predecessor.

Item 1150
This item used "BaseJob == Job_Rune_Knight" in its check which fails because BaseJob
(Now BaseSecond) checks for 2nd jobs. Changing to BaseThird allows proper checking like how the
example for item 1100 shows.

Item 1185
This item requires the player to be a trans job like Lord Knight or Crusader at minimum to equip
this item. But the successor jobs Rune Knight and Royal Guard can equip it no matter if your the
base/trans/baby version and gets inherited to Dragon Knight and Imperial Guard. The Upper setting
for classes now supports checking for primary 3rd jobs which removes the need for the All_Third
and Fourth classes settings since the permission is inherited by 3rd jobs and their successors.
No longer do you need to have 3 different classes settings anymore and if we get more successor
jobs in the future we won't need to come back to add other settings due to the inheriting.

Item 1376
Bad checks were used here in the script. A bunch of different jobs can equip this but being a
certain job like Whitesmith or Creator give you a bonus. This should be given to their successors
but a incorrect check checks for Mechanic, Genetic, and their trans versions seprately. This
makes the baby versions and their successor the Meister and Biolo not inherit the bonus. The use
of BaseThird fixes all of this to be handled like with item example 1100.

Item 1549
Requires you to be level 99 but also be a Mechanic. Because Blacksmith, Whitesmith, and Mechanic
can be level 99, a classes check is needed to make it require a minimum job of Mechanic. But the
Fourth classes setting was also needed for it to work with Meisters. Changing it to only need the
Renewal_Job keeps it to require being a Mechanic by doing a base check which also allows trans/baby
versions and successors like the Meister to inherit access to the item. Future successor jobs will
also inherit access thanks to this check, future proofing this so we don't have to revisit this
item.

Item 21049
Same example as item 1549. Less classes settings needed and future proofs the item.

Item 21058
Requires you to be a level 150 Rune Knight or higher. If you think about what the level cap is
for Knight/Lord Knight its 99. So you have to be a Rune Knight to go past 99 to reach level 150.
Because of this, using classes settings is pointless since the required level setting is enough
to stop predecessor jobs from equipping it. Now if the equip level was 99 like in item example 1549
then yes we would need a classes check like Renewal_Job.

Item 21063
Similar example to item example 21058. But this one requires a level 220 Dragon Knight. Rune Knight
can't go past level 200 and you have to be a Dragon Knight to go past 200. So classes checks are
pointless here since the equip level is enough.

Item 24792 and 24793
These items are what got me started on all this rework with classes check, BaseThird/BaseFourth,
reworking the job MAPID system, and the past few updates I did on rAthena. These 2 items have a
very special requirement of a base level of 200 and the player be as the description says,
"4th jobs and 2nd upper expanded jobs". This wasn't possible before since not all 2nd upper
expanded jobs are 4th jobs and we only had third/fourth job classes checks. But now we can
check for this using the classes setting Trait_Job. This setting checks for the base of these jobs
and future proofs things to allow future successor jobs to inherit access to the item.
Your welcome Atemo. =)

Item 600013
A weapon that requires level 200 Dragon Knight. But Rune Knight level cap is 200. So to restrict
access to the Dragon Knight and its future successors we use the classes setting Trait_Job.

Final Thoughts
With these example showing how to use the new job checks in scripts and the new classes settings,
we can now future proof things so we don't have to update classes settings or checks in scripts
anymore when more successor jobs come out in the future. If something like 5th jobs and 3rd upper
expanded jobcome out in the future we won't have to worry about updating any of this.

### DIFF
--- a/src/map/script_constants.hpp
+++ b/src/map/script_constants.hpp
@@ -273,8 +273,10 @@
 	export_constant2("EAJL_UPPER",JOBL_UPPER);
 	export_constant2("EAJL_BABY",JOBL_BABY);
 
-	export_constant2("EAJ_BASEMASK",MAPID_FIRSTMASK);// Update name in future update for item/scripts. (Rytech)
-	export_constant2("EAJ_UPPERMASK",MAPID_SECONDMASK);// Update name in future update for item/scripts. (Rytech)
+	export_constant2("EAJ_BASEMASK",MAPID_FIRSTMASK);// Use EAJ_FIRSTMASK instead.
+	export_constant2("EAJ_UPPERMASK",MAPID_SECONDMASK);// Use EAJ_SECONDMASK instead.
+	export_constant2("EAJ_FIRSTMASK",MAPID_FIRSTMASK);
+	export_constant2("EAJ_SECONDMASK",MAPID_SECONDMASK);
 	export_constant2("EAJ_THIRDMASK",MAPID_THIRDMASK);
 	export_constant2("EAJ_FOURTHMASK",MAPID_FOURTHMASK);
 
@@ -593,8 +595,8 @@
 	export_parameter("MaxSp",SP_MAXSP);
 	export_parameter("Ap",SP_AP);
 	export_parameter("MaxAp",SP_MAXAP);
-	export_parameter("BaseJob",SP_BASESECOND);// Note: Deprecated. Use BaseFirst instead.
-	export_parameter("BaseClass",SP_BASEFIRST);// Note: Deprecated. Use BaseSecond instead.
+	export_parameter("BaseJob",SP_BASESECOND);// Use BaseSecond instead.
+	export_parameter("BaseClass",SP_BASEFIRST);// Use BaseFirst instead.
 	export_parameter("BaseFirst",SP_BASEFIRST);
 	export_parameter("BaseSecond",SP_BASESECOND);
 	export_parameter("BaseThird",SP_BASETHIRD);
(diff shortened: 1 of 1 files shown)


Lance0720:
npc\custom-->jobmaster can help? It's not working!

Rytech2:
> npc\custom-->jobmaster can help? It's not working!

I didn't touch anything that would break it but it might be related to a previous pull request. What jobs are you having issues changing into?

Lance0720:
> > npc\custom-->jobmaster can help? It's not working!
> 
> I didn't touch anything that would break it but it might be related to a previous pull request. What jobs are you having issues changing into?

Fourth Job can't change I try rune knight T , not PR I use master .

Rytech2:
I got job group check support added for scripts and added a test script for it. Also added EAJ_FIRSTMASK and EAJ_SECONDMASK which works just like the older EAJ_BASEMASK and EAJ_UPPERMASK but just a clearer name on what it does.

idk-whoami:
it works

PR #9728, opened 2026-01-08
Split up Priest skills

Lemongrass3110:

* **Server Mode**: Both

### DIFF
--- a/src/map/skills/acolyte/skill_factory_acolyte.cpp
+++ b/src/map/skills/acolyte/skill_factory_acolyte.cpp
@@ -24,6 +24,7 @@
 #include "kyrieeleison.hpp"
 #include "lexdivina.hpp"
 #include "magnificat.hpp"
+#include "magnusexorcismus.hpp"
 #include "ruwach.hpp"
 #include "sanctuary.hpp"
 #include "statusrecovery.hpp"
@@ -134,6 +135,8 @@ std::unique_ptr<const SkillImpl> SkillFactoryAcolyte::create(const e_skill skill
 			return std::make_unique<SkillLexDivina>();
 		case PR_MAGNIFICAT:
 			return std::make_unique<SkillMagnificat>();
+		case PR_MAGNUS:
+			return std::make_unique<SkillMagnusExorcismus>();
 		case PR_SANCTUARY:
 			return std::make_unique<SkillSanctuary>();
 		case PR_SLOWPOISON:
--- a/src/map/battle.cpp
+++ b/src/map/battle.cpp
@@ -8489,10 +8489,6 @@ struct Damage battle_calc_magic_attack(block_list *src,block_list *target,uint16
 						else
 							skillratio += 20 * skill_lv - 20; //Monsters use old formula
 						break;
-					case PR_MAGNUS:
-						if (battle_check_undead(tstatus->race, tstatus->def_ele) || tstatus->race == RC_DEMON)
-							skillratio += 30;
-						break;
 					case BA_DISSONANCE:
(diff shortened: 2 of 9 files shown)


src/map/skills/acolyte/skill_factory_acolyte.cpp
```diff
@@ -107,18 +119,35 @@ std::unique_ptr<const SkillImpl> SkillFactoryAcolyte::create(const e_skill skill
...
+		case PR_BENEDICTIO:
+			return std::make_unique<SkillBenedictioSanctissimiSacramenti>();
+		case PR_GLORIA:
+			return std::make_unique<SkillGloria>();
 		case PR_IMPOSITIO:
-			return std::make_unique<StatusSkillImpl>(skill_id);
-#endif
+			return std::make_unique<SkillImpositioManus>();
+		case PR_KYRIE:
+			return std::make_unique<SkillKyrieEleison>();
```
Atemo, line 131: StatusSkillImpl ?

src/map/skills/acolyte/skill_factory_acolyte.cpp
```diff
@@ -107,18 +119,35 @@ std::unique_ptr<const SkillImpl> SkillFactoryAcolyte::create(const e_skill skill
...
+		case PR_ASPERSIO:
+			return std::make_unique<SkillAspersio>();
+		case PR_BENEDICTIO:
+			return std::make_unique<SkillBenedictioSanctissimiSacramenti>();
+		case PR_GLORIA:
+			return std::make_unique<SkillGloria>();
 		case PR_IMPOSITIO:
-			return std::make_unique<StatusSkillImpl>(skill_id);
-#endif
+			return std::make_unique<SkillImpositioManus>();
```
Atemo, line 129: How about
```suggestion
#ifndef RENEWAL
			return std::make_unique<StatusSkillImpl>(skill_id);
#else
			return std::make_unique<SkillImpositioManus>();
#endif
```

src/map/skills/acolyte/skill_factory_acolyte.cpp
```diff
@@ -107,18 +119,35 @@ std::unique_ptr<const SkillImpl> SkillFactoryAcolyte::create(const e_skill skill
...
+			return std::make_unique<SkillSanctuary>();
 		case PR_SLOWPOISON:
 			return std::make_unique<StatusSkillImpl>(skill_id);
-#ifndef RENEWAL
+		case PR_STRECOVERY:
+			return std::make_unique<SkillStatusRecovery>();
 		case PR_SUFFRAGIUM:
-			return std::make_unique<StatusSkillImpl>(skill_id);
-#endif
+			return std::make_unique<SkillSuffragium>();
```
Atemo, line 147: How about
```suggestion
#ifndef RENEWAL
			return std::make_unique<StatusSkillImpl>(skill_id);
#else
			return std::make_unique<SkillSuffragium>();
#endif
```

src/map/skills/acolyte/skill_factory_acolyte.cpp
Lemongrass3110: StatusSkillImpl uses clif_skill_nodamage(src, target, ...)
Kyrie and some other skills use clif_skill_nodamage(target, target, ...)

src/map/skills/acolyte/skill_factory_acolyte.cpp
Lemongrass3110: If both would be a default impl then it is okay, but if things get more complex i think it is cleaner to have both in the respective file.

Issue #9729, opened 2026-01-08
Nidhoggr Instance MVP can fail to spawn


Singe-Horizontal:
### rAthena Hash

bae0e2e6c

### Client Date

N/A

### Server Mode

Pre-Renewal, Renewal

### Result

When reaching the last arena of Nidhoggr's instance, the mvp can fail to spawn if a party member besides the leader enters first the area that is used as trigger while being followed by the leader.

### Relevant Log Output

```Shell

```

### Expected Result

The Mvp should always spawn once the party leader enters the triggering area.

### How to Reproduce

1. Put both party leader and another member at  2@nyd,198,259
2. As non party leader, move one pannel north, followed by the party leader
3. The MVP spawning script doesn't activate.

### Official Information

None

PR #9730, opened 2026-01-08
Fixed Nidhoggr Instance MVP's spawn

Singe-Horizontal:
* **Addressed Issue(s)**: 
#9729

* **Server Mode**: 
Both

Currently, if the party leader enters the boss triggering area while another party member is already inside, his script won't activate due to it having an OnTouch_ event.
By replacing OnTouch_ by OnTouch, the event will always be fired.

### DIFF
--- a/npc/instances/NydhoggsNest.txt
+++ b/npc/instances/NydhoggsNest.txt
@@ -1849,7 +1849,7 @@ OnTouch:
 }
 
 2@nyd,199,268,0	script	nyd_2f_boss_enter	-1,8,8,{
-OnTouch:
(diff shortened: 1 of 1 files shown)


npc/instances/NydhoggsNest.txt
```diff
@@ -1849,7 +1849,7 @@ OnTouch:
 }
 
 2@nyd,199,268,0	script	nyd_2f_boss_enter	-1,8,8,{
-OnTouch_:
+OnTouch:
```
Atemo:
```suggestion
OnTouch:	// !Fix me: official script uses OnTouch_
```

Atemo (approved):

PR #9734, opened 2026-01-10
Split up Blacksmith skills

Lemongrass3110:

* **Server Mode**: Both


src/map/skill.cpp
```diff
@@ -8785,10 +8781,7 @@ int32 skill_castend_nodamage_id (block_list *src, block_list *bl, uint16 skill_i
 		}
 		break;
 
-	case BS_ADRENALINE:
 	case BS_ADRENALINE2:
```
Atemo: BS_ADRENALINE2 need to be removed

PR #9739, opened 2026-01-12
Split up Hunter skills

Lemongrass3110:

* **Server Mode**: Both

### DIFF
--- a/src/map/skills/archer/landmine.cpp
+++ b/src/map/skills/archer/landmine.cpp
@@ -8,6 +8,13 @@
 SkillLandMine::SkillLandMine() : SkillImpl(HT_LANDMINE) {
 }
 
+void SkillLandMine::castendPos2(block_list* src, int32 x, int32 y, uint16 skill_lv, t_tick tick, int32& flag) const {
+	// Set flag to 1 to prevent deleting ammo (it will be deleted on group-delete).
+	flag |= 1;
+
+	skill_unitsetting(src,getSkillId(),skill_lv,x,y,0);
+}
+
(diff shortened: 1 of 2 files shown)


src/map/skills/archer/landmine.cpp
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#include "landmine.hpp"
+
+#include "map/status.hpp"
+
+SkillLandMine::SkillLandMine() : SkillImpl(HT_LANDMINE) {
+}
+
```
Atemo, line 10: Missing castendPos2
Maybe we could add calculateSkillRatio in battle_calc_misc_attack as well ?

src/map/skills/archer/skill_factory_archer.cpp
```diff
@@ -47,6 +48,8 @@ std::unique_ptr<const SkillImpl> SkillFactoryArcher::create(const e_skill skill_
 			return std::make_unique<SkillAnkleSnare>();
 		case HT_BLASTMINE:
 			return std::make_unique<SkillBlastMine>();
+		case HT_BLITZBEAT:
+			return std::make_unique<SkillBlitzBeat>();
```
Atemo, line 56: We could also use `return std::make_unique<SkillImplRecursiveDamageSplash>(skill_id);` unless we add more line in the file later

Lemongrass3110: Yeah we will add more to it

src/map/skills/archer/landmine.cpp
Lemongrass3110: Thx!
Either this or we add another function to the class. We will have to analyze that first.

PR #9743, opened 2026-01-14
Split up Alchemist skills

Lemongrass3110:

* **Server Mode**: Both

### DIFF
--- a/src/map/map-server-generator.vcxproj
+++ b/src/map/map-server-generator.vcxproj
@@ -345,6 +345,7 @@
     <ClInclude Include="skills/merchant/syntheticarmor.hpp" />
     <ClInclude Include="skills/merchant/twilightalchemy1.hpp" />
     <ClInclude Include="skills/merchant/twilightalchemy2.hpp" />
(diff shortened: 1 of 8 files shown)


src/map/skills/merchant/aidpotion.cpp
```diff
@@ -0,0 +1,149 @@
...
+		clif_skill_nodamage(nullptr,*target,MG_SRECOVERY,sp);
+	if (tsc) {
+#ifdef RENEWAL
+		if (tsc->getSCE(SC_EXTREMITYFIST))
+			sp = 0;
+#endif
+		if (tsc->getSCE(SC_NORECOVER_STATE)) {
+			hp = 0;
+			sp = 0;
+		}
```
Danil0v3s, line 146: NIT: some are `if (foo)` and some are `if( foo )`. Could we settle? xD

Danil0v3s (approved):

src/map/skills/merchant/aidpotion.cpp
Atemo: It easier to compare this code with the original code with minimal change. We could clean it up in another PR?

PR #9755, opened 2026-01-17
Fixed full strip which, when used by monster, bypassed FCP

Atemo:
* **Addressed Issue(s)**: NA

* **Server Mode**: both

### DIFF
--- a/src/map/skills/thief/divestall.cpp
+++ b/src/map/skills/thief/divestall.cpp
@@ -17,9 +17,11 @@ void SkillDivestAll::castendNoDamageId(block_list *src, block_list *target, uint
 	bool i;
 
 	//Special message when trying to use strip on FCP [Jobbie]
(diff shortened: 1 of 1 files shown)


Lemongrass3110 (approved):

Playtester:
Is this issue caused by the refactor? I'm asking because it's pushing into the refactor and I don't see it being linked to any reports.

Atemo:
The base branch is [refactor/stalker_skills](https://github.com/rathena/rathena/tree/refactor/stalker_skills) for reading. The issue  was found during the refactor. I didn't find any open issues about it, and indeed, I haven't opened any myself. Perhaps we could check and do more test, if you'd like to take care of it.

Playtester:
> The base branch is [refactor/stalker_skills](https://github.com/rathena/rathena/tree/refactor/stalker_skills) for reading. The issue was found during the refactor. I didn't find any open issues about it, and indeed, I haven't opened any myself. Perhaps we could check and do more test, if you'd like to take care of it.

I tested for a while by tanking Moonlight Flower. It seems she only strips me when I don't have FCP. So seems FCP protects from strip even against MVPs.

PR #9742, opened 2026-01-14
Fixed custom job master npc

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9733

* **Server Mode**: 

* **Description of Pull Request**:

PR #9765, opened 2026-01-24
Druid / Karnos / Alitea Class Implementation

datawulf:
* **Addressed Issue(s)**:  New Class Druid / Karnos / Alitea

* **Server Mode**:  Renewal

This is server side implementation of Druid class.
All skills match the first rebalance state (January 7th).
Implementation is based on in-game descriptions, available showcase videos and common sense.

If you find anything wrong, please report it (ideally with a reference) and I will adjust accordingly.

This was my first time touching the new skill_factory so bear with me if some parts are still messy or chaotic. Let's bring this together into a good state.

Jobmaster should be checked after https://github.com/rathena/rathena/pull/9742

Showcase of the added class can be seen here: https://youtu.be/PJ14yU85z0g

Tested / developed with client 20251217

Client-side support is not included and cannot be provided by me.


Atemo (commented):
Please use `sc->hasSCE` instead of `sc->getSCE` (when relevant)

src/map/skills/druid/aroundflower.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+void SkillAroundFlower::castendDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32& flag) const {
+	if (!(flag & 1)) {
+		clif_skill_nodamage(src, *target, getSkillId(), skill_lv);
+	}
+
+	SkillImplRecursiveDamageSplash::castendDamageId(src, target, skill_lv, tick, flag);
+}
+
+void SkillAroundFlower::calculateSkillRatio(const Damage* wd, const block_list* src, const block_list* target, uint16 skill_lv, int32& base_skillratio, int32 mflag) const {
```
Atemo: Rename `base_skillratio` into `skillratio` and update the code, when `RE_LVL_DMOD` is defined

src/map/skills/druid/aroundflower.cpp
```diff
@@ -0,0 +1,34 @@
...
+	castendDamageId(src, target, skill_lv, tick, flag);
+}
+
+void SkillAroundFlower::castendDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32& flag) const {
+	if (!(flag & 1)) {
+		clif_skill_nodamage(src, *target, getSkillId(), skill_lv);
+	}
+
+	SkillImplRecursiveDamageSplash::castendDamageId(src, target, skill_lv, tick, flag);
+}
```
Atemo:
```suggestion
void SkillAroundFlower::splashSearch(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32 flag) const {
	clif_skill_nodamage(src, *target, getSkillId(), skill_lv);

	SkillImplRecursiveDamageSplash::splashSearch(src, target, skill_lv, tick, flag);
}
```
`splashSearch` seems more relevant to me

src/map/skills/druid/aroundflower.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
+// For more information, see LICENCE in the main folder
+
+#include "aroundflower.hpp"
+
```
Atemo, line 5: Should we add `#include <config/const.hpp>` when `RE_LVL_DMOD` is used, for clarity?

src/map/skills/druid/aroundflower.cpp
```diff
@@ -0,0 +1,34 @@
...
+void SkillAroundFlower::calculateSkillRatio(const Damage* wd, const block_list* src, const block_list* target, uint16 skill_lv, int32& base_skillratio, int32 mflag) const {
+	const status_change* sc = status_get_sc(src);
+	const status_data* sstatus = status_get_status_data(*src);
+
+	int32 skillratio = 100 * skill_lv;
+	if (sc && sc->hasSCE(SC_TRUTH_OF_EARTH)) {
+		skillratio += sstatus->int_; // TODO - unknown scaling [munkrej]
+		RE_LVL_DMOD(100);
+	}
```
Atemo:
```suggestion
	}
	RE_LVL_DMOD(100);
```
Baselevel 227 without equipment, without buff, skill_lv 10 (base skillratio 1000) vs mob with 0 def, 25 softmdef :

Damage on KRO : 27782
By calculation :
MATK = MATK + MATK * SMATK / 100 = 901 + 901 * 36 / 100 = 1225
Dmg = skillratio * MATK / 100 - SoftMDef = 12225
Dmg doesn't match

But if we include `RE_LVL_DMOD(100);` :
skillratio = skillratio * Baselevel / 100 = 2270
Dmg = skillratio * MATK / 100 - SoftMDef = 27782
Dmg matches

(at least for karnos / alitea) unlike what the description say `RE_LVL_DMOD(100);` is not part of `SC_TRUTH_OF_EARTH` condition. It would be interesting to check for Baselevel < 100 (druid)

src/map/skills/druid/aroundflower.cpp
```diff
@@ -0,0 +1,34 @@
...
+	SkillImplRecursiveDamageSplash::castendDamageId(src, target, skill_lv, tick, flag);
+}
+
+void SkillAroundFlower::calculateSkillRatio(const Damage* wd, const block_list* src, const block_list* target, uint16 skill_lv, int32& base_skillratio, int32 mflag) const {
+	const status_change* sc = status_get_sc(src);
+	const status_data* sstatus = status_get_status_data(*src);
+
+	int32 skillratio = 100 * skill_lv;
+	if (sc && sc->hasSCE(SC_TRUTH_OF_EARTH)) {
+		skillratio += sstatus->int_; // TODO - unknown scaling [munkrej]
```
Atemo:
```suggestion
		skillratio += 5 * sstatus->int_;
```
I can provide the calculation if needed

Atemo (commented):
Quick review

datawulf:
> I cant put any skill tree after changing to druid class when using Job Changer NPC. also there is a New Basic Skill of doram in ETC skill tab.

are you sure you applied the changes correctly and are using a supported client. Also jobmaster is not included in this PR yet. @job works fine on clean rathena, I can't reproduce your issue.

src/map/skills/druid/icepillar.cpp
```diff
@@ -0,0 +1,49 @@
...
+// For more information, see LICENCE in the main folder
+
+#include "icepillar.hpp"
+
+#include <config/const.hpp>
+
+#include "map/clif.hpp"
+#include "map/status.hpp"
+
+SkillIcePillar::SkillIcePillar() : SkillImplRecursiveDamageSplash(KR_ICE_PILLAR) {
```
Atemo, line 10: The implementation seems incorrect.
This is a ground skill casted on position.
There is no `clif_skill_nodamage` for this skill.
On success the skill deals `DMG_MULTI_HIT` div 2 (init damage) and deals the first hit of `DMG_SPLASH` div 1 (fog damage) instantly.
With each iteration of fog damage (500ms between the 1st and 2nd, then 1 second), EFST_ICE_PILLAR is removed (when active) then re-started for the caster for 5 seconds (even if the damage doesn't hit). The fog damage is dealt before EFST_ICE_PILLAR begins, meaning the 1st fog damage is always without buff.
There is a total of 8 "fog damages" max.
The last EFST_ICE_PILLAR lasts 5 seconds (not removed).

datawulf: Adjusted the skill based on your comment

db/re/skill_db.yml
```diff
@@ -49135,8 +49135,8 @@ Body:
     TargetType: Attack
     Range: 7
     GiveAp: 2
-    Hit: Single
-    HitCount: 1
+    Hit: Multi_Hit
+    HitCount: 2
```
datawulf: @Atemo is this really HitCount: 2? Not HitCount: -2? I don't have a source, just wondering.

Atemo: You're right I made a mistake in my note and I didn't compare ig,
it should be `HitCount: -2`

src/map/skills/druid/groundbloom.cpp
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if (stacks > 12)
+		stacks = 12;
+
+	// Note: official gives the status change regardless of the stacks amount then casts the skill and removes the sc when the conditions are met (same result)
+	status_change_start(src, src, SC_GROUND_GROW, 10000, 0, 0, stacks, 0, 10000, SCSTART_NOAVOID);
+
+	if (cast_ground_bloom)
+		skill_castend_nodamage_id(src, src, KR_GROUND_BLOOM, skill_lv, tick, 0);
```
Atemo: The skill works like this
1. casting a earth skill increases the sc value (can be above max stack 12)
2. immediatly after the stacks are checked and KR_GROUND_BLOOM is casted (it may be slightly before or after the triggering skill, I don't know, but it's at the same timer of the triggering skill)

If I read correctly your change increases sc and doesn't cast immediatly KR_GROUND_BLOOM when sc value is increased to 12+ which is incorrect

datawulf: Not sure I misunderstood your note then. But it would never reach max stacks before. Here you can see 12 Stacks in the clip: https://youtu.be/7vjSsONtDMQ?si=whmWoKWM0SPC_3VN&t=72
The next cast causes ground bloom.
So it should not trigger at 12 stacks, that's sure. Not sure how it works if it would go from 11 + 2

Atemo: I see you're right there was a bug, but it was on the sc value. The skill is casted immediatly at 13+ then, the max stack is 12

> Not sure how it works if it would go from 11 + 2

Same mecanic, sc is updated and KR_GROUND_BLOOM is immediatly casted if sc value > 12

datawulf: reverted my change and applied just the fix to the condition

amm0nition:
It seems that Alitea cannot allocate Trait Status points because it's not recognized as a Fourth Job equivalent.
A quick change I did to make it work is by adding `(class_ & MAPID_THIRDMASK) == MAPID_ALITEA` into `#define pc_is_upper_expanded_second(class_)` in `pc.hpp`. 

This was tested in 20250716 client so I can't say for sure if clientdate has anything to do with this.

fmendezy:
> It seems that Alitea cannot allocate Trait Status points because it's not recognized as a Fourth Job equivalent. A quick change I did to make it work is by adding `(class_ & MAPID_THIRDMASK) == MAPID_ALITEA` into `#define pc_is_upper_expanded_second(class_)` in `pc.hpp`.
> 
> This was tested in 20250716 client so I can't say for sure if clientdate has anything to do with this.

Support for the Druid job branch requires a 2025-12-17 client or newer. For weapon/shield sprites and other fixes to work Is recommend the 2026-01-21 client.

rabbit-lunatic:
Could you please update this commit?

fmendezy:
Hi, I recommend using this client which includes the first rebalancing of these classes on the same day as the executable, so that the effects of the skills work perfectly.
[2026-01-07-ragexe](https://rathena.org/board/topic/149414-2026-01-07-ragexe-clientinfo-warp/)

Acknowledgments [VictorHug0](https://rathena.org/board/profile/22192-victorhug0/)

llchrisll:
I didn't check any videos but I read the skill descriptions (obv).
I thought the Primal Claw, Feral Claw and Alpha Claw behave like combo skills, so you have to press each skill on it's own.
But noticing that now, I guess that's not the case? or is that a bug?

First I pressed the skills on it's own hotkey as I expected and it also worked, then I pressed a wrong key by accident but it still followed up with the last skill, Alpha Claw.
See in the clip, what I mean.
https://github.com/user-attachments/assets/c24051f0-5deb-4fd7-ac71-8a3ee52c1058

And correct me if I'm wrong.

Atemo:
> I didn't check any videos but I read the skill descriptions (obv). I thought the Primal Claw, Feral Claw and Alpha Claw behave like combo skills, so you have to press each skill on it's own. But noticing that now, I guess that's not the case? or is that a bug?
> 
> First I pressed the skills on it's own hotkey as I expected and it also worked, then I pressed a wrong key by accident but it still followed up with the last skill, Alpha Claw. See in the clip, what I mean. https://github.com/user-attachments/assets/c24051f0-5deb-4fd7-ac71-8a3ee52c1058
> 
> And correct me if I'm wrong.

Official behaviour, handled by the client

llchrisll:
> > I didn't check any videos but I read the skill descriptions (obv). I thought the Primal Claw, Feral Claw and Alpha Claw behave like combo skills, so you have to press each skill on it's own. But noticing that now, I guess that's not the case? or is that a bug?
> > First I pressed the skills on it's own hotkey as I expected and it also worked, then I pressed a wrong key by accident but it still followed up with the last skill, Alpha Claw. See in the clip, what I mean. https://github.com/user-attachments/assets/c24051f0-5deb-4fd7-ac71-8a3ee52c1058
> > And correct me if I'm wrong.
> 
> Official behaviour, handled by the client

Alright, thanks for the clarification :).

mikiya1991-max:
Anyone have the equipment detail for the Karnos,Druid and Alitea?

hendra814:
this pull have conflict with new rathena update at skills/skill_factory.cpp

gasstation51:
SC_GLACIER_SHEILD is not generated when using Glacial Monolith. Therefore, Glacial Stomp cannot be used either.

Atemo:
Glacial Monolith and Glacial Stomp are working fine for me please re-check.

Minato448:
I don't know if it's a bug or not. I tested the Druid class and found that when changing into a wolf and a bird, it alternated back and forth. And sometimes I see the outfit as a novice.

PR #9772, opened 2026-01-30
Split up Shadow Cross skills

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9464

* **Server Mode**: 


src/map/skills/thief/crossslash.cpp
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) rAthena Dev Teams - Licensed under GNU GPL
```
Lemongrass3110, line 1: File and class name need shc removed

src/map/skill.cpp
```suggestion
		if (skill_id == IQ_MASSIVE_F_BLASTER || skill_id == MT_AXE_STOMP || skill_id == ABC_ABYSS_DAGGER)
```

src/map/skills/thief/crossslash.cpp
Atemo: There is another skill from kagerou using the same skill name, it looks like the compiler doesn't like it. I can change this one if you want

Lemongrass3110: Ay ay ay. Ok then ignore my comment.
Strange that they did that though.

Issue #9784, opened 2026-02-03
Extend SkillImpl to handle damage display logic


jezztify:
### Is your request related to a missing official feature?

No

### Describe the solution you'd like

@Atemo has done a lot of amazing work in splitting the skills into their own logic.  I am thinking that we can further extend this by handling the damage display logic in these files as well. This will help ease up the massive switch statement that is bundled within the `skill_attack` function.

PR #9785, opened 2026-02-03
Extend SkillImpl to handle damage display logic

jezztify:
* **Addressed Issue(s)**: #9784

* **Server Mode**: Both

This PR adds the ability for individual skill implementations to control their own damage display, isolating display logic from the monolithic switch statement in `skill_attack` into their respective skill implementation files.

- Added `SD_NODISPLAY` flag to enable splitted skills to handle their own damage display
- Added example implementations for `souldestroy.cpp` and `sightblaster.cpp`

### DIFF
--- a/src/map/skill.cpp
+++ b/src/map/skill.cpp
@@ -3602,6 +3602,8 @@ int64 skill_attack (int32 attack_type, block_list* src, block_list *dsrc, block_
 	skill_combo(src,dsrc,bl,skill_id,skill_lv,tick);
 
 	//Display damage.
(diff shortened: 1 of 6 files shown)


Lemongrass3110:
We will add more functionality once this "first round" has been completed.

We can look into your suggestion later on, but something like this is definitely planned.

jezztify:
> We will add more functionality once this "first round" has been completed.
> 
> We can look into your suggestion later on, but something like this is definitely planned.

Sweet. Thanks @Lemongrass3110.

PR #9797, opened 2026-02-09
Unity or amalgamated build for skills

Lemongrass3110:
* **Addressed Issue(s)**: Compile time

* **Server Mode**: Both

All files that were split up for the purpose of #9464 now are built via an unity build or amalgamated build.
For this purpose the skill related files are included directly into skill.cpp
The files are excluded from being build separately in MSVS, Make and CMake.
This restores the original compile time, while keeping the new structure.


YlenXWalker:
Neat! I don't have to wait too long for compilation.  Thanks!

PR #9827, opened 2026-02-14
Split up Wanderer skills

Danil0v3s:
* **Addressed Issue(s)**: 

* **Server Mode**: 


src/map/skills/archer/skill_factory_archer.cpp
```diff
@@ -16,7 +16,6 @@
 #include "battletheme.cpp"
 #include "beaststrafing.cpp"
 #include "blastmine.cpp"
-#include "blitzbeat.cpp"
```
Atemo, line 20: Please revert

Danil0v3s: I think this is the same issue with Rebellion, let me re-generate

PR #9837, opened 2026-02-17
Fix compiler errors after skill split

vstumpf:
* **Addressed Issue(s)**: N/A

* **Server Mode**: Both

* **Description of Pull Request**: These files fail to compile if you don't do the amalgamation. This also causes issues with intellisense/clangd, because some of the used functions/macros/variables aren't declared.

### DIFF
--- a/src/map/skills/swordman/trample.cpp
+++ b/src/map/skills/swordman/trample.cpp
@@ -3,6 +3,8 @@
 
 #include "trample.hpp"
 
+#include <common/nullpo.hpp>
+
 #include "map/battle.hpp"
 #include "map/clif.hpp"
 #include "map/map.hpp"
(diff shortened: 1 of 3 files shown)


src/map/skills/swordman/trample.cpp
vstumpf, line 23: This macro is undefined without nullpo

src/map/skills/thief/chasingbreak.cpp
vstumpf, line 34: `dirx` and `diry` don't exist!

src/map/skills/thief/chasingshot.cpp
vstumpf, line 34: `dirx` and `diry` don't exist!

PR #9846, opened 2026-02-20
prometheus metrics collector

Danil0v3s:
* **Addressed Issue(s)**: 
* ND

* **Server Mode**: 
* Both

* [Prometheus](https://github.com/prometheus/prometheus), a [Cloud Native Computing Foundation](https://cncf.io/) project, is a systems and service monitoring system. It collects metrics from configured targets at given intervals, evaluates rule expressions, displays the results, and can trigger alerts when specified conditions are observed.

It allows server owners to trace and keep track of their resource usages, such as, but not limited to, total cycle time (the amount of time it takes for the server to complete everything once, and also the specifics cycle. In this PR i propose a couple of metrics I could think of but anything could be in handy, really.

The current setup feeds the metrics to a file, which is then used by the docker setup to feed it into prometheus and then collected by grafana to display them nicely

Here's an example:

image

An interesting thing about this setup is that I can track my NPCs running time (Not sure if npcs using sleep or some fancy delays trick would report correctly or what) but ideally this setup would allow server owners to easily spot obvious performance hoggers


reunite-ro:
Current pull, I got a lot of errors compiling at windows, will compile in a ubuntu test server soon.

PR #9850, opened 2026-02-21
Improve docker setup

artemmikhalitsin:
**Addressed Issue(s)**: 
- Had a few painpoints setting up regarding SELinux, and figuring out how to create an admin

**Server Mode**: 

Both

- Fix dockerfile - use ":z" to properly share directories. Contributors should not be required to disable SELinux
- Add Makefile with commands for common operations
- Account management commands (one-command admin account)
- Update README with simplified setup instructions
- Update to use docker compose 2 command syntax

### DIFF
--- a/db/re/item_db_etc.yml
+++ b/db/re/item_db_etc.yml
@@ -83035,6 +83035,15 @@ Body:
       hateffect FOOTPRINT_EF_BASE,true;
     UnEquipScript: |
       hateffect FOOTPRINT_EF_BASE,false;
+  - Id: 313066
+    AegisName: aegis_313066
+    Name: Whirlwind Footprint    # !todo check english name
+    Type: Card
+    SubType: Enchant
+    Script: |
+      hateffect FOOTPRINT_EF_STR_BASE,true;
+    UnEquipScript: |
+      hateffect FOOTPRINT_EF_STR_BASE,false;
   - Id: 313067
     AegisName: aegis_313067
     Name: Purple Star Footprint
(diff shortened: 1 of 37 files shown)


Danil0v3s:
I recently had a go with docker again and decided to try removing that builder service and instead build the image and have each of the 4 services living in them without the source code... you can check at #9718

I would say going forward we should have only 1 dev container and then the final/production container should be one where we strip down the sources.

Also, we could add a service with something like pgadmin but mariadb/mysql on the compose file

What do you think?

artemmikhalitsin:
@Danil0v3s hey thanks for taking a look

Everything I say comes with the disclaimer that I'm not familiar with your development process and what works best for you. My comments are based on personal work experience of onboarding to new projects.

For development, that will be a super annoying setup that causes a lot of friction. You really don't want to re-compile everything and rebuild the image if you're just working on `char`, so you should try to leverage the power of `make` in recompiling only specific targets, and volumes so your changes get reflected into the container immediately without rebuild the image.

My changes in this PR only seek to improve experience with the current setup, but I have some thoughts about the best ways to handle docker-based development and production.

**For development**
Since the separate server binaries share some common code, you can either:
- use the builder container, then just mount the compiled binary + any assets required to run the server, then trigger a restart so the container starts with the new binary. each service's image should contain only the dependencies required to run that binary.
- mount src into each of the service containers and provide commands for developers to rebuild the service's binary + restart the application. Each image will need to have dependencies to build + run its binary.

Both approaches are valid, depending how you want developers to interact with the docker-compose stack, and what kind of production-development parity you want.

You definitely should not be mounting the whole repository into each of your containers. It works but it's lazy lol.

**For production**
- The production setup should provide a way to create a separate image for each of the services. They might share a common "buider" first step, but in the end each production image should contain only its own services binary + any assets required for the server, in order to cut down the image size as much as possible.

Regarding the `patch.sh` script from your draft PR: imho including the char_conf.txt, inter_conf.txt, already configured to be mounted for the docker setup is a better. The closer you can bring the new contributor experience to "Step 1: pull, Step 2: `docker compose up`, Step 3: you're developing", the better. 
The best implementation would be for each server to accept the addresses it cares about through env

Question about pgadmin - what kind of things do server admins usually do? You can theoretically add the pgadmin panel as another container (https://docs.docker.com/guides/pgadmin/), but I'm more curious why server admins have to dick around with their database lol

PR #9899, opened 2026-03-06
Corrected refineable restrictions of item ID 19240 (Toy Syringe)

idk-whoami:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

* **Description of Pull Request**: Corrected refineable restrictions of item ID 19240 (Toy Syringe)


db/re/item_db_equip.yml
```diff
@@ -96388,7 +96388,6 @@ Body:
     Locations:
       Head_Low: true
     ArmorLevel: 1
-    Refineable: true
     Script: |
```
Atemo, line 96393:
```suggestion
#   Bugfixes: can be refined according to official data
#   Refineable: true
    Script: |
```

idk-whoami: Thanks!

PR #9902, opened 2026-03-07
Corrected BO_ADVANCE_PROTECTION

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

The skill couldn't fail when casted by non-pc

### DIFF
--- a/src/map/skills/merchant/advanceprotection.cpp
+++ b/src/map/skills/merchant/advanceprotection.cpp
@@ -16,9 +16,10 @@ void SkillAdvanceProtection::castendNoDamageId(block_list* src, block_list* targ
 		if (map_session_data* sd = BL_CAST(BL_PC, src); sd != nullptr) {
 			clif_skill_fail(*sd, getSkillId());
 
-			// Don't consume item requirements
-			flag |= SKILL_NOCONSUME_REQ;
 		}
+
+		// Don't consume item requirements
+		flag |= SKILL_NOCONSUME_REQ;
 		return;
 	}
 


src/map/skills/merchant/advanceprotection.cpp
```diff
@@ -9,15 +9,16 @@
...
-		clif_skill_fail( *sd, getSkillId() );
-		// Don't consume item requirements
-		flag |= SKILL_NOCONSUME_REQ;
+	if (dstsd == nullptr || pc_checkequip(dstsd, EQP_SHADOW_GEAR) < 0) {
+		if (map_session_data* sd = BL_CAST(BL_PC, src); sd != nullptr) {
+			clif_skill_fail(*sd, getSkillId());
+
+			// Don't consume item requirements
+			flag |= SKILL_NOCONSUME_REQ;
+		}
```
Lemongrass3110:
```suggestion
		}

		// Don't consume item requirements
		flag |= SKILL_NOCONSUME_REQ;
```

Only the clif call should be in the if(sd) block

Atemo: The change doesn't matter to me but I made it like that because `SKILL_NOCONSUME_REQ` is (only) used for `skill_consume_requirement`, which always check sd. The flag is not used for non-pc
```
	if( sd && !(flag&1) )
	{// ensure that the skill last-cast tick is recorded
		sd->canskill_tick = gettick();

		if( sd->state.arrow_atk )
		{// consume arrow on last invocation to this skill.
			battle_consume_ammo(sd, skill_id, skill_lv);
		}
		skill_onskillusage(sd, bl, skill_id, tick);
		// perform skill requirement consumption
		if (!(flag&SKILL_NOCONSUME_REQ))
			skill_consume_requirement(sd,skill_id,skill_lv,2);
	}
```

PR #9720, opened 2026-01-07
Converted (almost) all StatusSkillImpl

Lemongrass3110:
* **Addressed Issue(s)**: #9464 

* **Server Mode**: Both

* **Description of Pull Request**: 

Converted (almost) all skills from the StatusSkillImpl cases in the old code to the new structure.
This will make it easier to review later on.

Identified additional cases (=skills) which are also StatusSkillImpls:

1. MER_AUTOBERSERK
2. SA_REVERSEORCISH
3. HAMI_BLOODLUST
4. HFLI_FLEET, HFLI_SPEED, MH_ANGRIFFS_MODUS, MH_GOLDENE_FERSE
5. GC_WEAPONBLOCKING
6. LG_REFLECTDAMAGE
7. SR_GENTLETOUCH_CHANGE, SR_GENTLETOUCH_ENERGYGAIN, SR_GENTLETOUCH_REVITALIZE
8. SU_HIDE
9. SU_STOOP
10. NW_AUTO_FIRING_LAUNCHER, NW_HIDDEN_CARD

This PR is based on #9719 and the other PR has to be merged first.

PR #9904, opened 2026-03-07
Updated SkillVanishingSlash

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

Moved the status change in applyAdditionalEffects

### DIFF
--- a/src/map/skills/ninja/vanishingslash.cpp
+++ b/src/map/skills/ninja/vanishingslash.cpp
@@ -7,11 +7,11 @@
 
 #include "map/status.hpp"
 
-SkillVanishingSlash::SkillVanishingSlash() : StatusSkillImpl(NJ_KASUMIKIRI) {
+SkillVanishingSlash::SkillVanishingSlash() : WeaponSkillImpl(NJ_KASUMIKIRI) {
 }
 
 void SkillVanishingSlash::applyAdditionalEffects(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32 attack_type, enum damage_lv dmg_lv) const {
-	sc_start(src, src, SC_HIDING, 100, skill_lv, skill_get_time(getSkillId(), skill_lv));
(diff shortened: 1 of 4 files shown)


src/map/skills/ninja/vanishingslash.cpp
```diff
@@ -10,16 +10,18 @@
 SkillVanishingSlash::SkillVanishingSlash() : StatusSkillImpl(NJ_KASUMIKIRI) {
```
Lemongrass3110: NJ_KASUMIKIRI was converted to a StatusSkillImpl as part of #9720.
However in skill_db.yml it has no status change entry, neither in pre-renewal nor in renewal.
So currently it cannot start any status change.
I think we should simply change it to WeaponSkillImpl.
This was an issue in the old code, as it also not defined as nodamage skill the old logic was most likely never triggered.

src/map/skills/ninja/vanishingslash.cpp
```diff
@@ -10,16 +10,18 @@
...
 #else
 	base_skillratio += 10 * skill_lv;
 #endif
 }
 
-void SkillVanishingSlash::castendDamageId(block_list *src, block_list *target, uint16 skill_lv, t_tick tick, int32& flag) const {
-	// TODO: refactor into applyAdditionalEffects
-	if (skill_attack(BF_WEAPON,src,src,target,getSkillId(),skill_lv,tick,flag) > 0)
-		sc_start(src,src,SC_HIDING,100,skill_lv,skill_get_time(getSkillId(),skill_lv));
+void SkillVanishingSlash::castendDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32& flag) const {
```
Lemongrass3110: Can be removed if we switch to WeaponSkillImpl

Lemongrass3110 (changes requested):

PR #9906, opened 2026-03-07
Refactor dragontail.cpp using SkillImplRecursiveDamageSplash

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: 

### DIFF
--- a/src/map/skills/gunslinger/dragontail.cpp
+++ b/src/map/skills/gunslinger/dragontail.cpp
@@ -6,8 +6,6 @@
 #include <config/core.hpp>
 
 #include "map/clif.hpp"
-#include "map/map.hpp"
-#include "map/pc.hpp"
 #include "map/status.hpp"
 
 SkillDragonTail::SkillDragonTail() : SkillImplRecursiveDamageSplash(RL_D_TAIL) {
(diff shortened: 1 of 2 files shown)


src/map/skills/gunslinger/dragontail.cpp
```diff
@@ -10,20 +10,32 @@
...
 
-// TODO: refactor to SkillImplRecursiveDamageSplash
-SkillDragonTail::SkillDragonTail() : SkillImpl(RL_D_TAIL) {
+SkillDragonTail::SkillDragonTail() : SkillImplRecursiveDamageSplash(RL_D_TAIL) {
 }
 
-void SkillDragonTail::castendDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32& flag) const {
-	const map_session_data* sd = BL_CAST(BL_PC, src);
-	const status_change* tsc = status_get_sc(target);
+void SkillDragonTail::calculateSkillRatio(const Damage* wd, const block_list* src, const block_list* target, uint16 skill_lv, int32& skillratio, int32 mflag) const {
```
Lemongrass3110: Please change back the order of the functions.
Like this it is hard to compare.

src/map/skills/gunslinger/dragontail.cpp
```diff
@@ -10,20 +10,22 @@
...
+SkillDragonTail::SkillDragonTail() : SkillImplRecursiveDamageSplash(RL_D_TAIL) {
 }
 
-void SkillDragonTail::castendDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32& flag) const {
+int32 SkillDragonTail::getSplashTarget(block_list* src) const {
+	return BL_CHAR;
+}
+
+int64 SkillDragonTail::splashDamage(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32 flag) const {
+	return skill_attack(skill_get_type(getSkillId()), src, src, target, getSkillId(), skill_lv, tick, flag | SD_ANIMATION);
```
Lemongrass3110: Since we already override splashSearch, you could put skill_area_temp[0] |= SD_ANIMATION into it?

src/map/skills/gunslinger/dragontail.cpp
```diff
@@ -33,15 +35,16 @@ void SkillDragonTail::castendDamageId(block_list* src, block_list* target, uint1
...
+	SkillImplRecursiveDamageSplash::splashSearch(src, target, skill_lv, tick, flag);
 }
 
 void SkillDragonTail::calculateSkillRatio(const Damage* wd, const block_list* src, const block_list* target, uint16 skill_lv, int32& skillratio, int32 mflag) const {
-	const map_session_data* sd = BL_CAST(BL_PC, src);
-
 	skillratio += -100 + 500 + 200 * skill_lv;
-	if (sd && (wd->miscflag & 8)) {
+
+	if (const map_session_data* sd = BL_CAST(BL_PC, src); sd != nullptr && (wd->miscflag & 8)) {
```
Lemongrass3110:
```suggestion
	if (wd->miscflag & 8) {
```
The miscflag is only set based on the logic of another function.
No need to double check.

We could consider using SKILL_ALTDMG_FLAG instead of hardcoding 8.

src/map/skills/gunslinger/dragontail.cpp
```diff
@@ -10,20 +10,22 @@
...
 	const map_session_data* sd = BL_CAST(BL_PC, src);
 	const status_change* tsc = status_get_sc(target);
 
-	if (flag & 1) {
-		skill_attack(skill_get_type(getSkillId()), src, src, target, getSkillId(), skill_lv, tick, flag | SD_ANIMATION);
-		return;
-	}
-
-	if (sd && tsc && tsc->getSCE(SC_C_MARKER)) {
+	if (sd != nullptr && tsc != nullptr && tsc->hasSCE(SC_C_MARKER)) {
```
Lemongrass3110: Using val2 (src->id of the caster, see crimsonmarker.cpp) instead of this lookup would be faster.
It would also allow to remove the pc specific checks.

PR #9810, opened 2026-02-13
Split up Sky Emperor skills

Lemongrass3110:
* **Addressed Issue(s)**: #9464

* **Server Mode**: Both

* **Description of Pull Request**:

PR #9914, opened 2026-03-12
Cleanup and fix SKE_DAWN_BREAK

Lemongrass3110:
* **Addressed Issue(s)**: No damage by SKE_DAWN_BREAK

* **Server Mode**: Both

The WeaponSkillImpl was lost in #9810 and now no damage is dealt.

Thanks to @Jason59762

### DIFF
--- a/src/map/skills/taekwon/dawnbreak.cpp
+++ b/src/map/skills/taekwon/dawnbreak.cpp
@@ -10,16 +10,12 @@
 #include "map/pc.hpp"
 #include "map/status.hpp"
 
-SkillDawnBreak::SkillDawnBreak() : SkillImpl(SKE_DAWN_BREAK) {
+SkillDawnBreak::SkillDawnBreak() : SkillImplRecursiveDamageSplash(SKE_DAWN_BREAK) {
 }
 
 void SkillDawnBreak::castendNoDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32& flag) const {
-	int32 starget = BL_CHAR|BL_SKILL;
-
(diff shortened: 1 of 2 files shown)


src/map/skills/taekwon/dawnbreak.cpp
```diff
@@ -10,16 +10,12 @@
...
 
-SkillDawnBreak::SkillDawnBreak() : SkillImpl(SKE_DAWN_BREAK) {
+SkillDawnBreak::SkillDawnBreak() : SkillImplRecursiveDamageSplash(SKE_DAWN_BREAK) {
 }
 
 void SkillDawnBreak::castendNoDamageId(block_list* src, block_list* target, uint16 skill_lv, t_tick tick, int32& flag) const {
-	int32 starget = BL_CHAR|BL_SKILL;
-
-	skill_area_temp[1] = 0;
 	clif_skill_nodamage(src,*target,getSkillId(),skill_lv);
```
Atemo, line 17: with `skill_area_temp[1] = 0;` SD_ANIMATION is used in skill_attack, shouldn't it be added in splashSearch ?

Lemongrass3110: It looked okay, when I tested it ingame.

Since it is a self targeted skill, the original target (=source) is not a valid target/enemy and therefore skill_attack is not triggered (=>no clif call).

But otherwise I would agree.
Maybe you can test it also ingame and let me know if you think it is displayed incorrectly now?

Atemo: Tested, good to me

PR #9915, opened 2026-03-12
Corrected refineable and trade restrictions of item ID 19267 (Survivor's Orb)

idk-whoami:
* **Addressed Issue(s)**: N/A

* **Server Mode**: renewal

* **Description of Pull Request**: Corrected refineable and trade restrictions of item ID 19267 (Survivor's Orb)


db/re/item_db_equip.yml
```diff
@@ -96726,9 +96726,15 @@ Body:
...
       Head_Low: true
+    Trade:
+      NoDrop: true
+      NoSell: true
+      NoMail: true
+      NoAuction: true
     ArmorLevel: 1
     EquipLevelMin: 1
-    Refineable: true
+#   Bugfixes: can be refined according to official data
```
Atemo: It's refineable on official data?
I don't have this item in my data

idk-whoami: i think there's a flaw in the Aegis data  

do you want the comment deleted?

Atemo: hmm just remove the comment if you don't have official data of the item

idk-whoami: Thanks!

PR #9928, opened 2026-03-15
Initial implementation of NPC_LOCKON_LASER

Lemongrass3110:
* **Addressed Issue(s)**: #1525

* **Server Mode**: Both

Added a skill flag to allow ignoring RES/MRES
Enabled the skill for MD_SWEETY


db/re/skill_db.yml
```diff
@@ -18072,6 +18072,34 @@ Body:
...
     Status: Relieve_off
+  - Id: 773
+    Name: NPC_LOCKON_LASER
+    Description: Lockon Laser
+    MaxLevel: 5
+    TargetType: Attack
+    DamageFlags:
+      NoDamage: true
+    Flags:
+      IsNpc: true
```
Atemo, line 18083:
```suggestion
      IgnoreGtb: true
      IsNpc: true
```

db/re/skill_db.yml
```diff
@@ -18072,6 +18072,34 @@ Body:
...
+    HitCount: 1
+    Duration1: 12000
+    Status: Lockon_Laser
+  - Id: 774
+    Name: NPC_LOCKON_LASER_ATK
+    Description: Lockon Laser Attack
+    MaxLevel: 5
+    Type: Magic
```
Atemo, line 18091: Probably misc? with misc we don't need ignore mres and ignore mdef if I read the code correctly

db/re/skill_db.yml
```diff
@@ -18072,6 +18072,34 @@ Body:
...
+    Name: NPC_LOCKON_LASER
+    Description: Lockon Laser
+    MaxLevel: 5
+    TargetType: Attack
+    DamageFlags:
+      NoDamage: true
+    Flags:
+      IsNpc: true
+    HitCount: 1
+    Duration1: 12000
```
Atemo, line 18085:
```suggestion
    Duration1:
      - Level: 1
        Time: 4000
      - Level: 2
        Time: 6000
      - Level: 3
        Time: 8000
      - Level: 4
        Time: 10000
      - Level: 5
        Time: 12000
```

db/re/skill_db.yml
```diff
@@ -18072,6 +18072,34 @@ Body:
...
     Duration1: 60000
     Status: Relieve_off
+  - Id: 773
+    Name: NPC_LOCKON_LASER
+    Description: Lockon Laser
+    MaxLevel: 5
+    TargetType: Attack
+    DamageFlags:
+      NoDamage: true
+    Flags:
```
Atemo, line 18082: there is probably a range, I don't know it but `Range: 9` is often used

fab10Carvalho:
Cicada Skin Shed (NJ_UTSUSEMI) Can block damage of this skill, Tested from KRO

crazy-arashi:
Just an addition to what Atemo said about 
`NPC_LOCKON_LASER give the sc. 1 second later and every 1 second until the end of the sc a ground unit is used on player with sc`

The ground sc unit also do a distance/wall check before the damage actually applies.

PR #9936, opened 2026-03-17
Split Skill Compilation Per Job

reunite-ro:
* **Addressed Issue(s)**: None.

* **Server Mode**: Not related.

* **Description of Pull Request**: Split skill compilation into per-job which can reduce memory used during compiling skill.cpp

Currently, 2GB ram isn't enough to compile skill, it needs around 4.5GB RAM to successfully compile. For 4GB RAM and lower most probably having issue compiling skill right now.

Adding swapfile around 4GB most probably helps without this PR.

My Visual Studio Build Test using  Ryzen 5 3500 :

Clean rathena
map only
========== Rebuild All: 4 succeeded, 0 failed, 0 skipped ==========
========== Rebuild completed at 8:19 AM and took 53.118 seconds ==========
full
========== Rebuild All: 15 succeeded, 0 failed, 0 skipped ==========
========== Rebuild completed at 8:21 AM and took 01:43.516 minutes ==========

PR
map only
========== Rebuild All: 4 succeeded, 0 failed, 0 skipped ==========
========== Rebuild completed at 8:23 AM and took 56.889 seconds ==========
full
========== Rebuild All: 15 succeeded, 0 failed, 0 skipped ==========
========== Rebuild completed at 8:25 AM and took 01:52.193 minutes ==========

### DIFF
--- a/src/map/CMakeLists.txt
+++ b/src/map/CMakeLists.txt
@@ -13,6 +13,13 @@ file(GLOB_RECURSE MAP_HEADERS ${MAP_SOURCE_DIR}/*.hpp)
 list(FILTER MAP_HEADERS EXCLUDE REGEX ".*/skills/.*")
 file(GLOB_RECURSE MAP_SOURCES ${MAP_SOURCE_DIR}/*.cpp)
 list(FILTER MAP_SOURCES EXCLUDE REGEX ".*/skills/.*")
+# Add skill factory/impl files as separate translation units per job category
+# Individual skill files are NOT compiled directly - they are #included by their job factory TUs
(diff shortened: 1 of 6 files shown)


kaninhot004:
I think this is good PR because currently skill.cpp take around 15+ minutes to compile with swap 4G (Test on ubuntu 1G ram).

reunite-ro:
> I think this is good PR because currently skill.cpp take around 15+ minutes to compile with swap 4G (Test on ubuntu 1G ram).

Have u tried this PR and compare the time taken for compile with it ?

By the way, this PR is actually for compilation to be possible at lower RAM and lower swap file, not really for the compilation speed.

As the test above, I could assume the time to compile most probably gonna be a bit longer, but makes the compilation possible even with 2GB RAM only.

kaninhot004:
Oops, I see. My mistake haha.
(Also not tried this PR yet)

Lemongrass3110 (approved):

Issue #9945, opened 2026-03-21
Visual bug for Berserk State for other player's perspective


gidzdlcrz:
### rAthena Hash

Latest

### Client Date

2025

### Server Mode

Pre-Renewal, Renewal

### Result

This has been reported to rA's discord. Players won’t be able to see if another player uses a berserk skill if it is used outside of their perspective.

Context:

<img width="1238" height="744" alt="Image" src="https://github.com/user-attachments/assets/7199bf0e-c664-4e67-baf1-ac4ee8322dc9" />

### Relevant Log Output

```Shell
N/A
```

### Expected Result

Other players should be able to see the option3 (bersek effect)

### How to Reproduce

1. Use the berserk skill and move out of Player B’s perspective.
2. Return to Player B’s perspective.
3. From Player B’s perspective, check the view.

Screen Record:

https://github.com/user-attachments/assets/263c5e8a-4cd6-414c-aeb1-40c396af1a25

### Official Information

N/A

### Modifications that may affect results

N/A

Daraen1:
I just want to say, it could be affect other skills using opt3. I only test with LK_BERSERK

gidzdlcrz:
> I just want to say, it could be affect other skills using opt3. I only test with LK_BERSERK

You are correct. Alright.

Daraen1:
After testing HP_ASSUMPTIO, LK_CONCENTRATION, MO_STEELBODY, and Gidz tried MO_EXPLOSIONSPIRITS, we can easily confirm every Opt3 skills have the  visual bug

PR #9946, opened 2026-03-21
Added missing flags to fix visual effects not reappearing after re-entering another player’s view

gidzdlcrz:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9945

* **Server Mode**: Both

Status effects that use opt3 and have a corresponding client EFST were not being stored in the display list, causing their visual effects to disappear when a player leaves and re-enters another player's view.

  This change ensures statuses that are both opt3 and have a defined EFST are correctly tracked and re-sent, keeping visual effects consistent across visibility updates.

  Statuses without a defined EFST are intentionally excluded, as they are not sent to the client.

  Thanks to @Daraen1 for pointing out other status effect that use ``op3``.

### DIFF
--- a/db/pre-re/status.yml
+++ b/db/pre-re/status.yml
@@ -305,6 +305,7 @@ Body:
     Flags:
       RequireWeapon: true
       RemoveOnHermode: true
+      DisplayPc: true
     Fail:
       Decreaseagi: true
   - Status: Concentrate
@@ -592,6 +593,7 @@ Body:
     Flags:
       MadoCancel: true
       RemoveOnHermode: true
+      DisplayPc: true
     Fail:
       Maxoverthrust: true
   - Status: Maximizepower
@@ -639,6 +641,7 @@ Body:
       EnergyCoat: true
     Flags:
       RemoveOnHermode: true
+      DisplayPc: true
   - Status: Brokenarmor
     Icon: EFST_BROKENARMOR
     DurationLookup: NPC_ARMORBRAKE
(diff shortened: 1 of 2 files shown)


Daraen1:
Sorry to bump this, but this, affect a lot the gameplay and need to be checked as soon as possible, please.

Playtester:
I'm a bit confused. If we are saying all Opt3 status changes need this, why don't we just do this automatically based on the Opt3 flag rather than having to manually specify it for each?

If we need an extra (redundant) flag each time, the risk is high that someone forgets to set this.

gidzdlcrz:
> I'm a bit confused. If we are saying all Opt3 status changes need this, why don't we just do this automatically based on the Opt3 flag rather than having to manually specify it for each?
> 
> If we need an extra (redundant) flag each time, the risk is high that someone forgets to set this.

I just found out about this as well. I agree with you. So what will becomes of the DisplayPc flag?

Daraen1:
> I'm a bit confused. If we are saying all Opt3 status changes need this, why don't we just do this automatically based on the Opt3 flag rather than having to manually specify it for each?
> 
> If we need an extra (redundant) flag each time, the risk is high that someone forgets to set this.

Can we use this as temporary fix? It's really annoying to not see players statuses tbh

Lemongrass3110:
> I'm a bit confused. If we are saying all Opt3 status changes need this, why don't we just do this automatically based on the Opt3 flag rather than having to manually specify it for each?
> 
> If we need an extra (redundant) flag each time, the risk is high that someone forgets to set this.

The condition seems to be opt3 + a defined EFST. We can catch that via the buildbot.

In this case it is now set incorrectly for some entries, since they dont have an EFST they are not sent to the client.
For example: Incatkrate, Ska, ...

@gidzdlcrz could you please update your PR with this information?

PR #9986, opened 2026-04-09
6th rebalance - Cardinal - Part 1

Atemo:
* **Addressed Issue(s)**: 

* **Server Mode**: renewal

Updated
----------------------------------------
Arbitrium
Reworks skill, removes single-target damage, area damage will deal damage for 2 hits.
- Changes base damage from 10000% MATK to 9500% MATK per hit based on level 10.
- Reduces skill cooldown from 1 second to 0.75 seconds.
- Since single-target damage is removed, the status ailment [quiet] also be removed from the skill. 

Petitio
- Increases base damage from 10500% ATK to 12000% ATK based on level 10.

Effligo
- When equipping mace, damage type is changed to long ranged physical.
- Increases base damage from 16500%/18000%(demon, undead) ATK to 18000%/20000%(demon, undead) ATK based on level 10.

Pneumaticus Procella
- Reduces AP consumption from 20 to 15 based on level 10. 

Informations
----------------------------------------

https://ro.gnjoy.com/news/notice/View.asp?BBSMode=10001&seq=8211&curpage=2
https://www.divine-pride.net/forum/index.php?/topic/3723-kro-jobs-improvement-project/page/15/

### DIFF
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -37113,7 +37113,7 @@ Body:
     FixedCastTime: 1500
     Requires:
       SpCost: 150
-      ApCost: 20
+      ApCost: 15
     Unit:
       Id: Pneumaticus_Procella
       Range:
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -36949,14 +36949,14 @@ Body:
     Type: Magic
     TargetType: Attack
     Range: 9
-    Hit: Single
-    HitCount: 1
+    Hit: Multi_Hit
+    HitCount: 2
     Element: Holy
     GiveAp: 2
     CastTime: 4000
     AfterCastActDelay: 500
     Duration1: 20000
-    Cooldown: 1000
+    Cooldown: 750
     FixedCastTime: 1500
     Requires:
       SpCost:
@@ -43545,7 +43545,7 @@ Body:
     Type: Weapon
     TargetType: Attack
     Range: 2
-    Hit: Multi_hit
+    Hit: Multi_Hit
     HitCount: -2
     GiveAp: 3
     Element: Weapon
--- a/db/re/skill_db.yml
+++ b/db/re/skill_db.yml
@@ -36948,10 +36948,13 @@ Body:
     MaxLevel: 10
     Type: Magic
     TargetType: Attack
+    DamageFlags:
+      Splash: true
     Range: 9
     Hit: Multi_Hit
     HitCount: 2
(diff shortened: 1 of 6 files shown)


src/map/battle.cpp
```diff
@@ -2705,10 +2705,11 @@ static int32 battle_range_type(const block_list* src, const block_list* target,
 		case SS_SHIMIRU: // 11 cell cast range.
 		case SKE_STAR_LIGHT_KICK: // 7 cell cast range.
 			return BF_SHORT;
-		case CD_PETITIO: { // Skill range is 2 but damage is melee with books and ranged with mace.
-			const map_session_data* sd = BL_CAST(BL_PC,src);
+		case CD_EFFLIGO:	// Skill range is 2 but damage is melee with books and ranged with mace.
```
Lemongrass3110, line 2708: Comment does not look correct (range+books)

Lemongrass3110 (approved):

src/map/battle.cpp
Atemo: Not sure what you mean, but 
CD_PETITIO and CD_EFFLIGO skill range is 2 => skills are melee by default.

```
  - Id: 5277
    Name: CD_EFFLIGO
...
    Requires:
      SpCost: 75
      ApCost: 12
      Weapon:
        Mace: true
        2hMace: true
        Book: true
```
^ CD_PETITIO and CD_EFFLIGO can only be used when maces type or book are equipped.

In the code when maces type are equipped, the skills are long range.
Otherwise they are melee by default (so with book, the only weapon remaining)

db/re/skill_db.yml
```diff
@@ -36949,14 +36949,14 @@ Body:
     Type: Magic
     TargetType: Attack
```
Jason59762, line 36950: DamageFlags:
      Splash: true

maybe?

db/re/skill_db.yml
```diff
@@ -36949,14 +36949,14 @@ Body:
...
     Range: 9
-    Hit: Single
-    HitCount: 1
+    Hit: Multi_Hit
+    HitCount: 2
     Element: Holy
```
Jason59762, line 36956: SplashArea: 4

maybe?

db/re/skill_db.yml
Jason59762: it can only deal 2 instances of single-target damage, not area damage.
lost DamageFlags and SplashArea maybe?

Issue #9652, opened 2025-11-19
Life potion effects not being stacked


eppc0330:
### rAthena Hash

master

### Client Date

Its serverside issue

### Server Mode

Renewal

### Result

Now all life potions(`M_LifePotion`, `L_LifePotion`, `S_ManaPotion`, etc) are using `bonus2 bRegenPercentHP` or `bonus2 bRegenPercentSP`
and this itemscript cannot be stacked

### Relevant Log Output

```Shell

```

### Expected Result

Life potion effects must be able to stack

### How to Reproduce

Just use variable life potions

### Official Information

Life potion effects must be able to stack

eppc0330:
There's two ways to fix this issue:
 - Revert #8831, lifepotion part, hardcode again.
 - create new bonus script like `bonus2 bRegenPercentHP2` which is able to be stacked

The first one is better in my opinion, because when Ive done the second way, I found that there's unnatural problem with `bRegenPercent`. This bonus script regenerates based on server time, not a remaining time of the status change.

For example, even if I use `S_LifePotion` and `S_Manapotion` with a gap of 2 or 3 seconds, they regenerate HP and SP on the exactly same time. 
If it works normally, they must regererate HP and SP on different time because both items regenerate HP and SP every 5 seconds 

Since I don't know if this is an official behavior, I'll leave the second method as an option for now until it's confirmed.

PR #10007, opened 2026-04-20
Corrected item regen bonus

Atemo:
* **Addressed Issue(s)**: https://github.com/rathena/rathena/issues/9652

* **Server Mode**: 

The bonus are now stackable like on official server

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -3551,6 +3551,19 @@ static void pc_bonus_addele(map_session_data* sd, unsigned char ele, int16 rate,
 	wd->addele2.push_back(entry);
 }
 
+static void pc_bonus_regen(std::vector<s_regen> &bonus, int16 value, int32 rate) {
+	if (rate < 1)
+		return;
+
+	s_regen entry = {};
+
+	entry.value = value;
+	entry.rate = rate;
+	entry.tick = 0;
+
+	bonus.push_back(entry);
+}
+
 /**
  * Reduce element bonus to player when attacking
  * @param sd: Player data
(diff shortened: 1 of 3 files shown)


Playtester:
Is there any practical pre-re scenario I could test?

Atemo:
For example equip 2 items that use `bHPRegenRate` bonus and note the total hp gained. Before this change the total depended only on the last equipped item that used the bonus. With this change the bonus are cumulative. I only tested the Life potions on renewal for now.

`Refined Bloody Roar` (id 1273) : `bonus2 bHPRegenRate,3,5000;` (+3 hp every 5 seconds)
`Vital Tree Shoes` (id 2450) : `bonus2 bHPRegenRate,30,10000;` (+30 hp every 10 seconds)
`General Egnigem Cenia Card` (id 4352) : `bonus2 bHPRegenRate,50,10000;` (+50 hp every 10 seconds)

Playtester:
> For example equip 2 items that use `bHPRegenRate` bonus and note the total hp gained. Before this change the total depended only on the last equipped item that used the bonus. With this change the bonus are cumulative. I only tested the Life potions on renewal for now.
> 
> `Refined Bloody Roar` (id 1273) : `bonus2 bHPRegenRate,3,5000;` (+3 hp every 5 seconds) `Vital Tree Shoes` (id 2450) : `bonus2 bHPRegenRate,30,10000;` (+30 hp every 10 seconds) `General Egnigem Cenia Card` (id 4352) : `bonus2 bHPRegenRate,50,10000;` (+50 hp every 10 seconds)

That combination wasn't possible in pre-re since Vital Tree Shoes are not slotted.

But I found another test: General Egnigem Cenia Card + Valorous Recovery Wand (https://ro.kokotewa.com/db/itm_info?id=BF_Staff4).

It seems to create two different instances of regular recovery. Instead of the recovery becoming 15 SP / 10 seconds, it's actually a tick for 10 SP and another tick for 5 SP, starting at the time you equipped the item.

The current implementation seems to add them together instead of giving them two independent timers. Otherwise seems to work fine.

Lemongrass3110 (approved):

PR #9624, opened 2025-11-05
Add check if the script is still waiting for input before freeing it

gidzdlcrz:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/9623
https://github.com/rathena/rathena/issues/9653
https://github.com/rathena/rathena/issues/9666
https://github.com/rathena/rathena/issues/9675
https://github.com/rathena/rathena/issues/9834
https://github.com/rathena/rathena/issues/10034

* **Server Mode**: Both

* **Description of Pull Request**: 
  * the script state (sd->st) should remain alive until the player responds. This PR fixes it.

PR #10045, opened 2026-06-19
Enhance item usage logic to prevent script state leaks

KWDe-v:
* **Addressed Issue(s)**:

N/A

* **Server Mode**:

Both

Fixes item scripts that enter a suspended state (e.g. `mes`, `next`, `select`, `input`) being prematurely destroyed after `run_script()` execution.

Previously, `pc_useitem()` always freed `sd->st` immediately after running an item script. This prevented item scripts from properly resuming when waiting for client interaction, causing item dialogs to become unresponsive.

Example:

```yaml
- Id: 12221
  AegisName: Megaphone_
  Name: Megaphone
  Type: Usable
  Script: |
    input @megaphone$;
    announce strcharinfo(0) + ": " + @megaphone$, bc_all, 0xFF0000;
```

When using this item, the `input` command suspends the script while waiting for the player's response. However, the script state is immediately destroyed by `pc_useitem()`, causing the item to stop functioning after the input dialog appears.

This change preserves the script state when the item script is suspended and awaiting client interaction, allowing item dialogs to resume correctly after user input. The original cleanup behavior is preserved for scripts that finish execution immediately, and item scripts are still discarded when another NPC script is already attached to the player to avoid script-state conflicts.

### DIFF
--- a/src/map/pc.cpp
+++ b/src/map/pc.cpp
@@ -6557,7 +6557,11 @@ int32 pc_useitem(map_session_data *sd,int32 n)
 
 	run_script( script, 0, sd->id, fake_nd->id );
 
-	if( sd->st != nullptr ){
+	if( sd->st != nullptr &&
+		( previous_st != nullptr ||
+		 ( sd->st->state != STOP &&
+		   sd->st->state != RERUNLINE &&
+		   sd->st->state != CLOSE ) ) ){
 		script_free_state( sd->st );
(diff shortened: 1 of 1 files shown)


gidzdlcrz:
I actually submitted an earlier PR for the same issue, specifically to address item scripts being freed immediately even when they are waiting for user input.

https://github.com/rathena/rathena/pull/9624

PR #10060, opened 2026-07-23
Gate EXP notify to 2009-10-27

kivutar:
- eAthena: db/packet_db.txt adds 0x07f6 at 2009-10-27aRagexeRE.
- Hercules: src/map/pc.c only calls displayexp for PACKETVER >= 20091027.


Lemongrass3110 (changes requested):
The gate should be only once (inside the clif_displayexp function).
The compiler will optimize away the useless function calls anyway!

Please change this, then it looks good!

Lemongrass3110 (approved):

PR #10064, opened 2026-07-23
Fix 2008 RE packetdb gates

kivutar:
This is the most important change I'd like to contribute: rathena was sending RE packets, while I was in pre-re with an old Sakexe.

- eAthena: db/packet_db.txt has separate 2008-09-10aSakexe and Renewal Client sections.
- Sabine: src/Shared/Network/PacketTable.2500_sakEP12.cs maps 0x0436..0x0439 for 2008-09-10aSakexe.
- Hercules: 2008 Sakray and RE packet tables both define 0x0436..0x0439 with the same sizes.

### DIFF
--- a/src/map/clif_packetdb.hpp
+++ b/src/map/clif_packetdb.hpp
@@ -1192,8 +1192,10 @@
 #endif
 
 // Renewal Clients
+#ifdef PACKETVER_RE_NUM
+
 // 2008-08-27aRagexeRE
-#if PACKETVER >= 20080827
+#if PACKETVER_RE_NUM >= 20080827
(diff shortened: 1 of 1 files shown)


Lemongrass3110 (changes requested):
I understand your issue, but I think this is too early to fix it like this.

I am unsure about the consequences for other clients, since we often only check for PACKETVER instead of the correct client specific one (MAIN, RE, SAK, etc).
We could end up with missing packet length definitions for outgoing packets, as well as missing packet handlers for incoming packets.

This one needs a larger cleanup.
However you could move only the shuffle packets to clif_shuffle and add the checks correctly there?

kivutar:
Yes, thanks for the review.

PR #10065, opened 2026-07-23
Fixed skill cast ACK PACKETVER definition

kivutar:
- eAthena: db/packet_db.txt maps 2008-09-10aSakexe CZ_USE_SKILL to 0x0438.
- eAthena: src/map/clif.c uses 0x013e for ZC_USESKILL_ACK before 2009-11-24.
- Sabine: docs/network/packet_tables/packet_table_2500.md lists CZ_USE_SKILL 0x0438 and ZC_USESKILL_ACK 0x013e.

### DIFF
--- a/src/map/packets_struct.hpp
+++ b/src/map/packets_struct.hpp
@@ -3975,7 +3975,8 @@ struct PACKET_ZC_USESKILL_ACK {
 	uint8 disposable;
 } __attribute__((packed));
 DEFINE_PACKET_HEADER(ZC_USESKILL_ACK, 0x07fb);
-#elif PACKETVER_MAIN_NUM >= 20090406 || PACKETVER_SAK_NUM >= 20080618 || PACKETVER_RE_NUM >= 20080827 || defined(PACKETVER_ZERO)
+// 2008-09-10aSakexe uses shuffled CZ skill packets but still expects the classic cast ACK.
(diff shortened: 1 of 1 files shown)


src/map/packets_struct.hpp
```diff
@@ -3975,7 +3975,8 @@ struct PACKET_ZC_USESKILL_ACK {
 	uint8 disposable;
 } __attribute__((packed));
 DEFINE_PACKET_HEADER(ZC_USESKILL_ACK, 0x07fb);
-#elif PACKETVER_MAIN_NUM >= 20090406 || PACKETVER_SAK_NUM >= 20080618 || PACKETVER_RE_NUM >= 20080827 || defined(PACKETVER_ZERO)
+// 2008-09-10aSakexe uses shuffled CZ skill packets but still expects the classic cast ACK.
+#elif PACKETVER_MAIN_NUM >= 20090406 || PACKETVER == 20080910 || PACKETVER_SAK_NUM >= 20080618 || PACKETVER_RE_NUM >= 20080827 || defined(PACKETVER_ZERO)
```
Lemongrass3110: Any Sakexe >= 2008-06-18 should already use 0x13e
This change looks strange?

NOTE:
Was this packet really only added in 2008?
It rather looks like it has been there from the beginning.

kivutar: I'll double check, thanks for the review.

kivutar: Mh you're right, ZC_USESKILL_ACK = 0x013e is there even in beta 2. It should be fine if we always enable it then?

Lemongrass3110 (approved):

Issue #10042, opened 2026-06-09
debugmes/errormes/mes reject numeric arguments despite conv_str supporting int→string conversion


Stingor:
### rAthena Hash

51ba7e844b40ae8c436cecc0195200ac458c1501

### Client Date

2025-07-16

### Server Mode

Pre-Renewal

### Result

Passing a numeric variable (or literal) to debugmes, errormes, or mes raises a build-in argument type warning and discards the value, even though the underlying string accessor (script_getstr → conv_str_) already converts integers to strings transparently.

### Relevant Log Output

```
[06:43:57] Unexpected type for argument 1. Expected string.
[06:43:57] Data: variable name='.@batch_qty' index=0
[06:43:57] Function: debugmes
```

### Expected Result

Numeric arguments should be accepted and converted to their string representation, as is already done for variant ('v') parameters elsewhere — the workaround `"" + .@n` should not be required, especially for debug/error output commands.

### How to Reproduce

```
.@n = 15;
debugmes .@n;      // warning, value not printed as expected
errormes .@n;      // same
mes .@n;           // same (requires the "" + .@n idiom)
```

### Official Information

The buildin parameter type for these commands is 's' (strict string):

```
BUILDIN_DEF(mes,"s*")
BUILDIN_DEF(debugmes,"s")
BUILDIN_DEF(errormes,"s")
```

The pre-call argument validation (src/map/script.cpp, case 's') rejects an integer-variable reference:

```
case 's':
    if( !data_isstring(data) && !( data_isreference(data) && is_string_variable(name) ) ){
        ShowWarning("Unexpected type for argument %d. Expected string.\n", idx-1);
        ...
    }
```

…yet conv_str_ (used by script_getstr inside these buildins) already handles int → string:

```
else if( data_isint(data) ){ // int -> string
    snprintf(p, ITEM_NAME_LENGTH, "%" PRId64 "", data->u.num);
    ...
}
```

### Modifications that may affect results

Change the parameter type from 's' to 'v' (variant) for these output commands; no body change is needed since script_getstr already converts:

```
BUILDIN_DEF(mes,"v*"),
BUILDIN_DEF(debugmes,"v"),
BUILDIN_DEF(errormes,"v"),
```

Lemongrass3110:
Fixed the formatting for you.

Do you mind opening a PR for this too?

Stingor:
Sure thing

PR #10076, opened 2026-08-03
Changed *mes script command declarations

Stingor:
* **Addressed Issue(s)**: 
https://github.com/rathena/rathena/issues/10042

* **Server Mode**:
Both

Fix mes/logmes/debugmes/errormes rejecting numeric arguments despite conv_str supporting int->string conversion.

### DIFF
--- a/src/map/script.cpp
+++ b/src/map/script.cpp
@@ -27927,7 +27927,7 @@ BUILDIN_FUNC(preg_match) {
 /// for an explanation on args, see add_buildin_func
 struct script_function buildin_func[] = {
 	// NPC interaction
-	BUILDIN_DEF(mes,"s*"),
+	BUILDIN_DEF(mes,"v*"),
 	BUILDIN_DEF(next,""),
 	BUILDIN_DEF(clear,""),
 	BUILDIN_DEF(close,""),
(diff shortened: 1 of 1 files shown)


src/map/script.cpp
```diff
@@ -27859,7 +27859,7 @@ BUILDIN_FUNC(mesemotion){
 	// Validates emotion range
 	if (id < ET_SURPRISE || id >= ET_MAX) {
 		ShowError("buildin_mesemotion: Emotion ID %d is invalid.\n", id);
-		st->state = END;
+		st->state = END;
```
Lemongrass3110: Whatever happened here, please revert?

src/map/script.cpp
```diff
@@ -27870,8 +27870,8 @@ BUILDIN_FUNC(mesemotion){
 	return SCRIPT_CMD_SUCCESS;
 #else
 	ShowError( "buildin_mesemotion: This command requires PACKETVER 2023-03-02 or newer.\n" );
-	st->state = END;
-	return SCRIPT_CMD_FAILURE;
+	st->state = END;
+	return SCRIPT_CMD_FAILURE;
```
Lemongrass3110: Whatever happened here, please revert?

Lemongrass3110 (approved):

src/map/script.cpp
Stingor: sorry whitespaces converted to tab

src/map/script.cpp
Stingor: turn out it was vscode converting pesky crlf to lf

